PR #36, opened 2024-11-26
Update Llama example to Fabric architecture

jakmro:
## Description
Updated Llama example to new React Native architecture.

### Type of change
- [x] Example Update

### Tested on
- [x] iOS
- [x] Android

### Checklist


examples/llama/ios/llama/llama.entitlements
NorbertKlockiewicz, line 1: Add increased memory limit to llama demo app so it doesn't crash when too much ram is consumed

NorbertKlockiewicz (changes requested):

PR #38, opened 2024-11-27
feat: ExecuTorch bindings for android

NorbertKlockiewicz:
## Description
Added TurboModule(ETModule) which let's user use ExecuTorch Module methods such as:
- loadMethod
- loadModule
- loadForward
- forward

### Type of change

### Tested on
- [ ] iOS
- [x] Android

### Checklist


src/ETModule.ts
```diff
@@ -0,0 +1,143 @@
+import { useEffect, useState } from 'react';
+import { Image } from 'react-native';
+import { ETModule } from './native/RnExecutorchModules';
+
+export enum ETError {
+  //react-native-ExecuTorch errors
```
mkopcins: add link to Error.h on which the enums are based

src/ETModule.ts
```diff
@@ -0,0 +1,143 @@
...
+  forward: (input: ETInput, shape: number[]) => Promise<number[]>;
+  loadMethod: (methodName: string) => Promise<void>;
+  loadForward: () => Promise<void>;
+}
+
+const getError = (e: unknown): string => {
+  const error = e as Error;
+  const errorCode = parseInt(error.message, 10);
+  if (errorCode in ETError) return ETError[errorCode] as string;
+  return ETError[-3] as string;
```
mkopcins: -3? maybe this should be -255?

src/ETModule.ts
```diff
@@ -0,0 +1,143 @@
...
+  // Resource errors
+  NotFound = 32,
+  MemoryAllocationFailed = 33,
+  AccessFailed = 34,
+  InvalidProgram = 35,
+
+  // Delegate errors
+  DelegateInvalidCompatibility = 48,
+  DelegateMemoryAllocationFailed = 49,
+  DelegateInvalidHandle = 50,
```
mkopcins: We can use hex here

src/ETModule.ts
```diff
@@ -0,0 +1,143 @@
...
+  }, [modulePath]);
+
+  const forward = async (input: ETInput, shape: number[]) => {
+    if (isModelLoading) {
+      throw new Error(ETError[-2]);
+    }
+
+    const inputType = getTypeIdentifier(input);
+    if (inputType === -1) {
+      throw new Error(ETError[18]);
```
mkopcins: I'm not a fan of this approach as the number can change but the string remain the same. Maybe let's do `ETError[ETError.InvalidArgument]` instead for better readibility.
Or maybe let's do a separate function:
```
const getETError = (errorCode: number | string) : Error =>
    typeof x == 'number' ? new Error(ETError[x]) : new Error(ETError[ETError[x]]);
```

then we can do `throw getETError(ETError.InvalidArgument)`

mkopcins: Now I see you already have `getError` function, maybe allow it to handle string as arguments as well. Then we can just move all of that error logic to Error.ts file

mkopcins (changes requested):

mkopcins (approved):

PR #41, opened 2024-12-02
fix: wrong handle of assets files in release apk

NorbertKlockiewicz:
## Description
The file loaded from React-Native assets is served in a different way in debug and release mode, for now in release mode it was blocking applications from working

### Type of change

### Tested on
- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
+++ b/android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
@@ -1,8 +1,6 @@
 package com.swmansion.rnexecutorch
 
-import android.os.Build
 import android.util.Log
-import androidx.annotation.RequiresApi
 import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
 import com.swmansion.rnexecutorch.utils.Fetcher
@@ -72,7 +70,6 @@ class RnExecutorchModule(reactContext: ReactApplicationContext) :
     promise.resolve("Model loaded successfully")
   }
 
-  @RequiresApi(Build.VERSION_CODES.TIRAMISU)
   override fun loadLLM(
     modelSource: String,
     tokenizerSource: String,
--- a/android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
+++ b/android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
@@ -47,12 +47,13 @@ class RnExecutorchModule(reactContext: ReactApplicationContext) :
   }
 
   private fun downloadResource(
-    url: URL,
+    url: String,
     resourceType: ResourceType,
-    callback: (path: String?, error: Exception?) -> Unit
+    isLargeFile: Boolean = false,
+    callback: (path: String?, error: Exception?) -> Unit,
   ) {
     Fetcher.downloadResource(
-      reactApplicationContext, client, url, resourceType,
+      reactApplicationContext, client, url, resourceType, isLargeFile,
       { path, error -> callback(path, error) },
       object : ProgressResponseBody.ProgressListener {
         override fun onProgress(bytesRead: Long, contentLength: Long, done: Boolean) {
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
chmjkb, line 1: Can we change the name of this file? It is responsible for llama, so current name is super confusing

android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
```diff
@@ -120,7 +119,6 @@ class RnExecutorchModule(reactContext: ReactApplicationContext) :
     }
   }
 
-  @RequiresApi(Build.VERSION_CODES.N)
```
chmjkb, line 123: Is that no longer needed?

android/src/main/java/com/swmansion/rnexecutorch/utils/Fetcher.kt
```diff
@@ -31,14 +31,14 @@ class Fetcher {
       return file
     }
 
-    private fun hasValidExtension(fileName: String, resourceType: ResourceType): Boolean {
+    private fun hasValidExtension(extension: String, resourceType: ResourceType): Boolean {
```
chmjkb: Rename the method, hasValidExtension sounds like a filename should be passed

android/src/main/java/com/swmansion/rnexecutorch/utils/Fetcher.kt
```diff
@@ -132,85 +124,132 @@ class Fetcher {
       return response
     }
 
+    private fun getIdOfResource(context: Context, resourceName: String): Int {
+      return context.resources.getIdentifier(resourceName, "raw", context.packageName)
```
chmjkb: defType can be a default parameter to `getIdOfResource`

android/src/main/java/com/swmansion/rnexecutorch/utils/Fetcher.kt
```diff
@@ -132,85 +124,132 @@ class Fetcher {
...
-        fileName = extractFileName(url)
-      } catch (e: Exception) {
-        onComplete(null, e)
-        return
-      }
+        if (!url.contains("://")) {
+          //The provided file is from react-native assets folder in release mode
+          val resId = getIdOfResource(context, url)
+          val resName = context.resources.getResourceEntryName(resId)
+          val fileExtension = if (resourceType == ResourceType.TOKENIZER) "bin" else "pte"
```
chmjkb: maybe we can move this to a function, i think it is likely that we will re-use this, also this might be a problem if we add any other resource type

android/src/main/java/com/swmansion/rnexecutorch/utils/Fetcher.kt
```diff
@@ -132,85 +124,132 @@ class Fetcher {
...
-      }
+        /*
+          The provided file is either a remote file or a local file
+          - local file: file://path/to/file
+          - remote file: https://path/to/file || http://10.0.2.2:8080/path/to/file
+        */
+        val resUrl = URL(url)
+        if (resUrl.protocol == "file") {
+          // The provided file is a local file, get rid of the file:// prefix and return path
+          val localPath = resUrl.toString().split("://")[1]
```
chmjkb: whats the point of calling `.toString()` right after you turned that string into an URL? I believe you can do the same thing on the URL object. you can use `.path`, for example

android/src/main/java/com/swmansion/rnexecutorch/utils/Fetcher.kt
chmjkb, line 1: Overall, maybe we can move some of functionalities in `downloadResource` to other functions? Its pretty long at this point. I feel like separating some of those if cases to functions would make the code more readable

android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchModule.kt
NorbertKlockiewicz: To change it on android we also have to change it in typescript spec and on ios, as this is a pull request to fix the problem with release builds I think it will be better to change it in separate pull request

chmjkb: sure, we can do it in another PR. When we're at it - would be nice to do the same for iOS 👍🏻

chmjkb (approved):

PR #43, opened 2024-12-05
fix: Change iOS native bindings so they return an array of outputs

chmjkb:
This PR fixes the issue in iOS ExecuTorch bindings where if the model returned multiple output arrays, only the first one was considered.

### Type of change

### Tested on
- [x] iOS
- [ ] Android

### Checklist


third-party/ios/ExecutorchLib/ExecutorchLib/Utils.hpp
```diff
@@ -12,7 +12,14 @@
 #import <Foundation/Foundation.h>
 #endif
 
+template <typename T> struct DataPtrWithNumel {
```
msluszniak: I think you can think of utilizing here `std::span` to avoid raw pointers with their sizes

third-party/ios/ExecutorchLib/ExecutorchLib/Utils.mm
```diff
@@ -62,42 +80,60 @@
...
-    numel = outputTensor.numel();
-    return outputTensor.const_data_ptr<T>();
+    std::vector<DataPtrWithNumel<T>> outputVec;
+
+    for (const auto &currentResult : *result) {
+      DataPtrWithNumel<T> currentDataPtr;
+      Tensor currentTensor = currentResult.toTensor();
+      currentDataPtr.dataPtr = currentTensor.const_data_ptr<T>();
+      currentDataPtr.numel = currentTensor.numel();
+      outputVec.push_back(currentDataPtr);
```
msluszniak: I think there you should use `std::move` in this push_back as otherwise it will create copy and as you dont use `result` anymore, you can transfer ownership

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -34,38 +34,37 @@ - (NSArray *)forward:(NSArray *)input
...
   std::vector<int> shapes = NSArrayToIntVector(shape);
-  ssize_t outputNumel = 0;
   @try {
     switch (inputTypeIntValue) {
     case 0: {
       // Int8Array
-      const int8_t *output =
-          runForwardFromNSArray<int8_t>(input, outputNumel, shapes, _model);
-      return arrayToNSArray<int8_t>(output, outputNumel);
+      std::vector<DataPtrWithNumel<int8_t>> output =
```
msluszniak: I don't know how invasive this change would be, but changing `inputTypeIntValue` to enum type seems reasonable here

msluszniak: This way, you would get rid of this magic numbers

chmjkb: Good point, thanks for contributing:D

third-party/ios/ExecutorchLib/ExecutorchLib/Utils.mm
```diff
@@ -48,6 +47,21 @@
...
+NSArray *arrayToNSArray(const std::vector<std::span<const T>> &dataPtrVec) {
+    NSMutableArray *nsArray = [NSMutableArray array];
+    for (const auto &span : dataPtrVec) {
+        const T *dataPtr = span.data();
+        ssize_t numel = span.size();
+        NSMutableArray *innerArray = [NSMutableArray arrayWithCapacity:numel];
+        for (ssize_t i = 0; i < numel; ++i) {
+            [innerArray addObject:@(dataPtr[i])];
+        }
+        [nsArray addObject:[innerArray copy]];
```
mkopcins:
```suggestion
        NSMutableArray *innerArray = [NSMutableArray arrayWithCapacity:span.size()];
        for(auto x : span) {
            [innerArray addObject:@(x)];
        }
        [nsArray addObject:[innerArray copy]];
    }
```

third-party/ios/ExecutorchLib/ExecutorchLib/Utils.mm
```diff
@@ -62,42 +76,58 @@
...
+template NSArray *arrayToNSArray<int64_t>(
+    const std::vector<std::span<const int64_t>> &dataPtrVec);
+template NSArray *
+arrayToNSArray<float>(const std::vector<std::span<const float>> &dataPtrVec);
+template NSArray *
+arrayToNSArray<double>(const std::vector<std::span <const double>> &dataPtrVec);
+
 template NSArray *arrayToNSArray<int8_t>(const void *array, ssize_t numel);
 template NSArray *arrayToNSArray<int32_t>(const void *array, ssize_t numel);
 template NSArray *arrayToNSArray<int64_t>(const void *array, ssize_t numel);
```
mkopcins: this is getting a bit out of hand, let's just move all this code to the header file

mkopcins (changes requested):

mkopcins (approved):

PR #44, opened 2024-12-05
feat: BaseModel abstract class, primitive Style Transfer, functions for retrieving model metadata

NorbertKlockiewicz:
## Description
This pull request introduces Model base class on top of which every other model should be built, there's also a primitive implementation of StyleTransferModel and functions for retrieving model metadata. Fix for bindings allowing to return multiple outputs.

**__During code review don't focus on StyleTransfer Module as it will change and it's for test purposes right now__**

Don't review:
- StyleTransfer.kt
- StyleTransferModel.kt
- StyleTransfer.h
- StyleTransfer.mm
- StyleTransferModel.mm
- StyleTransferModel.h
- BitmapUtils.kt
- Imageprocessor.h/Imageprocessor.m
- TensorUtils.kt

### Type of change

### Tested on
- [x] iOS
- [x] Android

### Checklist


android/build.gradle
```diff
@@ -97,6 +97,6 @@ dependencies {
   //noinspection GradleDynamicVersion
   implementation "com.facebook.react:react-native:+"
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-  implementation 'com.github.software-mansion:react-native-executorch:main-SNAPSHOT'
```
jakmro: I think there should be no change here.

android/libs/executorch-llama.aar
jakmro, line 1: Instead of adding this file, the `third-party/android/libs/executorch.aar` should be updated.

jakmro (changes requested):

android/build.gradle
NorbertKlockiewicz: In my opinion on main we should have a version of library that have every feature working. As for now we don't want to make a new release for android lib as it can change before 0.2.0, that's why it's linked locally.

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.swmansion.rnexecutorch.utils.ProgressResponseBody
+import com.swmansion.rnexecutorch.utils.ResourceType
+import okhttp3.OkHttpClient
+import org.pytorch.executorch.EValue
+import org.pytorch.executorch.Module
+
+
+abstract class BaseModel<Input, Output>(val context: Context) {
+  protected lateinit var module: Module
+  private val client = OkHttpClient()
```
chmjkb: Does each model need to have his own client? can't we share the client across models?

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
```diff
@@ -0,0 +1,66 @@
...
+
+  private fun downloadModel(
```
chmjkb: same here, wouldn't it make more sense to move the downlaoding logic outside of the Model class? Notice that you're defining the exact same function in ETModule

chmjkb: We could pass a fetcher instance to this class and just call a method instead

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
```diff
@@ -0,0 +1,66 @@
...
+        modelSource
+      ) { path, error ->
+        if (error != null) {
+          throw Error(error.message!!)
+        }
+
+        module = Module.load(path)
+      }
+    } catch (e: Exception) {
+      throw Error(e.message!!)
```
chmjkb: If the `.load()` throws an exception, is the `.message` the ExecuTorch error code?

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  protected fun forward(input: EValue): Array<EValue> {
+    try {
+      val result = module.forward(input)
+      return result
+    } catch (e: IllegalArgumentException) {
+      //The error is thrown when transformation to Tensor fails
+      throw Error("18")
```
chmjkb: Why 18? Kind of a magic number, can we do sth about it? Maybe we could make use of some enum here

examples/llama/ios/llama.xcodeproj/project.pbxproj
```diff
@@ -363,15 +357,21 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = llama/llama.entitlements;
+				CODE_SIGN_IDENTITY = "Apple Development";
+				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
+				DEVELOPMENT_TEAM = B357MU264T;
```
chmjkb, line 363: Should this be here?

examples/llama/ios/llama.xcodeproj/project.pbxproj
```diff
@@ -399,10 +396,16 @@
...
```
chmjkb, line 402: Should this be here?

examples/llama/package.json
```diff
@@ -16,7 +16,7 @@
     "metro-config": "^0.81.0",
     "react": "18.3.1",
     "react-native": "0.76.3",
-    "react-native-executorch": "^0.1.2",
+    "react-native-executorch": "^0.1.3",
```
chmjkb, line 19: just a note, there is no need to do that, since the caret will automatically get the latest patch here

ios/RnExecutorch/models/BaseModel.mm
```diff
@@ -0,0 +1,35 @@
...
+
+@implementation BaseModel
+
+- (NSArray *)forward:(NSArray *)input shape:(NSArray *)shape inputType:(NSNumber *)inputType {
+    @try {
+        NSArray *result = [module forward:input shape:shape inputType:inputType];
+        return result;
+    } @catch (NSException *exception) {
+        NSLog(@"Exception encountered: %@", [exception reason]);
+        return @[@10, @10, @10]; // Error handling logic here
```
chmjkb: im guessing this is isn\t finished

ios/RnExecutorch/models/BaseModel.mm
```diff
@@ -0,0 +1,35 @@
...
+#import "../utils/Fetcher.h"
+
+@implementation BaseModel
+
+- (NSArray *)forward:(NSArray *)input shape:(NSArray *)shape inputType:(NSNumber *)inputType {
+    @try {
+        NSArray *result = [module forward:input shape:shape inputType:inputType];
+        return result;
+    } @catch (NSException *exception) {
+        NSLog(@"Exception encountered: %@", [exception reason]);
```
chmjkb: Redundant log

ios/RnExecutorch/models/BaseModel.mm
```diff
@@ -0,0 +1,35 @@
...
+        return @[@10, @10, @10]; // Error handling logic here
+    }
+}
+
+- (void)loadModel:(NSURL *)modelURL completion:(void (^)(BOOL success, NSNumber* code))completion {
+    module = [[ETModel alloc] init];
+    [Fetcher fetchResource:modelURL resourceType:ResourceType::MODEL completionHandler:^(NSString *filePath, NSError *error) {
+        if (error) {
+            NSLog(@"Error loading model: %@", [error localizedDescription]);
```
chmjkb: Redundant log

src/StyleTransfer.ts
chmjkb, line 1: Overall, this is very similar to ETModule.ts. The `loadModel` function is basically the same. Forward looks very similar as well, i think we should do something about it.

src/StyleTransfer.ts
```diff
@@ -0,0 +1,62 @@
...
+  isModelRunning: boolean;
+  forward: (input: string) => Promise<string>;
+}
+
+export const useStyleTransfer = ({
+  modulePath,
+}: Props): StyleTransferModule => {
+  const [error, setError] = useState<null | string>(null);
+  const [isModelLoading, setIsModelLoading] = useState(true);
+  const [isModelRunning, setIsModelRunning] = useState(false);
```
chmjkb: `isModelGenerating` would be better i think. Analogously to what we did with LLaMA. This change would also need to be done in ETModule.ts

src/native/NativeStyleTransfer.ts
```diff
@@ -0,0 +1,10 @@
+import type { TurboModule } from 'react-native';
+import { TurboModuleRegistry } from 'react-native';
+
+export interface Spec extends TurboModule {
+  loadModule(modelSource: string): Promise<number>;
+
+  forward(input: string): Promise<string>;
```
chmjkb, line 7: Is string input the only possible option? Maybe we should allow passing plain arrays as well

ios/RnExecutorch/models/BaseModel.h
```diff
@@ -0,0 +1,14 @@
...
+#import "ETModel.h"
+
+@interface BaseModel : NSObject
+{
+@protected
+    ETModel *module;
+}
+
+- (NSArray *)forward:(NSArray *)input shape:(NSArray *)shape inputType:(NSNumber *)inputType;
+- (void)loadModel:(NSURL *)modelURL completion:(void (^)(BOOL success, NSNumber *code))completion;
```
chmjkb: Why don't we have the preprocessing and postprocessing functions here?

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
 }
 
+- (NSNumber *)getNumberOfInputs {
+  const auto method_meta = _model->method_meta("forward");
+  
+  if (method_meta.ok()) {
+    return @(method_meta->num_inputs());
+  }
+  
+  return @-1;
```
chmjkb: `method_meta.error()` should be returned instead of a random -1. Actually, an error should be thrown with a message from `method_meta.error()`. Then that error should be caught within the ETModule.mm, and the promise should be rejected with a proper message.

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+  const auto method_meta = _model->method_meta("forward");
+  
+  if(method_meta.ok()){
+    const auto input_meta = method_meta->input_tensor_meta([index unsignedLongValue]);
+    if(input_meta.ok()){
+      return [self getTypeAsNumber:input_meta->scalar_type()];
+    }
+  }
+  
+  return @-1;
```
chmjkb: same as above

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+  
+  if (method_meta.ok()) {
+    return @(method_meta->num_inputs());
+  }
+  
+  return @-1;
+}
+
+- (NSNumber *)getInputType:(NSNumber *)index {
+  const auto method_meta = _model->method_meta("forward");
```
chmjkb, line 46: Maybe we could parametrise the method name

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+      
+      for(int i = 0; i < shape.size(); i++) {
+        [nsShape addObject:@(shape[i])];
+      }
+      
+      return [nsShape copy];
+    }
+  }
+  
+  return nil;
```
chmjkb: I don't think we should do that

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+};
+
+- (NSNumber *)getNumberOfOutputs {
+  const auto method_meta = _model->method_meta("forward");
+  
+  if (method_meta.ok()) {
+    return @(method_meta->num_outputs());
+  }
+  
+  return @-1;
```
chmjkb: same as above

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+  const auto method_meta = _model->method_meta("forward");
+  
+  if(method_meta.ok()){
+    const auto output_meta = method_meta->output_tensor_meta([index unsignedLongValue]);
+    if(output_meta.ok()){
+      return [self getTypeAsNumber:output_meta->scalar_type()];
+    }
+  }
+  
+  return @-1;
```
chmjkb: same as above

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+      
+      for(int i = 0; i < shape.size(); i++) {
+        [nsShape addObject:@(shape[i])];
+      }
+      
+      return [nsShape copy];
+    }
+  }
+  
+  return nil;
```
chmjkb: same as above

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+    case ScalarType::Int: return @1;
+    case ScalarType::Long: return @2;
+    case ScalarType::Float: return @3;
+    case ScalarType::Double: return @4;
+      
+    default:
+      return @-1;
+  }
+}
+
```
chmjkb: I introduced a new enum called InputType:
https://github.com/software-mansion/react-native-executorch/pull/43/commits/bcff4224af407a184085c6841eac6e023f4f1271,
maybe we could make use of it here, for example return that enum instead of the numbers

chmjkb (changes requested):

ios/RnExecutorch/models/BaseModel.h
NorbertKlockiewicz: To write an abstract functions in objective-c++ it's required to first write class in pure c++ and then bridge it between c++ and objective-c++. We tried this and it introduced a lot of complexity which in my opinion is unecessary here.

src/native/NativeStyleTransfer.ts
NorbertKlockiewicz: I believe that we talked about it and decided to let users pass plain arrays with bindings and for dedicated hooks it should accept more "high level" input so in this case url or base64.

ios/RnExecutorch/models/BaseModel.h
chmjkb: If so, then I'm wondering what is even the purpose of this here? How will we use the BaseModel?

src/native/NativeStyleTransfer.ts
chmjkb: ah, i thought we were doing both, but we can stick to strings then

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
NorbertKlockiewicz: From what I remember the android implementation of load doesn't throw any erorrs, so every error thrown from here is "FetcherError" which is handled in the class which calls this method

ios/RnExecutorch/models/BaseModel.h
NorbertKlockiewicz: It's implementing forward and loadModel functions, you don't have to implement them for example in StyleTransferModel. You can just use [self forward: someInput], if we remove BaseModel you will have to implement forward in every model class.

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
mkopcins: Let's use similar solution as in ts

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
mkopcins: Could we actually implement a Singleton for this? Sth like this
```
object OkHttpClientSingleton {
    val instance: OkHttpClient = OkHttpClient.Builder()
        // Configure the client as needed
        ...
        .build()
}
```

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
```diff
@@ -0,0 +1,66 @@
...
+      client,
+      url,
+      ResourceType.MODEL,
+      false,
+      { path, error -> callback(path, error) },
+      object : ProgressResponseBody.ProgressListener {
+        override fun onProgress(bytesRead: Long, contentLength: Long, done: Boolean) {
+        }
+      })
+  }
```
mkopcins: Let's move the callback inside the `downloadModel` method, as it is right now is a bit convoluted for no apparent reason (unless there is one I'm missing)

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
```diff
@@ -0,0 +1,66 @@
...
+  protected fun forward(input: EValue): Array<EValue> {
+    try {
+      val result = module.forward(input)
+      return result
+    } catch (e: IllegalArgumentException) {
+      //The error is thrown when transformation to Tensor fails
+      throw Error("18")
+    } catch (e: Exception) {
+      //Executorch forward method throws an exception with a message: "Method forward failed with code XX"
+      val exceptionCode = e.message!!.substring(e.message!!.length - 2)
```
mkopcins: I think this a bit of a hurdle to maintain, what if the error format from ET changes? Let's check if the message matches a regex first or match it against the pattern typed out in the comment, if it doesn't let's then return the whole error message unchanged

ios/RnExecutorch/models/BaseModel.h
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+#import "ETModel.h"
+
+@interface BaseModel : NSObject
+{
+@protected
+    ETModel *module;
```
mkopcins, line 8: Just a thought, shouldn't this be `model`? `module` seems confusing since we have `ETModule` which is an abstraction over the `ETModel`?

ios/RnExecutorch/models/ImageProcessor.h
mkopcins, line 1: move this to `Utils` directory

ios/RnExecutorch/models/ImageProcessor.m
mkopcins, line 1: move this to `Utils` directory

src/StyleTransfer.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    try {
+      setIsModelRunning(true);
+      const output = await StyleTransfer.forward(input);
+      return output;
+    } catch (e) {
+      throw new Error(getError(e));
+    } finally {
+      setIsModelRunning(false);
+    }
```
mkopcins: I would shy away from returning from try block like this, while this is valid cod it is slightly less readable than if it was placed outside the `try/catch/finally` block
```suggestion
    try {
      setIsModelRunning(true);
      const output = await StyleTransfer.forward(input);
    } catch (e) {
      throw new Error(getError(e));
    } finally {
      setIsModelRunning(false);
    }
    return output;
```

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+      const auto shape = input_meta->sizes();
+      NSMutableArray *nsShape = [[NSMutableArray alloc] init];
+      
+      for(int i = 0; i < shape.size(); i++) {
+        [nsShape addObject:@(shape[i])];
+      }
+      
+      return [nsShape copy];
+    }
+  }
```
mkopcins: avoid such nesting
```suggestion
  const auto input_meta = method_meta->input_tensor_meta([index unsignedLongValue]);
  if(!method_meta.ok()){
    //handler error here
    return ...;
  }
    
  if(!input_meta.ok()){
    //handler error here
    return ...;
  }
  const auto shape = input_meta->sizes();
  NSMutableArray *nsShape = [[NSMutableArray alloc] init];
      
  for(int i = 0; i < shape.size(); i++) {
    [nsShape addObject:@(shape[i])];
  }
      
  return [nsShape copy];
```

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
mkopcins: agreed, let's return the `meta` error

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+
+- (NSNumber *)getOutputType:(NSNumber *)index {
+  const auto method_meta = _model->method_meta("forward");
+  
+  if(method_meta.ok()){
+    const auto output_meta = method_meta->output_tensor_meta([index unsignedLongValue]);
+    if(output_meta.ok()){
+      return [self getTypeAsNumber:output_meta->scalar_type()];
+    }
+  }
```
mkopcins: refactor if statements as above

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
```diff
@@ -29,6 +29,105 @@ - (NSNumber *)loadForward {
...
+      const auto shape = output_meta->sizes();
+      NSMutableArray *nsShape = [[NSMutableArray alloc] init];
+      
+      for(int i = 0; i < shape.size(); i++) {
+        [nsShape addObject:@(shape[i])];
+      }
+      
+      return [nsShape copy];
+    }
+  }
```
mkopcins: refactor if statements as above

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/ETModel.mm
mkopcins: Yes, lets do that

src/StyleTransfer.ts
mkopcins: hmm, not so sure about that, this implies we would do also `isModelSegmenting` and `isModelDetecting` and so on, unless you suggest we should use `isModelGenerating` in all cases

src/StyleTransfer.ts
mkopcins: Hmm, yes this is something we should consider changing, but I think this is a bigger change so let's not do that right now and focus on the release, then we can do a bit of clean up with the New Year (new year, new codebase xD)

NorbertKlockiewicz: I am aware of this issue, however I didn't want to introduce another change in this pr as it's already a bit big, but we can create something like ETModel which extends BaseModel, that way we will reduce redundant implementation of loadModel and forward methods

src/StyleTransfer.ts
NorbertKlockiewicz: I had same thoughts as @mkopcins but we can stick to `isModelGenerating` as it should be understandable for everyone

ios/RnExecutorch/models/BaseModel.h
NorbertKlockiewicz: I am not sure about it but that's how it was called in ETModel.mm so I wanted to keep it the same here

android/src/main/java/com/swmansion/rnexecutorch/models/BaseModel.kt
NorbertKlockiewicz: We are controlling what error message is thrown from executorch, I will modify it to return only code so we overcome overhead with processing whole message

ios/RnExecutorch/models/BaseModel.mm
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+- (void)loadModel:(NSURL *)modelURL completion:(void (^)(BOOL success, NSNumber* code))completion {
+    module = [[ETModel alloc] init];
+    [Fetcher fetchResource:modelURL resourceType:ResourceType::MODEL completionHandler:^(NSString *filePath, NSError *error) {
+        if (error) {
+          completion(NO, @-1);
+          return;
+        }
```
mkopcins: indentation
```suggestion
      if (error) {
        completion(NO, @-1);
        return;
      }
```

mkopcins (approved):

chmjkb (approved):

PR #48, opened 2024-12-10
feat: style transfer with openCV(ios)

NorbertKlockiewicz:
## Description
This PR introduces implementation for style transfer module which is using opencv under the hood, it allows to load images from external sources, device storage and base64.

### Type of change

### Tested on
- [x] iOS
- [ ] Android

### Checklist


src/StyleTransfer.ts
```diff
@@ -30,23 +30,27 @@ export const useStyleTransfer = ({
...
 
       try {
-        setIsModelLoading(true);
+        setIsModelReady(false);
         await StyleTransfer.loadModule(path);
       } catch (e) {
         setError(getError(e));
       } finally {
-        setIsModelLoading(false);
+        setIsModelReady(true);
```
mkopcins: This looks a bit off to me, why do we set `isModelReady` to `true` even if an error occurs? I think it is more intuitive to treat `isModelReady` like a superset of `isModelLoadedSuccessfully`.
Also on this subject, change default value for `isModelReady` to false and only set it to `true` after loading succeeds.
Also perhaps we should move `setIsModelReady(false)` outside of the `async` block to avoid any possible weird race conditions when forward is called before `isModelReady` is set to `false`

src/StyleTransfer.ts
```diff
@@ -30,23 +30,27 @@ export const useStyleTransfer = ({
...
   const forward = async (input: string) => {
-    if (isModelLoading) {
+    if (!isModelReady) {
       throw new Error(getError(ETError.ModuleNotLoaded));
     }
 
+    if (error) {
+      throw new Error(error);
+    }
+
```
mkopcins: This shouldn't be necessary if we keep invariant that: isModelReady is true only if error is null.

src/StyleTransfer.ts
```diff
@@ -30,23 +30,27 @@ export const useStyleTransfer = ({
...
+        setIsModelReady(true);
       }
     };
 
     loadModel();
   }, [modulePath]);
 
   const forward = async (input: string) => {
-    if (isModelLoading) {
+    if (!isModelReady) {
```
mkopcins, line 45: Considering this is async, perhaps we should also check if `isModelGenerating` is `false`

src/StyleTransfer.ts
```diff
@@ -30,23 +30,27 @@ export const useStyleTransfer = ({
...
       throw new Error(getError(ETError.ModuleNotLoaded));
     }
 
+    if (error) {
+      throw new Error(error);
+    }
+
     try {
       setIsModelGenerating(true);
       const output = await StyleTransfer.forward(input);
```
mkopcins, line 54: Below this line as I can't comment there, move `setIsModelGenerating(false)` to `finally`

ios/RnExecutorch/utils/ImageProcessor.mm
```diff
@@ -0,0 +1,96 @@
...
+        @throw [NSException exceptionWithName:@"readImage_error"
+                                   reason:[NSString stringWithFormat:@"%ld", (long)InvalidArgument]
+                                 userInfo:nil];
+    }
+    NSString *encodedString = parts[1];
+    NSData *data = [[NSData alloc] initWithBase64EncodedString:encodedString options:NSDataBase64DecodingIgnoreUnknownCharacters];
+    cv::Mat encodedData(1, [data length], CV_8UC1, (void *)data.bytes);
+    inputImage = cv::imdecode(encodedData, cv::IMREAD_COLOR);
+  }
+  if([[url scheme] isEqualToString: @"file"]){
```
mkopcins:
```suggestion
 else if([[url scheme] isEqualToString: @"file"]){
```

ios/RnExecutorch/models/StyleTransferModel.mm
```diff
@@ -1,88 +1,46 @@
...
-  NSMutableArray *array = [NSMutableArray arrayWithCapacity:length];
+- (NSArray *)preprocess:(cv::Mat &)input {
+  self->originalSize = cv::Size(input.cols, input.rows);
   
-  for (size_t i = 0; i < length; ++i) {
-    NSNumber *number = [NSNumber numberWithFloat:floatArray[i]];
-    [array addObject:number];
-  }
+  cv::Size modelImageSize = [self getModelImageSize];
+  cv::resize(input, input, modelImageSize);
```
mkopcins: This seems wrong, what if the `input` is smaller than `modelImageSize`?

ios/RnExecutorch/models/StyleTransferModel.mm
```diff
@@ -1,88 +1,46 @@
...
-  return [ImageProcessor resizeImage:input toSize:targetSize];
-}
-
-- (UIImage *)postprocess:(UIImage *)input {
-  // Assume any necessary format conversions or adjustments
-  return input;
+- (cv::Mat)postprocess:(NSArray *)output {
+  cv::Size modelImageSize = [self getModelImageSize];
+  cv::Mat processedImage = [ImageProcessor arrayToMat: output width:modelImageSize.width height:modelImageSize.height];
+  cv::resize(processedImage, processedImage, originalSize);
```
mkopcins: Again, what if the `originalSize` is larger than  `modelImageSize`?

mkopcins (changes requested):

ios/RnExecutorch/models/StyleTransferModel.mm
mkopcins: never mind this, I thought initially this is an array pointer and not a `cv::Mat` object. Anyway since this is not performed 'inplace' due to the above reason I would consider creating another `cv::Mat` object `output` for better readability

src/StyleTransfer.ts
```diff
@@ -18,7 +18,7 @@ export const useStyleTransfer = ({
   modulePath,
 }: Props): StyleTransferModule => {
   const [error, setError] = useState<null | string>(null);
-  const [isModelLoading, setIsModelLoading] = useState(true);
+  const [isModelReady, setIsModelReady] = useState(true);
```
mkopcins:
```suggestion
  const [isModelReady, setIsModelReady] = useState(false);
```

src/StyleTransfer.ts
```diff
@@ -30,33 +30,34 @@ export const useStyleTransfer = ({
       }
 
       try {
-        setIsModelLoading(true);
         await StyleTransfer.loadModule(path);
```
mkopcins, line 34: add `setIsModelReady(false);` here, in case `modulePath` prop changes

mkopcins (approved):

PR #50, opened 2024-12-12
refactor: change naming of RnExecutorchModule to LLM

NorbertKlockiewicz:
## Description
Turbo module which was handling running of llama models was named RnExecutorchModule which wasn't really clear and conflicted with name of library, it's now changed to LLM, matching the style of StyleTransfer, ObjectDetection etc.

### Type of change

### Tested on
- [x] iOS
- [x] Android

### Checklist


mkopcins (changes requested):
Considering this is not a general LLM utility but rather Llama interface I would suggest using other name like LLamaModule or something closer to that. Especially that we are using function naming like `initializeLlamaModule`

mkopcins (approved):

PR #64, opened 2024-12-18
chore:refactor ts cv hooks

mkopcins:
### Type of change

### Tested on
- [ ] iOS
- [x] Android

### Checklist


src/types/types.ts
chmjkb, line 1: maybe we can rename this to common or something like this? seems more intuitive to me

src/types/object_detection_types.ts
chmjkb, line 1: Maybe we can get rid of the `types` part in the file name? the parent directory kind of says that the types are in that file

src/useModule.ts
```diff
@@ -1,45 +1,49 @@
 import { useEffect, useState } from 'react';
 import { Image } from 'react-native';
-import { StyleTransfer } from './native/RnExecutorchModules';
 import { ETError, getError } from './Error';
+import { _ClassificationModule, _ObjectDetectionModule, _StyleTransferModule } from './native/RnExecutorchModules';
 
 interface Props {
   modelSource: string | number;
+  module: _ClassificationModule | _StyleTransferModule | _ObjectDetectionModule;
```
chmjkb: I think this type should be moved somewhere, as it might grow

chmjkb (commented):
Can we maybe somehow use ETModule.ts for this?

examples/computer-vision/assets/ssdlite320-mobilenetv3-large.pte
chmjkb, line 1: 👎🏻

chmjkb (changes requested):

chmjkb (approved):

Issue #42, opened 2024-12-04
Bug: Response Generation Completes Without Proper Indicator


ajmalp5:
### Description

There is a case where llma.response returns data and isModelGenerating is false, but the model is still generating a response. Due to this inconsistency, it becomes impossible to determine when the response generation has fully completed, making it challenging to handle the response appropriately in the database.

### Steps to reproduce

1. Start with an initial model where llama.response and llama.isModelGenerating work as expected.
2. Switch to a different model.
3. Generate responses multiple times after switching models.
4. Observe that the issue occurs where llama.response and llama.isModelGenerating is false, even though the response is still being generated

**Expected Behavior**
The response generation process should have a clear indicator when it has fully completed, ensuring data is stored to the database only after generation finishes.
**Actual Behavior**
llama.response and !llama.isModelGenerating trigger prematurely, leading to saving the response multiple times during generation.

```
useEffect(() => {
  const { response, isModelGenerating } = llama;
  console.log(response, isModelGenerating);
  if (llama.response && !llama.isModelGenerating) {
    console.log('Message is stored to DB after finishing generating');
    addMessage(db, parseInt(chatIdRef.current), {
      content: llama.response,
      role: Role.Bot,
    });
  }
}, [llama.response, llama.isModelGenerating]);

```

### React Native Executorch version

0.1.2

### React Native version

0.76.3

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

Iphone 14 pro

### AI model

llama-3.2-1B/QLoRA/llama3_2_qat_lora.pte , v0.1.0/llama-3.2-3B/spinquant/llama3_2_3B_spinquant.pte

### Acknowledgements

Yes

NorbertKlockiewicz:
Hello,

Thank you for reporting this issue. We are aware of it and plan to resolve it in an upcoming release.

ajmalp5:
You're doing a great job. Thanks!

PR #73, opened 2024-12-19
fix: issue with isModelGenerating when switching between multiple models

NorbertKlockiewicz:
## Description
There was a problem when user was switching between multiple llms in one component, to fix this issue I removed code related to handling strict mode problems(strict mode was causing problems with event listeners also so I think we should ignore it as it isn't the best with background tasks), I've also remove deleteModule native function as it wasn't really doing anything, now user can download multiple llms within one component and seamlessly switch between them without bugs.
The problem was mentioned in issue #42 

### Type of change

### Tested on
- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
+++ b/android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
@@ -64,6 +64,7 @@ class LLM(reactContext: ReactApplicationContext) :
   private fun initializeLlamaModule(modelPath: String, tokenizerPath: String, promise: Promise) {
     llamaModule = LlamaModule(1, modelPath, tokenizerPath, 0.7f)
     isFetching = false
+    this.tempLlamaResponse.clear()
     promise.resolve("Model loaded successfully")
   }
 
@@ -74,8 +75,8 @@ class LLM(reactContext: ReactApplicationContext) :
     contextWindowLength: Double,
     promise: Promise
   ) {
-    if (llamaModule != null || isFetching) {
-      promise.reject("Model already loaded", "Model is already loaded or fetching")
+    if (isFetching) {
+      promise.reject("Model already loaded", "Model is already fetching")
       return
     }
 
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
```diff
@@ -74,8 +75,8 @@ class LLM(reactContext: ReactApplicationContext) :
     contextWindowLength: Double,
     promise: Promise
   ) {
-    if (llamaModule != null || isFetching) {
-      promise.reject("Model already loaded", "Model is already loaded or fetching")
+    if (isFetching) {
+      promise.reject("Model already loaded", "Model is already fetching")
```
chmjkb: The first argument seems wrong

ios/RnExecutorch/LLM.mm
```diff
@@ -54,8 +54,8 @@ - (void)updateDownloadProgress:(NSNumber *)progress {
 - (void)loadLLM:(NSString *)modelSource tokenizerSource:(NSString *)tokenizerSource systemPrompt:(NSString *)systemPrompt contextWindowLength:(double)contextWindowLength resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject {
   NSURL *modelURL = [NSURL URLWithString:modelSource];
   NSURL *tokenizerURL = [NSURL URLWithString:tokenizerSource];
-    
-  if(self->runner || isFetching){
+  
+  if(isFetching){
     reject(@"model_already_loaded", @"Model and tokenizer already loaded", nil);
```
chmjkb: Is this error message correct? you're not checking if its loaded

android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
```diff
@@ -148,10 +149,6 @@ class LLM(reactContext: ReactApplicationContext) :
     llamaModule!!.stop()
   }
 
-  override fun deleteModule() {
```
chmjkb: Why are we getting rid of this?

NorbertKlockiewicz: I noticed that it doesn't really change memory usage, but let's leave it for now as it's not breaking anything

chmjkb (approved):

PR #96, opened 2025-02-13
feat: ocr(android)

NorbertKlockiewicz:
### Type of change

### Tested on
- [ ] iOS
- [x] Android

### Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/rnexecutorch/models/ocr/RecognitionHandler.kt
+++ b/android/src/main/java/com/swmansion/rnexecutorch/models/ocr/RecognitionHandler.kt
@@ -1,0 +1,133 @@
+package com.swmansion.rnexecutorch.models.ocr
+
+import android.util.Log
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.WritableArray
+import com.swmansion.rnexecutorch.models.ocr.utils.BBoxPoint
+import com.swmansion.rnexecutorch.models.ocr.utils.CTCLabelConverter
+import com.swmansion.rnexecutorch.models.ocr.utils.OCRbBox
+import com.swmansion.rnexecutorch.models.ocr.utils.RecognizerUtils
+import com.swmansion.rnexecutorch.utils.ImageProcessor
+import org.opencv.core.Core
+import org.opencv.core.Mat
+
+const val modelHeight = 64
+const val largeModelWidth = 512
+const val mediumModelWidth = 256
+const val smallModelWidth = 128
+const val lowConfidenceThreshold = 0.3
+const val adjustContrast = 0.2
+
+class RecognitionHandler(
+  symbols: String,
+  languageDictPath: String,
+  reactApplicationContext: ReactApplicationContext
+) {
+  private val recognizerLarge = Recognizer(reactApplicationContext)
+  private val recognizerMedium = Recognizer(reactApplicationContext)
+  private val recognizerSmall = Recognizer(reactApplicationContext)
+  private val converter = CTCLabelConverter(symbols, mapOf(languageDictPath to "key"))
+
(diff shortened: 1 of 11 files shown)


NorbertKlockiewicz:
I am wondering if I should add comments for functions similarly to iOS version

android/src/main/java/com/swmansion/rnexecutorch/models/ocr/RecognitionHandler.kt
```diff
@@ -0,0 +1,115 @@
...
+    onComplete: (Int, Exception?) -> Unit
+  ) {
+    try {
+      recognizerLarge.loadModel(largeRecognizerPath)
+      recognizerMedium.loadModel(mediumRecognizerPath)
+      recognizerSmall.loadModel(smallRecognizerPath)
+      onComplete(0, null)
+    } catch (e: Exception) {
+      onComplete(1, e)
+    }
```
mkopcins, line 49: Can we tell from error which model failed?

android/src/main/java/com/swmansion/rnexecutorch/models/ocr/RecognitionHandler.kt
```diff
@@ -0,0 +1,115 @@
...
+      var confidenceScore = result.second
+
+      if (confidenceScore < Constants.LOW_CONFIDENCE_THRESHOLD) {
+        Core.rotate(croppedImage, croppedImage, Core.ROTATE_180)
+        val rotatedResult = runModel(croppedImage)
+        val rotatedConfidenceScore = rotatedResult.second
+        if (rotatedConfidenceScore > confidenceScore) {
+          result = rotatedResult
+          confidenceScore = rotatedConfidenceScore
+        }
```
mkopcins, line 93: Shouldn't we skip processing if confidence is still below `Constants.LOW_CONFIDENCE_THRESHOLD`?

android/src/main/java/com/swmansion/rnexecutorch/models/ocr/utils/CTCLabelConverter.kt
```diff
@@ -0,0 +1,75 @@
...
+      val subArray = textIndex.subList(index, index + segmentLength)
+
+      val text = StringBuilder()
+      var lastChar: Int? = null
+      val isNotRepeated = mutableListOf(true)
+      val isNotIgnored = mutableListOf<Boolean>()
+
+      subArray.forEachIndexed { i, currentChar ->
+        if (i > 0) {
+          isNotRepeated.add(lastChar != currentChar)
```
mkopcins, line 57: Why are we ignoring repeated characters? if we haev text like `strawberry` does this ignore double r's?

mkopcins (approved):

android/src/main/java/com/swmansion/rnexecutorch/models/ocr/RecognitionHandler.kt
NorbertKlockiewicz: Sometimes the confidence score is still low and the result is correct, we are returning a confidence score to user so I think we should leave handling those cases for him

android/src/main/java/com/swmansion/rnexecutorch/models/ocr/utils/CTCLabelConverter.kt
NorbertKlockiewicz: Simply the model is scanning the image column by column where column has width of I believe 1 pixel, so the same character will be scanned more than once, that's why we are ignoring repeated characters

android/src/main/java/com/swmansion/rnexecutorch/models/ocr/RecognitionHandler.kt
NorbertKlockiewicz: Without separate try catches for every model I don't think so

PR #101, opened 2025-02-20
feat: Whisper

chmjkb:
## Description
This PR provides an almost-ready implementation of Whisper. The only thing left to do should be to align the TS implementation with Moonshine, Hookless API and to create a hook for this.

### Type of change

### Tested on
- [x] iOS
- [x] Android

### Checklist


android/src/main/java/com/swmansion/rnexecutorch/SpeechToText.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.swmansion.rnexecutorch.utils.ArrayUtils
+import com.swmansion.rnexecutorch.utils.ETError
+
+class SpeechToText(reactContext: ReactApplicationContext) :
+  NativeSpeechToTextSpec(reactContext) {
+  private var whisperPreprocessor = WhisperPreprocessor(reactContext)
+  private var whisperEncoder = WhisperEncoder(reactContext)
+  private var whisperDecoder = WhisperDecoder(reactContext)
+  private var START_TOKEN = 50257
+  private var EOS_TOKEN = 50256
```
mkopcins, line 18: I think we cshould move those inside Whisper model, as they can vary from tokenizer to tokenizer. Or move it to another config file tokenizer based, but I like the first one better right now

mkopcins (approved):

PR #113, opened 2025-03-05
feat: Image segmentation for ios

JakubGonera:
## Description
Add semantic image segmentation with iOS native code. The result of a model run is a map containing masks of per-pixel probability for specified classes and an 'argmax' mask containing the indices of max value labels for each pixel.

### Type of change

### Tested on
- [x] iOS
- [ ] Android

### Checklist


ios/RnExecutorch/models/image_segmentation/ImageSegmentationModel.h
```diff
@@ -0,0 +1,22 @@
...
+NSArray* matToNSArray(const cv::Mat& mat) {
+    std::size_t numPixels = mat.rows * mat.cols;
+    NSMutableArray *arr = [[NSMutableArray alloc] initWithCapacity:numPixels];
+
+    for (std::size_t x = 0; x < mat.rows; ++x) {
+        for (std::size_t y = 0; y < mat.cols; ++y) {
+            arr[x * mat.cols + y] = @(mat.at<T>(x, y));
+        }
+    }
+    return arr;
```
chmjkb: I think this can be moved to utils of some kind

src/modules/computer_vision/ImageSegmentationModule.ts
```diff
@@ -0,0 +1,19 @@
+import { BaseModule } from '../BaseModule';
+import { _ImageSegmentationModule } from '../../native/RnExecutorchModules';
+import { getError } from '../../Error';
+
+export class ImageSegmentationModule extends BaseModule {
+  static module = new _ImageSegmentationModule();
+
+  static async forward(input: string, classesOfInteres?: string[]) {
+    console.log('# classes: ', classesOfInteres?.length);
```
chmjkb: Redundant log

ios/RnExecutorch/models/image_segmentation/ImageSegmentationModel.mm
```diff
@@ -0,0 +1,137 @@
...
+  return modelInput;
+}
+
+std::vector<cv::Mat> rescaleResults(NSArray *result, std::size_t numLabels, 
+                cv::Size modelImageSize, cv::Size originalSize) {
+  std::size_t numModelPixels = modelImageSize.height * modelImageSize.width;
+
+  std::vector<cv::Mat> resizedLabelScores(numLabels);
+  for (std::size_t label = 0; label < numLabels; ++label) {
+    cv::Mat labelMat = cv::Mat(modelImageSize, CV_64F);
```
msluszniak, line 50: Direct initialization instead of copy initialization
```suggestion
    cv::Mat labelMat(modelImageSize, CV_64F);
```

ios/RnExecutorch/models/image_segmentation/ImageSegmentationModel.mm
```diff
@@ -0,0 +1,137 @@
...
+
+void adjustScoresPerPixel(std::vector<cv::Mat>& labelScores, cv::Mat& maxArg,
+                cv::Size originalSize, std::size_t numLabels) {
+  std::size_t numOriginalPixels = originalSize.height * originalSize.width;
+  for (std::size_t pixel = 0; pixel < numOriginalPixels; ++pixel) {
+    int row = pixel / originalSize.width;
+    int col = pixel % originalSize.width;
+    std::vector<double> scores;
+    scores.reserve(numLabels);
+    for (const cv::Mat& mat : labelScores) {
```
msluszniak:
```suggestion
    for (const auto& mat : labelScores) {
```

ios/RnExecutorch/models/image_segmentation/ImageSegmentationModel.mm
```diff
@@ -0,0 +1,137 @@
...
+  std::size_t numLabels = deeplabv3_resnet50_labels.size();
+
+  NSAssert((std::size_t)output.count ==  numLabels * modelImageSize.height * modelImageSize.width, 
+        @"Model generated unexpected output size.");
+
+  // For each label extract it's matrix and rescale it to the original size
+  std::vector<cv::Mat> resizedLabelScores = 
+        rescaleResults(output, numLabels, modelImageSize, originalSize);
+
+  cv::Mat maxArg = cv::Mat(originalSize, CV_32S);
```
msluszniak: Same as above

ios/RnExecutorch/models/image_segmentation/ImageSegmentationModel.mm
JakubGonera: Both cases are triggering move construction https://en.cppreference.com/w/cpp/language/copy_elision. Happy to change if it's just syntax preference, but then uniform/brace {} initialization is the preferred way of initializing since () can cause most vexing parse https://en.wikipedia.org/wiki/Most_vexing_parse.

msluszniak: You're using Clang that supports e.g. `std::span` so the copy-ellision is guarateed and hence there is even no moving constructor. So this change definitely doesn't give performance boost. If I remember correctly list initialization with curly braces, which is ofc preffered, is not avaliable for `cv::Mat`. So there are only copy or direct initialization to consider. In this particular case, the most vaxing parse will not have a place as there is a call with explicit constructor parameters. Direct initialization is better e.g. because it can use explicit and user defined constructors, while copy initialization uses non-explicit copy constructor. Yeah, so  with 3rd party code like OpenCV there is no much change here. I always thought it is a preference but since c++17 and if both works correct (no explicit constructors in `cv::Mat`) then whatever 😜.

src/hooks/computer_vision/useImageSegmentation.ts
```diff
@@ -0,0 +1,65 @@
...
+export const useImageSegmentation = ({
+  modelSource,
+}: Props): {
+  error: string | null;
+  isReady: boolean;
+  isGenerating: boolean;
+  downloadProgress: number;
+  forward: (
+    input: string,
+    classesOfInterest?: string[]
```
chmjkb: I think this param could be redefined. Currently it is just `string[]`, which may be very easily misused by user. Instead, i would suggest defining such enum:
```
enum Classes {
  Cow,
  Dog,
  ...
}
```
And then this hook would accept:
`classesOfInterest?: (keyof typeof Classes)[]`

src/hooks/computer_vision/useImageSegmentation.ts
```diff
@@ -0,0 +1,65 @@
...
+  ) => Promise<{ [category: string]: number[] }>;
```
chmjkb: The same goes for the key in the return type 👍🏻

src/hooks/computer_vision/useImageSegmentation.ts
```diff
@@ -0,0 +1,65 @@
...
+        const fileUri = await fetchResource(modelSource, setDownloadProgress);
+        await module.loadModule(fileUri);
+        setIsReady(true);
+      } catch (e) {
+        setError(getError(e));
+      }
+    };
+
+    loadModel();
+  }, [modelSource, module]);
```
chmjkb: You can do useModule instead. Since useModule doesn't return a setter for isGenerating, you would have to create a state for isGenerating within this hook. However this would reduce the code duplication for loading the model.

src/hooks/computer_vision/useImageSegmentation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const forward = async (input: string, classesOfInterest?: string[]) => {
```
chmjkb: As above (classesOfInterest)

ios/RnExecutorch/models/image_segmentation/ImageSegmentationModel.mm
```diff
@@ -0,0 +1,137 @@
...
+  cv::Size modelImageSize = [self getModelImageSize];
+
+  std::size_t numLabels = deeplabv3_resnet50_labels.size();
+
+  NSAssert((std::size_t)output.count ==  numLabels * modelImageSize.height * modelImageSize.width, 
+        @"Model generated unexpected output size.");
+
+  // For each label extract it's matrix and rescale it to the original size
+  std::vector<cv::Mat> resizedLabelScores = 
+        rescaleResults(output, numLabels, modelImageSize, originalSize);
```
chmjkb: Maybe this should be used only when a flag is specified? For large images i expect that this introduces a lot of overhead for returning the results back to RN. Some people might not need that. WDYT?

src/constants/image_segmentation/image_segmentation.ts
chmjkb, line 1: This is not used anywhere

src/modules/computer_vision/ImageSegmentationModule.ts
```diff
@@ -0,0 +1,19 @@
+import { BaseModule } from '../BaseModule';
+import { _ImageSegmentationModule } from '../../native/RnExecutorchModules';
+import { getError } from '../../Error';
+
+export class ImageSegmentationModule extends BaseModule {
+  static module = new _ImageSegmentationModule();
+
+  static async forward(input: string, classesOfInteres?: string[]) {
```
chmjkb: Consider adjusting types as mentioned in the hook

src/constants/image_segmentation/image_segmentation.ts
chmjkb: However, this could be used if we wanted to make the aforementioned types in TS. Then we could get rid of the mapping on the native side.

chmjkb (changes requested):

src/hooks/computer_vision/useImageSegmentation.ts
JakubGonera: My concern is that this either:
* binds the implementation to a single model (or set of classes recognized by this model) if all the enum members would need to be a valid class
or
* would create the same opportunity for misuse if an enum would contain a union of all sets of classes of models allowed

Right now if the user specifies a wrong class they will just not see it in the result dict.

src/constants/image_segmentation/image_segmentation.ts
JakubGonera: I'm intending on using this in the example app; the argmax mask returns an index per pixel, so to decode any useful information the user needs a way to lookup the meaning of the indices. If classesOfInterest would be passed via strings the docs could also refer the user for a set of valid labels.

src/hooks/computer_vision/useImageSegmentation.ts
chmjkb: 1. True, but not sure if we're ever gonna touch this type of segmentation again. Even if we introduce some other models, we will most likely need to refactor this. I think the typehints here are pretty cool though. If you opt against it, we have to document it pretty well though. Maybe the middleground would be to run console.warn when the specified class is not an actual class that the model recognizes.
2. Kind of bound to 1.

src/hooks/computer_vision/useImageSegmentation.ts
```diff
@@ -16,8 +17,8 @@ export const useImageSegmentation = ({
   downloadProgress: number;
   forward: (
     input: string,
-    classesOfInterest?: string[]
-  ) => Promise<{ [category: string]: number[] }>;
+    classesOfInterest?: DeeplabLabel[]
```
chmjkb: Wondering if we should use this or keyof typeof DeeplabLabel. Using keyof typeof would make it consistent, since it is used in object detection but I think i like this one better. WDYT?

src/types/image_segmentation.ts
chmjkb, line 1: Can we make members of this enum uppercase to remain consistent?

chmjkb (commented):
Overall, LGTM 🚀

src/hooks/computer_vision/useImageSegmentation.ts
JakubGonera: I haven't notice the use in object detection, but I would still make the case for this. In object detection the enum is used as a return value, so we're just interested in the string content, whereas here the enum is supplied by the user (in both cases, either as the filter for classes in a list, or as the key to query results in the dict), so we care more about correctness. I think that autocompletion for the enum values when writing is worth more than consistency in this case.

chmjkb: Yeah that is what i thought :)

chmjkb (approved):

PR #141, opened 2025-03-19
fix: remove redundant text from speech-to-text example app

ArjunPatidarRadix:
… in Text component

There was an unnecessary Text zRXW within View in this file. That was causing the error like Text should be rendered within the Text Component.

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


chmjkb (approved):
Thanks for catching this!

PR #150, opened 2025-03-21
fix: handle local files

jakmro:
## Description

fix: handle local files

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/src/utils/fetchResource.ts
+++ b/src/utils/fetchResource.ts
@@ -38,6 +38,12 @@ export const fetchResource = async (
 ) => {
   const uri =
     typeof source === 'number' ? Asset.fromModule(source).uri : source;
+
+  // Handle local files
+  if (uri.startsWith('file://')) {
+    return uri;
+  }
+
   const filename = getFilenameFromUri(uri);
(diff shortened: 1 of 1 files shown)


src/utils/fetchResource.ts
```diff
@@ -38,6 +38,12 @@ export const fetchResource = async (
 ) => {
   const uri =
     typeof source === 'number' ? Asset.fromModule(source).uri : source;
+
+  // Handle local files
+  if (uri.startsWith('file://')) {
```
NorbertKlockiewicz, line 43: Remove file:// prefix returned path should look like this:
/private/var/Application/.../model.pte

NorbertKlockiewicz (changes requested):

src/utils/fetchResource.ts
jakmro: Why is that?

NorbertKlockiewicz: nvm, I thought we didn't parse it on native side

NorbertKlockiewicz (approved):

PR #162, opened 2025-03-27
feat: add option to pass skip special tokens flag to decode

NorbertKlockiewicz:
## Description

This PR adds skipSpecialTokens flag to argument list of decode method

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/third-party/ios/ExecutorchLib/ExecutorchLib/Exported/HuggingFaceTokenizer.mm
+++ b/third-party/ios/ExecutorchLib/ExecutorchLib/Exported/HuggingFaceTokenizer.mm
@@ -44,12 +44,18 @@ std::string loadBytesFromFile(const std::string &path) {
 }
 
 - (NSString *)decode:(NSArray<NSNumber *> *)tokenIds {
+  return [self decode:tokenIds skipSpecialTokens:NO];
+}
+
+- (NSString *)decode:(NSArray<NSNumber *> *)tokenIds
+    skipSpecialTokens:(BOOL)skipSpecialTokens {
   std::vector<int32_t> stdTokenIds;
   stdTokenIds.reserve([tokenIds count]);
   for (NSNumber *tokenId in tokenIds) {
     stdTokenIds.push_back([tokenId intValue]);
   }
-  std::string decodedString = _tokenizer->Decode(stdTokenIds);
+  std::string decodedString =
+      _tokenizer->Decode(stdTokenIds, skipSpecialTokens);
   return [NSString stringWithUTF8String:decodedString.c_str()];
 }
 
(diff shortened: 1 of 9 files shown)


third-party/ios/ExecutorchLib/ExecutorchLib/Exported/HuggingFaceTokenizer.mm
```diff
@@ -44,12 +44,18 @@ - (instancetype)initWithTokenizerPath:(NSString *)tokenizerPath {
 }
 
 - (NSString *)decode:(NSArray<NSNumber *> *)tokenIds {
+  return [self decode:tokenIds skipSpecialTokens:NO];
```
chmjkb, line 47: Is this method needed now? I think we can get rid of this

chmjkb (changes requested):
Also, I think there should be a method within the tokenizers that returns BOS/EOS tokens. We might need this, can you please export them as well? 🙏🏻

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/HuggingFaceTokenizer.mm
NorbertKlockiewicz: In some cases I think we want to decode text with special tokens and I wanted this method to behave as method with default parameter. I think it doesn't introduce unnecessary complexity and can offer a bit simpler code in some places. However, right now it's not used anywhere so it won't be a problem to remove it. WDYT?

NorbertKlockiewicz:
> Also, I think there should be a method within the tokenizers that returns BOS/EOS tokens. We might need this, can you please export them as well? 🙏🏻

Yeah, I am aware of it, however those methods require loading the tokenizer-config file, which is not supported by rust -> c bridge I am using, so it requires more complex solution. I am planning to add this but not sure if I will write it myself or modify bridge(if it's possible). Depends on how fast you want those special tokens to be merged I can add it in this PR in few days or merge this PR and add it in another one.

third-party/ios/ExecutorchLib/ExecutorchLib/Exported/HuggingFaceTokenizer.mm
chmjkb: This is not exported to TS, hence I thought that you left it here by accident. If you're thinking about using it this way in the native side, then you're right 👍🏻

chmjkb:
> > Also, I think there should be a method within the tokenizers that returns BOS/EOS tokens. We might need this, can you please export them as well? 🙏🏻
> 
> Yeah, I am aware of it, however those methods require loading the tokenizer-config file, which is not supported by rust -> c bridge I am using, so it requires more complex solution. I am planning to add this but not sure if I will write it myself or modify bridge(if it's possible). Depends on how fast you want those special tokens to be merged I can add it in this PR in few days or merge this PR and add it in another one.

In this case I'll just approve this, and just submit another PR when you're done with the config files.

chmjkb (approved):

PR #163, opened 2025-03-27
feat: text embeddings

jakmro:
## Description
Add text embeddings

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


android/src/main/java/com/swmansion/rnexecutorch/models/TextEmbeddings/TextEmbeddingsModel.kt
```diff
@@ -0,0 +1,49 @@
...
+  private lateinit var tokenizer: HuggingFaceTokenizer
+
+  fun loadTokenizer(tokenizerSource: String) {
+    tokenizer = HuggingFaceTokenizer(URL(tokenizerSource).path)
+  }
+
+  fun preprocess(input: String): Array<LongArray> {
+    val inputIds = tokenizer.encode(input).map { it.toLong() }.toLongArray()
+    val attentionMask = inputIds.map { if (it != 0L) 1L else 0L }.toLongArray()
+    return arrayOf(inputIds, attentionMask) // Shape: [2, max_length]
```
chmjkb, line 21: Where is max_length specified? I think mentioning it here would be nice

src/hooks/natural_language_processing/useTextEmbeddings.ts
```diff
@@ -0,0 +1,65 @@
...
+import { TextEmbeddings } from '../../native/RnExecutorchModules';
+import { fetchResource } from '../../utils/fetchResource';
+import { ETError, getError } from '../../Error';
+
+export const useTextEmbeddings = ({
+  modelSource,
+  tokenizerSource,
+}: {
+  modelSource: string | number;
+  tokenizerSource: string | number;
```
chmjkb: use `ResourceSource` instead of `string | number`

src/hooks/natural_language_processing/useTextEmbeddings.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    try {
+      setIsGenerating(true);
+      return await TextEmbeddings.forward(input);
+    } catch (e) {
+      throw new Error(getError(e));
+    } finally {
+      setIsGenerating(false);
+    }
+  };
```
chmjkb: Why not use `useModule()` instead?

src/modules/natural_language_processing/TextEmbeddingsModule.ts
```diff
@@ -0,0 +1,36 @@
...
+  static async forward(input: string) {
+    try {
+      return await TextEmbeddings.forward(input);
+    } catch (err) {
+      throw new Error((err as Error).message);
+    }
+  }
+
+  static onDownloadProgress(callback: (downloadProgress: number) => void) {
+    this.onDownloadProgressCallback = callback;
```
chmjkb: Same here, maybe it is possible to use the convention similar to the one used in ObjectDetectionModule?

src/hooks/natural_language_processing/useTextEmbeddings.ts
jakmro: This hook accepts modelSource and tokenizerSource

android/src/main/java/com/swmansion/rnexecutorch/models/TextEmbeddings/TextEmbeddingsModel.kt
jakmro: `max_length` is specified inside tokenizer.json

ios/RnExecutorch/models/BaseModel.mm
```diff
@@ -16,6 +16,22 @@ - (NSArray *)forward:(NSArray *)input {
...
+  NSMutableArray *inputTypes = [NSMutableArray new];
+  NSNumber *numberOfInputs = [module getNumberOfInputs];
+
+  for (NSUInteger i = 0; i < [numberOfInputs intValue]; i++) {
+    [shapes addObject:[module getInputShape:[NSNumber numberWithInt:i]]];
+    [inputTypes addObject:[module getInputType:[NSNumber numberWithInt:i]]];
+  }
+
+  NSArray *result = [module forward:inputs shapes:shapes inputTypes:inputTypes];
+
```
mkopcins: This does not look correct, wouldn't it be better to actually check if length on `inputs` is the same as `numberOfInputs` and if not abort? Then we don't need to create another method just to copy 99% of the code. You can either add if sattement here checking if input is array of arrays or change it to only work with array of inputs

mkopcins: and then change the name

ios/RnExecutorch/models/text_embeddings/TextEmbeddingsModel.mm
```diff
@@ -0,0 +1,34 @@
+#import "TextEmbeddingsModel.h"
+#import "TextEmbeddingsUtils.h"
+
+@implementation TextEmbeddingsModel
+
+- (NSArray *)preprocess:(NSString *)input {
+  NSArray *input_ids = [self->tokenizer encode:input];
+  NSMutableArray *attention_mask = [NSMutableArray new];
+  for (int i = 0; i < [input_ids count]; i++) {
+    [attention_mask addObject:@([input_ids[i] intValue] != 0 ? 1 : 0)];
```
mkopcins:
```suggestion
    [attention_mask addObject:@((int)([input_ids[i] intValue] != 0))];
```

examples/text-embeddings/App.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const [inputSentence, setInputSentence] = useState('');
+  const [sentencesWithEmbeddings, setSentencesWithEmbeddings] = useState<
+    { sentence: string; embedding: number[] }[]
+  >([]);
+  const [topMatches, setTopMatches] = useState<
+    { sentence: string; similarity: number }[]
+  >([]);
+
+  const dotProduct = (a: number[], b: number[]) =>
+    a.reduce((sum, val, i) => sum + val * b[i], 0);
```
mkopcins, line 35: This looks a bit more readable imo
```suggestion
  const dotProduct = (a: number[], b: number[]) => a.reduce((sum, _, i) => sum + a[i] * b[i], 0);
```

src/hooks/natural_language_processing/useTextEmbeddings.ts
```diff
@@ -0,0 +1,66 @@
+import { useEffect, useState } from 'react';
```
mkopcins: Change this to use `useModule` inside like in `src/hooks/computer_vision/useClassification.ts`. Regarding tokenizer source just add it as an optional prop. This code is almost the same as in `useModule`, `forward` method is exactly the same as `useModule::forwardImage` (maybe change this name :D)

src/native/RnExecutorchModules.ts
```diff
@@ -143,6 +133,18 @@ const Tokenizer = TokenizerSpec
...
+  ? TextEmbeddingsSpec
+  : new Proxy(
+      {},
+      {
+        get() {
+          throw new Error(LINKING_ERROR);
+        },
+      }
+    );
+
```
mkopcins: add class for this like for other modules, e.g.
```
class _ClassificationModule {
  async forward(input: string): ReturnType<ClassificationInterface['forward']> {
    return await Classification.forward(input);
  }
  async loadModule(
    modelSource: string | number
  ): ReturnType<ClassificationInterface['loadModule']> {
    return await Classification.loadModule(modelSource);
  }
}
```

src/modules/natural_language_processing/TextEmbeddingsModule.ts
```diff
@@ -0,0 +1,36 @@
+import { TextEmbeddings } from '../../native/RnExecutorchModules';
```
mkopcins: Change it to use the same convention as in other modules (like cv modules)

mkopcins (changes requested):

examples/text-embeddings/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "dependencies": {
+    "expo": "~52.0.42",
+    "expo-status-bar": "~2.0.1",
+    "react": "18.3.1",
+    "react-native": "0.76.8"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.25.2",
+    "@types/react": "~18.3.12",
+    "react-native-executorch": "file:../../react-native-executorch-20250401133031.tgz",
```
mkopcins, line 20: cleanup

jakmro: This will be cleaned up once v0.4.0 will be released on npm

ios/RnExecutorch/models/BaseModel.mm
jakmro: This will be changed in another PR as mentioned in `BaseModel.h`. This change requires changes in multiple files and I decided to not include them here.

examples/text-embeddings/App.tsx
jakmro: I would leave it or use a for loop. What do you think?

src/hooks/natural_language_processing/useTextEmbeddings.ts
jakmro: I can do it, but I think we need to change a few things in our codebase to make it more maintainable.

examples/text-embeddings/App.tsx
mkopcins: No point discussing this, both are good so it was just a suggestion :)

src/hooks/natural_language_processing/useTextEmbeddings.ts
mkopcins: Let's chat about it in the office

docs/docs/hookless-api/TextEmbeddingsModule.md
```diff
@@ -0,0 +1,47 @@
+---
+title: TextEmbeddingsModule
+sidebar_position: 8
+---
+
```
mkopcins: Also add the description and the caution banner like in `useTextEmbeddings` page

mkopcins (changes requested):

mkopcins (approved):

PR #166, opened 2025-03-31
feat: Add multilingual Whisper 

chmjkb:
### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


src/controllers/SpeechToTextController.ts
```diff
@@ -185,65 +163,116 @@ export class SpeechToTextController {
...
+      langTokenId,
+      transcribeTokenId,
+      noTimestampsTokenId,
+    ];
+    return startingTokenIds;
+  }
 
-    if (!waveform) {
-      this.isGeneratingCallback(false);
+  // TODO: think about narrowing the type of speakerLanguage
```
mkopcins: Add chceck for supported langs, lang token should be in range `(<|startoftranscript|>, <|translate|>)` and add warning that quality for some langs can be bad

src/controllers/SpeechToTextController.ts
```diff
@@ -185,65 +163,116 @@ export class SpeechToTextController {
...
-    let prevseq: number[] = [];
+    let prevSeq: number[] = [];
     for (let chunkId = 0; chunkId < this.chunks.length; chunkId++) {
-      let lastToken = this.config.tokenizer.sos;
       let prevSeqTokenIdx = 0;
       let finalSeq: number[] = [];
-      let seq = [lastToken];
+      let seq;
+
+      seq = await this.getStartingTokenIds(speakerLanguage);
```
mkopcins:
```suggestion
      let seq = await this.getStartingTokenIds(speakerLanguage);
```

src/controllers/SpeechToTextController.ts
```diff
@@ -185,65 +163,116 @@ export class SpeechToTextController {
...
+        lastToken = seq[seq.length - 1];
         try {
-          lastToken = await this.nativeModule.decode(seq);
+          // Returns a single predicted token
+          lastToken = await this.nativeModule.decode(seq, encoderOutput);
         } catch (error) {
-          this.onErrorCallback?.(`Decode ${error}`);
+          this.onErrorCallback?.(
+            `An error has ocurred while decoding: ${error}`
+          );
```
mkopcins, line 274: Unify this wwith encoding error, so either revert this or change encoding error

src/controllers/SpeechToTextController.ts
```diff
@@ -185,65 +163,116 @@ export class SpeechToTextController {
...
+    let prevSeq: number[] = [];
     for (let chunkId = 0; chunkId < this.chunks.length; chunkId++) {
-      let lastToken = this.config.tokenizer.sos;
       let prevSeqTokenIdx = 0;
       let finalSeq: number[] = [];
-      let seq = [lastToken];
+      let seq;
+
+      seq = await this.getStartingTokenIds(speakerLanguage);
+      let numSpecialTokens = seq.length;
```
mkopcins:
```suggestion
      const numSpecialTokens = seq.length;
```

src/controllers/SpeechToTextController.ts
```diff
@@ -185,65 +163,116 @@ export class SpeechToTextController {
...
+          this.chunks!.at(chunkId)!
+        );
       } catch (error) {
         this.onErrorCallback?.(`Encode ${error}`);
         return '';
       }
+
+      let lastToken;
       while (lastToken !== this.config.tokenizer.eos) {
+        lastToken = seq[seq.length - 1];
```
mkopcins:
```suggestion
      let lastToken = seq.at(-1);
      while (lastToken !== this.config.tokenizer.eos) {
```

src/controllers/SpeechToTextController.ts
```diff
@@ -253,63 +282,53 @@ export class SpeechToTextController {
...
+        seqs = [seq.slice(0, -(numSpecialTokens + 3))];
+      } else if (
+        seqs.length ===
+        Math.ceil(waveform.length / this.windowSize) - 1
+      ) {
+        seqs.push(seq.slice(numSpecialTokens + 3));
       } else {
-        seqs = [...seqs, seq.slice(4, -4)];
+        seqs.push(seq.slice(numSpecialTokens + 3, -(numSpecialTokens + 3)));
       }
```
mkopcins: I know this is from old code, but move the `3` to a local const variable here

mkopcins: Or maybe const in sttDefaults

src/controllers/SpeechToTextController.ts
```diff
@@ -253,63 +282,53 @@ export class SpeechToTextController {
...
-      .join('')
-      .replaceAll(this.config.tokenizer.specialChar, ' ');
-  }
-
-  public async encode(waveform: number[]) {
-    return await this.nativeModule.encode(waveform);
-  }
-
-  public async decode(seq: number[], encodings?: number[]) {
-    return await this.nativeModule.decode(seq, encodings);
```
mkopcins, line 313: restore this

src/modules/natural_language_processing/SpeechToTextModule.ts
```diff
@@ -51,16 +52,9 @@ export class SpeechToText {
...
   ): ReturnType<SpeechToTextController['transcribe']> {
-    return await this.module.transcribe(waveform);
-  }
-
-  static async encode(waveform: number[]) {
-    return await this.module.encode(waveform);
-  }
-
-  static async decode(seq: number[], encodings?: number[]) {
-    return await this.module.decode(seq, encodings);
```
mkopcins: restore this

mkopcins (changes requested):

src/controllers/SpeechToTextController.ts
chmjkb: Tbf I'm not sure if it is a right way to go. It still doesn't fix the fact that the user doesn't know which languages are supported. Instead, I introduced an enum that a user has to pass to this function. This way, there is no way of passing a wrong language, without breaking TS rules. I'll add a check on top of that 👍🏻

src/controllers/SpeechToTextController.ts
```diff
@@ -185,65 +165,121 @@ export class SpeechToTextController {
...
+      const numSpecialTokens = seq.length;
+      let encoderOutput;
       try {
-        await this.nativeModule.encode(this.chunks!.at(chunkId)!);
+        encoderOutput = await this.nativeModule.encode(
+          this.chunks!.at(chunkId)!
+        );
       } catch (error) {
-        this.onErrorCallback?.(`Encode ${error}`);
+        this.onErrorCallback?.(`An error has ocurred while decoding ${error}`);
```
mkopcins:
```suggestion
        this.onErrorCallback?.(`An error has ocurred while encoding ${error}`);
```

PR #168, opened 2025-04-01
@md/s2t streaming

mdydek:
### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist

### DIFF
--- a/src/controllers/SpeechToTextController.ts
+++ b/src/controllers/SpeechToTextController.ts
@@ -8,6 +8,7 @@ import {
   MODEL_CONFIGS,
   ModelConfig,
   MODES,
+  STREAMING_ACTION,
 } from '../constants/sttDefaults';
 
 const longCommonInfPref = (seq1: number[], seq2: number[]) => {
@@ -46,6 +47,11 @@ export class SpeechToTextController {
   public isReady = false;
   public isGenerating = false;
   private modelName!: 'moonshine' | 'whisper';
+  private seqs: number[][] = [];
+  private prevSeq: number[] = [];
+  private streamWaveform: number[] = [];
+  private isDecodingChunk = false;
+  private numberOfDecodedChunks = 0;
 
   // tokenizer tokens to string mapping used for decoding sequence
   private tokenMapping!: { [key: number]: string };
@@ -190,6 +196,44 @@ export class SpeechToTextController {
     }
   }
 
+  private async decodeChunk(chunk: number[]): Promise<number[]> {
+    let lastToken = this.config.tokenizer.sos;
+    let seq = [lastToken];
+    let prevSeqTokenIdx = 0;
+    try {
+      await this.nativeModule.encode(chunk);
+    } catch (error) {
+      this.onErrorCallback?.(`Encode ${error}`);
+      return [];
+    }
+    while (lastToken !== this.config.tokenizer.eos) {
+      try {
+        lastToken = await this.nativeModule.decode(seq);
+      } catch (error) {
+        this.onErrorCallback?.(`Decode ${error}`);
+        return [];
+      }
+      seq = [...seq, lastToken];
+      if (
+        this.seqs.length > 0 &&
+        seq.length < this.seqs.at(-1)!.length &&
+        seq.length % 3 !== 0
+      ) {
+        this.prevSeq = [...this.prevSeq, this.seqs.at(-1)![prevSeqTokenIdx++]!];
+        this.decodedTranscribeCallback(this.prevSeq);
+      }
+    }
+    return seq;
+  }
+
+  private handleOverlaps(seqs: number[][]): number[] {
+    const maxInd = longCommonInfPref(seqs.at(-2)!, seqs.at(-1)!);
+    let finalSeq = [...this.sequence, ...seqs.at(-2)!.slice(0, maxInd)];
+    this.sequence = finalSeq;
+    this.decodedTranscribeCallback(finalSeq);
+    return finalSeq;
+  }
+
   public async transcribe(waveform: number[]): Promise<string> {
     if (!this.isReady) {
       this.onErrorCallback?.(new Error('Model is not yet ready'));
@@ -200,53 +244,17 @@ export class SpeechToTextController {
       return '';
     }
     this.onErrorCallback?.(undefined);
+    this.decodedTranscribeCallback([]);
     this.isGeneratingCallback(true);
 
     this.sequence = [];
-
-    if (!waveform) {
-      this.isGeneratingCallback(false);
-
-      this.onErrorCallback?.(
-        new Error(
-          `Nothing to transcribe, perhaps you forgot to call this.loadAudio().`
-        )
-      );
-    }
-
+    this.seqs = [];
+    this.prevSeq = [];
     this.chunkWaveform(waveform);
 
-    let seqs: number[][] = [];
-    let prevseq: number[] = [];
     for (let chunkId = 0; chunkId < this.chunks.length; chunkId++) {
-      let lastToken = this.config.tokenizer.sos;
-      let prevSeqTokenIdx = 0;
+      let seq = await this.decodeChunk(this.chunks!.at(chunkId)!);
       let finalSeq: number[] = [];
-      let seq = [lastToken];
-      try {
-        await this.nativeModule.encode(this.chunks!.at(chunkId)!);
-      } catch (error) {
-        this.onErrorCallback?.(`Encode ${error}`);
-        return '';
-      }
-      while (lastToken !== this.config.tokenizer.eos) {
-        try {
-          lastToken = await this.nativeModule.decode(seq);
-        } catch (error) {
-          this.onErrorCallback?.(`Decode ${error}`);
-          return '';
-        }
-        seq = [...seq, lastToken];
-        if (
-          seqs.length > 0 &&
-          seq.length < seqs.at(-1)!.length &&
-          seq.length % 3 !== 0
-        ) {
-          prevseq = [...prevseq, seqs.at(-1)![prevSeqTokenIdx++]!];
-          this.decodedTranscribeCallback(prevseq);
-        }
-      }
-
       if (this.chunks.length === 1) {
         finalSeq = seq;
         this.sequence = finalSeq;
@@ -254,33 +262,105 @@ export class SpeechToTextController {
         break;
       }
       // remove sos/eos token and 3 additional ones
-      if (seqs.length === 0) {
-        seqs = [seq.slice(0, -4)];
-      } else if (seqs.length === this.chunks.length - 1) {
-        seqs = [...seqs, seq.slice(4)];
+      if (this.seqs.length === 0) {
+        this.seqs = [seq.slice(0, -4)];
+      } else if (this.seqs.length === this.chunks.length - 1) {
+        this.seqs = [...this.seqs, seq.slice(4)];
       } else {
-        seqs = [...seqs, seq.slice(4, -4)];
+        this.seqs = [...this.seqs, seq.slice(4, -4)];
       }
-      if (seqs.length < 2) {
+      if (this.seqs.length < 2) {
         continue;
       }
 
-      const maxInd = longCommonInfPref(seqs.at(-2)!, seqs.at(-1)!);
-      finalSeq = [...this.sequence, ...seqs.at(-2)!.slice(0, maxInd)];
-      this.sequence = finalSeq;
-      this.decodedTranscribeCallback(finalSeq);
-      prevseq = finalSeq;
+      this.prevSeq = this.handleOverlaps(this.seqs);
 
       //last sequence processed
-      if (seqs.length === this.chunks.length) {
-        finalSeq = [...this.sequence, ...seqs.at(-1)!];
+      if (this.seqs.length === this.chunks.length) {
+        finalSeq = [...this.sequence, ...this.seqs.at(-1)!];
         this.sequence = finalSeq;
         this.decodedTranscribeCallback(finalSeq);
-        prevseq = finalSeq;
+        this.prevSeq = finalSeq;
       }
     }
     const decodedSeq = this.decodeSeq(this.sequence);
     this.isGeneratingCallback(false);
+    return decodedSeq;
+  }
+
+  public async streamingTranscribe(
+    waveform: number[],
+    streamAction: STREAMING_ACTION
+  ): Promise<string> {
+    if (!this.isReady) {
+      this.onErrorCallback?.(new Error('Model is not yet ready'));
+      return '';
+    }
+    this.onErrorCallback?.(undefined);
+
+    if (streamAction == STREAMING_ACTION.START) {
+      this.sequence = [];
+      this.seqs = [];
+      this.streamWaveform = [];
+      this.prevSeq = [];
+      this.numberOfDecodedChunks = 0;
+      this.decodedTranscribeCallback([]);
+      this.isGeneratingCallback(true);
+    }
+    this.streamWaveform = [...this.streamWaveform, ...waveform];
+    this.chunkWaveform(this.streamWaveform);
+    if (!this.isDecodingChunk && streamAction != 2) {
+      this.isDecodingChunk = true;
+      while (
+        this.chunks.at(this.numberOfDecodedChunks)?.length ==
+          2 * this.overlapSeconds + this.windowSize ||
+        (this.numberOfDecodedChunks == 0 &&
+          this.chunks.at(this.numberOfDecodedChunks)?.length ==
+            this.windowSize + this.overlapSeconds)
+      ) {
+        let seq = await this.decodeChunk(
+          this.chunks.at(this.numberOfDecodedChunks)!
+        );
+        // remove sos/eos token and 3 additional ones
+        if (this.numberOfDecodedChunks == 0) {
+          this.seqs = [seq.slice(0, -4)];
+        } else {
+          this.seqs = [...this.seqs, seq.slice(4, -4)];
+          this.prevSeq = this.handleOverlaps(this.seqs);
+        }
+        this.numberOfDecodedChunks++;
+        if (this.seqs.length < 2) {
+          continue;
+        }
+      }
+      this.isDecodingChunk = false;
+    }
+    while (
+      this.numberOfDecodedChunks < this.chunks.length &&
+      streamAction == STREAMING_ACTION.STOP
+    ) {
+      let seq = await this.decodeChunk(
+        this.chunks.at(this.numberOfDecodedChunks)!
+      );
+      if (this.numberOfDecodedChunks == 0) {
+        this.sequence = seq;
+        this.decodedTranscribeCallback(seq);
+        this.isGeneratingCallback(false);
+        break;
+      }
+      //last sequence processed
+      if (this.numberOfDecodedChunks == this.chunks.length - 1) {
+        let finalSeq = [...this.sequence, ...seq];
+        this.sequence = finalSeq;
+        this.decodedTranscribeCallback(finalSeq);
+        this.isGeneratingCallback(false);
+      } else {
+        this.seqs = [...this.seqs, seq.slice(4, -4)];
+        this.handleOverlaps(this.seqs);
+      }
+      this.numberOfDecodedChunks++;
+    }
+    const decodedSeq = this.decodeSeq(this.sequence);
     return decodedSeq;
   }
 
--- a/src/hooks/natural_language_processing/useSpeechToText.ts
+++ b/src/hooks/natural_language_processing/useSpeechToText.ts
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
 import { SpeechToTextController } from '../../controllers/SpeechToTextController';
 import { ResourceSource } from '../../types/common';
+import { STREAMING_ACTION } from '../../constants/sttDefaults';
 
 interface SpeechToTextModule {
   isReady: boolean;
@@ -12,6 +13,10 @@ interface SpeechToTextModule {
   transcribe: (
     input: number[]
   ) => ReturnType<SpeechToTextController['transcribe']>;
+  streamingTranscribe: (
+    input: number[],
+    streamAction: STREAMING_ACTION
+  ) => ReturnType<SpeechToTextController['streamingTranscribe']>;
 }
 
 export const useSpeechToText = ({
@@ -77,5 +82,7 @@ export const useSpeechToText = ({
     sequence,
     error,
     transcribe: (waveform: number[]) => model.transcribe(waveform),
+    streamingTranscribe: (waveform: number[], streamAction: STREAMING_ACTION) =>
+      model.streamingTranscribe(waveform, streamAction),
   };
 };
--- a/examples/speech-to-text/screens/SpeechToTextScreen.tsx
+++ b/examples/speech-to-text/screens/SpeechToTextScreen.tsx
@@ -1,4 +1,4 @@
-import { useSpeechToText } from 'react-native-executorch';
+import { useSpeechToText, STREAMING_ACTION } from 'react-native-executorch';
 import {
   Text,
   View,
@@ -51,7 +51,13 @@ export const SpeechToTextScreen = () => {
     sequence,
     error,
     transcribe,
-  } = useSpeechToText({ modelName: 'moonshine', streamingConfig: 'balanced' });
+    streamingTranscribe,
+  } = useSpeechToText({
+    modelName: 'moonshine',
+    streamingConfig: 'balanced',
+    windowSize: 5,
+    overlapSeconds: 1.2,
+  });
 
   const loadAudio = async (url: string) => {
     const audioContext = new AudioContext({ sampleRate: 16e3 });
@@ -72,16 +78,18 @@ export const SpeechToTextScreen = () => {
   const onChunk = (data: string) => {
     const float32Chunk = float32ArrayFromPCMBinaryBuffer(data);
     audioBuffer.current?.push(...float32Chunk);
+    streamingTranscribe(audioBuffer.current, STREAMING_ACTION.DATA);
   };
 
   const handleRecordPress = async () => {
     if (isRecording) {
       LiveAudioStream.stop();
       setIsRecording(false);
-      await transcribe(audioBuffer.current);
+      await streamingTranscribe(audioBuffer.current, STREAMING_ACTION.STOP);
       audioBuffer.current = [];
     } else {
       setIsRecording(true);
+      streamingTranscribe(audioBuffer.current, STREAMING_ACTION.START);
       startStreamingAudio(audioStreamOptions, onChunk);
     }
   };
@@ -133,6 +141,7 @@ export const SpeechToTextScreen = () => {
             setModalVisible(visible);
             if (audioUrl) {
               await transcribe(await loadAudio(audioUrl));
+              setAudioUrl('');
             }
           }}
           onChangeText={setAudioUrl}
@@ -156,6 +165,7 @@ export const SpeechToTextScreen = () => {
                   setModalVisible(true);
                 } else {
                   await transcribe(await loadAudio(audioUrl));
+                  setAudioUrl('');
                 }
               }}
             >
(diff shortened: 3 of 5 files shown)


src/hooks/natural_language_processing/useSpeechToText.ts
```diff
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
```
mkopcins: You forgot about `react-native-executorch/src/modules/natural_language_processing/SpeechToTextModule.ts`

src/controllers/SpeechToTextController.ts
```diff
@@ -211,6 +230,52 @@ export class SpeechToTextController {
...
+    try {
+      encoderOutput = await this.nativeModule.encode(chunk);
+    } catch (error) {
+      this.onErrorCallback?.(`An error has ocurred while encoding ${error}`);
+      return [];
+    }
+    let lastToken = seq.at(-1) as number;
+    while (lastToken !== this.config.tokenizer.eos) {
+      try {
+        lastToken = await this.nativeModule.decode(seq, encoderOutput);
```
mkopcins: Don't call this like this, when you pass `encoderOutput` as a parameter it is being passed via json over the react native bridge, which in this case is a huge overhead. Instead you can simply skip this parameter and the most recent result of `nativeModul.decode` function call is cached and used.
```suggestion
        lastToken = await this.nativeModule.decode(seq);
```

src/controllers/SpeechToTextController.ts
```diff
@@ -211,6 +230,52 @@ export class SpeechToTextController {
...
+    const maxInd = longCommonInfPref(
+      seqs.at(-2)!,
+      seqs.at(-1)!,
+      HAMMING_DIST_THRESHOLD
+    );
+    let finalSeq = [...this.sequence, ...seqs.at(-2)!.slice(0, maxInd)];
+    this.sequence = finalSeq;
+    this.decodedTranscribeCallback(finalSeq);
+    return finalSeq;
+  }
```
mkopcins: lets remove `finalSeq` local var
```suggestion
    this.sequence = [...this.sequence, ...seqs.at(-2)!.slice(0, maxInd)];
    this.decodedTranscribeCallback(this.sequence);
    return this.sequence;
  }
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+    this.seqs = [];
+    this.prevSeq = [];
     this.chunkWaveform(waveform);
 
-    this.sequence = [];
-    let seqs: number[][] = [];
-    let prevSeq: number[] = [];
     for (let chunkId = 0; chunkId < this.chunks.length; chunkId++) {
-      let prevSeqTokenIdx = 0;
+      let seq = await this.decodeChunk(this.chunks!.at(chunkId)!);
```
mkopcins:
```suggestion
      const seq = await this.decodeChunk(this.chunks!.at(chunkId)!);
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
-        seqs.at(-2)!,
-        seqs.at(-1)!,
-        HAMMING_DIST_THRESHOLD
-      );
-      finalSeq = [...this.sequence, ...seqs.at(-2)!.slice(0, maxInd)];
-      this.sequence = finalSeq;
-      this.decodedTranscribeCallback(finalSeq);
-      prevSeq = finalSeq;
+      finalSeq = this.handleOverlaps(this.seqs);
+      this.prevSeq = finalSeq;
```
mkopcins:
```suggestion
      this.prevSeq = this.handleOverlaps(this.seqs);
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
-      // last sequence processed
-      if (seqs.length === this.chunks.length) {
-        finalSeq = [...this.sequence, ...seqs.at(-1)!];
+      //last sequence processed
+      if (this.seqs.length === this.chunks.length) {
+        finalSeq = [...this.sequence, ...this.seqs.at(-1)!];
         this.sequence = finalSeq;
         this.decodedTranscribeCallback(finalSeq);
-        prevSeq = finalSeq;
+        this.prevSeq = finalSeq;
```
mkopcins: remove `finalSeq` here too

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+
+    if (streamAction == STREAMING_ACTION.START) {
+      this.sequence = [];
+      this.seqs = [];
+      this.streamWaveform = [];
+      this.prevSeq = [];
+      this.indexOfCurrentlyDecodingChunk = 0;
+      this.isChunkDeleted = false;
+      this.decodedTranscribeCallback([]);
+      this.isGeneratingCallback(true);
```
mkopcins: lets move this to a sth like a `clearState`,`initState` or simillar and move those lines from `transcribe too

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+    this.streamWaveform = [...this.streamWaveform, ...waveform];
+    this.chunkWaveform(this.streamWaveform, this.isChunkDeleted);
+    if (!this.isDecodingChunk && streamAction != STREAMING_ACTION.STOP) {
+      this.isDecodingChunk = true;
+      while (
+        this.chunks.at(this.indexOfCurrentlyDecodingChunk)?.length ==
+          2 * this.overlapSeconds + this.windowSize ||
+        (this.indexOfCurrentlyDecodingChunk == 0 &&
+          this.chunks.at(0)?.length == this.windowSize + this.overlapSeconds)
+      ) {
```
mkopcins: Lets change it to a helper function like `expectedChunkLength()` and then do 
```suggestion
      while (
        this.chunks.at(this.indexOfCurrentlyDecodingChunk)?.length === expectedChunkLength()
      ) {
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+        } else {
+          this.seqs = [
+            ...this.seqs,
+            seq.slice(
+              numSpecialTokens + NUM_TOKENS_TO_SLICE,
+              -(numSpecialTokens + NUM_TOKENS_TO_SLICE)
+            ),
+          ];
+          this.prevSeq = this.handleOverlaps(this.seqs);
+        }
```
mkopcins: lets move it to another function (i.e. `trimSequenceTokens` or sth like that) too

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+    streamAction: STREAMING_ACTION,
+    audioLanguage?: SpeechToTextLanguage
+  ): Promise<string> {
+    if (!this.isReady) {
+      this.onErrorCallback?.(new Error('Model is not yet ready'));
+      return '';
+    }
+    this.onErrorCallback?.(undefined);
+
+    if (streamAction == STREAMING_ACTION.START) {
```
mkopcins:
```suggestion
    if (streamAction === STREAMING_ACTION.START) {
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+      this.streamWaveform = [];
+      this.prevSeq = [];
+      this.indexOfCurrentlyDecodingChunk = 0;
+      this.isChunkDeleted = false;
+      this.decodedTranscribeCallback([]);
+      this.isGeneratingCallback(true);
+    }
+    this.streamWaveform = [...this.streamWaveform, ...waveform];
+    this.chunkWaveform(this.streamWaveform, this.isChunkDeleted);
+    if (!this.isDecodingChunk && streamAction != STREAMING_ACTION.STOP) {
```
mkopcins:
```suggestion
    if (!this.isDecodingChunk && streamAction !== STREAMING_ACTION.STOP) {
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +305,149 @@ export class SpeechToTextController {
...
+          this.chunks.at(0)?.length == this.windowSize + this.overlapSeconds)
+      ) {
+        let seq = await this.decodeChunk(
+          this.chunks.at(this.indexOfCurrentlyDecodingChunk)!,
+          audioLanguage
+        );
+        const numSpecialTokens = (await this.getStartingTokenIds(audioLanguage))
+          .length;
+        // remove sos/eos token and 3 additional ones
+        if (this.indexOfCurrentlyDecodingChunk == 0) {
```
mkopcins:
```suggestion
        if (this.indexOfCurrentlyDecodingChunk === 0) {
```

src/controllers/SpeechToTextController.ts
mkopcins: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Equality_comparisons_and_sameness

src/controllers/SpeechToTextController.ts
```diff
@@ -211,6 +230,52 @@ export class SpeechToTextController {
...
+  private handleOverlaps(seqs: number[][]): number[] {
+    const maxInd = longCommonInfPref(
+      seqs.at(-2)!,
+      seqs.at(-1)!,
+      HAMMING_DIST_THRESHOLD
+    );
+    let finalSeq = [...this.sequence, ...seqs.at(-2)!.slice(0, maxInd)];
+    this.sequence = finalSeq;
+    this.decodedTranscribeCallback(finalSeq);
+    return finalSeq;
```
mkopcins:
```suggestion
    return finalSe.slice();
```

src/controllers/SpeechToTextController.ts
```diff
@@ -240,98 +303,163 @@ export class SpeechToTextController {
...
+          this.seqs = [
+            ...this.seqs,
+            seq.slice(
+              numSpecialTokens + NUM_TOKENS_TO_SLICE,
+              -(numSpecialTokens + NUM_TOKENS_TO_SLICE)
+            ),
+          ];
+          this.prevSeq = this.handleOverlaps(this.seqs);
+        }
+        this.indexOfCurrentlyDecodingChunk++;
```
mkopcins: I don't think we need this variable at all, instead let's treat `this.chunks` like a queue, which means we always remove first element after processing, and so at each loop iteration we decode `this.decodeChunk(this.chunks[0])` and the check for proper length becomes sth like `this.chunks[0].length === (1 + !this.isChunkDeleted) * this.overlapSeconds + this.windowSize`

mkopcins: also change `isChunkDeleted` to sth like `firstChunk` or simmillar

mkopcins (changes requested):

src/controllers/SpeechToTextController.ts
```diff
@@ -364,97 +403,97 @@ export class SpeechToTextController {
...
         this.decodedTranscribeCallback(seq);
         this.isGeneratingCallback(false);
         break;
       }
       //last sequence processed
-      const numSpecialTokens = (await this.getStartingTokenIds(audioLanguage))
-        .length;
-      if (this.indexOfCurrentlyDecodingChunk == this.numOfChunks - 1) {
-        this.seqs.push(seq.slice(numSpecialTokens + NUM_TOKENS_TO_SLICE));
+      if (indexOfCurrentlyDecodingChunk === this.numOfChunks - 1) {
```
mkopcins: Is this working as intended? `indexOfCurrentlyDecodingChunk` is either 0 or 1, so this seems like it can trigger false positives

mkopcins (changes requested):
Resolve conflicts, I will test it tomorrow, but looks ok in general

src/controllers/SpeechToTextController.ts
mdydek: hypothetically speaking, if the speed of decoding would be lower than filling up windows we can have multiple chunks not decoded not last one or two, am I right?

docs/docs/speech-to-text/useSpeechToText.md
```diff
@@ -63,6 +63,58 @@ if (error) {
 }
 ```
 
+You can also get microphone data however you want, but we provide example using `react-native-live-audio-stream` library.
```
NorbertKlockiewicz: This line here is a repetition from the text in line 34

NorbertKlockiewicz: Also in line 24 there's an info block which I guess isn't valid anymore

docs/docs/speech-to-text/useSpeechToText.md
```diff
@@ -63,6 +63,58 @@ if (error) {
...
+
+:::caution
+Before trying to obtain microphone input remember to add proper permissions in your app. [Details](https://www.npmjs.com/package/react-native-live-audio-stream#add-microphone-permissions)
+:::
+
+```typescript
```
NorbertKlockiewicz: I think the code snippet should also include react native executorch code

docs/docs/speech-to-text/useSpeechToText.md
```diff
@@ -92,15 +144,16 @@ Specifies config for both `overlapSeconds` and `windowSize` values. Three option
...
+| `transcribeStreaming` | `(streamAction: STREAMING_ACTION, input?: number[], audioLanguage?: SpeechToTextLanguage) => Promise<string>`            | Allows optimal transcribing when data is generated real-time e.g microphone data. Options for `STREAMING_ACTION` parameter: <ul> <li> `STREAMING_ACTION.START` resets previous remembered state and prepare for upcoming data.</li><li>`STREAMING_ACTION.DATA` should be invoked with new portions of input, which will be processed after filling the certain size. If input not provided error is thrown.</li> <li>`STREAMING_ACTION.STOP` decodes rest of the data and resolves promise. </li></ul> For multilingual models, you have to specify the audioLanguage flag, which is the language of the spoken language in the audio. |
```
mkopcins:
```suggestion
| `transcribeStreaming` | `(streamAction: STREAMING_ACTION, input?: number[], audioLanguage?: SpeechToTextLanguage) => Promise<string>`            | Starts a transcription process for a chunked data stream (e.g. microphone data). `Input` should be a waveform at 16kHz. Options for `STREAMING_ACTION` parameter: <ul> <li> `STREAMING_ACTION.START` prepares module for processing data stream.</li><li>`STREAMING_ACTION.DATA` should be invoked with new data chunks. Transcription is triggered when inner buffer is filled. If input not provided error is thrown.</li> <li>`STREAMING_ACTION.STOP` decodes rest of the inner buffer and resolves the promise. </li></ul> For multilingual models, you have to specify the audioLanguage flag, which is the language of the spoken language in the audio. |
```

docs/docs/speech-to-text/useSpeechToText.md
```diff
@@ -204,7 +257,7 @@ await model.transcribe(mySpanishAudio, SpeechToTextLanguage.Spanish);
 // Rest of your code...
 ```
 
-## Example
+## File example
```
mkopcins:
```suggestion
## `transcribe` example
```

docs/docs/speech-to-text/useSpeechToText.md
```diff
@@ -227,7 +280,48 @@ function App() {
         title="Transcribe"
       />
       <Text>{error ? error : sequence}</Text>
-    </View>****
+    </View>
+  );
+}
+```
+
+## Microphone example
```
mkopcins:
```suggestion
## `transcribeStreaming` example
```

docs/docs/speech-to-text/useSpeechToText.md
NorbertKlockiewicz:
```suggestion
## Example with a file
```

docs/docs/speech-to-text/useSpeechToText.md
NorbertKlockiewicz:
```suggestion
## Example with a microphone
```

NorbertKlockiewicz (changes requested):

docs/docs/speech-to-text/useSpeechToText.md
mdydek: Do you mean "however you want" part? Only leave part about example using react-native-live-audio-stream library?

src/controllers/SpeechToTextController.ts
```diff
@@ -242,93 +342,154 @@ export class SpeechToTextController {
...
+          indexOfCurrentlyDecodingChunk === this.numOfChunks - 1
+            ? SLICING_PLACE.END
+            : SLICING_PLACE.MIDDLE,
+          seq,
+          audioLanguage
+        )
+      );
+      this.handleOverlaps(this.seqs);
+      if (indexOfCurrentlyDecodingChunk === this.numOfChunks - 1) {
+        let finalSeq = [...this.sequence, ...this.seqs.at(-1)!];
```
mkopcins:
```suggestion
        const finalSeq = [...this.sequence, ...this.seqs.at(-1)!];
```

src/controllers/SpeechToTextController.ts
```diff
@@ -242,93 +342,154 @@ export class SpeechToTextController {
...
+            : this.streamWaveform.slice(this.windowSize - this.overlapSeconds);
+          this.isChunkDeleted = true;
+          break;
+        }
+      }
+      this.isDecodingChunk = false;
+    }
+    let indexOfCurrentlyDecodingChunk = this.firstChunk ? 0 : 1;
+    while (streamAction === STREAMING_ACTION.STOP) {
+      let seq = await this.decodeChunk(
```
mkopcins:
```suggestion
      const seq = await this.decodeChunk(
```

src/controllers/SpeechToTextController.ts
```diff
@@ -242,93 +342,154 @@ export class SpeechToTextController {
...
+      this.streamWaveform = [...this.streamWaveform, ...waveform!];
+    }
+    this.chunkWaveform(this.streamWaveform, this.isChunkDeleted);
+    if (!this.isDecodingChunk && streamAction !== STREAMING_ACTION.STOP) {
+      this.isDecodingChunk = true;
+      while (
+        this.chunks.at(this.firstChunk ? 0 : 1)?.length ==
+        this.expectedChunkLength()
+      ) {
+        let seq = await this.decodeChunk(
```
mkopcins:
```suggestion
        const seq = await this.decodeChunk(
```

mkopcins (changes requested):

examples/llm/screens/LLMScreen.tsx
pweglik, line 1: Do we need audio streaming in <ins>LLM</ins> example app? Maybe it would be more suitable in `speech-to-text`  app? Right now, over half of the file has nothing to do with our LLM feature that it should showcase

src/controllers/SpeechToTextController.ts
```diff
@@ -210,138 +210,242 @@ export class SpeechToTextController {
...
+    let prevSeqTokenIdx = 0;
+    this.prevSeq = this.sequence.slice();
     try {
-      this.checkCanTranscribe();
-    } catch (e) {
-      this.onErrorCallback?.(e);
-      return '';
+      await this.encode(chunk);
+    } catch (error) {
+      this.onErrorCallback?.(new Error(getError(error) + ' encoding error'));
```
pweglik: This regards all `this.onErrorCallback` calls in this file. What if dev doesn't pass `onErrorCallback` in constructor? It would make all those errors disappear into thin air. We may use something like suggested here:
https://github.com/software-mansion/react-native-executorch/issues/183
Maybe even define this.errorCallback in constructor like:
```
this.errorCallback = () => {
if (this.errorCallback) {
    this.errorCallback(getError(e));
  } else {
    throw new Error(getError(e));
  }
```
or some variation. We may also require to pass errorCallback or ignore this issue altogheter but I think it may result in poor DX

src/controllers/SpeechToTextController.ts
```diff
@@ -210,138 +210,242 @@ export class SpeechToTextController {
...
+        this.sequence = [...this.sequence, ...this.seqs.at(-1)!];
+      }
+      this.decodedTranscribeCallback(this.sequence);
+      this.isGeneratingCallback(false);
+      this.streaming = false;
+    }
+
+    const decodedText = await this.tokenIdsToText(this.sequence);
+
+    return decodedText;
```
pweglik, line 443: Maybe move it inside STOP conditional and only `return [];` here instead? It would make it clear that decoding only happens on STOP action

src/controllers/SpeechToTextController.ts
```diff
@@ -210,138 +210,242 @@ export class SpeechToTextController {
...
+        this.seqs[idx] = this.seqs[idx]!.slice(0, -numOfTokensToSlice);
+    }
+  }
+
+  private async trimSequences(audioLanguage?: string) {
+    const numSpecialTokens = (await this.getStartingTokenIds(audioLanguage))
+      .length;
+    this.trimLeft(numSpecialTokens + NUM_TOKENS_TO_SLICE);
+    this.trimRight(numSpecialTokens + NUM_TOKENS_TO_SLICE);
+  }
```
pweglik: We call `trimSeqence` in loop where we push new `seq`. Wouldn't it be enough then:
```typescript
private trimLeft(newSeq, numOfTokensToSlice: number) {
      if (newSeq[0] === this.config.tokenizer.bos)
        newSeq = newSeq.slice(numOfTokensToSlice);
  }

  private trimRight(newSeq, numOfTokensToSlice: number) {
      if (newSeq.at(-1) === this.config.tokenizer.eos)
        newSeq = newSeq.slice(0, -numOfTokensToSlice);
  }

  private async trimSequences(newSeq: number[], audioLanguage?: string) {
    const numSpecialTokens = (await this.getStartingTokenIds(audioLanguage))
      .length;
    this.trimLeft(newSeq, numSpecialTokens + NUM_TOKENS_TO_SLICE);
    this.trimRight(newSeq, numSpecialTokens + NUM_TOKENS_TO_SLICE);
  }
  ```
  Almost feel like you can do without `trimLeft` and `trimRight` - just put it all in one function

File: src/controllers/SpeechToTextController.ts [hunk 27]
```diff
@@ -210,138 +210,242 @@ export class SpeechToTextController {
...
+      const chunk = this.waveform.slice(
+        0,
+        this.windowSize +
+          this.overlapSeconds * (1 + Number(this.seqs.length > 0))
+      );
+      this.chunks = [chunk]; //save last chunk for STREAMING_ACTION.STOP
+      this.waveform = this.waveform.slice(
+        this.windowSize - this.overlapSeconds * Number(this.seqs.length === 0)
+      );
+      const seq = await this.decodeChunk(chunk, audioLanguage);
```
pweglik, line 405: We decode it online. What happens if decoding is slower than new samples coming in? Is it realistic case or it never happens?

src/controllers/SpeechToTextController.ts
mkopcins: But this is not the case at all, decoding happens in on-line fashion so it is expected to have something returned from other `STREAMING_ACTION` values

src/controllers/SpeechToTextController.ts
mkopcins: this code is not strictly correct since we trimRight the second to last seq and trim left the last seq, but yes, I can simplify this

src/controllers/SpeechToTextController.ts
mkopcins: Shouldn't happen, with any reasonable settings (haven't really tested unreasonable settings :)), but the inference speed is sufficient for this to never happen. Also we will fix this with c++ implementation, so let's treat this as a known issue for now

src/controllers/SpeechToTextController.ts
pweglik: Ah ok, I see `this.sequence` is also set in `handleOverlaps` - that's what I missed, it's fine then!

examples/llm/ios/Podfile.lock
```diff
@@ -1399,7 +1401,7 @@ PODS:
     - React-jsiexecutor
     - React-RCTFBReactNativeSpec
     - ReactCommon/turbomodule/core
-  - react-native-executorch (0.3.2):
+  - react-native-executorch (0.3.1-stt-12):
```
pweglik: ?

examples/llm/screens/VocieChatScreen.tsx
pweglik, line 1: Misspell in filename

src/controllers/SpeechToTextController.ts
```diff
@@ -64,7 +71,13 @@ export class SpeechToTextController {
       this.isGenerating = isGenerating;
       isGeneratingCallback?.(isGenerating);
     };
-    this.onErrorCallback = onErrorCallback;
+    this.onErrorCallback = (error) => {
+      if (onErrorCallback) {
+        return onErrorCallback(error ? new Error(getError(error)) : undefined);
```
pweglik: I think it's enough to call `onErrorCallback`. We're never going to assign the returned value to anything, so no point in returning result of callback

examples/llm/screens/VocieChatScreen.tsx
```diff
@@ -0,0 +1,221 @@
...
+      visible={!llm.isReady || !speechToText.isReady}
+      textContent={`Loading the model ${(llm.downloadProgress * 100).toFixed(0)} %\nLoading the speech model ${(speechToText.downloadProgress * 100).toFixed(0)} %`}
+    />
+  ) : (
+    <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
+      <KeyboardAvoidingView
+        style={styles.keyboardAvoidingView}
+        behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
+        keyboardVerticalOffset={Platform.OS === 'android' ? 30 : 0}
+      >
```
pweglik: No need for it here, there's no keyboard

examples/llm/screens/LLMScreen.tsx
```diff
@@ -30,39 +33,40 @@ export default function LLMScreen({
...
+      textContent={`Loading the model ${(llm.downloadProgress * 100).toFixed(0)} %\n`}
     />
   ) : (
     <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
-      <View style={styles.container}>
+      <KeyboardAvoidingView
+        style={styles.keyboardAvoidingView}
+        behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
+        keyboardVerticalOffset={Platform.OS === 'android' ? 30 : 0}
+      >
```
pweglik: We already wrap everything in this view in App.tsx. No need to double this

examples/llm/screens/LLMScreen.tsx
pweglik, line 1: Why are we changing this file in this PR? Some of those changes seem like leftover from a previous edit. If there's something that's improvement over previous version leave it, in other cases drop the changes

pweglik (changes requested):

src/controllers/SpeechToTextController.ts
mkopcins: we can still return from void functions, I don't think this is a problem

pweglik: Yeah, but reading this code, I got an impression we want to use value returned by users callback here, in Controller. To make sure this return value is not used I'd need to check all callsites of `this.errorCallback`. But we're not doing anything liek that, so there's no need to cause this confusion

mkopcins (approved):

docs/docs/natural-language-processing/useSpeechToText.md
```diff
@@ -63,6 +65,72 @@ if (error) {
...
+    messageRecorded.current = true;
+    await speechToText.streamingTranscribe(STREAMING_ACTION.STOP);
+  } else {
+    setIsRecording(true);
+    startStreamingAudio(audioStreamOptions, onChunk);
+    await speechToText.streamingTranscribe(STREAMING_ACTION.START);
+  }
+};
+```
+
```
pweglik: It feels like a very verbose example for docs. Maybe we should guide users to an example app to learn about details of how to set up streaming with `LiveAudioStream`? Or move it to Examples section at the bottom of page

pweglik (approved):

PR #181, opened 2025-04-08
feat: support tokenizer config

pweglik:
### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


pweglik:
CI fails due to changes in public API - fixes to CI will solve it. I'd appreciate review of code changes anyway, so it's ready for merge

examples/llama/screens/ChatScreen.tsx
```diff
@@ -14,51 +14,43 @@ import { SafeAreaView } from 'react-native-safe-area-context';
...
   const sendMessage = async () => {
-    appendToMessageHistory(userInput.trim(), 'user');
     setUserInput('');
     textInputRef.current?.clear();
     try {
-      await llama.generate(userInput);
+      console.log('userInput:', userInput);
+      await chat.sendMessage(userInput, TOOL_DEFINITIONS_PHONE);
     } catch (e) {
+      console.log('ERROR');
```
NorbertKlockiewicz: I think this console log here can be removed as the console.error under it, will also be logged

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,232 @@
...
-      await LLM.runInference(input);
-    } catch (err) {
-      throw new Error((err as Error).message);
+      this.responseCallback('');
+      this.isGeneratingCallback(true);
+      await this.nativeModule.runInference(input);
+      this.isGeneratingCallback(false);
+    } catch (e) {
+      this.isGeneratingCallback(false);
+      this.handleError(e);
```
NorbertKlockiewicz: We can use finally here to set isGeneratingCallback to false

NorbertKlockiewicz (approved):

src/modules/natural_language_processing/LLMModule.ts
pweglik: 👍

src/constants/llmDefaults.ts
chmjkb, line 1: I see you're deleting the EOT token here. Can you also make changes to sttDefaults and the stt module to dynamically use BOS/EOS tokens form the tokenizer config? Currently they are hardcoded 🙏🏻

src/constants/modelUrls.ts
chmjkb, line 1: can we also do `LLAMA3_2_1B_TOKENIZER = LLAMA3_2_TOKENIZER`? You're introducing a breaking change which can be avoided very easily, we've done that before (look at the end of this file)

src/types/common.ts
```diff
@@ -54,7 +42,10 @@ export type Module =
...
   | typeof ETModule;
 
+export type MessageRole = 'user' | 'assistant' | 'system';
 export interface MessageType {
-  role: 'user' | 'assistant';
+  role: MessageRole;
   content: string;
 }
+// unfortunately there's no one standard so it's hard to type it better
+export type LLMTool = Object;
```
chmjkb: Not sure if those types are _common_, how about we move them to `types/llms.ts`?

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
+  private _isGenerating = false;
+  private _messageHistory: MessageType[] = [];
 
-  static async load(
+  // User callbacks
+  private responseCallback: (response: string) => void;
+  private messageHistoryCallback: (messageHistory: MessageType[]) => void;
+  private isReadyCallback: (isReady: boolean) => void;
+  private isGeneratingCallback: (isGenerating: boolean) => void;
+  private modelDownloadProgressCallback:
```
chmjkb: Any reason for changing the name of this? It used to be `onDownloadProgressCallback`, as is in every other module.

src/modules/natural_language_processing/LLMModule.ts
chmjkb, line 1: I think all of those methods should be also static

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
+    isReadyCallback,
+    isGeneratingCallback,
+    modelDownloadProgressCallback,
+    errorCallback,
+    chatConfig = {
+      systemPrompt: DEFAULT_SYSTEM_PROMPT,
+      initialMessageHistory: DEFAULT_MESSAGE_HISTORY,
+      contextWindowLength: DEFAULT_CONTEXT_WINDOW_LENGTH,
+    },
+  }: {
```
chmjkb: As I mentioned in the other comment, I think we should do what is done in the LLMModule, we call a static `load()`, instead of a constructor.

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
+
+export const SPECIAL_TOKENS = [
+  'bos_token',
+  'eos_token',
+  'unk_token',
+  'sep_token',
+  'pad_token',
+  'cls_token',
+  'mask_token',
+];
```
chmjkb: Can we move this to `types/llm.ts` as well?

src/types/common.ts
chmjkb: Also, if we don't type this it would be nice to give some context in the comment here. I'm not sure what that is :D

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
+      this._messageHistory = messageHistory;
+      messageHistoryCallback?.(messageHistory);
+    };
+    this.isReadyCallback = (isReady) => {
+      this._isReady = isReady;
+      isReadyCallback?.(isReady);
+    };
+    this.isGeneratingCallback = (isGenerating) => {
+      this._isGenerating = isGenerating;
+      isGeneratingCallback?.(isGenerating);
```
chmjkb: I'm not sure I understand what is going on here. Why not `this.isGeneratingCallback = isGeneratingCallback`, etc?

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
-  static async generate(input: string) {
+  public delete() {
+    this.onToken?.remove();
+    this.onToken = null;
+    this.nativeModule.deleteModule();
+  }
+
+  public async runInference(input: string) {
+    if (!this._isReady) {
+      throw new Error('Model is not loaded!');
```
chmjkb:
```suggestion
      throw new Error(getError(ETError.ModuleNotLoaded));

```

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
+    }
     try {
-      await LLM.runInference(input);
-    } catch (err) {
-      throw new Error((err as Error).message);
+      this.responseCallback('');
+      this.isGeneratingCallback(true);
+      await this.nativeModule.runInference(input);
+    } catch (e) {
+      this.handleError(e);
```
chmjkb: I am not sure if we should set the error callback here. IIRC we agreed on a convention that we should throw errors when they occur during inference (they are somehow not related to the lifecycle of a module). Possibly something that we should re-discuss.

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,225 @@
...
+    ];
+
+    const renderedChat: string = this.applyChatTemplate(
+      messageHistoryWithPrompt,
+      this.tokenizerConfig,
+      tools
+    );
+
+    await this.runInference(renderedChat);
+
```
chmjkb: I'm wondering if we should somehow cap the number of tokens passed to this function. Since there is a new flag `numContextLen` within the ET runtime, I suspect that this might crash the app on long conversations.

examples/llama/screens/ChatScreen.tsx
```diff
@@ -14,51 +14,40 @@ import { SafeAreaView } from 'react-native-safe-area-context';
...
-    modelSource: LLAMA3_2_1B_QLORA,
-    tokenizerSource: LLAMA3_2_1B_TOKENIZER,
-    contextWindowLength: 6,
+  const llm = useLLM({
+    modelSource:
+      'https://huggingface.co/nklockiewicz/ocr/resolve/main/hammer/hammer-1_5b_bf16.pte',
+    tokenizerSource:
+      'https://huggingface.co/nklockiewicz/ocr/resolve/main/hammer/tokenizer-hammer.json',
+    tokenizerConfigSource:
+      'https://huggingface.co/MadeAgents/Hammer2.1-1.5b/resolve/main/tokenizer_config.json',
```
chmjkb: Use constants from the library

examples/llama/screens/ChatScreen.tsx
chmjkb, line 1: I'm not sure if its a dev version or not. If it's not, then we shouldn't kill the conversational LLM functionality. Maybe add a toggle for function calling.

chmjkb (changes requested):

examples/llama/screens/ChatScreen.tsx
pweglik: Good suggestion, maybe we should do something similar to CV demo apps - two tabs with two different models loading? Or some kind of switch when you start the app? What would you suggest? I like the setup screen before showing ChatScreen

src/constants/modelUrls.ts
pweglik: Good suggestion 👍

src/constants/llmDefaults.ts
pweglik: Can you create an issue for it and assign me? I think it's better to keep it to separate PR.

src/modules/natural_language_processing/LLMModule.ts
pweglik: Is it handled yet anywhere in the library? I'd keep such changes to the PR that solves that problem. If it's not handled anywhere yet, I'd keep it that way here as well, is it okay?

src/modules/natural_language_processing/LLMModule.ts
pweglik: I like the idea, I'll change it accordingly

pweglik:
Temporarily changing to draft to apply chagnes and wait for merge of https://github.com/software-mansion/react-native-executorch/pull/197 and https://github.com/software-mansion/react-native-executorch/pull/163

src/modules/natural_language_processing/LLMModule.ts
chmjkb: AFAIK It is not yet handled anywhere in our code, that is because this is a new change which was introduced in the v0.6.0 ET runtime. It is strictly related to the LLM runner, you can see the line that causes it [here](https://github.com/pytorch/executorch/blob/a4be2e4e26284306e03c4b716317eb0fdf8eac09/examples/models/llama/runner/runner.cpp#L49). The `kMaxContextLen` is taken from the metadata of the model _export_. Our exports currently have this parameter set to 2048, so if the conversation is any longer the app will most likely crash. I think we should discuss how we approach this, I'll open up an issue.

pweglik:
Ready for review (don't mind most of the docs - they are preemptive merge of docs changed structure from @jakmro) I've only modified `useLLM.md` and `LLMModule.md`

pweglik:
Force push after massive rebase

android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
NorbertKlockiewicz, line 1: Do you have ktlint installed, as formatting in this file looks odd to me

android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
```diff
@@ -27,69 +19,42 @@ class LLM(
...
-      // We want to add the LLM response to the conversation once all the tokens are generated.
-      // Each token is appended to the tempLlamaResponse StringBuilder in onResult callback.
-      this.conversationManager.addResponse(this.tempLlamaResponse.toString(), ChatRole.ASSISTANT)
-      this.tempLlamaResponse.clear()
-      Log.d("ExecutorchLib", this.conversationManager.getConversation())
-    }.start()
-
-    promise.resolve("Inference completed successfully")
+              llmModule!!.generate(input, this)
+              promise.resolve("Inference completed successfully")
```
chmjkb: Maybe it's not within the scope of what you're doing here, but we should get rid of that weird resolve.

docs/docs/natural-language-processing/useLLM.md
chmjkb, line 1: There is also a quick start guide in the main readme, make sure to update that as well

examples/llama/screens/ChatScreen.tsx
```diff
@@ -15,50 +15,75 @@ import SWMIcon from '../assets/icons/swm_icon.svg';
...
+  LLAMA3_2_TOKENIZER,
+  LLAMA3_2_TOKENIZER_CONFIG,
+  LLMType,
   useLLM,
 } from 'react-native-executorch';
 import PauseIcon from '../assets/icons/pause_icon.svg';
 import ColorPalette from '../colors';
 import Messages from '../components/Messages';
-import { MessageType, SenderType } from '../types';
+import { LLMTool } from 'react-native-executorch/lib/typescript/types/llm';
```
NorbertKlockiewicz: The import should be from 'react-native-executorch'

ios/RnExecutorch/LLM.mm
chmjkb, line 1: It seems like you forgot to delete the following from the `releaseResources` function:
```objective-c
self->conversationManager = nil;
self->tempLlamaResponse = nil;
```

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -15,11 +15,13 @@ keywords:
...
     text generation,
+    tool calling,
+    function calling,
   ]
-description: "Learn how to use Llama models in your React Native applications with React Native ExecuTorch's useLLM hook."
+description: "Learn how to use LLMs in your React Native applications with React Native ExecuTorch's useLLM hook."
 ---
 
-React Native ExecuTorch supports Llama 3.2 models, including quantized versions. Before getting started, you’ll need to obtain the .pte binary—a serialized model—and the tokenizer. There are various ways to accomplish this:
+React Native ExecuTorch supports Llama 3.2 models, including quantized versions (and few other LLMs as well, checkout our [HuggingFace repository](https://huggingface.co/software-mansion)). Before getting started, you’ll need to obtain the .pte binary—a serialized model and the tokenizer and tokenizer config JSON files. There are various ways to accomplish this:
```
chmjkb: I think we should point to the list of supported models here instead and rephrase this to say React Native ExecuTorch supports a variety of LLMs including LLama 3.2, ...
maybe this is better for SEO 😅, also more readable to me

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -15,11 +15,13 @@ keywords:
...
 
 - For your convenience, it's best if you use models exported by us, you can get them from our [HuggingFace repository](https://huggingface.co/software-mansion/react-native-executorch-llama-3.2). You can also use [constants](https://github.com/software-mansion/react-native-executorch/tree/main/src/constants/modelUrls.ts) shipped with our library.
```
chmjkb: This bullet point is kind of outdated at this point, since it points to a single repository. Maybe we should point to the model urls here without pointing to any repository.

chmjkb: Maybe we could also mention using optimum-executorch somewhere 👍🏻

examples/llama/screens/ChatScreen.tsx
```diff
@@ -15,50 +15,75 @@ import SWMIcon from '../assets/icons/swm_icon.svg';
...
 import Spinner from 'react-native-loading-spinner-overlay';
 import {
+  HAMMER2_1_1_5B,
+  HAMMER2_1_TOKENIZER,
+  HAMMER2_1_TOKENIZER_CONFIG,
   LLAMA3_2_1B_QLORA,
-  LLAMA3_2_1B_TOKENIZER,
+  LLAMA3_2_TOKENIZER,
+  LLAMA3_2_TOKENIZER_CONFIG,
+  LLMType,
```
chmjkb, line 24: It seems like you're not exporting that type

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -102,63 +112,78 @@ Given computational constraints, our architecture is designed to support only on
...
+
+```typescript
+return (
+  <View>
+    {llm.chatHistory.map((message) => (
+      <Text>{message.content}</Text>
+    ))}
+  </View>
+)
+```
```
chmjkb: You're using both chatHistory and messageHistory, not sure which is correct.

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -102,63 +112,78 @@ Given computational constraints, our architecture is designed to support only on
...
 **`modelSource`** - A string that specifies the location of the model binary. For more information, take a look at [loading models](../fundamentals/loading-models.md) section.
 
-**`tokenizerSource`** - URL to the binary file which contains the tokenizer
+**`tokenizerSource`** - URL to the JSON file which contains the tokenizer
 
-**`systemPrompt`** - Often used to tell the model what is its purpose, for example - "Be a helpful translator"
+**`tokenizerConfigSource`** - URL to the JSON file which contains the tokenizer config
 
-**`messageHistory`** - An array of `MessageType` objects that represent the conversation history. This can be used to provide context to the model.
+**`chatConfig`** - Object configuring chat management:
```
chmjkb:
```suggestion
**`chatConfig`** - Object configuring chat management, contains following properties:
```

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -102,63 +112,78 @@ Given computational constraints, our architecture is designed to support only on
...
+| `sendMessage`      | `(message: string, tools?: LLMTool[]) => Promise<void>` | Method to add user message to conversation. After model responds, `messageHistory` will be updated with both user message and model response.                                                                                                                                                                                                                          |
+| `runInference`     | `(input: string) => Promise<void>`                      | Runs model inference with raw input string. You need to provide entire conversation and prompt (in correct format and with special tokens!) in input string to this method. It doesn't manage conversation context. It is intended for users that need access to the model itself without any wrapper. If you want simple chat with model consider using `sendMessage` |
```
chmjkb, line 142: I think that this is pretty raw. Maybe for now we should just let users pass list of message objects (messageHistory, in fact), or something like this, wdyt?

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -102,63 +112,78 @@ Given computational constraints, our architecture is designed to support only on
...
+const llm = useLLM({
+  modelSource: LLAMA3_2_1B,
+  tokenizerSource: LLAMA3_2_TOKENIZER,
+  tokenizerConfigSource: LLAMA3_2_TOKENIZER_CONFIG,
 });
 
 ...
 const message = 'Hi, who are you?';
-await llama.generate(message);
+await llm.generate(message);
```
chmjkb:
```suggestion
await llm.sendMessage(message);
```

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -102,63 +112,78 @@ Given computational constraints, our architecture is designed to support only on
...
 ...
 ```
 
 ## Listening for the response
 
-As you might've noticed, there is no return value from the `runInference` function. Instead, the `.response` field of the model is updated with each token.
+As you might've noticed, there is no return value from the `runInference` function. Instead, the `response` field of the model is updated with each token.
```
chmjkb:
```suggestion
As you might've noticed, there is no return value from the `runInference` and `sendMessage` functions. Instead, the `response` field of the model is updated with each token.
```

docs/docs/typescript-api/LLMModule.md
```diff
@@ -1,35 +1,40 @@
 ---
 title: LLMModule
+sidebar_position: 3
 ---
 
+<!-- TODO change it!! -->
```
chmjkb: Not sure if it's relevant or here by accident, but if is still relevant please mention what needs to be done :D

docs/docs/natural-language-processing/useLLM.md
chmjkb, line 1: Btw the benchmarks table is now outdated, so is model list

docs/docs/typescript-api/LLMModule.md
```diff
@@ -1,35 +1,40 @@
...
+    onDownloadProgressCallback: printDownloadProgress,
+  }
+);
 
 // Running the model
-LLMModule.generate('Hello, World!');
+await LLMModule.sendMessage('Hello, World!');
 
-// Interrupting the model
+// Interrupting the model (to actually interrupt the generatoin it would have to be called when sendMessage or runInference is running)
```
chmjkb:
```suggestion
// Interrupting the model (to actually interrupt the generation it would have to be called when sendMessage or runInference is running)
```

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
+| `sendMessage`  | `(message: string, tools?: LLMTool[]) => Promise<void>`                                                                                                                                                                                                                                                                                 | Method to add user message to conversation. After model responds it will call`messageHistoryCallback()`containing both user message and model response.                                                                                                                                                                                                               |
```
chmjkb: Wrong return type, it returns `Promise<MessageType[]>`.

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
+| `runInference` | `(input: string) => Promise<void>`                                                                                                                                                                                                                                                                                                      | Runs model inference with raw input string. You need to provide entire conversation and prompt (in correct format and with special tokens!) in input string to this method. It doesn't manage conversation context. It is intended for users that need access to the model itself without any wrapper. If you want simple chat with model consider using`sendMessage` |
```
chmjkb: Same here, the return type is `Promise<string>` in the code, not sure if that's even correct.

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
+| `delete`       | `() => void`                                                                                                                                                                                                                                                                                                                            | Method to delete the model from memory.                                                                                                                                                                                                                                                                                                                               |
```
chmjkb, line 49: I think this method is not defined in the `LLM.mm`

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
+
+- **`initialMessageHistory`** - An array of `MessageType` objects that represent the conversation history. This can be used to provide initial context to the model.
+
+- **`contextWindowLength`** - The number of messages from the current conversation that the model will use to generate a response. The higher the number, the more context the model will have. Keep in mind that using larger context windows will result in longer inference time and higher memory usage.
+
+- **`onDownloadProgressCallback`** - Function that will be called on download progess.
+
+- **`responseCallback`** - Function that will be called on every generated token.
+
+- **`messageHistoryCallback`** - Function that will be called on every finished message. Returns entire message history.
```
chmjkb: That looks wrong

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
 
-To subscribe to the download progress event, you can use the `onDownloadProgress` method. It accepts a callback function that will be called whenever the download progress changes.
+To subscribe to the download progress event, you can pass the `modelDownloadProgressCallback` functions to constructor. This function will be called whenever the download progress changes.
 
 ## Running the model
 
-To run the model, you can use the `generate` method. It accepts one argument, which is the input string. The method returns a promise, which can resolve to an error or void.
-
-## Listening for token
+To run the model, you can use the `sendMessage` method. It accepts two argument, which are the user message and, optionally tools for model to use. After model responds it will return new message history containing both user message and model response.. Additionally, it will call `messageHistoryCallback`
```
chmjkb:
```suggestion
To run the model, you can use the `sendMessage` method. It accepts two arguments, which are the user message and, optionally tools for model to use. After model responds it will return new message history containing both user message and model response.. Additionally, it will call `messageHistoryCallback`
```

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
 
-To subscribe to the token event, you can use the `onToken` method. It accepts a callback function that will be called whenever a token is generated.
+Alternatively, you can use `runInference`. It provides direct access to the model, without any wrapper, so the input string is passed straight into the model. If you're not sure what are implications of that, you're better of with `sendMessage`
```
chmjkb:
```suggestion
Alternatively, you can use `runInference`. It provides direct access to the model, without any wrapper, so the input string is passed straight into the model. If you're not sure what are implications of that, you're better off with `sendMessage`
```

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
 
-## Interrupting the model
```
chmjkb, line 88: I don't think you should delete that

docs/docs/typescript-api/LLMModule.md
```diff
@@ -38,56 +43,74 @@ LLMModule.delete();
...
+## Listening for token
 
-In order to interrupt the model, you can use the `interrupt` method.
+To subscribe to the token event, you can pass `responseCallback` or `messageHistoryCallback` functions to constructor. `responseCallback` is called on every token and contains only the most recent model response and `messageHistoryCallback` is called whenever model finishes generation and contains all message history including user's and model's last masseges.
```
chmjkb: To subscribe to the `token event` - I think you should elaborate on what you mean by token event.

examples/llama/App.tsx
```diff
@@ -1,16 +1,88 @@
...
+    switch (selectedMode) {
+      case ModelType.LLM:
+        return <ChatScreenLLM />;
+
+      case ModelType.LLM_TOOL_CALLING:
+        return <ChatScreenLLMToolCalling />;
+
+      default:
+        return <ChatScreenLLM />;
+    }
```
chmjkb, line 37: Maybe a cleaner solution would be to pass different props to something like ChatScreen, since those screens don't differ. The `ChatScreenLM` and `ChatScreenLLMToolCalling` differ just by the props passed to useLLM.

examples/llama/App.tsx
```diff
@@ -1,16 +1,88 @@
...
+  };
 
   return (
     <SafeAreaProvider>
-      <ChatScreen />
+      <SafeAreaView style={styles.container}>
+        <View style={styles.topContainer}>
+          <SWMIcon width={45} height={45} />
+          <View style={styles.wheelPickerContainer}>
+            <ScrollPicker
```
chmjkb, line 46: I'm not sure if it's the exact same picker as in the CV app but if it is, let's not use it as it offers very poor UX.

examples/llama/components/Messages.tsx
```diff
@@ -1,13 +1,13 @@
 import { useRef } from 'react';
 import { ScrollView, StyleSheet, View, Text } from 'react-native';
-import { MessageType } from '../types';
 import AnimatedChatLoading from './AnimatedChatLoading';
 import LlamaIcon from '../assets/icons/llama_icon.svg';
 import ColorPalette from '../colors';
 import MessageItem from './MessageItem';
+import { MessageType } from 'react-native-executorch/lib/typescript/types/llm';
```
chmjkb: Refactor this so its just:
`import { MessageType } from 'react-native-executorch';`

examples/llama/screens/ChatScreen.tsx
```diff
@@ -15,50 +15,75 @@ import SWMIcon from '../assets/icons/swm_icon.svg';
...
+  tools?: boolean;
+}) {
+  const [isTextInputFocused, setIsTextInputFocused] = useState(false);
+  const [userInput, setUserInput] = useState('');
+
   useEffect(() => {
-    if (llama.response && !llama.isGenerating) {
-      appendToMessageHistory(llama.response, 'assistant');
+    if (llm.error) {
+      console.log('LLM error:', llm.error);
```
chmjkb, line 64: I think it would be nice to display the error message in the app instead.

src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -1,62 +1,58 @@
...
   }
 
-  static onDownloadProgress(callback: (downloadProgress: number) => void) {
-    this.onDownloadProgressCallback = callback;
+  static async runInference(input: string): Promise<string> {
+    await this.controller.runInference(input);
+    return this.controller.response;
   }
 
-  static onToken(callback: (data: string | undefined) => void) {
```
chmjkb, line 51: Why are we getting rid of this?

chmjkb (changes requested):

NorbertKlockiewicz (approved):

android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
pweglik: You're right, it looks weird. I'm not sure why hook didn't fix it

android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
pweglik: We need it, so the promise in RN side is resolved.
In LLM.mm we also resolve promises:
```
dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
        NSError *error = nil;
        [self->runner generate:input
             withTokenCallback:^(NSString *token) {
               [self onResult:token prompt:input];
             }
                         error:&error];

        if (error) {
          reject(@"error_in_generation", error.localizedDescription, nil);
          return;
        }
        resolve(@"Inference completed successfully");
        return;
      }
      ```

chmjkb: Yeah what i meant is that resolving a promise with this message doesn't make much sense (for example we could return the generated text).

docs/docs/natural-language-processing/useLLM.md
pweglik: I'll create seprate issue for it

docs/docs/natural-language-processing/useLLM.md
pweglik: Do we have a list of supported models? I feel like creating one is a little bit out of scope in this already overgrown PR, and it's constantly growing. I've rephrased it though

docs/docs/natural-language-processing/useLLM.md
pweglik: Changed URL, but I'm not really sure what could I write about optimum-executorch as I haven't worked with it

docs/docs/natural-language-processing/useLLM.md
pweglik: I disagree. That's the point of this method. It's for using raw model. For managed conversation, users are advised to use `sendMessage` that appends the single message of the user to chat history

docs/docs/natural-language-processing/useLLM.md
pweglik: `messageHistory` I forgot to change this one

docs/docs/typescript-api/LLMModule.md
pweglik: Yeah, I copied them over from useLLM.ts and in there they don't return anything. Module has to be static so the functions has to return results for it to make sense. That's why there's difference. Fixed  now

docs/docs/typescript-api/LLMModule.md
pweglik: It's actually fixed in next PRs. One of @jakmro  PRs changed `delete` -> `releaseResources` in LLM.mm without anyone noticing. I'd let it be here and let next PR fix it

docs/docs/typescript-api/LLMModule.md
pweglik: Fixed now

docs/docs/typescript-api/LLMModule.md
pweglik: Changed it to `token generation event`

examples/llama/App.tsx
pweglik: Yeah, I've used the same thing but I'd rather extract it to separate PR. We can change it in all of example apps, if it's bad

examples/llama/screens/ChatScreen.tsx
pweglik: In all of the demo apps we only log errors, they are demo apps after all. We can create PR for demo apps improvement, but I feel like adding extra components for nice error is out of scope here

src/modules/natural_language_processing/LLMModule.ts
pweglik: We simply pass `responseCallback` to `load()`

android/src/main/java/com/swmansion/rnexecutorch/LLM.kt
pweglik: To return generated text, we would need to accumulate it somewhere, but that's moved to the RN side now. We could resolve without any text, but this way it's kinda self-explanatory what happended and doesn't require comment.

chmjkb: I mean resolving with no text doesn't make much sense as well, pretty much the same as this. It's just that I think resolving with the result string would make sense here. Let's keep this as a future consideration tho.

docs/docs/natural-language-processing/useLLM.md
chmjkb: Yeah, we do have such lists, for example:
https://docs.swmansion.com/react-native-executorch/docs/computer-vision/useClassification#supported-models
or
https://docs.swmansion.com/react-native-executorch/docs/speech-to-text/useSpeechToText#supported-models

docs/docs/natural-language-processing/useLLM.md
chmjkb: I think our current API design has two extremes:
- `sendMessage` is very high-level and handles everything for the user and doesn't let them handle the history.
- `runInference` is extremely low-level and requires users to implement their own conversation management. (including the `<assistant>`, etc. tokens part, which I don't think is going to benefit a lot of people)`

While it makes sense to keep both, I feel we’re missing a middle-ground option for users who want a bit more flexibility without having to manage everything manually.
One idea could be to extend sendMessage with an optional argument (like history or other config params) so users can optionally control parts of the conversation flow while still benefiting from the managed setup.
Not saying we should drop runInference — it’s useful for edge cases — but introducing this intermediary layer could make the library more approachable for a broader range of users.

I feel like the current design doesn't really solve the problem we were initially trying to solve. 

Also, this is some indicator of what the devs are looking for:
https://discord.com/channels/464786597288738816/1250038860453380176/1365472600985440339

docs/docs/natural-language-processing/useLLM.md
pweglik: I'll create issue for it. We should create a list like this for LLMs and then include it here.

examples/llama/App.tsx
pweglik: Created issue for it

docs/docs/natural-language-processing/useLLM.md
pweglik: One way to solve it would be to introduce another stateless function (just like `runInference` it shouldn't modify the history). Call it `runInferenceOnMessageHistory` and that would do basically that: 
- accept message history
- apply chat template to it
- send it to model
It'd be exactly middle way between `runInference` and `sendMessage`

Another way would be to create getter for tokenizerConfig and make `applyChatTemplate` a public method. Then users would be able to do whatever they want with those building blocks. It would be the same pipeline, they would pass array of messages, `applyChatTemplate` turns it into one big string that can be passed into `runInference`. They could basically reimplement `sendMessage` themselves but do it history handling however they want.

What do you think about those two methods?

docs/docs/typescript-api/LLMModule.md
pweglik: Fixed now, I've changed my mind and fixed it here as well

docs/docs/natural-language-processing/useLLM.md
pweglik: I think I prefer the first way of adding another function. I'd move it to separate PR though, as it simply add functionality and doesn't modify any existing code. It'll be easier to review this way.
I think correct naming would be:
`runInference` - for the new function - the one that accepts Message[]
`runInferenceRaw` - new name for current function - the one that simply accepts string

examples/llama/App.tsx
pweglik: I actually like this composition pattern here. It creates clean `App.tsx` file. And creates more abstract `ChatScreen.tsx` that simply accepts LLM, but doesn't care about specifics. In next PR, tools are enclosed in LLM config so it is even cleaner without `tools` props to ChatScreen

chmjkb (approved):

PR #182, opened 2025-04-09
refactor: apply the new useModule to the remaining modules, refactor BaseModule 

chmjkb:
## Description

This PR aims to reduce code complexity by removing some redundant abstractions (for example the wrappers of native modules). It also introduces some changes to the baseModule and applies the new useModule hook to the remaining hooks. 

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


src/native/RnExecutorchModules.ts
```diff
@@ -12,137 +12,43 @@ const LINKING_ERROR =
...
-  ? TokenizerSpec
-  : new Proxy(
-      {},
-      {
-        get() {
-          throw new Error(LINKING_ERROR);
-        },
-      }
-    );
+function returnSpecOrThrowLinkingError(source: any) {
```
jakmro:
```suggestion
const returnSpecOrThrowLinkingError = (spec: any) => {
  return spec
    ? spec
    : new Proxy(
        {},
        {
          get() {
            throw new Error(LINKING_ERROR);
          },
        }
      );
};
```

src/controllers/OCRController.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '../utils/fetchResource';
 
 export class OCRController {
-  private nativeModule: _OCRModule;
+  private nativeModule: typeof OCRNativeModule
```
JakubGonera: This name is inconsistent with other controllers, but we could just get rid of the field altogether and use `OCRNativeModule` directly?

src/controllers/SpeechToTextController.ts
```diff
@@ -36,7 +36,7 @@ const longCommonInfPref = (seq1: number[], seq2: number[]) => {
 };
 
 export class SpeechToTextController {
-  private nativeModule: _SpeechToTextModule;
+  private speechToTextNativeModule: typeof SpeechToTextNativeModule;
```
JakubGonera, line 39: Same as above.

src/controllers/VerticalOCRController.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '../utils/fetchResource';
 
 export class VerticalOCRController {
-  private nativeModule: _VerticalOCRModule;
+  private nativeModule: typeof VerticalOCRNativeModule;
```
JakubGonera: Same as above.

JakubGonera (approved):

src/controllers/SpeechToTextController.ts
chmjkb: I made that change due to the fact that we also have a `nativeTokenizer` class property, which is also a native module, therefore I though that this naming might be a bit vague/confusing in this case. 

Maybe i can change the naming in OCRController to `ocrNativeModule` or something like that to unify the naming across controllers. Does that make sense?

src/controllers/OCRController.ts
chmjkb: I'm not a fan of using `OCRNativeModule` directly, as it is required to override the variable when extending BaseModule, so as I mentioned in my other comment I can just change the naming here.

JakubGonera: But you are not extending BaseModule here, or am I missing something?

chmjkb: yep, I just wanted it to remain consistent across modules and controllers. 
If we have to do this for modules, we might do this for controllers as well to avoid inconsistencies.

src/modules/BaseModule.ts
```diff
@@ -1,38 +1,29 @@
...
+    () => {};
 
+  static async load(...sources: ResourceSource[]): Promise<void> {
     try {
-      const fileUri = await fetchResource(
-        modelSource,
-        this.onDownloadProgressCallback
+      const modelFileUris = await Promise.all(
+        sources.map((source) =>
+          fetchResource(source, this.onDownloadProgressCallback)
```
jakmro: `onDownloadProgressCallback` should be applied to only one source or you can use `calculateDownloadProgres` to multiple sources, depending on the experience you want to deliver

jakmro (changes requested):

src/controllers/OCRController.ts
JakubGonera: Sure

pweglik (approved):
Let's merge it as it is and then make another PR to [@jakmro/code-refactor](https://github.com/software-mansion/react-native-executorch/tree/%40jakmro/code-refactor) according to what we've discussed at a meeting:
1. Move implementation from Controllers to Modules.
2. isReady and isGenerating should be implemented in Module

PR #184, opened 2025-04-11
feat: Add C++ JSI interface and Executorch dependency

JakubGonera:
## Description

Add C++ JSI interface and Executorch dependency for both Android and iOS.

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


android/build.gradle
```diff
@@ -21,7 +23,14 @@ buildscript {
 
 def reactNativeArchitectures() {
   def value = rootProject.getProperties().get("reactNativeArchitectures")
-  return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
+  def defaultArchitectures = ["x86_64", "arm64-v8a"]
```
chmjkb, line 28: Can you add a comment here that would explain that we only have executorch binaries built for those architectures?

android/build.gradle
```diff
@@ -21,7 +23,14 @@ buildscript {
...
+  if(!value) {
+    return defaultArchitectures
+  }
+  else { 
+    def architectures = value.split(",")
+    return architectures.findAll { it in defaultArchitectures }
```
chmjkb: You can get rid of the else statement

android/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+  ReactAndroid::reactnative
+)
+else()
+set(RN_VERSION_LINK_LIBRARIES
+  ReactAndroid::folly_runtime
+  ReactAndroid::react_nativemodule_core
+  ReactAndroid::glog
+  ReactAndroid::reactnativejni
+)
+endif()
```
chmjkb: Not sure if we should do this check if we don't support old arch

src/modules/general/ETInstallerModule.ts
```diff
@@ -0,0 +1,9 @@
+import { _ETInstallerModule } from '../../native/RnExecutorchModules';
+
+export class ETInstallerModule {
+  static module = new _ETInstallerModule();
+
+  static install() {
+    return this.module.install();
```
chmjkb: Lets get rid of this wrapper class

src/native/RnExecutorchModules.ts
```diff
@@ -296,6 +310,12 @@ class _TokenizerModule {
   }
 }
 
+class _ETInstallerModule {
+  install(): ReturnType<ETInstallerInterface['install']> {
+    return ETInstaller.install();
+  }
+}
+
```
chmjkb: As above, lets not use that wrapper here

chmjkb (changes requested):

msluszniak:
C++ part looks good :))

android/build.gradle
JakubGonera: What if someone wants to build only for a single architecture?

chmjkb: what i meant is that you could do:
```
if (!value) {
    return defaultArchitectures
}
def architectures = value.split(",")
return architectures.findAll { it in defaultArchitectures }
```
but ofc its a nitpick though

android/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.13)
+project(RnExecutorch)
+
+set (CMAKE_VERBOSE_MAKEFILE ON)
+set (CMAKE_CXX_STANDARD 20)
+
+include("${REACT_NATIVE_DIR}/ReactAndroid/cmake-utils/folly-flags.cmake")
+add_compile_options(${folly_FLAGS})
+
+if(${IS_NEW_ARCHITECTURE_ENABLED})
```
NorbertKlockiewicz: Shouldn't we just omit such checks if we only support new architecture?

android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchPackage.kt
```diff
@@ -47,6 +49,7 @@ class RnExecutorchPackage : TurboReactPackage() {
           LLM.NAME,
           false, // canOverrideExistingModule
           false, // needsEagerInit
+          true, // hasConstants
```
NorbertKlockiewicz, line 50: Why are you changing it? I know that we will get rid of this but just out of curiosity

third-party/include/executorch/extension/tensor/tensor.h
NorbertKlockiewicz, line 1: I think a good idea would be to include those files directly from executorch(submodule) so we won't have duplicates in repository. Also it should be simpler to bump to newer versions

android/build.gradle
```diff
@@ -21,7 +23,16 @@ buildscript {
...
+  }
 }
 
 def isNewArchitectureEnabled() {
```
NorbertKlockiewicz: I think we can also get rid of this function here and just change it to true whenever its used.

android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchPackage.kt
JakubGonera: According to react-native-audio-api people this was causing hard to trace problems, even though the flag is deprecated.

third-party/include/executorch/extension/tensor/tensor.h
JakubGonera: Executorch doesn't have those headers in a single folder like here, they are generated (or rather separated from the .cpp files) during build. You can see them already in our repo [here](https://github.com/software-mansion/react-native-executorch/tree/main/third-party/ios/ExecutorchLib/frameworks/executorch.xcframework/ios-arm64-simulator/Headers). Of course we could have a separate submodule just for headers that we build by hand but I don't know if that makes anything easier.

We can explore an automated setup in the future -- right now the frameworks already have the headers inside them, and if the Android archive would have them as well, we could fetch ET for both platforms from remote, and then unpack them, but that's not the case right now (see the ET discord discussion from few days ago https://discord.com/channels/1334270993966825602/1349854760299270284/1358966828336152706)

android/build.gradle
JakubGonera: I don't think I understand. Can you elaborate?

JakubGonera: Nvm, I see what you mean.

third-party/include/executorch/extension/tensor/tensor.h
JakubGonera: It seems we should leave the includes here as they are, so for posterity I'll list the reasons:
- we could have a script that builds them which is required to run when cloning the repo, but it would be easy to forget to rebuild the headers each time we would switch between branches with different versions of ET
- they could be included as the whole ET submodule, but then we have to pack the 100mb+ repo into our package
- if executorch.aar would include headers, we could extract the headers from a repository at build time, but as of now executorch.aar does not contain them

android/src/main/java/com/swmansion/rnexecutorch/ETInstaller.kt
```diff
@@ -0,0 +1,44 @@
...
+    } catch (exception: UnsatisfiedLinkError) {
+      throw RuntimeException("Could not load native module Install", exception)
+    }
+  }
+
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  override fun install(): Boolean {
+    injectJSIBindings()
+    return true
+  }
```
mkopcins, line 65: is it possible for this to fail and if so how is this error handled?

common/Log.h
```diff
@@ -0,0 +1,70 @@
...
+#ifdef __ANDROID__
+#include <android/log.h>
+#endif
+#ifdef __APPLE__
+#include <os/log.h>
+#endif
+
+namespace rnexecutorch {
+
+enum LOG_LEVEL { INFO_LVL, ERROR_LVL, DEBUG_LVL };
```
mkopcins: nit pick, can we change to `lvl` to `level`? since the naming is inconsitent between enum name and values, or ski the `lvl` altogether

common/Log.h
```diff
@@ -0,0 +1,70 @@
...
+  va_start(args, fmt);
+
+  // Maximum length of a log message.
+  static constexpr size_t kMaxLogMessageLength = 256;
+  char buf[kMaxLogMessageLength];
+  size_t len = vsnprintf(buf, kMaxLogMessageLength, fmt, args);
+  if (len >= kMaxLogMessageLength - 1) {
+    buf[kMaxLogMessageLength - 2] = '$';
+    len = kMaxLogMessageLength - 1;
+  }
```
mkopcins: Change the number to something that makes more sense other than what ET decided, their logging is not really SotA, and change `$` to ellipsis

mkopcins (changes requested):

common/Log.h
JakubGonera: I added "_LVL" because "DEBUG" was conflicting with something in the global namespace for one of the platforms, but enum classes prevent this problem from happening, so they can be plain "DEBUG" etc..

android/src/main/java/com/swmansion/rnexecutorch/ETInstaller.kt
JakubGonera: If you are asking whether this function is dependant on any kind of input that could be wrong then no, all the input along the consecutive calls will be given at compile time (with class, function names etc.). Right now if an exception is thrown somewhere along the way the app will crash at startup, I don't know if a more sensibile approach is possible.

mkopcins (approved):

PR #197, opened 2025-04-17
fix: error handling

pweglik:
## Description

When non-ET error was passed to `getError()` function it resulted in Undefined error. I've met this example when I had not enough space on the device to download the model

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/src/Error.ts
+++ b/src/Error.ts
@@ -43,5 +43,5 @@ export const getError = (e: unknown | ETError): string => {
   const error = e as Error;
   const errorCode = parseInt(error.message, 10);
   if (errorCode in ETError) return ETError[errorCode] as string;
-  return ETError[ETError.UndefinedError] as string;
+  return error.toString();
 };
--- a/src/Error.ts
+++ b/src/Error.ts
@@ -40,8 +40,12 @@ export const getError = (e: unknown | ETError): string => {
     return ETError[ETError.UndefinedError] as string;
   }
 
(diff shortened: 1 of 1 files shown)


src/Error.ts
```diff
@@ -43,5 +43,5 @@ export const getError = (e: unknown | ETError): string => {
   const error = e as Error;
   const errorCode = parseInt(error.message, 10);
   if (errorCode in ETError) return ETError[errorCode] as string;
-  return ETError[ETError.UndefinedError] as string;
+  return error.toString();
```
NorbertKlockiewicz: To be consistent I think we should also change it in line 40

NorbertKlockiewicz (changes requested):

src/Error.ts
pweglik: It depends, my understanding was, that if the error is just a number, it is something specific to executorch, so we should return undefined executorch error if we get unknown (not defiend yet) error code.
On contrary if it is a string we may return it to user in plaintext

mkopcins: well, if it is something from Executorch it wouldn't get to this line in the first place. I think best case would be to return sth like `ETError[ETError.UndefinedError] + e`

mkopcins (changes requested):

mkopcins (approved):

PR #204, opened 2025-04-22
feat: Add markdown spell-checker

pweglik:
## Description

Adding cspell for spellchecking docs. Eslint plugin couldn't do it as it is made for checking code, not plaintext.
Also fixed some minor misspells we had in repo

### Type of change

### Checklist


package.json
```diff
@@ -68,6 +68,7 @@
     "@types/jest": "^29.5.5",
     "@types/react": "^18.2.44",
     "commitlint": "^17.0.2",
+    "cspell": "^8.19.2",
```
kopcion, line 71: Just a question, we already have `@cspell/eslint-plugin`, is this something we can work with? or maybe we only need one of those

.cspell-wordlist.txt
```diff
@@ -1,6 +1,7 @@
 swmansion
 executorch
 execu
+Execu
```
kopcion, line 4: perhaps also add Torch and ExecuTorch

package.json
pweglik: We need it, "@cspell/eslint-plugin" is necessary for correctly checking files with code, but doesn't really work for Markdown. They mention the differrence here:
https://www.npmjs.com/package/@cspell/eslint-plugin#:~:text=In%20Combination%20with%20CSpell

.cspell-wordlist.txt
pweglik: Not needed, ExecuTorch would be broken down into Execu and Torch and torch is a valid word.
I think, it's best to keep the custom wordlist minimal and only append words to it, when we receive errors

pweglik: We can add "Executorch" - that wouldn't be broken down, BUT it's also not the official way of spelling it, so I guess it's better to keep it off list

mkopcins (approved):

PR #207, opened 2025-04-23
fix: prevent exceeding kMaxContextLen from crashing the app

### LINKED ISSUE #201: Handle large context lengths in the native LLM runner
In the v0.6.0 version of the ExecuTorch runtime, a new flag `kMaxContextLen` was introduced to the LLM runner. You can see it [here](https://github.com/pytorch/executorch/blob/a4be2e4e26284306e03c4b716317eb0fdf8eac09/examples/models/llama/runner/runner.cpp#L49) This flag specifies the length of the context window, so basically the whole context passed to the `generate` function along with the user prompt. This parameter defaults to 128, meaning that if the conversation exceeds 128 tokens, the app will most likely crash. The value of this parameter is overriden if the exported model does not specify that parameter in its metadata. However, we shouldn't rely on that and the default parameter is pretty low. The scope of this issue is to handle it, so the context window is larger and eventually the error is gracefully handled.

chmjkb:
## Description

This PR deletes the use of macro which aborted the ET runtime when the length of the prompt exceeds the kMaxContextLen. Instead of aborting an error is returned to the end user and ET_LOG is made as well. Fixes #201 

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


third-party/ios/ExecutorchLib/ExecutorchLib/runner/runner.cpp
```diff
@@ -187,16 +187,19 @@ Error Runner::generate(const std::string &prompt,
...
+  if (num_prompt_tokens < 1) {
+    ET_LOG(Error,
+           "num_prompt_tokens %d < 1, expected at least 1 token to be passed "
+           "to generate()!",
+           num_prompt_tokens);
+    return Error::InvalidArgument;
+  } else if (num_prompt_tokens >= seq_len) {
+    ET_LOG(Error,
+           "num_prompt_tokens %d >= seq_len %d, Sequence length exceeded - "
+           "please increate the seq_len value papsed to generate()!",
```
NorbertKlockiewicz:
```suggestion
           "please increase the seq_len value passed to generate()!",
```

NorbertKlockiewicz (approved):

NorbertKlockiewicz:
Fix typo, besides that LGTM

mkopcins (approved):

Issue #227, opened 2025-04-28
Port style transfer model to C++


JakubGonera:
Port style transfer capabilities to C++ as a POC for C++/JSI integration. This involves replicating the utilities needed for computer vision processing in C++, as well as adding OpenCV dependency to Android C++ (iOS can use the same dependency as for Objective C++).

PR #229, opened 2025-04-29
feat: port style transfer implementation to C++

JakubGonera:
Port native implementation of style transfer to C++ #227. This is a part of a larger effort to merge native code from Kotlin and Objective C to a single implementation in C++.

# Changes by sections
## Typescript
Old style transfer modules have been replaced by new, stateful implementation. For this purpose non-static versions of useModule and BaseModule have been introduced.
## Objective C & Kotlin
Old implementations of style transfer have been removed. Added methods for handling fetching data via https that are passed on as function objects to C++.
## C++
- Ported image processing capabilities equivalent to the ObjC/Kotlin implementations.
- `ModelHostObject` serves as an automatic interface between model implementations and JSI. This is done by template metaprogramming; defining methods in `JsiConversions.h` for types used by the model is necessary for it to work.
- A factory method for loading a style transfer model is registered in `RnExecutorchInstaller`.

# Code attribution
- `common/rnexecutorch/jsi` originates from react-native-audio-api
- `base64.*` is used for converting from base64 due to https://renenyffenegger.ch/notes/development/Base64/Encoding-and-decoding-base-64-with-cpp/
- `ada` is a header-only library for parsing urls due to https://github.com/ada-url/ada
# Dependencies introduced/updated
- headers for Executorch 0.6 have been updated
- OpenCV 4.11.0 dependency is introduced for Android/C++. For iOS the version is bumped via Cocoapods to 4.11.0.
- C10 headers are the dependency of OpenCV.

# Known issues
- Exceptions in native C++ do not result in promise getting rejected when forwarding yet.
- Garbage collection can malfunction for host objects. C++ host objects need to notify JS runtime about their size via external memory pressure for the garbage collector to know to free them. This is not yet done.


common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,122 @@
...
+  for (int i = 0; i < pixelCount; i++) {
+    int row = i / matSize.width;
+    int col = i % matSize.width;
+    float r = 0, g = 0, b = 0;
+
+    r = buffer[0 * pixelCount + i];
+    g = buffer[1 * pixelCount + i];
+    b = buffer[2 * pixelCount + i];
+
+    cv::Vec3b color((uchar)(b * 255), (uchar)(g * 255), (uchar)(r * 255));
```
msluszniak: Prefer static_cast over c-style casts

common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -0,0 +1,60 @@
...
+
+  JSI_HOST_FUNCTION(forward) {
+
+    auto promise = promiseVendor.createPromise(
+        [this, count, args, &runtime](std::shared_ptr<Promise> promise) {
+          std::thread([this, promise = std::move(promise), count, args,
+                       &runtime]() {
+            constexpr std::size_t forwardArgCount =
+                jsiconversion::getArgumentCount(&Model::forward);
+            if (forwardArgCount != count) {
```
msluszniak, line 30: Maybe add more detailed info on mismatched count

common/rnexecutorch/models/BaseModel.cpp
```diff
@@ -0,0 +1,43 @@
...
+  if (!method_meta.ok()) {
+    throw std::runtime_error("Failed to load forward");
+  }
+
+  auto input_meta = method_meta->input_tensor_meta(0);
+  if (!input_meta.ok()) {
+    throw std::runtime_error("Failed to load forward input");
+  }
+
+  auto shape = input_meta->sizes();
```
msluszniak: Maybe `std::copy` here

common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  return mat;
+}
+
+std::string getTimeID() {
+  return std::to_string(std::chrono::duration_cast<std::chrono::milliseconds>(
+                            std::chrono::system_clock::now().time_since_epoch())
+                            .count());
+}
```
chmjkb: We might use this somewhere else, can we move it?

common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,122 @@
...
+// to this variable. It's done to not handle SSL intricacies manually, as it is
+// done automagically in ObjC++/Kotlin.
+extern FetchUrlFunc_t fetchUrlFunc;
+namespace imageprocessing {
+std::vector<float> colorMatToVector(const cv::Mat &mat) {
+  return colorMatToVector(mat, cv::Scalar(0.0, 0.0, 0.0),
+                          cv::Scalar(1.0, 1.0, 1.0));
+}
+
+std::vector<float> colorMatToVector(const cv::Mat &mat, cv::Scalar mean,
```
chmjkb, line 25: You can add a comment that this is channel-first so I don't have to think too much when using it, likewise for bufferToColorMat

chmjkb: I'm also thinking if the normalization part can be flagged, wdyt?

common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+std::string saveToTempFile(const cv::Mat &image) {
+  std::string filename = "rn_executorch_" + getTimeID() + ".png";
+
+  std::filesystem::path tempDir = std::filesystem::temp_directory_path();
+  std::filesystem::path filePath = tempDir / filename;
+
+  if (!cv::imwrite(filePath.string(), image)) {
+    throw std::runtime_error("Failed to save the image: " + filePath.string());
```
chmjkb, line 73: Btw, how are those exceptions resolved on the JS side?

src/modules/computer_vision/StyleTransferModule.ts
```diff
@@ -1,17 +1,15 @@
...
-  static override async load(modelSource: ResourceSource) {
-    return await super.load(modelSource);
-  }
-
-  static override async forward(
-    input: string
-  ): ReturnType<typeof this.nativeModule.forward> {
-    return await this.nativeModule.forward(input);
+export class StyleTransferModule extends BaseNonStaticModule {
+  static async load(
```
chmjkb: I think this shouldnt be static anymore

src/modules/BaseNonStaticModule.ts
```diff
@@ -0,0 +1,23 @@
+import { ResourceFetcher } from '../utils/ResourceFetcher';
+import { getError } from '../Error';
+import { ResourceSource } from '../types/common';
+
+export class BaseNonStaticModule {
+  protected nativeModule: any;
+  static onDownloadProgressStub: (downloadProgress: number) => void = () => {};
+
+  static async loadFiles(
```
chmjkb: Make it non-static

src/modules/BaseNonStaticModule.ts
```diff
@@ -0,0 +1,23 @@
+import { ResourceFetcher } from '../utils/ResourceFetcher';
+import { getError } from '../Error';
+import { ResourceSource } from '../types/common';
+
+export class BaseNonStaticModule {
+  protected nativeModule: any;
```
chmjkb: Do we need this variable?

common/rnexecutorch/data_processing/ImageProcessing.cpp
NorbertKlockiewicz: I think that the function in line 19 makes up for it, so if you want to normalize you use function with mean and variance arguments and if not you just pass Matrix

common/rnexecutorch/models/BaseModel.cpp
```diff
@@ -0,0 +1,40 @@
...
+          modelSource, Module::LoadMode::MmapUseMlockIgnoreErrors)),
+      runtime(runtime) {
+  Error loadError = module->load();
+  if (loadError != Error::Ok) {
+    throw std::runtime_error("Couldn't load the model, error: " +
+                             std::to_string(static_cast<uint32_t>(loadError)));
+  }
+}
+
+std::vector<int32_t> BaseModel::getInputShape() {
```
NorbertKlockiewicz: Probably we should modify it to return shapes for every input tensor instead of shape of the first one, some models accepts for example two tensors(LLMs)

common/rnexecutorch/models/StyleTransfer.cpp
```diff
@@ -0,0 +1,58 @@
...
+  cv::Size modelImageSize =
+      cv::Size(modelInputShape[modelInputShape.size() - 1],
+               modelInputShape[modelInputShape.size() - 2]);
+
+  cv::Mat resizedInput;
+  cv::resize(input, resizedInput, modelImageSize);
+
+  std::vector<float> inputVector =
+      imageprocessing::colorMatToVector(resizedInput);
+  auto tensor =
```
NorbertKlockiewicz: I believe that up to this point the code should be in preprocess method which will return tensor ready to be put into forward function.

common/rnexecutorch/models/StyleTransfer.cpp
```diff
@@ -0,0 +1,58 @@
...
+      executorch::extension::make_tensor_ptr(modelInputShape, inputVector);
+
+  auto forwardResult = module->forward(tensor);
+  if (!forwardResult.ok()) {
+    throw std::runtime_error(
+        "Failed to forward, error: " +
+        std::to_string(static_cast<uint32_t>(forwardResult.error())));
+  }
+
+  auto resultTensor = forwardResult->at(0).toTensor();
```
NorbertKlockiewicz: Similary to previous comment code below should be in postprocess function.

NorbertKlockiewicz (changes requested):

common/rnexecutorch/data_processing/ImageProcessing.cpp
JakubGonera: All of these have to be caught and currently they are then used to reject a promise -- if they are not caught we will get a runtime error that might crash the app. If the code throwing an exception is not called via a promise, but we still want the error to be pushed to JS, a JSError needs to be created and registered with the JS runtime. I could add a helper function to throw JSError with a C++ exception, the question is whether to do this in this PR. Thoughts?

common/rnexecutorch/data_processing/ImageProcessing.cpp
JakubGonera: I agree, having the overload should be sufficient.

src/hooks/useNonStaticModule.ts
```diff
@@ -13,38 +18,41 @@ export const useNonStaticModule = <
...
-      try {
-        setIsReady(false);
-        modelRef.current = await module.load(...loadArgs, setDownloadProgress);
-        setIsReady(true);
-      } catch (err) {
-        setError((err as Error).message);
-      }
-    };
-    loadModule();
+    if (!doNotLoad) {
```
chmjkb: Sync the name of this flag with Przemek's PR

src/modules/computer_vision/StyleTransferModule.ts
```diff
@@ -1,15 +1,24 @@
...
-  ): Promise<any> {
-    const loadedPaths = await super.loadFiles(
+    onDownloadProgressCallback: (_: number) => void = () => {}
+  ): Promise<void> {
+    const paths = await ResourceFetcher.fetchMultipleResources(
       onDownloadProgressCallback,
       modelSource
     );
-    return global.loadStyleTransfer(loadedPaths[0] || '');
+    this.nativeModule = global.loadStyleTransfer(paths[0] || '');
```
chmjkb, line 16: Make sure this actually works when loading fails

chmjkb (approved):

common/rnexecutorch/Log.cpp
```diff
@@ -1,7 +1,7 @@
-#pragma once
+#include "Log.h"
 
 #include <cstdarg>
-#include <string>
+#include <cstdio>
```
mkopcins, line 4: why the change?

common/rnexecutorch/jsi/JsiHostObject.cpp
```diff
@@ -0,0 +1,108 @@
...
+                  std::placeholders::_1, std::placeholders::_2,
+                  std::placeholders::_3, std::placeholders::_4);
+
+    return hostFunctionCache
+        .emplace(nameAsString, jsi::Function::createFromHostFunction(
+                                   runtime, name, 0, dispatcher))
+        .first->second.asFunction(runtime);
+  }
+
+  return jsi::Value::undefined();
```
mkopcins: For better readability inverese this statement
```suggestion
  if (function == functions_->end()) {
    return jsi::Value::undefined();
  }
    auto dispatcher =
        std::bind(function->second, reinterpret_cast<JsiHostObject *>(this),
                  std::placeholders::_1, std::placeholders::_2,
                  std::placeholders::_3, std::placeholders::_4);

    return hostFunctionCache
        .emplace(nameAsString, jsi::Function::createFromHostFunction(
                                   runtime, name, 0, dispatcher))
        .first->second.asFunction(runtime);
 


```

common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <rnexecutorch/Log.h>
+#include <rnexecutorch/RnExecutorchInstaller.h>
+#include <rnexecutorch/data_processing/FileUtils.h>
+#include <rnexecutorch/data_processing/base64.h>
+
+namespace rnexecutorch {
+// This is defined in RnExecutorchInstaller.cpp. This function fetches data
+// from a url address. It is implemented in Kotlin/ObjectiveC++ and then bound
+// to this variable. It's done to not handle SSL intricacies manually, as it is
+// done automagically in ObjC++/Kotlin.
```
mkopcins, line 17: add similar comment next to varaible declaration too

common/rnexecutorch/data_processing/ImageProcessing.cpp
mkopcins: sounds like a useful helper function to me

common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,129 @@
...
+    // base64
+    std::stringstream uriStream(imageURI);
+    std::string stringData;
+    std::size_t segmentIndex{0};
+    while (std::getline(uriStream, stringData, ',')) {
+      if (segmentIndex == 1)
+        break;
+      ++segmentIndex;
+    }
+    if (segmentIndex != 1) {
```
mkopcins: looking at this, am I right that if uriStream is malformed so that it contains multiple comas this breaks? Shouldn't we parse whole uriStream?

mkopcins: right now segmentIndex is either 0 or 1

common/rnexecutorch/data_processing/ImageProcessing.cpp
```diff
@@ -0,0 +1,129 @@
...
+    // local file
+    auto url = ada::parse(imageURI);
+    image = cv::imread(std::string{url->get_pathname()}, cv::IMREAD_COLOR);
+  } else {
+    // remote file
+    std::vector<std::byte> imageData = fetchUrlFunc(imageURI);
+    image = cv::imdecode(
+        cv::Mat(1, imageData.size(), CV_8UC1, (void *)imageData.data()),
+        cv::IMREAD_COLOR);
+  }
```
mkopcins, line 106: are we sure about this? shouldn't we validate this in any way? for some random input this would be ran, doesn't seem correct/safe to just fetch some random string

common/rnexecutorch/data_processing/base64.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+      for (i = 0; (i < 4); i++)
+        ret += base64_chars[char_array_4[i]];
+      i = 0;
+    }
+  }
+
+  if (i) {
+    for (j = i; j < 3; j++)
+      char_array_3[j] = '\0';
```
mkopcins: nit pick but please add {} in any statements that are not a single line, debugging errors like missing brackets is a pain in the ***

mkopcins: https://www.imperialviolet.org/2014/02/22/applebug.html

common/rnexecutorch/data_processing/base64.cpp
```diff
@@ -0,0 +1,101 @@
...
+        ((char_array_3[0] & 0x03) << 4) + ((char_array_3[1] & 0xf0) >> 4);
+    char_array_4[2] =
+        ((char_array_3[1] & 0x0f) << 2) + ((char_array_3[2] & 0xc0) >> 6);
+    char_array_4[3] = char_array_3[2] & 0x3f;
+
+    for (j = 0; (j < i + 1); j++)
+      ret += base64_chars[char_array_4[j]];
+
+    while ((i++ < 3))
+      ret += '=';
```
mkopcins: here too

src/modules/computer_vision/StyleTransferModule.ts
mkopcins: isn't `global.loadStyleTransfer` possibly null?

third-party/include/executorch/runtime/core/error.h
```diff
@@ -79,6 +79,12 @@ enum class Error : error_code_t {
   /// Error caused by the contents of a program.
   InvalidProgram = 0x23,
 
+  /// Error caused by the contents of external data.
+  InvalidExternalData = 0x24,
+
+  /// Does not have enough resources to perform the requested operation.
+  OutOfResources = 0x25,
+
```
mkopcins, line 87: add this to `src/Error.ts`

mkopcins (changes requested):

common/rnexecutorch/Log.cpp
msluszniak: `vsnprintf` is used from `<cstdio>`. No other functionality is used from heavier `<string>` so the change seems to be completely resonable.

common/rnexecutorch/data_processing/ImageProcessing.cpp
msluszniak: `std::getline` will fail if there are two or more consecutive separators.

msluszniak: So that's a good point, that we need to somehow check for some malformed uris.

common/rnexecutorch/data_processing/ImageProcessing.cpp
JakubGonera: It turns out that it's sufficient to throw a JSError with the runtime attached to make the exception readable in JS. The only place where we call the native code without a promise is `RnExecutorchInstaller`, so right now we catch there any exceptions and rethrow them as JSErorrs.

src/modules/computer_vision/StyleTransferModule.ts
JakubGonera: Indeed it could be. The question is whether we could do anything about this. Right now the user will have an exception thrown at startup with a message indicating this. If this is null then this means that all the native code went out the window.

common/rnexecutorch/data_processing/ImageProcessing.cpp
JakubGonera: Right now if the user supplies a string that has two commas next to each other we will pass an empty string to OpenCV which will return an empty image which in turn will result in an exception "Read image error: invalid argument" being thrown. The same holds for other cases where the payload does not represent a valid image. The exception will then result in a rejected promise.

I don't think that we should do any more work here on parsing. As far as I understand URI specification everything after the comma should be a base64 payload, and if the user is not conforming to this standard it shouldn't make any difference whether we fail while parsing a part of the intended payload or while parsing the whole.

common/rnexecutorch/data_processing/ImageProcessing.cpp
JakubGonera: Why not? If the url is wrong then the function will fail and result in an exception which will reject the promise, and otherwise OpenCV should be free to try to parse the bytes.

What kind of validation would you see here?

common/rnexecutorch/data_processing/ImageProcessing.cpp
msluszniak: SGTM

mkopcins (approved):

Issue #226, opened 2025-04-28
Allow users to manage conversation history in LLM


pweglik:
@chmjkb touched on important issue here:
https://github.com/software-mansion/react-native-executorch/pull/181#discussion_r2060147882

The plan is to  add another function to LLMController.
I think correct naming would be:
runInference - for the new function - the one that accepts Message[]
runInferenceRaw - new name for current function - the one that simply accepts string

PR #243, opened 2025-05-06
feat: Improved LLM public API 

pweglik:
Added generate function, and some little re…formats

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Related issues
 #226 

### Checklist


docs/docs/typescript-api/LLMModule.md
```diff
@@ -126,7 +128,9 @@ To subscribe to the download progress event, you can pass the `modelDownloadProg
 
 To run the model, you can use the `sendMessage` method. It accepts the user message. After model responds it will return new message history containing both user message and model response.. Additionally, it will call `messageHistoryCallback`.
 
-Alternatively, you can use `runInference`. It provides direct access to the model, without any wrapper, so the input string is passed straight into the model. If you're not sure what are implications of that, you're better off with `sendMessage`
+Alternatively, you can use `generate` method. It allows you to simply pass chat messages and receive completion from the model. It doesn't provide any message history management.
+
+If you need raw model, without any wrappers, you can use `forward`. It provides direct access to the model, so the input string is passed straight into the model. It may be useful to work with models that aren't finetuned for chat completions. If you're not sure what are implications of that, you're better off with `sendMessage`
```
NorbertKlockiewicz:
```suggestion
If you need raw model, without any wrappers, you can use `forward`. It provides direct access to the model, so the input string is passed straight into the model. It may be useful to work with models that aren't finetuned for chat completions. If you're not sure what are implications of that, you're better off with `sendMessage`.
```

examples/llm/App.tsx
```diff
@@ -8,7 +8,7 @@ import { useState } from 'react';
 import ColorPalette from './colors';
 import { View, StyleSheet, Text } from 'react-native';
 
-enum ModelType {
+enum ModeType {
```
NorbertKlockiewicz: Maybe just Mode instead of ModeType

src/controllers/LLMController.ts
```diff
@@ -141,14 +141,15 @@ export class LLMController {
...
-  public async runInference(input: string) {
+  public async forward(input: string) {
     if (!this._isReady) {
       throw new Error(getError(ETError.ModuleNotLoaded));
     }
     try {
       this.responseCallback('');
       this.isGeneratingCallback(true);
-      await this.nativeModule.runInference(input);
+      console.log('INPUT:', input);
```
NorbertKlockiewicz: Remove console.log

NorbertKlockiewicz:
Can you also bump huggingface/jinja to newest version(0.5.0), as the current has problems with qwen 3 tokenizer.

README.md
chmjkb, line 1: I believe that this is wrong, you're supposed to pass all the special tokens to forward(), but the above prompt definition doesn't do that. Change the call of this method to `generate()` (not sure if this one is ok)

docs/docs/typescript-api/LLMModule.md
chmjkb: I think you can mention here that you need to add special tokens to make it work

chmjkb (changes requested):

src/controllers/LLMController.ts
```diff
@@ -222,7 +231,7 @@ export class LLMController {
   }
 
   private applyChatTemplate(
-    messages: MessageType[],
+    messages: Message[],
     tokenizerConfig: any,
     tools?: LLMTool[],
     template_flags?: Object
```
chmjkb: Can you make `template_flags` camel case?

pweglik: Sorry, missed it, maybe we should enforce camel case with eslint? Can you create issue for it?

chmjkb: Yep, will do

NorbertKlockiewicz (approved):

PR #244, opened 2025-05-06
fix: fix exception handling in C++ native code

JakubGonera:
## Description

Fix exception handling in C++ native code so that it results in JS promise getting rejected with an appropriate message. There are two fixes here:
1. Runtime type information in the current version of React Native is disabled because one of its dependencies, which results in C++ not being able to recognize that `std::runtime_error` is inheriting from `std::exception` so to extract the exception message we need to match the type exactly. A fix has been introduced in RN repo, but not yet present in any release: https://github.com/facebook/react-native/commit/3132cc88dd46f95898a756456bebeeb6c248f20e.
2. When rejecting a promise the old code was accessing JS runtime in an unsafe way, this is fixed now.

### Type of change

### Tested on

- [x] iOS
- [x] Android

### DIFF
--- a/android/src/main/java/com/swmansion/rnexecutorch/ETInstaller.kt
+++ b/android/src/main/java/com/swmansion/rnexecutorch/ETInstaller.kt
@@ -21,16 +21,20 @@ class ETInstaller(
     @JvmStatic
     @DoNotStrip
     @Throws(Exception::class)
-    fun fetchByteDataFromUrl(source: String): ByteArray {
-      val url = URL(source)
-      val connection = url.openConnection()
-      connection.connect()
+    fun fetchByteDataFromUrl(source: String): ByteArray? {
+      try {
+        val url = URL(source)
+        val connection = url.openConnection()
+        connection.connect()
 
-      val inputStream: InputStream = connection.getInputStream()
-      val data = inputStream.readBytes()
-      inputStream.close()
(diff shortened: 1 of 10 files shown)


android/src/main/java/com/swmansion/rnexecutorch/ETInstaller.kt
```diff
@@ -21,16 +21,20 @@ class ETInstaller(
...
-      val data = inputStream.readBytes()
-      inputStream.close()
+        val inputStream: InputStream = connection.getInputStream()
+        val data = inputStream.readBytes()
+        inputStream.close()
 
-      return data
+        return data
+      } catch (exception: Throwable) {
+        return null
```
NorbertKlockiewicz, line 36: Why `return null` here, in objective c we throw an error, what happens here?

JakubGonera: In Objective C we are able to throw a C++ exception immidiately, here if we throw a java exception the application will not be able to handle this. This null is then tested in the immidiate C++ code and an exception is thrown there.

NorbertKlockiewicz: Okay, fair enough

mkopcins (approved):

PR #277, opened 2025-05-11
Bump EXPO version to 53

pweglik:
## Description

- [x] Release 0.3.3
- [ ] Let people know in issues asking about it


.github/workflows/build-ios-llama-example.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       group: ios-${{ github.ref }}
       cancel-in-progress: true
     steps:
+      - uses: maxim-lobanov/setup-xcode@v1
```
chmjkb, line 25: Why is it a part of this release?

app.json
chmjkb, line 1: do we need the app.json in the root?

.github/workflows/build-ios-llama-example.yml
pweglik: So the CI doesn't fail, but maybe it isn't needed in the patch. Same with examples changes. To fix the issue in the library, we only need to do changes in the library, but i think it's better to keep apps consistent with it on the same branch

app.json
pweglik: Yeah, that's weird

pweglik: Removed

Issue #264, opened 2025-05-07
Fix llm demo app crash


pweglik:
App is crashing if you change tabs without stopping inference and it's still running. Changing tabs should stop inference - calling `interrupt` should be enough

Issue #241, opened 2025-05-06
Start download of models after user agrees


pweglik:
See: https://github.com/software-mansion/react-native-executorch/issues/237#issuecomment-2852005285
Calling `useLLM` starts model download automatically. We want to prevent by either:
- making parent component call `load` function explicitly
- adding `preventDownload` props to the hook, so it will only start loading when component calling it allows it

Issue #237, opened 2025-05-05
Fix state initlialilzation in hooks using controllers


pweglik:
In `useLLM` and `useSpeechToText` we use hook props in state initialization function. This is bad practice and results in controller to initializing with new configuration.

Creating state for variables that can be recalculated using other states and/or props is bad too so OCR hooks should be reformatted as well probably.

List
- [x] useLLM 
- [x] Other hooks

kuriel-dev:
i noticed several problems with current useLLM implementation, it's not fully compatible with React sctructure (by re-renders)
it's very very hard to find a place where useLLM is not re-setted by a render, i would like to suggest some improvements:

`const model = useLLM();`

then in a better moment like useEffect() or onPress to switch models (by example)

```javascript
model.setSources({
    modelSource,
    tokenizerSource,
});
```

so we can also load messageHistory first since almost all loads are async:

```javascript
const messageHistory = await getMessagesFromChat();
model.setMessageHistory(messageHistory);
```

We also should find the best moment to ask the user to download model:

UI:  Hi user, we are gonna download the required resources, we recommend to switch to a wifi or unlimited data connection.
[Continue | Later (i'm offline or anything)]

```javascript
// nowadays the download starts no advice, once useLLM() is called :( no control
model.download({
    onProgress: (progress)=>//...
    onError: (error)=> //...
    onCancel: ()=> //... interrupt by connection lost or user 
    onComplete: () => //...
});

// user interrupt the download, nowadays the only way is closing the app
model.downloadCancel();
```

These are some suggestions, and I humbly request your consideration of these approaches to better control the process, specifically regarding implementation and UI/UX.

pweglik:
Hey @kuriel-dev thanks for feedback! The first issue you mentioned will be resolved when fixing this issue. In 0.4.0 we did big refactor of `useLLM`.
I'll create a separate issue for the second problem you mention. Calling `useLLM` won't start download automatically, user will need to call it themselves or will add argument to useLLM to prevent automatic download if you'd rather do it later - we need to discuss it with a team.
Thanks again for the feedback!

Issue #235, opened 2025-05-03
How to reset llama = useLLM() ?


kuriel-dev:
### Description

How do you guys reset llama = useLLM()  ?? 

when a llama.error exists and i choose another model in the same useLLM (i dont think i should use another new)

the llama.error persists across the other models, that makes my UI buggy, how to reset that value or if possibe, the llama value.

thanks.

### Steps to reproduce

1. select one model that produces error
2. select another model 

the error persists across the models.

### React Native Executorch version

^0.3.2

### React Native version

0.76.9

### Platforms

Android

### AI model

across default models

pweglik:
This should be fixed in 0.4.0 version releasing in few weeks. If you need this feature immediately, you can try `v0.4.0-rc` branch (note it can be unstable though).

PR #278, opened 2025-05-12
fix: Improvements and fixes in `useLLM`

pweglik:
## Description

Solving few issues:
- Example app crashed when model was deleted during generation. This is prevented now and we throw error if library users try it.
- state of hook and controller is properly reseted when changing models.
- `preventLoad` props in hook so you can postpone loading model until you're ready
- fixes in state management in hook

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Screenshots

Preventing model swap during generation:

https://github.com/user-attachments/assets/0cfec0e9-02fc-414c-ac3a-5f6e767ea7ea

### Related issues

https://github.com/software-mansion/react-native-executorch/issues/235
https://github.com/software-mansion/react-native-executorch/issues/237
https://github.com/software-mansion/react-native-executorch/issues/241
https://github.com/software-mansion/react-native-executorch/issues/264

### Checklist


docs/docs/natural-language-processing/useLLM.md
```diff
@@ -58,14 +53,12 @@ const useLLM: ({
   modelSource,
   tokenizerSource,
   tokenizerConfigSource,
-  chatConfig,
-  toolsConfig,
+  preventLoad = false,
```
chmjkb, line 56: Maybe preventAutomaticLoad or something like this sounds better?

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -129,9 +129,33 @@ Given computational constraints, our architecture is designed to support only on
 
 **`modelSource`** - A string that specifies the location of the model binary. For more information, take a look at [loading models](../fundamentals/loading-models.md) section.
 
-**`tokenizerSource`** - URL to the JSON file which contains the tokenizer
+**`tokenizerSource`** - URL to the JSON file which contains the tokenizer.
+
+**`tokenizerConfigSource`** - URL to the JSON file which contains the tokenizer config.
 
-**`tokenizerConfigSource`** - URL to the JSON file which contains the tokenizer config
+**`[Optional]preventLoad`** - flag providing a way to control when to start loading model (and more importantly - downloading model data, when you load it first time).
```
chmjkb: I think we should just add a questionmark at the end of the name of the parameter instead of [Optional].

chmjkb: Also I think this could just be:
`Whether to model is instantly downloaded / loaded after running the hook` or sth like this

examples/llm/App.tsx
```diff
@@ -55,20 +58,26 @@ export default function App() {
...
+                  highlightColor={ColorPalette.primary}
+                  wrapperBackground="#fff"
+                  highlightBorderWidth={3}
+                  itemHeight={40}
+                  activeItemTextStyle={styles.activeScrollItem}
+                />
+              </View>
+            ) : (
+              <Text style={styles.placeholderText}>
+                Model is generating. Interrupt to swap models!
```
chmjkb, line 78: Maybe we should handle it internally?

src/hooks/natural_language_processing/useLLM.ts
chmjkb, line 1: Not the scope of this PR, but we also need to apply this changes to the rest of the hooks.

docs/docs/natural-language-processing/useLLM.md
pweglik: Load is always automatic, that's why you're using hook. In my opinion makeing this argument's name longer doesn't really add explainability

docs/docs/natural-language-processing/useLLM.md
pweglik: I rephrased it a little bit

examples/llm/App.tsx
pweglik: This is an interesting issue. We can't simply do :
```
public delete() {
    this.interrupt();
    ...
  }
 ```
 Because all that `interrupt` does is setting flag in native runner. The flag is the checked in runner generation loop.
 We'd need to somehow await in `delete()` for `forward()` method to finish running, so it's safe to delete model (it sets `isGenerating` to false). I think it's generally fine for user's to handle it. We throw a meaningful error if someone's try to delete model when it runs.
 We could handle it, but it'd require some unelegant code in LLMController

Comment (chmjkb) [reply examples/llm/App.tsx:78]: I'm not sure why that is a problem, I mean if you call interrupt(), then the native runner will stop generating, then forward() will stop executing, meaning you can safely switch between models, right?

File: src/hooks/natural_language_processing/useLLM.ts
Comment (pweglik) [reply src/hooks/natural_language_processing/useLLM.ts:1]: Done: 
* #283

File: docs/docs/natural-language-processing/useLLM.md (hunk 1 above)
Comment (mkopcins) [reply docs/docs/natural-language-processing/useLLM.md:56]: I think the confusing part in this naming is the `load`, does this prevent loading the model to memory or downloading the model? I think changing this to `lazyLoad` or similar is much better

File: docs/docs/natural-language-processing/useLLM.md [hunk 4]
```diff
@@ -129,9 +129,33 @@ Given computational constraints, our architecture is designed to support only on
...
+| ------------------ | --------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `messageHistory`   | `Message[]`                                                                 | State of the generated response. This field is updated with each token generated by the model                                                                                                                                                                                                                                                                          |
```
mkopcins: is it tho? I think this is the state of the whole conversation

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -129,9 +129,33 @@ Given computational constraints, our architecture is designed to support only on
 
 **`modelSource`** - A string that specifies the location of the model binary. For more information, take a look at [loading models](../fundamentals/loading-models.md) section.
 
-**`tokenizerSource`** - URL to the JSON file which contains the tokenizer
+**`tokenizerSource`** - URL to the JSON file which contains the tokenizer.
+
+**`tokenizerConfigSource`** - URL to the JSON file which contains the tokenizer config.
 
-**`tokenizerConfigSource`** - URL to the JSON file which contains the tokenizer config
+**`preventLoad?`** - Boolean that can prevent automatic model loading (and downloading the data if you load it for the first time) after running the hook
```
NorbertKlockiewicz:
```suggestion
**`preventLoad?`** - Boolean that can prevent automatic model loading (and downloading the data if you load it for the first time) after running the hook.
```

docs/docs/natural-language-processing/useLLM.md
```diff
@@ -129,9 +129,33 @@ Given computational constraints, our architecture is designed to support only on
...
+
+### Returns
```
NorbertKlockiewicz, line 138: In the table the sentences in Description column sometimes end with "." and sometimes don't. Make it the same for each.

mkopcins (changes requested):

docs/docs/natural-language-processing/useLLM.md
pweglik: `lazyLoad` is event more confusing term, as this flag doesn't enable some kind of lazy loading, it blocks any kind of model loading. 

It prevents both downloading and loading model. It is also explained in docs

docs/docs/natural-language-processing/useLLM.md
pweglik: Wow, huge fuckup on my part

docs/docs/natural-language-processing/useLLM.md
mkopcins: right, thought about this too after writing this comment, let's leave it as it is then

PR #305, opened 2025-05-20
fix: Make generate respect chatConfig within LLMModule

chmjkb:
### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist

### DIFF
--- a/src/modules/natural_language_processing/LLMModule.ts
+++ b/src/modules/natural_language_processing/LLMModule.ts
@@ -63,7 +63,7 @@ export class LLMModule {
   }
 
   static async deleteMessage(index: number): Promise<Message[]> {
-    await this.controller.deleteMessage(index);
+    this.controller.deleteMessage(index);
     return this.controller.messageHistory;
   }
 
--- a/src/modules/natural_language_processing/LLMModule.ts
+++ b/src/modules/natural_language_processing/LLMModule.ts
@@ -62,7 +62,7 @@ export class LLMModule {
     return this.controller.messageHistory;
   }
 
-  static async deleteMessage(index: number): Promise<Message[]> {
+  static deleteMessage(index: number): Message[] {
     this.controller.deleteMessage(index);
     return this.controller.messageHistory;
   }
--- a/src/controllers/LLMController.ts
+++ b/src/controllers/LLMController.ts
@@ -185,8 +185,15 @@ export class LLMController {
     if (!this._isReady) {
       throw new Error(getError(ETError.ModuleNotLoaded));
     }
+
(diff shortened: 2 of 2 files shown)


src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -62,8 +62,8 @@ export class LLMModule {
     return this.controller.messageHistory;
   }
 
-  static async deleteMessage(index: number): Promise<Message[]> {
-    await this.controller.deleteMessage(index);
+  static deleteMessage(index: number): Message[] {
+    this.controller.deleteMessage(index);
```
pweglik, line 66: Change this type in docs as well

pweglik (changes requested):
One thing I thought about now is: 
- we can configure chat and then use generate, but generate only uses systemPrompt and contextWindowLenght, but doesn't care about initial message history. Maybe we should mark it in docs? 
It feels like we're creating a weird case, where some of config is used and some is not.
Maybe we should keep chatConfig only to managed chat feature (the one that works with `sendMessage`) and require user to include prompt themselves when they're using chat completions (similarly to Open AI: https://platform.openai.com/docs/guides/text?api-mode=chat#message-roles-and-instruction-following)

chmjkb:
> One thing I thought about now is:
> 
> * we can configure chat and then use generate, but generate only uses systemPrompt and contextWindowLenght, but doesn't care about initial message history. Maybe we should mark it in docs?
>   It feels like we're creating a weird case, where some of config is used and some is not.
>   Maybe we should keep chatConfig only to managed chat feature (the one that works with `sendMessage`) and require user to include prompt themselves when they're using chat completions (similarly to Open AI: https://platform.openai.com/docs/guides/text?api-mode=chat#message-roles-and-instruction-following)

Does OpenAI require to pass system prompts as well within the message history? If so then its fine but we need to make sure to mention in the docs the fact that the `chatConfig` is discarded when using generate.

I'm also wondering if it's even worth to maintain the state internally, or just offload it to the user.

pweglik:
Yeah, when you're doing state completions (using Chat Completions API - not new Responses API). You need to pass system prompt as message with "system" role (recently changed to "developer" role I think).

I think, the state management should be offloaded to separate component that would allow users to use chat with llm feature instantly, as this was our goal: to make it easier. But we've decided against additional hooks and it was placed in LLMController as internal state + sendMessage method. It's okay solution, if we keep forward and generate methods stateless. Only sendMessage would interact with state in this case.

pweglik:
Closing as #309 implements it

PR #330, opened 2025-05-23
feat: monorepo

NorbertKlockiewicz:
## Description

This PR introduces new structure of repository:
- apps directory - contains standalone applications which are using package from `packages` directory
- packages - contains react native executorch library
### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


.eslintrc.js
pweglik, line 1: We're missing camelcase rule here

pweglik, line 1: We're missing a camelcase rule here

tsconfig.json
pweglik, line 1: Do we need this? Should it look like this? I can see it looks different here:
[json](https://github.com/software-mansion/react-native-reanimated/blob/main/tsconfig.json)
and here:
https://github.com/cdierkens/yarn-monorepo-example

.eslintrc.js
```diff
@@ -0,0 +1,33 @@
...
+      },
+    ],
+    '@cspell/spellchecker': [
+      'warn',
+      {
+        customWordListFile: path.resolve(__dirname, '.cspell-wordlist.txt'),
+      },
+    ],
+  },
+  plugins: ['prettier'],
```
pweglik, line 36: it used to be:
```
 "plugins": [
        "eslint-plugin-prettier"
    ],
```

I'm not sure which way is correct, but this suggests that neither :D 
https://www.npmjs.com/package/eslint-plugin-prettier/v/5.4.0#configuration-legacy-eslintrc

.yarnrc.yml
```diff
@@ -1,10 +1,7 @@
-nodeLinker: node-modules
-nmHoistingLimits: workspaces
+compressionLevel: mixed
```
pweglik, line 1: Are we sure we want `mixed`?
https://yarnpkg.com/configuration/yarnrc#compressionLevel
It suggests that default 0 results in faster install times

.yarnrc.yml
```diff
@@ -1,10 +1,7 @@
...
+
+enableGlobalCache: false
```
pweglik, line 3: Isn't it just going to waste space? Why do we need this?

.yarnrc.yml
pweglik, line 1: I've did some reading: https://yarnpkg.com/features/caching
I think in our case we can go with global caching as we're not going to commit yarn cache anyway.

README.md
```diff
@@ -10,15 +10,16 @@
...
+- [🦙 **Quickstart - Running Llama**](#-quickstart---running-llama)
+  - [1️⃣ **Installation**](#1️⃣-installation)
+  - [2️⃣ **Setup \& Initialization**](#2️⃣-setup--initialization)
+  - [3️⃣ **Run the model!**](#3️⃣-run-the-model)
+  - [Minimal supported versions](#minimal-supported-versions)
+  - [Examples 📲](#examples-)
+    - [Warning](#warning)
+  - [License](#license)
+  - [What's next?](#whats-next)
+  - [React Native ExecuTorch is created by Software Mansion](#react-native-executorch-is-created-by-software-mansion)
```
pweglik: I don't get it? It creates hierarchical table of contents and looks weird. Maybe decrease header leavel of Quickstart - Running Llama to 2 (##) instead of 1 (#)? And leave ToC as it is: linking to all 2 level headers

package.json
```diff
@@ -1,165 +1,30 @@
...
-  },
-  "dependencies": {
-    "@huggingface/jinja": "^0.5.0",
-    "expo": "^53.0.7",
-    "expo-asset": "~11.1.5",
-    "expo-file-system": "~18.1.10",
-    "react-native-live-audio-stream": "^1.1.1"
+    "prettier": "^3.3.3",
+    "prettier-plugin-jsdoc": "^1.3.0",
+    "typescript": "~5.3.0"
```
pweglik: Let's bump it, most of our apps are using higher version anyway

lefthook.yml
```diff
@@ -19,6 +19,3 @@ pre-commit:
     format-other:
       glob: '*.{json,yml,watchmanconfig,md,css}'
       run: npx prettier --write {staged_files} && git add {staged_files}
-    spell-check-other:
-      glob: '*.{md,css}'
-      run: cspell {staged_files} --config .cspell.json --quiet && git add {staged_files}
```
pweglik: Why are we removing it?

packages/react-native-executorch/package.json
```diff
@@ -0,0 +1,153 @@
+{
+  "name": "react-native-executorch",
+  "version": "0.4.2",
+  "description": "An easy way to run AI models in react native with ExecuTorch",
```
pweglik:
```suggestion
  "description": "An easy way to run AI models in React Native with ExecuTorch",
```

packages/react-native-executorch/package.json
pweglik, line 1: It feels liek there's a lot of unnecessary configs in this file that we don't use

packages/react-native-executorch/package.json
```diff
@@ -0,0 +1,153 @@
...
+    "example"
+  ],
+  "packageManager": "yarn@4.1.1",
+  "jest": {
+    "preset": "react-native",
+    "modulePathIgnorePatterns": [
+      "<rootDir>/example/node_modules",
+      "<rootDir>/lib/"
+    ]
+  },
```
pweglik: Do we even use jest? Maybe we should remove it until we actually have some unit testing

packages/react-native-executorch/package.json
```diff
@@ -0,0 +1,153 @@
...
+  "commitlint": {
+    "extends": [
+      "@commitlint/config-conventional"
+    ]
+  },
```
pweglik: Again, we're not using it. Maybe we should start? But for now, let's delete it from here and from deps if it's unused

packages/react-native-executorch/package.json
```diff
@@ -0,0 +1,153 @@
...
+    },
+    "github": {
+      "release": true
+    },
+    "plugins": {
+      "@release-it/conventional-changelog": {
+        "preset": "angular"
+      }
+    }
+  },
```
pweglik: Same here, for last publishes I've just used `npm publish`. I think we don't use it

packages/react-native-executorch/package.json
```diff
@@ -0,0 +1,153 @@
...
+          "project": "tsconfig.json"
+        }
+      ]
+    ]
+  },
+  "create-react-native-library": {
+    "type": "module-new",
+    "languages": "kotlin-objc",
+    "version": "0.41.2"
+  },
```
pweglik: Do we need it here?

pweglik (changes requested):
Mostly nitpicks and clarifications in configs

.yarnrc.yml
NorbertKlockiewicz: Hmm, however with the lower compression level the project will weigh more, with `compressionLevel` set to 0 the weigh of added packages is 964 MB and with `mixed` it's 326 MB, also I didn't notice any improvement on the side of installation time(probably matters more in CI) but I think we should stick with mixed

.yarnrc.yml
NorbertKlockiewicz: right, it was taken from reanimated XD

README.md
NorbertKlockiewicz: It's how markdown works, at least in my editor, I think that msluszniak's PR fixes it, basically the warning is treated as a subsection of example because Examples starts with ## and warning starts with ###

.yarnrc.yml
NorbertKlockiewicz: I tested it without enableGlobalCache: false and every yarn install was very long so I guess we should leave it as it is

tsconfig.json
NorbertKlockiewicz: I was trying to configure it like in the reanimated and I got tons of problems cause our apps are using expo ts config as a base and there was a lot of conflicting rules between apps and package so I decided to do it this way. Probably can be done better but currently I have no idea on how to improve it. I will look more into it now but I also don't want to spent too much time on optimizing typescript config when it works fine.

Issue #254, opened 2025-05-07
Notify the garbage collector of the size of loaded models in C++ native code


JakubGonera:
Notify the garbage collector of the size of loaded models in C++ native code. Host objects' memory is managed by the JS runtime garbage collector, which needs to know the real size of the objects to use a correct strategy when to garbage collect them. This means we need to set the size of a loaded model in JSI (via external memory pressure) to guarantee a proper memory behaviour.

JakubGonera:
The intended solution:
- Set the memory size as the size of the .pte file (this is an underestimate)
- Provide a method to unload the ET module from the memory

PR #337, opened 2025-05-26
feat: add memory pressure to models and allow freeing their memory

JakubGonera:
## Description

The garbage collector doesn't see the memory allocated on the heap by host objects, so we need to notify it of the memory occupied by ET modules, otherwise it will take a long time before freeing it. We set the memory pressure to the size of the .pte file of the model (an underestimate of the true size) and add a host function to release the module memory.

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Related issues

#254

### Checklist


common/rnexecutorch/models/BaseModel.cpp
```diff
@@ -40,4 +51,15 @@ std::vector<std::vector<int32_t>> BaseModel::getInputShape() {
...
 
+std::size_t BaseModel::getMemoryLowerBound() { return memorySizeLowerBound; }
+
+void BaseModel::unloadModule() { module.reset(nullptr); }
+
+Result<std::vector<EValue>> BaseModel::forwardET(const EValue &input_value) {
+  if (!module) {
+    throw std::runtime_error("Forward called on unloaded model");
+  }
+  return std::move(module->forward(input_value));
```
msluszniak, line 62: What is the reason of using `std::move` on returned value here?

JakubGonera: Indeed, it's unnecessary. This is fixed in a cherry-pick PR: https://github.com/software-mansion/react-native-executorch/pull/362

Issue #209, opened 2025-04-24
Refactor ExecuTorch bindings


pweglik:
Currently the forward function of useExecutorchModule accepts the following:
```typescript
forward: (   input: ETInput[],   shape: number[][] | number[] )
```
This is pretty confusing and hard to work with for the developer using it and for us. This task aims to change the API to resemble how ExecuTorch tensors work, meaning that we would pass an array of Tensors to the forward function. Each Tensor would have its own dataPtr, scalar type, and shape.

PR #345, opened 2025-05-28
feat: port + refactor ExecutorchModule to C++

chmjkb:
## Description

This PR aims to replace the existing ExecuTorch bindings for ones that leverage JSI and resemble the underlying runtime more accurately. Currently WIP. 

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Related issues

#209 
#338 


packages/react-native-executorch/common/rnexecutorch/utils/TypeConstraints.h
JakubGonera, line 1: 1. Maybe `TypeConcepts.h` for the file name would be more descriptive?
2. Could we think of some alternatives to having this in `utils/` folder? `utils/` usually ends up as a garbage for things that we don't know where to put, and so you don't know what to expect there (that was also the reason behind `data_processing/` instead of `utils/).

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -8,25 +8,34 @@
...
 template <typename Model> class ModelHostObject : public JsiHostObject {
 public:
   explicit ModelHostObject(const std::shared_ptr<Model> &model,
                            std::shared_ptr<react::CallInvoker> callInvoker)
-      : model(model), callInvoker(callInvoker) {
-    addFunctions(JSI_EXPORT_FUNCTION(ModelHostObject<Model>,
-                                     promiseHostFunction<&Model::forward>,
-                                     "forward"));
-    if constexpr (DerivedFromBaseModel<Model>) {
+      : JsiHostObject(), model(model), callInvoker(callInvoker) {
```
JakubGonera:
```suggestion
      : model(model), callInvoker(callInvoker) {
```

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <string>
+
+#include <ReactCommon/CallInvoker.h>
+#include <executorch/extension/module/module.h>
+#include <filesystem>
```
JakubGonera: Move to *.cpp, not used in the header.

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.h
```diff
@@ -0,0 +1,36 @@
...
+#include <jsi/jsi.h>
+#include <rnexecutorch/jsi/OwningArrayBuffer.h>
+#include <rnexecutorch/utils/JsiTensorView.h>
+
+using namespace facebook;
```
JakubGonera: Don't pollute the global scope and move inside `namespace rnexecutorch {...`.

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
```diff
@@ -15,9 +15,9 @@ class BaseModel {
 public:
   BaseModel(const std::string &modelSource,
             std::shared_ptr<react::CallInvoker> callInvoker);
-  std::vector<std::vector<int32_t>> getInputShape();
   std::size_t getMemoryLowerBound();
   void unload();
+  std::vector<std::vector<int32_t>> getAllInputShapes();
```
JakubGonera: Maybe `getForwardInputShapes` would be a better name, a bit long though.

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.h
```diff
@@ -0,0 +1,36 @@
...
+  // possibly we need to make ExecutorchModule a BaseModel.
+  std::size_t getMemoryLowerBound();
+  void unload();
+
+  ExecutorchModule(const std::string &modelSource,
+                   std::shared_ptr<facebook::react::CallInvoker> callInvoker);
+  std::vector<int32_t> getInputShape(std::string method_name, int index);
+  std::vector<std::shared_ptr<OwningArrayBuffer>>
+  forward(std::vector<JsiTensorView> tensorViewVec);
+  std::size_t memorySizeLowerBound{0};
```
JakubGonera: Could be a protected member, or even private (`BaseModel` has this protected, but I think that's wrong and it should be private).

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.cpp
```diff
@@ -0,0 +1,98 @@
+#include "ExecutorchModule.h"
+
+#include <executorch/extension/module/module.h>
+#include <rnexecutorch/Log.h>
+#include <sstream>
+
+namespace rnexecutorch {
+
+using ::executorch::extension::Module;
```
JakubGonera: This is redundant if you have `using namespace executorch::extension;`.

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.h
```diff
@@ -0,0 +1,36 @@
...
+  // note: this needs to be refactored due to the fact that both BaseModel and
+  // ExecuTorch module implement this separately, which is pretty error-prone.
+  // possibly we need to make ExecutorchModule a BaseModel.
+  std::size_t getMemoryLowerBound();
+  void unload();
+
+  ExecutorchModule(const std::string &modelSource,
+                   std::shared_ptr<facebook::react::CallInvoker> callInvoker);
+  std::vector<int32_t> getInputShape(std::string method_name, int index);
+  std::vector<std::shared_ptr<OwningArrayBuffer>>
```
JakubGonera: You probably need to pass information about dimensions of the return tensors? Something like:
```cpp
struct JSTensor {
  std::shared_ptr<OwningArrayBuffer> dataBuffer;
  std::vector<uint32_t> dimensions;
}
```
and then return:
```cpp
std::vector<JSTensor> forward(...);
```

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.cpp
```diff
@@ -0,0 +1,98 @@
...
+std::vector<std::shared_ptr<OwningArrayBuffer>>
+ExecutorchModule::forward(std::vector<JsiTensorView> tensorViewVec) {
+  std::vector<executorch::runtime::EValue> evalues;
+  evalues.reserve(tensorViewVec.size());
+  // Because EValue doesn't hold to the dynamic data and metadata from
+  // TensorPtr, we need to make sure that the TensorPtr for each EValue is valid
+  // as long as that EValue is in use. Therefore we create a vec solely for
+  // keeping references to the TensorPtr
+  std::vector<TensorPtr> tensorPtrs;
+  tensorPtrs.reserve(evalues.size());
```
JakubGonera: You can just store tensor pointers and they will be automatically converted to EValues when passed to `forward`. See `executorch/runtime/core/evalue.h`:
```cpp
  /****** Tensor Type ******/
  /*implicit*/ EValue(executorch::aten::Tensor t) : tag(Tag::Tensor) {
    // When built in aten mode, at::Tensor has a non trivial constructor
    // destructor, so regular assignment to a union field is UB. Instead we must
    // go through placement new (which causes a refcount bump).
    new (&payload.as_tensor) executorch::aten::Tensor(t);
  }

  // Template constructor that allows construction from types that can be
  // dereferenced to produce a type that EValue can be implicitly constructed
  // from.
  template <typename T,
            typename = typename std::enable_if<std::is_convertible<
                decltype(*std::forward<T>(std::declval<T>())), // declval to
                                                               // simulate
                                                               // forwarding
                EValue>::value>::type>
  /*implicit*/ EValue(T &&value) {
    ET_CHECK_MSG(value != nullptr, "Pointer is null.");
    // Note that this ctor does not initialize this->tag directly; it is set by
    // moving in the new value.
    moveFrom(*std::forward<T>(value));
  }
```

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+  for (size_t i = 0; i < tensorViewVec.size(); i++) {
+    const auto &currTensorView = tensorViewVec[i];
+    auto tensorPtr = make_tensor_ptr(currTensorView.shape,
+                                     currTensorView.dataPtr, ScalarType::Float);
```
JakubGonera: Should this always be `ScalarType::Float`?

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.cpp
```diff
@@ -0,0 +1,98 @@
...
+                                                     int index) {
+  auto method_meta = module->method_meta(method_name);
+  if (!method_meta.ok()) {
+    throw std::runtime_error("Failed to load method with name " + method_name);
+  }
+
+  std::vector<int32_t> input_shape;
+  auto input_meta = method_meta->input_tensor_meta(index);
+  if (!input_meta.ok()) {
+    throw std::runtime_error("Failed to load forward input " +
```
JakubGonera: This no longer needs to be the `forward` method input.

packages/react-native-executorch/common/rnexecutorch/utils/JsiTensorView.h
JakubGonera, line 1: This should go to either `jsi/` or to `host_objects/` or even something else, but this is definitely not a util. Maybe we could rename `host_objects/` to something else, because `ModelHostObject.h` might end up being the only actual host object file and at the same time I feel like we will need more files that are somewhat connected to our business logic but are mainly regarding JSI stuff.

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -30,6 +39,79 @@ inline std::string getValue<std::string>(const jsi::Value &val,
...
 }
 
+template <>
+inline JsiTensorView getValue<JsiTensorView>(const jsi::Value &val,
+                                             jsi::Runtime &runtime) {
+  jsi::Object obj = val.asObject(runtime);
+  JsiTensorView tensorView;
+
+  int scalarTypeInt =
+      getValue<int>(obj.getProperty(runtime, "scalarType"), runtime);
```
JakubGonera: NIT: you could use bare .asNumber() instead of getValue<int>, but feel free to ignore.

packages/react-native-executorch/src/index.tsx
```diff
@@ -7,16 +7,19 @@ declare global {
...
+  var loadExecutorchModule: (source: string) => any;
 }
 // eslint-disable no-var
-
-if (global.loadStyleTransfer == null) {
+if (
+  global.loadStyleTransfer == null ||
+  global.loadImageSegmentation == null ||
+  global.loadExecutorchModule == null
+) {
```
JakubGonera: This skips `loadClassification`.

packages/react-native-executorch/src/modules/general/NewExecutorchModule.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  async forward(inputTensor: TensorPtr[]): Promise<ArrayBuffer[]> {
+    return await this.nativeModule.forward(inputTensor);
+  }
+
+  async getInputShape(methodName: string, index: number): Promise<number[]> {
+    return this.nativeModule.getInputShape(methodName, index);
+  }
+}
```
JakubGonera: You want to implement `delete()` that calls `unload()` as well.

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
chmjkb: I think it makes sense to just leave the naming as is and pass a methodName param instead, so we can reuse this method if we ever have to.

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.cpp
```diff
@@ -0,0 +1,100 @@
...
+                             std::to_string(index));
+  }
+
```
msluszniak: Move creating `input_shape` after checking if `input_meta` is ok, and reserve `input_meta->sizes()` space for it.

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -19,6 +23,11 @@ inline double getValue<double>(const jsi::Value &val, jsi::Runtime &runtime) {
   return val.asNumber();
 }
 
+template <>
```
msluszniak: Consider replacing the specialization for int and double with 
```
template<typename T>
concept IsNumeric = std::is_arithmetic<T>::value;

template <typename T>
requires IsNumeric<T>
inline T getValue(const jsi::Value &val, jsi::Runtime &runtime) {
    static_assert(std::is_integral<T>::value || std::is_floating_point<T>::value,
                  "Only integral and floating-point types are supported");
    return static_cast<T>(val.asNumber());
}
```
This will enable for all floating and numeric which I don't know if it's desired, so feel free to skip it.

packages/react-native-executorch/common/rnexecutorch/bindings/ExecutorchModule.cpp
JakubGonera: As it has been pointed to me out, we use a vector on the outside here, so this is not a possibility.

packages/react-native-executorch/common/rnexecutorch/host_objects/JSTensorView.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace rnexecutorch {
+
+using executorch::aten::ScalarType;
+
+struct JSTensorView {
+  void *dataPtr;
+  ScalarType scalarType;
+  std::vector<int32_t> shape;
```
JakubGonera: We probably should name this `JSTensorViewIn` and unify the shape/sizes names with `JSTensorViewOut`.

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -30,6 +39,78 @@ inline std::string getValue<std::string>(const jsi::Value &val,
...
+                                           jsi::Runtime &runtime) {
+  jsi::Object obj = val.asObject(runtime);
+  JSTensorView tensorView;
+
+  int scalarTypeInt = obj.getProperty(runtime, "scalarType").asNumber();
+  tensorView.scalarType = static_cast<ScalarType>(scalarTypeInt);
+
+  jsi::Value shapeValue = obj.getProperty(runtime, "shape");
+  jsi::Array shapeArray = shapeValue.asObject(runtime).asArray(runtime);
+  size_t shapeDims = shapeArray.size(runtime);
```
JakubGonera: NIT: `shapeDims` -> `numShapeDims`

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.cpp
```diff
@@ -28,38 +28,140 @@ BaseModel::BaseModel(const std::string &modelSource,
...
+BaseModel::getTensorShape(const executorch::aten::Tensor &tensor) {
+  std::vector<int32_t> tensorShape;
+  auto sizes = tensor.sizes();
+  tensorShape.reserve(sizes.size());
+  for (auto size : sizes) {
+    tensorShape.push_back(static_cast<int32_t>(size));
   }
-  return module->forward(input_value);
+  return tensorShape;
 }
```
JakubGonera: Alternative:
```cpp
std::vector<int32_t>
BaseModel::getTensorShape(const executorch::aten::Tensor &tensor) {
  auto sizes = tensor.sizes();
  return std::vector<int32_t>(sizes.begin(), sizes.end());
}
```

packages/react-native-executorch/src/modules/BaseNonStaticModule.ts
```diff
@@ -1,5 +1,29 @@
...
+    this.nativeModule = global.loadExecutorchModule(paths[0] || '');
+  }
+
+  protected async forwardET(inputTensor: TensorPtr[]): Promise<TensorPtr[]> {
+    return await this.nativeModule.forward(inputTensor);
+  }
+
+  async getInputShape(methodName: string, index: number): Promise<number[]> {
+    return this.nativeModule.getInputShape(methodName, index);
+  }
```
JakubGonera: Two thoughts on this setup:
1. The `load` method contains logic that would only be used if used with `ExecutorchModule` (correct me if I'm wrong), so we don't gain anything on having this code here compared to in `ExecutorchModule`.
2. Adding `forwardET` to the base class requires each module inheriting to be a single model, so it excludes things like OCR which will not have a single `forwardET` method we could invoke. Maybe we should inherit from `ExecutorchModule` where possible (e.g. `StyleTransfer`), and have this be a true base class for all modules?

JakubGonera (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.cpp
```diff
@@ -50,7 +49,10 @@ std::vector<int32_t> BaseModel::getInputShape(std::string method_name,
         std::to_string(static_cast<int>(input_meta.error())));
   }
 
-  for (auto size : input_meta->sizes()) {
+  std::vector<int32_t> input_shape;
+  auto sizes = input_meta->sizes();
+  input_shape.reserve(sizes.size());
+  for (auto size : sizes) {
```
msluszniak: Sorry that I didn't add this comment at first place, but now you can create `input_shape` based on sizes namely:
```cpp
std::vector<int32_t> input_shape(sizes.begin(), sizes.end());
```

packages/react-native-executorch/src/modules/BaseNonStaticModule.ts
chmjkb: 1. That's correct, this setup was done strictly for testing purposes and I forgot to change it
2. Yeah, that makes sense but I'm not sure if this can be done now. Currently the native API accepts one source arg on load. So either we refactor the native API and then make the changes you mentioned, or we can do something like this:
```javascript
class EncoderDecoderModel { // doesn't inherit from base, kind of like a controller 
  constructor(encoderSource, decoderSource) {
    // they do inherit from base
    this.encoder = new SomeEncoder(encoderSource) 
    this.decoder = new SomeDecoder(decoderSource)
  }
}
```

JakubGonera (approved):

Issue #255, opened 2025-05-07
Enable native C++ and port current functionalities


JakubGonera:

darthez:
Thank you @JakubGonera 👍

PR #365, opened 2025-06-04
docs: update C++ changes

JakubGonera:
## Description

Change documentation that is influenced by C++ ports

### Type of change

### Related issues

#255


msluszniak:
@JakubGonera Why every single package version in `yarn.lock` now is prefixed with `npm`? This is the same as in the monorepo `yarn.lock`, so I guess it is correct, but I cannot get any information about this on the Internet.

JakubGonera:
> @JakubGonera Why every single package version in `yarn.lock` now is prefixed with `npm`? This is the same as in the monorepo `yarn.lock`, so I guess it is correct, but I cannot get any information about this on the Internet.

I don't know, but when you run yarn now it changes it automatically, and without it the CI fails because it also wants to modify yarn.lock, but is not allowed.

msluszniak (approved):
I checked the typescript API and the changed documentation correctly describes them now.

chmjkb (approved):

PR #375, opened 2025-06-05
fix: Added conversion from BGR to RGB in vision model. Images loaded with …

pweglik:
Added conversion from BGR to RGB in vision model. Images laoded with OpenCV are in BGR and we ignored it until now.

### Type of change

### Tested on

- [x] iOS
- [ ] Android


pweglik:
I think, it is not available in C++ unfortunately

PR #381, opened 2025-06-09
Enable passing extra args to load function in classes derived from BaseModule

pweglik:
### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

Tested in CV object detection - it works

### Checklist

### DIFF
--- a/packages/react-native-executorch/src/modules/BaseModule.ts
+++ b/packages/react-native-executorch/src/modules/BaseModule.ts
@@ -7,13 +7,16 @@ export class BaseModule {
   static onDownloadProgressCallback: (downloadProgress: number) => void =
     () => {};
 
-  static async load(...sources: ResourceSource[]): Promise<void> {
+  static async load(
+    sources: ResourceSource[],
+    ...loadArgs: any[] // this can be used in derived classes to pass extra args to load method
+  ): Promise<void> {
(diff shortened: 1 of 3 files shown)


packages/react-native-executorch/src/modules/BaseModule.ts
```diff
@@ -7,13 +7,16 @@ export class BaseModule {
...
+    sources: ResourceSource[],
+    ...loadArgs: any[] // this can be used in derived classes to pass extra args to load method
+  ): Promise<void> {
     try {
       const paths = await ResourceFetcher.fetchMultipleResources(
         this.onDownloadProgressCallback,
         ...sources
       );
-      await this.nativeModule.loadModule(...paths);
+      await this.nativeModule.loadModule(...paths, ...loadArgs);
```
mkopcins, line 19: I dont think nativeModule accepts `loadArgs`

pweglik: Well, in some cases it can? What if we want to pass an extra argument to the native side load function? I see no downside. When you call this in derived class:
```
await super.load([modelSource, tokenizerSource], someArg);
```
you can expect that everything goes to the native side as it should

PR #411, opened 2025-06-18
feat: port TextEmbeddings to C++

chmjkb:
## Description

This PR implements TextEmbeddings Module in C++, following a bigger effort of porting all the modules to C++. The hook now returns `Float32Array` instead of `number[]` for perf reasons. This might be refactored in the future if we want to support models in FP16 or other types. The PR also gets rid of some dead code from `TokenizerModule`, `ETModule` and now `TextEmbeddingsModule`.

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
@@ -1,6 +1,5 @@
 #include "TextEmbeddings.h"
 #include <executorch/extension/tensor/tensor_ptr_maker.h>
-#include <rnexecutorch/Log.h>
 #include <rnexecutorch/data_processing/Numerical.h>
 
 namespace rnexecutorch {


packages/react-native-executorch/common/rnexecutorch/data_processing/Numerical.cpp
```diff
@@ -16,4 +17,53 @@ void softmax(std::vector<float> &v) {
...
+  float sum = 0.0f;
+  for (float &x : v) {
+    sum += x * x;
+  }
+  // Preventing division by 0
+  float norm = std::max(std::sqrt(sum), 1e-9f);
+  for (float &x : v) {
+    x /= norm;
+  }
+}
```
msluszniak: I think that you can safely delete this function and use only version for `std::span` since for vector elements are continuous in memory. Hence span will be easy-constructible from vector. Try it by deleting this function

packages/react-native-executorch/common/rnexecutorch/data_processing/Numerical.cpp
```diff
@@ -16,4 +17,53 @@ void softmax(std::vector<float> &v) {
...
+
+std::vector<float> meanPooling(std::span<const float> modelOutput,
+                               std::span<const int64_t> attnMask) {
+  auto attnMaskLength = attnMask.size();
+  auto embeddingDim = modelOutput.size() / attnMaskLength;
+
+  float maskSum = 0;
+  for (auto &v : attnMask) {
```
msluszniak:
```suggestion
  for (const auto &v : attnMask) {
```

packages/react-native-executorch/common/rnexecutorch/data_processing/Numerical.cpp
```diff
@@ -16,4 +17,53 @@ void softmax(std::vector<float> &v) {
...
+  }
+  maskSum = std::max(maskSum, 1e-9f);
+
+  auto result = std::vector<float>();
+  result.reserve(embeddingDim);
+  for (size_t i = 0; i < embeddingDim; i++) {
+    float dimensionSum = 0;
+    for (size_t j = 0; j < attnMaskLength; j++) {
+      dimensionSum +=
+          (modelOutput[j * embeddingDim + i] * static_cast<float>(attnMask[j]));
```
msluszniak:
```suggestion
          modelOutput[j * embeddingDim + i] * static_cast<float>(attnMask[j]);
```

packages/react-native-executorch/common/rnexecutorch/data_processing/Numerical.cpp
chmjkb: wow, that's cool, thanks!

packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+TextEmbeddings::TextEmbeddings(const std::string &modelSource,
+                               const std::string &tokenizerSource,
+                               std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker),
+      tokenizer(
+          std::make_unique<TokenizerModule>(tokenizerSource, callInvoker)),
+      inputShapes(getAllInputShapes()) {}
+
+std::pair<std::vector<int64_t>, std::vector<int64_t>>
```
msluszniak: Maybe alias this type or even create a simple struct here. Right now you need to know that inputIds are the first element of pair and attentionMask is the second element. If you will refer to them via `<struct_name>.inputIds` it would be easier follow the code without much scrolling

packages/react-native-executorch/src/modules/natural_language_processing/TextEmbeddingsModule.ts
pweglik, line 1: I think we should stick to previous solution where meanPooling is passed through load function. It allows us to pass it with preset model and tokenizer and user won't need to remember about it (they may not even know whether their model uses meanPooling or not)

packages/react-native-executorch/src/hooks/natural_language_processing/useTextEmbeddings.ts
pweglik, line 1: Analogous to Module change  - meanPooling should be passed to hook

packages/react-native-executorch/src/modules/natural_language_processing/TextEmbeddingsModule.ts
```diff
@@ -1,28 +1,30 @@
...
 
-  static override async forward(input: string): Promise<number[]> {
-    return this.nativeModule.forward(input, this.meanPooling);
+  async forward(
+    input: string,
+    meanPooling: boolean = true
+  ): Promise<Float32Array> {
+    return new Float32Array(
+      await this.nativeModule.generate(input, meanPooling)
+    );
```
pweglik: Does it work correctly when model is in half precision?

packages/react-native-executorch/src/modules/natural_language_processing/TextEmbeddingsModule.ts
```diff
@@ -1,28 +1,37 @@
...
-    await super.load([modelSource, tokenizerSource]);
-    this.meanPooling = meanPooling;
   }
 
-  static override async forward(input: string): Promise<number[]> {
-    return this.nativeModule.forward(input, this.meanPooling);
+  async forward(input: string): Promise<Float32Array> {
+    return new Float32Array(
+      await this.nativeModule.generate(input, this.meanPooling)
+    );
```
pweglik, line 36: Does it work correctly when model is in half precision?

chmjkb: It doesn't, however I believe that we don't have any fp16 models now.

pweglik: I tried to solve it like this in imag embeddings PR:
```
export const getArrayConstructor = (
  scalarType: ScalarType
): ((arr: TensorBuffer) => Float32Array) => {
  if (scalarType === ScalarType.FLOAT)
    return (arr: TensorBuffer) => new Float32Array(arr as Float32Array);
  // todo add more types?

  throw Error("We don't handle this type of output array");
};
```

chmjkb: The native side here doesn't retain information about the underlying type, it doesn't return a `JSTensorViewOut`, but a simple `std::shared_ptr<OwningArrayBuffer>`, so this is a different case. I guess we can refactor this later on when we add support for FP16. I don't think this change makes much sense now because the native implementation is not really generic at all now (and it never was).

packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
```diff
@@ -0,0 +1,84 @@
...
+    auto buffer = std::make_shared<OwningArrayBuffer>(size);
+    std::memcpy(buffer->data(), data, size);
+    return buffer;
+  };
+
+  if (useMeanPooling) {
+    auto pooledOutput = numerical::meanPooling(modelOutput, attnMask);
+    numerical::normalize(pooledOutput);
+    return createBuffer(pooledOutput.data(),
+                        pooledOutput.size() * sizeof(float));
```
pweglik, line 79: You could use `size_bytes()` here as well

chmjkb: meanPooling returns std::vector, which doesn't have such method.

packages/react-native-executorch/android/src/main/java/com/swmansion/rnexecutorch/RnExecutorchPackage.kt
```diff
@@ -28,8 +28,6 @@ class RnExecutorchPackage : TurboReactPackage() {
       ETInstaller(reactContext)
     } else if (name == Tokenizer.NAME) {
```
mkopcins: Remove dead code here too

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
```diff
@@ -33,10 +33,10 @@ class BaseModel {
   // (unnecessary copies instead of working on JS memory). In this case
   // CallInvoker can be used to get jsi::Runtime, and use it in a safe manner.
   std::shared_ptr<react::CallInvoker> callInvoker;
+  std::unique_ptr<executorch::extension::Module> module;
```
msluszniak: Just FYI, there is a keyword `module` in C++ from C++20: https://en.cppreference.com/w/cpp/identifier_with_special_meaning/module so there is a little naming clash here. I don't have any obvious suggestion with a naming change here, so just leaving this info.

packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
```diff
@@ -0,0 +1,84 @@
...
+    inputIds64.push_back(static_cast<int64_t>(tok));
+  }
+
+  std::vector<int64_t> attentionMask;
+  attentionMask.reserve(inputIds.size());
+  for (const auto &tok : inputIds64) {
+    // Distinguishing between real tokens and padding, 0 is for paddings, while
+    // any other tokens are relevant
+    attentionMask.push_back(tok != 0);
+  }
```
mkopcins, line 32:
```suggestion
  std::vector<int64_t> attentionMask = inputIds64 | std::views::transform([](auto tok) { return !tok; }) | std::ranges::to<std::vector>();
```

packages/react-native-executorch/common/rnexecutorch/data_processing/Numerical.cpp
```diff
@@ -16,4 +17,41 @@ void softmax(std::vector<float> &v) {
...
+  for (const auto &val : span) {
+    sum += val * val;
+  }
+
+  // Preventing divison by 0
+  float norm = std::max(std::sqrt(sum), 1e-9f);
+  for (auto &val : span) {
+    val /= norm;
+  }
+}
```
pweglik: if norm is 0, is there even a point in normalizing? I think we should simply return early

packages/react-native-executorch/src/modules/natural_language_processing/TextEmbeddingsModule.ts
```diff
@@ -1,28 +1,37 @@
...
     if (meanPooling === undefined) {
       console.warn(
-        "You haven't passed meanPooling flag. It is defaulting to true. If your model doesn't require pooling it may misbehave."
+        "You haven't passed meanPooling flag. It is defaulting to true, if your model doesn't require mean pooling it may return wrong results."
       );
-      meanPooling = true;
+    } else {
+      this.meanPooling = meanPooling;
     }
```
pweglik, line 30: We need to set class field in both cases - when it was passed and when it was not
```suggestion
    } else {
      meanPooling = meanPooling;
    }
    this.meanPooling = meanPooling
```

packages/react-native-executorch/common/rnexecutorch/data_processing/Numerical.cpp
msluszniak: Yeap, here only a zero-vector can generate zero norm, so there is no need to dividing by a small epsilon. Also small nit, you can calculate inverse of square root of norm, and then multiply by it, it's faster than n consecutive divisions.

packages/react-native-executorch/src/modules/natural_language_processing/TextEmbeddingsModule.ts
chmjkb: Do we? It's set in line 6

pweglik: It doesn't protect us from case when someone calls load two times, they can set it to false first time, and don't set it the other time. It will warn them, that the default is true, even though it is set to false from previous run.

packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+TokenIdsWithAttentionMask TextEmbeddings::preprocess(const std::string &input) {
+  auto inputIds = tokenizer->encode(input);
+  // Tokenizers-cpp return tokens as int32, but text embedding models require
+  // int64 as input
+  std::vector<int64_t> inputIds64;
+  inputIds64.reserve(inputIds.size());
+  for (const auto &tok : inputIds) {
+    inputIds64.push_back(static_cast<int64_t>(tok));
+  }
```
mkopcins, line 24:
```suggestion
  std::vector<int64_t> inputIds64 = inputIds | std::views::transform([](auto tok) { return static_cast<int64_t>(tok); }) | std::ranges::to<std::vector>();
```

mkopcins: this should be a tiny bit more performant from what I read, but more of a interesting alternative than anything else

mkopcins (changes requested):

pweglik (approved):

packages/react-native-executorch/common/rnexecutorch/models/text_embeddings/TextEmbeddings.cpp
pweglik: much less readable imo

pweglik: Maybe if we stick to `tok != 0` and leave the comment, it should be fine

msluszniak: First of all `std::ranges::to` is available from c++23: https://en.cppreference.com/w/cpp/ranges/to.html, which disqualifies this. Also if you use `std::ranges`, you need to use `#include <ranges>`.

msluszniak: Also I performed benchmark with both approaches using the following code:
```c++
#include <iostream>
#include <vector>
#include <random>
#include <chrono>
#include <ranges>

int main()
{
    constexpr size_t count = 100;
    std::vector<int> inputIds;
    inputIds.reserve(count);

    std::random_device rd;
    std::mt19937 eng(rd());
    std::uniform_int_distribution<> distr(0, 100);

    for (size_t i = 0; i < count; ++i)
        inputIds.push_back(distr(eng));
        
    using namespace std::chrono;

    auto start = high_resolution_clock::now();
    std::vector<int64_t> inputIds64 = inputIds | std::views::transform([](auto tok) { return static_cast<int64_t>(tok); }) | std::ranges::to<std::vector>();
    auto end = high_resolution_clock::now();
    auto duration = duration_cast<nanoseconds>(end - start).count();
    std::cout << "Parsed " << inputIds.size() << " tokens in " << duration << " ns. using std::ranges\n";
    
    start = high_resolution_clock::now();
    std::vector<int64_t> attentionMask;
    attentionMask.reserve(inputIds.size());
    for (const auto &tok : inputIds64) {
      attentionMask.push_back(tok != 0);
    }
    end = high_resolution_clock::now();
    duration = duration_cast<nanoseconds>(end - start).count();
    std::cout << "Parsed " << attentionMask.size() << " tokens in " << duration << " ns. using classic approach\n";
}
```
and the results are the following:
```
Parsed 100 tokens in 9889 ns. using std::ranges
Parsed 100 tokens in 4069 ns. using classic approach

Parsed 1000 tokens in 84190 ns. using std::ranges
Parsed 1000 tokens in 59430 ns. using classic approach

Parsed 10000 tokens in 668049 ns. using std::ranges
Parsed 10000 tokens in 430970 ns. using classic approach

Parsed 100000 tokens in 6301817 ns. using std::ranges
Parsed 100000 tokens in 4638781 ns. using classic approach
```
So it doesn't seem to be faster either way.

pweglik (approved):

mkopcins (approved):

PR #415, opened 2025-06-24
feat: port LLMs to C++

chmjkb:
## Description

This PR ports the current LLM functionality to C++, getting rid of the `Runner` within the `ExecutorchLib` framework. I've also made some changes to the build system. 

`tokenizers-cpp`
- Previously the library was linked in `ExecutorchLib` in XCode via a build script, now it is completely removed from the frameworks. 
- I've prebuiilt static libraries from `tokenizers-cpp` repo, which I uploaded to `common/ios/libs/tokenizers-cpp`, similarly to the pre-build ExecuTorch binaries.  
- The includes for tokenizers-cpp are now in `react-native-executorch/third-party/include/tokenizers-cpp/tokenizers_cpp.h`
- Made some changes to the `libs` directory structure, please see the podspec for reference.
- These headers are then included from the llama runner source code.
- Since `tokenizers-cpp` for Android is pre-built with the `ExecuTorch` aar, I'm not making any changes here. This will need to be updated once we bump the ExecuTorch runtime and when we can safely get rid of the aar/jitpack setup. We can keep tokenizers-cpp as our submodule and then just reference it in Android's CMake. 

`runner`
- The runner source along with headers was moved from `ExecuTorchLib` to `common/runner`, similarly to `ada` and is compiled on the fly when our library compiles. 
- In the current situation, I think that we will be soon able to get rid of the ET fork and the submodule. 

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/runner/sampler.cpp
+++ b/packages/react-native-executorch/common/runner/sampler.cpp
@@ -32,7 +32,6 @@
  * SOFTWARE.
  */
 
-// #include <executorch/extension/llm/sampler/sampler.h>
 #include "sampler.h"
 #include <algorithm>
 


packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
```diff
@@ -0,0 +1,55 @@
...
+using executorch::runtime::Error;
+
+LLM::LLM(const std::string &modelSource, const std::string &tokenizerSource,
+         std::shared_ptr<react::CallInvoker> callInvoker)
+    : runner(std::make_unique<example::Runner>(modelSource, tokenizerSource)),
+      callInvoker(callInvoker) {
+  auto loadResult = runner->load();
+  if (loadResult != Error::Ok) {
+    throw std::runtime_error("Failed to load LLM runner");
+  }
```
msluszniak: I don't think loading model runner inside constructor is the perfect place for this. Apart from adding yet another responsibility to constructor, you also make the constructor throwable.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
```diff
@@ -0,0 +1,55 @@
...
+  auto nativeCallback = [this, callback](const std::string &token) {
+    callInvoker->invokeAsync([callback, token](jsi::Runtime &runtime) {
+      callback->call(runtime, jsi::String::createFromUtf8(runtime, token));
+    });
+  };
+
+  auto error = runner->generate(input, nativeCallback, {}, false);
+  if (error != executorch::runtime::Error::Ok) {
+    throw std::runtime_error("Failed to generate text, error: " +
+                             std::to_string(static_cast<int>(error)));
```
msluszniak: You do not provide error message, just exit code, maybe change this to `exit code:` or provide a real error message.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
chmjkb: This is exactly what is done for every other model, and also the native constructor is called on `.load()` call from JS, so I'll keep it this way, maybe a future consideration.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <executorch/extension/tensor/tensor.h>
+#include <filesystem>
+#include <rnexecutorch/Log.h>
+
+namespace rnexecutorch {
+using namespace facebook;
+using executorch::extension::TensorPtr;
+using executorch::runtime::Error;
+
+LLM::LLM(const std::string &modelSource, const std::string &tokenizerSource,
```
msluszniak, line 11: Mark constructor explicit, all functions that are suitable for it with `const`, `noexcept` keywords.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
```diff
@@ -0,0 +1,55 @@
+#include "LLM.h"
+
+#include <executorch/extension/tensor/tensor.h>
+#include <filesystem>
+#include <rnexecutorch/Log.h>
```
msluszniak: I guess this is an artifact from development phase, remove it.

packages/react-native-executorch/common/runner/sampler.h
```diff
@@ -19,18 +19,19 @@
 #endif
 
 #include <executorch/runtime/core/exec_aten/exec_aten.h>
+#include <executorch/runtime/platform/compiler.h>
 
 namespace executorch {
 namespace extension {
 namespace llm {
```
msluszniak, line 26: squash these namespaces with `::` to `namespace executorch::extension::llm {`, same with comment at the end

chmjkb: This code is a literal copy of ET code, so I'd rather keep it as is to avoid future misunderstandings

msluszniak: Ok, sure. It's just one of Clang tidy rules: https://clang.llvm.org/extra/clang-tidy/checks/modernize/concat-nested-namespaces.html so I mentioned this automatically ;p

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
msluszniak: Ok, so I'll keep that in mind for future cleaning.

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -32,6 +32,7 @@
  * SOFTWARE.
  */
 
+// #include <executorch/extension/llm/sampler/sampler.h>
```
mkopcins: ?

chmjkb: I'm not exactly sure why that line was here in the first place. Below you can see a line:
```cpp
#include "sampler.h"
````

Which literally includes the same thing. Also, we can't access `extension/llm/sampler/sampler.h` since there is no such header in our code, so it wouldn't compile.

PR #417, opened 2025-06-25
feat: Added generic vector log in C++

pweglik:
## Description

My proposition to log vector easier. I'm not sure if this is correct, give some feedback!

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/Log.h
+++ b/packages/react-native-executorch/common/rnexecutorch/Log.h
@@ -1,4 +1,6 @@
 #pragma once
+#include <sstream>
+#include <vector>
 
 namespace rnexecutorch {
 
@@ -6,5 +8,14 @@ enum class LOG_LEVEL { Info, Error, Debug };
 
 // const char* instead of const std::string& as va_start doesn't take references
 void log(LOG_LEVEL logLevel, const char *fmt, ...);
+template <typename T> void log(LOG_LEVEL logLevel, const std::vector<T> &vec) {
+  std::ostringstream oss;
+  oss << "Vector: [";
+  std::copy(vec.begin(), vec.end(), std::ostream_iterator<T>(oss, ", "));
+  std::string str = oss.str();
+  str = str.substr(0, str.length() - 2); // Remove the last comma and space
+  str += "]";
+  log(logLevel, "%s", str.c_str());
+}
 
 } // namespace rnexecutorch


packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,25 @@
...
+  oss << "Vector: [";
+  for (size_t i = 0; i < vec.size(); ++i) {
+    oss << vec[i];
+    if (i < vec.size() - 1)
+      oss << ", ";
+  }
+  oss << "]";
+
+  log(logLevel, "%s", oss.str().c_str());
+}
```
chmjkb: Can we instead do a function that would convert the vector to a `c_str`? I guess it would be a bit more flexible

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,25 @@
...
+#include <vector>
 
 namespace rnexecutorch {
 
 enum class LOG_LEVEL { Info, Error, Debug };
 
 // const char* instead of const std::string& as va_start doesn't take references
 void log(LOG_LEVEL logLevel, const char *fmt, ...);
+template <typename T>
+void logVector(LOG_LEVEL logLevel, const std::vector<T> &vec) {
```
msluszniak: Maybe instead of creating new name for vector just overload the `log` function?
`void log(LOG_LEVEL logLevel, const std::vector<T> &vec)`

pweglik: I like both your and Kuba solutions. Kuba argues that with conversion function we could do 
```
log(loglevel, "dupa", convertVectorToString(some_vector))
```
instead of two separate logs
What do you think?

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,25 @@
...
+  std::ostringstream oss;
+  oss << "Vector: [";
+  for (size_t i = 0; i < vec.size(); ++i) {
+    oss << vec[i];
+    if (i < vec.size() - 1)
+      oss << ", ";
+  }
+  oss << "]";
+
+  log(logLevel, "%s", oss.str().c_str());
```
msluszniak: I guess this approach might be better since we do not check every time if we have last element, just remove last redundant comma.
```suggestion
  std::copy(vec.begin(), vec.end(), std::ostream_iterator<T>(oss, ", "));
  std::string str = oss.str();
  str = str.substr(0, str.length() - 2); // Remove the last comma and space
  str += "]";

  log(logLevel, "%s", str.c_str());
```

packages/react-native-executorch/common/rnexecutorch/Log.h
pweglik: @msluszniak

chmjkb: Actually what I meant was:
```cpp
log(loglevel, "dupa: " + convertVectorToString(some_vector))
```

msluszniak: I guess it's up to you because you will use it mostly, so if you think that this case might be frequent then go with Kuba's approach :))

chmjkb: I don't have a strong opinion as well so feel free to choose whatever you want :D

msluszniak: For the best logging of every single structure I suggest to search some competitive programming templates of pro guys on e.g. CodeForces. They need to log everything very quickly, so I guess they mastered it.

msluszniak: I personally used this one on CP: https://github.com/Wolokin/cp-template/blob/main/debug.h. I need to check how this worked, but I think it might be more general.

pweglik:
I went with @msluszniak suggestions

msluszniak (approved):
This function looks good to me :)). In my free time, I'll dig around with more general logging, and send a PR with it to support e.g. all iterables like `std::array` `std::queue` etc.

msluszniak:
I have no rights to push my own branch to this repo, but here is a complex logging code (header only, to use it comment whole `Log.cpp`)
```cpp
#pragma once
#include <cstdarg>
#include <cstdio>
#include <iostream>
#include <map>
#include <queue>
#include <set>
#include <sstream>
#include <tuple>
#include <type_traits>
#include <unordered_map>
#include <unordered_set>
#include <utility>
#include <vector>
#ifdef __ANDROID__
#include <android/log.h>
#endif
#ifdef __APPLE__
#include <os/log.h>
#endif

namespace detail {
// To allow ADL with custom begin/end
using std::begin;
using std::end;

template <typename T>
concept is_iterable_impl = requires(T &t) {
  begin(t) != end(t);                     // begin/end and operator !=
  ++std::declval<decltype(begin(t)) &>(); // operator ++
  *begin(t);                              // operator*
};
} // namespace detail

template <typename T>
concept is_iterable = detail::is_iterable_impl<T>;

template <typename T>
concept is_streamable = requires(std::ostream &os, const T &t) {
  { os << t } -> std::convertible_to<std::ostream &>;
};

namespace rnexecutorch {

template <typename T>
  requires is_streamable<T>
void print_element(std::ostream &os, const T &value) {
  os << value;
}

// Forward declarations to guaratee template instantiations
template <typename T>
  requires is_iterable<T> && (!is_streamable<T>)
void print_element(std::ostream &os, const T &container);

template <typename T>
void print_element(std::ostream &os, const std::queue<T> &q);

template <typename... Args>
void print_element(std::ostream &os, const std::tuple<Args...> &tpl);

template <typename T, typename U>
void print_element(std::ostream &os, const std::pair<T, U> &p) {
  os << "(";
  print_element(os, p.first);
  os << ", ";
  print_element(os, p.second);
  os << ")";
}

template <typename T>
  requires is_iterable<T> && (!is_streamable<T>)
void print_element(std::ostream &os, const T &container) {
  os << "[";
  auto it = std::begin(container);
  if (it != std::end(container)) {
    print_element(os, *it++);
    for (; it != std::end(container); ++it) {
      os << ", ";
      print_element(os, *it);
    }
  }
  os << "]";
}

template <typename T>
void print_element(std::ostream &os, const std::queue<T> &q) {
  std::queue<T> temp = q;
  os << "[";
  if (!temp.empty()) {
    print_element(os, temp.front());
    temp.pop();
  }
  while (!temp.empty()) {
    os << ", ";
    print_element(os, temp.front());
    temp.pop();
  }
  os << "]";
}

template <typename... Args>
void print_element(std::ostream &os, const std::tuple<Args...> &tpl) {
  os << "<";
  std::apply(
      [&os](const auto &...args) {
        // Counter to apply commas correctly
        size_t count = 0;
        size_t total = sizeof...(args);

        (
            [&] {
              print_element(os, args);
              if (++count < total) {
                os << ", ";
              }
            }(),
            ...);
      },
      tpl);
  os << ">";
}

enum class LOG_LEVEL { Info, Error, Debug };

#ifdef __ANDROID__
android_LogPriority androidLogLevel(LOG_LEVEL logLevel) {
  switch (logLevel) {
  case LOG_LEVEL::Info:
    return ANDROID_LOG_INFO;
  case LOG_LEVEL::Error:
    return ANDROID_LOG_ERROR;
  case LOG_LEVEL::Debug:
    return ANDROID_LOG_DEBUG;
  default:
    return ANDROID_LOG_DEFAULT;
  }
}
#endif

template <typename... Args> void log(LOG_LEVEL logLevel, const Args &...args) {
  std::ostringstream oss;
  (print_element(oss, args),
   ...); // Fold expression used to handle all arguments

  constexpr size_t log_size = 1024;
  const std::string full_message = oss.str();
  bool is_message_longer_than_limit = full_message.size() > log_size;
  std::string buf = full_message.substr(0, log_size);
  if (is_message_longer_than_limit) {
    buf += "...";
  }
  const auto cbuf = buf.c_str();

#ifdef __ANDROID__
  __android_log_print(androidLogLevel(logLevel), "RnExecutorch", "%s", cbuf);
#elif defined(__APPLE__)
  switch (logLevel) {
  case LOG_LEVEL::Info:
    os_log_info(OS_LOG_DEFAULT, "%{public}s", cbuf);
    break;
  case LOG_LEVEL::Error:
    os_log_error(OS_LOG_DEFAULT, "%{public}s", cbuf);
    break;
  case LOG_LEVEL::Debug:
    os_log_debug(OS_LOG_DEFAULT, "%{public}s", cbuf);
    break;
  }
#else
  std::cout << cbuf
            << std::endl; // Default log to cout if none of the above platforms
#endif
}

} // namespace rnexecutorch
```

Example usage in `StyleTransfer.cpp`
```cpp

#include "../../Log.h"

StyleTransfer::StyleTransfer(const std::string &modelSource,
                             std::shared_ptr<react::CallInvoker> callInvoker)
    : BaseModel(modelSource, callInvoker) {
  auto inputShapes = getAllInputShapes();
  auto myPair = std::pair<std::vector<int>, std::vector<int>>{
      std::vector<int>{1, 2, 3}, std::vector<int>{4, 5}};
  // auto myMap = std::map<int, std::pair<int, float>>{{1, {1, 2}}, {2, {3, 4}}}; 
  // auto myTuple = std::make_tuple(1, 2.5, "hello", 'a');
  // auto myVector = std::vector<std::vector<int>>(1000, std::vector<int>(3, 7));
  log(LOG_LEVEL::Info, "Test pair: ", myPair);
  // AND SO ON
```
Ofc, I need to split implementation and declarations, add consistent naming, add some additional structures like std::stack, std::priority_queue, etc. So I guess I'll create a separate PR from fork.

pweglik:
Thanks for contribution @msluszniak! Feel free to open separate PR from fork just like you mentioned

pweglik:
Closing, as it is included in #422

Issue #387, opened 2025-06-10
Improve Resource Fetcher by adding missing functionalities


NorbertKlockiewicz:
- [ ] Pause/Resume download
- [ ] Cancel download in the middle of the process
- [ ] Add a way to get information about size of a file that is downloaded

eriCCsan:
I would Like to have this Feature too. Since the Files are very large and a User May doesn't wants IT the Download must pauseable.
Please please merge this :)

mlodyjesienin:
Hi @eriCCsan, this pull request introduced some breaking changes, therefore pause/resume/cancel features will be available in react-native-executorch v0.5.0. 
We plan to release v0.5.0 in the end of July, maybe even sooner. 
Sorry for the inconvenience.

PR #421, opened 2025-06-26
Resource Fetcher Improvement - PAUSE/RESUME/CANCEL

mlodyjesienin:
## Description

Implemented utils functions for ResourceFetcher module - mainly pause/resume/cancel  functions for remote downloads.
Refactored the module to make the code more readable. 
Made progress callback function more meaningful for fetching multiple resources. 

### Type of change

### Tested on

- [x] iOS
- [ ] Android
### Related issues

Closes Issue #387 

### Checklist

### Additional notes
 - Currently the Example References in docs are incomplete  and not generally true. However they give a rough idea about how it all works, but they definitely need a fix so I'd appreciate  if someone could take a look at it. 
 - I just noticed that Issue #387 mentions "Add a way to get information about size of a file that is downloaded".  It is not currently  implemented, but with current logic It can be easily done. I'll add this as soon as I can. 
 - Currently pausing/resuming/cancelling is separated between fetch and fetchMultipleResources. However I just noticed, that with current logic I can easily mix them together - that way there would not be two methods: `pauseFetching` and `pauseMultipleFetching` but a single method `pauseFetching` is going to be sufficient for all cases.
 - Although all this imperfections I think this code is in a state where it is ready for a review.


pweglik:
@mlodyjesienin  I think you need to fix types in callsites of resource fetcher - some types changed and it fails typescript transpliation checks

docs/docs/utilities/resource-fetcher.md
```diff
@@ -34,7 +34,10 @@ const localUri = await ResourceFetcher.fetch(
 
 ### Returns
 
-`Promise<string>` – A promise which resolves to a local file path (without file:// prefix) for the stored resource.
+`Promise<string | null>`
+
+- If the fetch was successful, a promise which resolves to a local file path (without file:// prefix) for the stored resource.
```
NorbertKlockiewicz:
```suggestion
- If the fetch was successful, it returns a promise which resolves to a local file path (without file:// prefix) for the stored resource.
```

docs/docs/utilities/resource-fetcher.md
```diff
@@ -34,7 +34,10 @@ const localUri = await ResourceFetcher.fetch(
...
+- If the fetch was interrupted by `pauseFetching` or `cancelFetching`, it returns `null`.
```
NorbertKlockiewicz:
```suggestion
- If the fetch was interrupted by `pauseFetching` or `cancelFetching`, it returns a promise which resolves to `null`.
```

docs/docs/utilities/resource-fetcher.md
```diff
@@ -63,7 +66,181 @@ const uris = await ResourceFetcher.fetchMultipleResources(
 
 ### Returns
 
-`Promise<string[]>` – A promise which resolves to an array of local file paths for the downloaded/stored resources (without file:// prefix).
+`Promise<string[] | null>`:
+
+- If the fetch was successful, A promise which resolves to an array of local file paths for the downloaded/stored resources (without file:// prefix).
```
NorbertKlockiewicz:
```suggestion
- If the fetch was successful, it returns a promise which resolves to an array of local file paths for the downloaded/stored resources (without file:// prefix).
```

docs/docs/utilities/resource-fetcher.md
```diff
@@ -63,7 +66,181 @@ const uris = await ResourceFetcher.fetchMultipleResources(
...
+- If the fetch was interrupted by `pauseMultipleFetching` or `cancelMultipleFetching`, it returns `null`.
```
NorbertKlockiewicz:
```suggestion
- If the fetch was interrupted by `pauseMultipleFetching` or `cancelMultipleFetching`, it returns a promise which resolves to `null`.
```

NorbertKlockiewicz: Similarly change the rest of similar sentences in this file

packages/react-native-executorch/src/utils/ResourceFetcher.ts
NorbertKlockiewicz, line 1: This file is very long, I am wondering if splitting it into single fetcher, multiple fetcher wouldn't be a better option. Leaving this comment here to start a discussion.

NorbertKlockiewicz: Also maybe moving some functions like calculatingProgress etc. to for example ResourceFetcherUtils namespace would be a great idea

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -9,85 +41,161 @@ import {
...
-    if (uri.startsWith('file://')) {
-      return this.removeFilePrefix(uri);
-    }
-
-    const filename = this.getFilenameFromUri(uri);
-    const fileUri = `${RNEDirectory}${filename}`;
-
-    if (await this.checkFileExists(fileUri)) {
-      return this.removeFilePrefix(fileUri);
+    const sourceType = await this.getType(source);
```
NorbertKlockiewicz: You are awaiting getType and it's not an async function

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -1,3 +1,35 @@
...
+ * - Downloaded file management:
+ *   - `.listDownloadedFiles()`, `.listDownloadedModels()`, `.deleteMultipleResources()`
+ *
+ * Remark: The pausing/resuming/canceling works only for fetching remote resources.
+ *
+ * Most exported functions accept either:
+ * - A single argument of type `ResourceSource` (union type of string, number, or object)
+ * - Multiple `ResourceSource` arguments
+ *
+ * Methods `.fetch()` and `fetchMultipleResources()` take optional argument as callback thar reports download progress.
```
NorbertKlockiewicz:
```suggestion
 * Methods `.fetch()` and `fetchMultipleResources()` take optional argument as callback that reports download progress.
```

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -100,19 +208,141 @@ export class ResourceFetcher {
...
+      );
+    }
+    await this.pauseFetching(source);
+  }
+
+  static async resumeMultipleFetching(...sources: ResourceSource[]) {
+    const source = this.findActive(sources);
+    if (source === null) {
+      throw new Error(
+        'None of given sources are currently during downloading process.'
```
NorbertKlockiewicz:
```suggestion
        'None of given sources are currently in a paused state.'
```

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -100,19 +208,141 @@ export class ResourceFetcher {
...
+    this.downloads.delete(source);
+  }
+
+  static async pauseMultipleFetching(...sources: ResourceSource[]) {
+    const source = this.findActive(sources);
+    if (source === null) {
+      throw new Error(
+        'None of given sources are currently during downloading process.'
+      );
+    }
```
NorbertKlockiewicz: This part is repeated many times, let's maybe extract it

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -100,19 +208,141 @@ export class ResourceFetcher {
...
+        resource.status = DownloadStatus.ONGOING;
+        const result = await resource.downloadResumable.resumeAsync();
+        if (
+          !this.downloads.has(source) ||
+          this.downloads.get(source)!.status === DownloadStatus.PAUSED
+        ) {
+          //if canceled or paused after earlier resuming.
+          return null;
+        }
+        if (!result || (result.status !== 200 && result.status !== 206)) {
```
NorbertKlockiewicz: Let's maybe create a constants at the top of file
```
const HTTP_OK = 200;
const HTTP_PARTIAL_CONTENT = 206;
```

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -128,7 +358,28 @@ export class ResourceFetcher {
...
+      return SourceType.OBJECT;
+    } else if (typeof source === 'number') {
+      const uri = Asset.fromModule(source).uri;
+      if (!uri.includes('://')) {
+        return SourceType.RELEASE_MODE_FILE;
+      }
+      return SourceType.DEV_MODE_FILE;
+    } else {
+      // typeof source == 'string'
+      if (source.startsWith('file:://')) {
```
NorbertKlockiewicz: I think that prefix should be `file://`

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -9,85 +41,161 @@ import {
...
+              totalLength,
+              info[0]!.previousFilesTotalLength,
+              info[0]!.length,
+              callback
+            )
+          : () => {},
+      results: [],
+    };
+
+    var node = head;
```
NorbertKlockiewicz: You shouldn't use `var`, in this case it's better to use `let`

NorbertKlockiewicz (changes requested):
I like the new concept, however the file is very large, I think we should focus on splitting it into multiple classes so it's easier to maintain and read :)

packages/react-native-executorch/src/utils/ResourceFetcher.ts
jakmro: Consider using [TS enums](https://www.typescriptlang.org/docs/handbook/enums.html)

packages/react-native-executorch/src/utils/ResourceFetcher.ts
mlodyjesienin: In my opinion it is not a good idea to split the fetcher into singleFetcher and multipleFetcher. Although a file is long and a bit messy, from a user perspective, the API is rather simple and not very different. Splitting it into two classes is suggesting that there are some key differences from user perspective, which is not the case. In fact, even from our perspective in the end we utilize same techniques to manage both cases so it is natural to keep those things together. Definitely though, we can move some utils functions to other file.

packages/react-native-executorch/src/utils/ResourceFetcher.ts
mlodyjesienin: The easiest way to extract it is just simply throw an error inside `findActive` function. I kept it outside `findActive` function on purpose, however maybe changing that is not a bad idea.

packages/react-native-executorch/src/utils/ResourceFetcher.ts
mlodyjesienin: In conflict with a next suggested change. (Extracting this part means same error message)

packages/react-native-executorch/src/utils/ResourceFetcher.ts
pweglik: Good idea imo

packages/react-native-executorch/src/utils/ResourceFetcher.ts
mlodyjesienin: I created a separate file `ResourceFetcherUtils.ts` and moved some functions, interfaces and enums from `ResourceFetcher` there. Not always was it easy to decide whether a function fits better in the utils file or main file, so I would like to see some opinions about it.

packages/react-native-executorch/src/controllers/OCRController.ts
```diff
@@ -48,13 +48,13 @@ export class OCRController {
...
 
-      const paths = await ResourceFetcher.fetchMultipleResources(
+      const paths = (await ResourceFetcher.fetchMultipleResources(
         this.modelDownloadProgressCallback,
         detectorSource,
         recognizerSources.recognizerLarge,
         recognizerSources.recognizerMedium,
         recognizerSources.recognizerSmall
-      );
+      ))!;
```
pweglik: I don't think it's a very good practice to use typescript ! as a non-null assertion operator

NorbertKlockiewicz: I feel like in this case we are sure that the function will either throw an error or return a path as it will never be paused so in my opinion utilising non-null assertion in this case is justified and better than null check.

pweglik: It's just really easy to omit while reading the code - if check - while more verbose is much clearer to me. 
NB: this can be paused - devs can call Module.load and then resourseFetcher.pause() on the same resource path that would cause it to stop - we've discussed such cases yesterday with @mlodyjesienin and decided to keep it as a feature 😄 We will also default to resuming download when load is called again on the paused resource. Lmk what you think

packages/react-native-executorch/src/controllers/OCRController.ts
```diff
@@ -48,14 +48,16 @@ export class OCRController {
...
-      const paths = await ResourceFetcher.fetchMultipleResources(
+      const paths = await ResourceFetcher.fetch(
         this.modelDownloadProgressCallback,
         detectorSource,
         recognizerSources.recognizerLarge,
         recognizerSources.recognizerMedium,
         recognizerSources.recognizerSmall
       );
-
+      if (paths === null) {
```
NorbertKlockiewicz: It's inconsistent, in llms you check the length of paths and in the rest you are only performing a null check, let's also check the length of this array here and in the rest of controllers

NorbertKlockiewicz (changes requested):
Please add check for an array length in all the controllers, then ready to merge. Good job!

NorbertKlockiewicz (approved):

PR #422, opened 2025-06-26
Add complex logging

msluszniak:
## Description

This PR adds complex logging with simple unit tests in gtest.

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/tests/run_all_tests.sh
+++ b/packages/react-native-executorch/common/rnexecutorch/tests/run_all_tests.sh
@@ -9,7 +9,6 @@ do
     if [ -f "$file" ]; then
         file_name=$(basename "$file")
         echo "Processing $file_name"
-        # Replace the following line with whatever command you wish to run on each file
         "$SCRIPT_DIR"/run_test.sh "$file"
     fi
 done


msluszniak:
@pweglik I think we can close #417 in favor of these changes. Please feel free to test it and add some ideas for some changes, improvements, and additional tests.

pweglik:
Sure, I'm closing #417

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,263 @@
...
+    container.pop();
+
+    while (!container.empty()) {
+      os << ", ";
+      if constexpr (FrontAccessible<T>) {
+        print_element(os, container.front());
+      } else if constexpr (TopAccessible<T>) {
+        print_element(os, container.top());
+      }
+      container.pop();
```
msluszniak: This will be refactored

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,263 @@
...
+  switch (logLevel) {
+  case LOG_LEVEL::Info:
+    os_log_info(OS_LOG_DEFAULT, "%{public}s", cbuf);
+    break;
+  case LOG_LEVEL::Error:
+    os_log_error(OS_LOG_DEFAULT, "%{public}s", cbuf);
+    break;
+  case LOG_LEVEL::Debug:
+    os_log_debug(OS_LOG_DEFAULT, "%{public}s", cbuf);
+    break;
```
msluszniak: This part will be refactored

packages/react-native-executorch/common/rnexecutorch/LogTest.cpp
```diff
@@ -0,0 +1,89 @@
+#include "Log.h"
+#include <gtest/gtest.h>
+
+TEST(PrintElementTest, HandlesIntegers) {
+  int number = 123;
+  std::ostringstream oss;
+  log_implementation::print_element(oss, number);
+  EXPECT_EQ(oss.str(), "123");
+}
```
msluszniak: I'll add full coverage for all STL containers available in C++20, custom structs which overloads `<<(std::ostream&, Sth)`

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,263 @@
...
+#include <map>
+#include <queue>
+#include <set>
+#include <sstream>
+#include <tuple>
+#include <type_traits>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
```
msluszniak: I'll remove redundant imports

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,263 @@
...
+ *
+ * The function takes logging level and arbitrary data type that implements `<<`
+ * operator for `std::ostream` or STL container with such data type.
+ *
+ * @param logLevel logging level - one of `LOG_LEVEL` enum class value: `Info`,
+ * `Error`, and `Debug`.
+ * @param args Data to be logged.
+ * @return Function does not return, only prints to console.
+ */
+template <typename... Args> void log(LOG_LEVEL logLevel, const Args &...args) {
```
msluszniak: I'm thinking about adding template on size for maximal logging message. For now, to extend log message above 1024 signs, you need to manually change `log_size`. If provided. then you will be able to call (at least I hope so) it like that: `log<2048>(LOG_LEVEL::Info, "Veeeeery long message ...");`

msluszniak: @chmjkb @mkopcins Let me know if you like this idea or not.

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,263 @@
...
+              if (++count < total) {
+                os << ", ";
+              }
+            }(),
+            ...);
+      },
+      tpl);
+  os << ">";
+}
+
```
msluszniak, line 237: Need to add overload for `print_element` that match objects that do not implement `<<` for `std::ostream` and throw and error for them

msluszniak:
Also, I'll change the styling of this file to match the rest of the files, e.g. camel case for variable and function names.

Edit: DONE

msluszniak:
For now changing to draft as there are quite a few ideas for improvement

packages/react-native-executorch/common/rnexecutorch/LogTest.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+TEST_F(EgdeCasesPrintTest, HandleEmptyContainer) {
+  std::vector<int> emptyVector{};
+  testContainerPrint(emptyVector, "[]");
+}
+
+int main(int argc, char **argv) {
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
```
msluszniak:
```suggestion
}

```

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,273 @@
...
+#ifdef __ANDROID__
+  high_level_log_implementation::handleAndroidLog(logLevel, cStyleBuffer);
+#elif defined(__APPLE__)
+  high_level_log_implementation::handleIosLog(logLevel, cStyleBuffer);
+#else
+  // Default log to cout if none of the above platforms
+  std::cout << cStyleBuffer << std::endl;
+#endif
+}
+} // namespace rnexecutorch
```
msluszniak:
```suggestion
} // namespace rnexecutorch

```

packages/react-native-executorch/common/rnexecutorch/Log.h
chmjkb: Makes sense to me :)

mkopcins: agreed

msluszniak:
@chmjkb @mkopcins PR is ready for review

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,423 @@
...
+};
+
+template <typename T>
+concept Iterable = HasBeginEnd<T> && requires(T &t) {
+  ++std::declval<decltype(begin(t)) &>(); // Support for increment
+  *begin(t);                              // Support for dereferencing
+};
+
+template <typename T>
+concept Sequencable = HasPop<T> && (FrontAccessible<T> || TopAccessible<T>);
```
mkopcins: Add support for deque

pweglik:
Why are my strings displayed like this?
Code:
```
  rnexecutorch::log(rnexecutorch::LOG_LEVEL::Debug, "prompt tokens:");
```
Output:
```
[p, r, o, m, p, t,  , t, o, k, e, n, s, :,
```

Also, I can't simply do, any workarounds? use two print statements?
```
rnexecutorch::log(rnexecutorch::LOG_LEVEL::Debug, "LOG Cpp prev_decoded %s", prev_decoded.c_str());
```

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,423 @@
...
+ * - Smart pointers, `std::variant`, and `std::optional`
+ * - `std::tuple` and `std::pair`
+ * - `std::error_code` and `std::exception_ptr`
+ * - `std::filesystem::path` and `std::filesystem::directory_iterator`
+ * - Every combination for mentioned above like `std::vector<std::set<int>>`
+ *
+ * You can manipulate size of the log message. By default it is set to 1024
+ * characters. To change this, specify the template argument like so:
+ * @code{.cpp}
+ * log<2048>(LOG_LEVEL::Info, longMsg);
```
mkopcins, line 409: What is the advantage of this approach? This will generate code for each declared length, right? Why not instead pass this as a simple function argument?

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,423 @@
...
+}
+#endif
+
+std::string getBuffer(std::ostringstream &oss, std::size_t maxLogMessageSize) {
+  const std::string fullMessage = oss.str();
+  bool isMessageLongerThanLimit = fullMessage.size() > maxLogMessageSize;
+  std::string buffer = fullMessage.substr(0, maxLogMessageSize);
+  if (isMessageLongerThanLimit) {
+    buffer += "...";
+  }
```
mkopcins: in this case shouldn't the buffer length be shortened by the length of elipsis?

packages/react-native-executorch/common/rnexecutorch/Log.h
mkopcins: Oh, it is Iterable, right

packages/react-native-executorch/common/rnexecutorch/Log.h
msluszniak: How would you distinguish between the length and the argument of type let's say `size_t`?

packages/react-native-executorch/common/rnexecutorch/Log.h
msluszniak: No, because it will generate case where when you pass string of size 1023 you will see 1023 characters, but when you pass 1025, you will see only 1021.

msluszniak:
@pweglik you do not pass `std::string` but string literal aka `const char[15]`, which is table of chars, if you what to print string pass string ;)

So it is outputted correctly. The only thing that is off is the `]` which is not here because strings literals are `\0` terminated. So this thing is additional argument to create a separate overload for `const char[N]` ;)

msluszniak:
Or I can do an overload to treat string literals more like a `std::string` if you'd like to

packages/react-native-executorch/common/rnexecutorch/Log.h
mkopcins: Wouldn't somethign liek this work?
```
log(LOG_LEVEL logLevel, size_t max_length, const Args &...args)
```

pweglik:
That would be super useful, then we souln't need to wrap every text we want to display with std::string `std::string("some text")`

msluszniak:
Yeah, makes sense, I'll add this

EDIT: DONE

packages/react-native-executorch/common/rnexecutorch/Log.h
msluszniak: But then, you'll be forced to pass this size every single time. Not a cool one I guess

msluszniak:
The one thing that I need to handle is to add spaces between separate object printed to `std::ostream`. Currently for
```cpp
  log(LOG_LEVEL::Debug, "ABC", std::vector<int>{1, 2, 3, 4}, std::pair<int, double>(1, 2.0));
```
the result is: `ABC[1, 2, 3, 4](1, 2)`

EDIT: DONE

packages/react-native-executorch/common/rnexecutorch/Log.h
mkopcins: fair point

mkopcins (approved):

msluszniak:
I need to test this on Android as well and guarantee that `LogTest.cpp` is not compiled because this will enforce additional google test dependency to the project.

msluszniak:
I can confirm that after excluding explicitly `LogTest.cpp` file from `CMakeList.txt` building process the logs appear successfully in `adb logcat`:
```
07-02 11:58:06.168 15854 15899 D RnExecutorch: ABC [1, 2, 3, 4] (1, 2)
```

I need to check the iOS building process the same way.

msluszniak:
For iOS I also excluded test file and logs are shown in logging box in Xcode:
image

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -68,13 +68,18 @@ Pod::Spec.new do |s|
     "ios/**/*.{m,mm,h}",
     "common/**/*.{cpp,c,h,hpp}",
   ]
+
+  # Exclude file with test to not introduce gtest dependency.
```
msluszniak:
```suggestion
  # Exclude file with tests to not introduce gtest dependency.
```

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,447 @@
...
+
+std::string getBuffer(const std::string &logMessage,
+                      std::size_t maxLogMessageSize) {
+  bool isMessageLongerThanLimit = logMessage.size() > maxLogMessageSize;
+  std::string buffer = logMessage.substr(0, maxLogMessageSize);
+  if (isMessageLongerThanLimit) {
+    buffer += "...";
+  }
+  return buffer;
+}
```
msluszniak:
```suggestion
std::string getBuffer(const std::string &logMessage,
                      std::size_t maxLogMessageSize) {
  if (logMessage.size() > maxLogMessageSize) {
    return logMessage.substr(0, maxLogMessageSize) + "...";
  }
  return logMessage;
}
```

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,447 @@
...
+  }
+}
+
+// A special function for C-style arrays deducing size via template
+template <typename T, std::size_t N>
+void printElement(std::ostream &os, T (&array)[N]) {
+  os << "[";
+  for (std::size_t i = 0; i < N; ++i) {
+    if (i > 0)
+      os << ", ";
```
msluszniak:
```suggestion
    if (i > 0) {
      os << ", ";
    }
```

packages/react-native-executorch/common/rnexecutorch/LogTest.cpp
chmjkb, line 1: I think the location of this file is a bit weird - can we move it somewhere else? Maybe create a separate directory like `rnexecutorch/tests` (just a suggestion, if you have better ideas then go ahead), and then just exclude all the files from this directory from compilation.

packages/react-native-executorch/android/src/main/cpp/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.13)
 
 file(GLOB_RECURSE ANDROID_CPP_SOURCES CONFIGURE_DEPENDS "${ANDROID_CPP_DIR}/*.cpp")
 file(GLOB_RECURSE COMMON_CPP_SOURCES CONFIGURE_DEPENDS "${COMMON_CPP_DIR}/*.cpp" "${COMMON_CPP_DIR}/*.c")
+list(REMOVE_ITEM COMMON_CPP_SOURCES "${COMMON_CPP_DIR}/rnexecutorch/LogTest.cpp")
```
chmjkb: Same as below, we can create a separate dir for tests

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,446 @@
...
+  case LOG_LEVEL::Error:
+    return ANDROID_LOG_ERROR;
+  case LOG_LEVEL::Debug:
+    return ANDROID_LOG_DEBUG;
+  default:
+    return ANDROID_LOG_DEFAULT;
+  }
+}
+
+void handleAndroidLog(LOG_LEVEL logLevel, const char *buffrer) {
```
chmjkb: typo - buffrer instead of buffer

packages/react-native-executorch/common/rnexecutorch/LogTest.cpp
chmjkb: Also some guidelines on how to run this would be handy too

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,446 @@
...
+namespace low_level_log_implementation {
+
+using namespace std::string_literals;
+
+namespace concepts {
+template <typename T>
+concept HasBeginEnd = requires(T t) {
+  { std::begin(t) } -> std::convertible_to<decltype(std::begin(t))>;
+  { std::end(t) } -> std::convertible_to<decltype(std::end(t))>;
+};
```
chmjkb: Can we do the following instead? Same goes for the other concepts 
```cpp
concept HasBeginEnd = requires(T t) {
  std::begin(t);
  std::end(t);
};
```

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,446 @@
...
+template <typename T, std::size_t N>
+void printElement(std::ostream &os, T (&array)[N]);
+
+template <typename T>
+  requires concepts::Iterable<T> && (!concepts::Streamable<T>)
+void printElement(std::ostream &os, const T &container);
+
+template <typename T>
+  requires concepts::Sequencable<T>
+void printElement(std::ostream &os, T container);
```
chmjkb, line 108: Why pass by value here?

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -68,13 +68,18 @@ Pod::Spec.new do |s|
...
+  # Exclude file with tests to not introduce gtest dependency.
   # Do not include the headers from common/rnexecutorch/jsi/ as source files. 
   # Xcode/Cocoapods leaks them to other pods that an app also depends on, so if 
   # another pod includes a header with the same name without a path by 
   # #include "Header.h" we get a conflict. Here, headers in jsi/ collide with 
   # react-native-skia. The headers are preserved by preserve_paths and 
   # then made available by HEADER_SEARCH_PATHS.
-  s.exclude_files = "common/rnexecutorch/jsi/*.{h,hpp}"
+  s.exclude_files = [
+    "common/rnexecutorch/LogTest.cpp",
```
chmjkb: Same as with CMake

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/Log.h
msluszniak: Because there is no way to iterate via these containers to not modify them, so copy is safe here.

packages/react-native-executorch/common/rnexecutorch/Log.h
```diff
@@ -1,10 +1,446 @@
...
+    return ANDROID_LOG_ERROR;
+  case LOG_LEVEL::Debug:
+    return ANDROID_LOG_DEBUG;
+  default:
+    return ANDROID_LOG_DEFAULT;
+  }
+}
+
+void handleAndroidLog(LOG_LEVEL logLevel, const char *buffrer) {
+  __android_log_print(androidLogLevel(logLevel), "RnExecutorch", "%s", buffrer);
```
msluszniak:
```suggestion
void handleAndroidLog(LOG_LEVEL logLevel, const char *buffer) {
  __android_log_print(androidLogLevel(logLevel), "RnExecutorch", "%s", buffer);
```

packages/react-native-executorch/common/rnexecutorch/Log.h
msluszniak: Yeah, type is trivially convertible to itself 😅, I'll fix that.

packages/react-native-executorch/common/rnexecutorch/LogTest.cpp
msluszniak: Both suggestions are on point, I'll add both separate directory with tests and guidelines how to run them.

packages/react-native-executorch/common/rnexecutorch/tests/run_all_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+# Loop through all .cpp files in that directory
+for file in "$SCRIPT_DIR"/*.cpp
+do
+    if [ -f "$file" ]; then
+        file_name=$(basename "$file")
+        echo "Processing $file_name"
+        # Replace the following line with whatever command you wish to run on each file
+        "$SCRIPT_DIR"/run_test.sh "$file"
+    fi
+done
```
msluszniak:
```suggestion
done

```

packages/react-native-executorch/common/rnexecutorch/tests/run_test.sh
```diff
@@ -0,0 +1,18 @@
...
+file_name="$1"
+
+# Compile the file with specified libraries
+g++ -std=c++20 -o test_executable "$file_name" -lgtest -lgtest_main -lpthread
+
+# Execute the binary
+./test_executable
+
+# Remove the executable
+rm -f test_executable
```
msluszniak:
```suggestion
rm -f test_executable

```

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -68,13 +68,18 @@ Pod::Spec.new do |s|
...
+  # Exclude file with tests to not introduce gtest dependency.
   # Do not include the headers from common/rnexecutorch/jsi/ as source files. 
   # Xcode/Cocoapods leaks them to other pods that an app also depends on, so if 
   # another pod includes a header with the same name without a path by 
   # #include "Header.h" we get a conflict. Here, headers in jsi/ collide with 
   # react-native-skia. The headers are preserved by preserve_paths and 
   # then made available by HEADER_SEARCH_PATHS.
-  s.exclude_files = "common/rnexecutorch/jsi/*.{h,hpp}"
+  s.exclude_files = [
+    "common/rnexecutorch/tests/*",
```
msluszniak:
```suggestion
    "common/rnexecutorch/tests/*.{cpp}",
```

packages/react-native-executorch/common/rnexecutorch/tests/run_all_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+# Get the directory where the script is located
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
+
+# Loop through all .cpp files in that directory
+for file in "$SCRIPT_DIR"/*.cpp
+do
+    if [ -f "$file" ]; then
+        file_name=$(basename "$file")
+        echo "Processing $file_name"
+        # Replace the following line with whatever command you wish to run on each file
```
msluszniak:
```suggestion
```

chmjkb (approved):
LGTM

PR #427, opened 2025-06-30
feat: Port STT to C++

chmjkb:
### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


packages/react-native-executorch/common/rnexecutorch/data_processing/FFT.cpp
```diff
@@ -0,0 +1,19 @@
+#include <rnexecutorch/data_processing/FFT.h>
+
+namespace rnexecutorch::dsp {
+
+FFT::FFT(int size) : size_(size) {
+  pffftSetup_ = pffft_new_setup(size_, PFFFT_REAL);
+  work_ = (float *)pffft_aligned_malloc(size_ * sizeof(float));
```
msluszniak: C-style casts are not preferred
```suggestion
  work_ = static_cast<float*>(_pffft_aligned_malloc(size_ * sizeof(float)));
```

packages/react-native-executorch/common/rnexecutorch/data_processing/FFT.cpp
```diff
@@ -0,0 +1,19 @@
...
+}
+
+FFT::~FFT() {
+  pffft_destroy_setup(pffftSetup_);
+  pffft_aligned_free(work_);
+}
+
+void FFT::doFFT(float *in, std::vector<std::complex<float>> &out) {
+  pffft_transform_ordered(pffftSetup_, in, reinterpret_cast<float *>(&out[0]),
```
msluszniak: This is the preferred way to passing STL containers to C-style APIs
```suggestion
  pffft_transform_ordered(pffftSetup_, in, reinterpret_cast<float *>(out.data()),
```

packages/react-native-executorch/common/rnexecutorch/data_processing/FFT.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+extern "C" {
+
+#include <pfft/pfft.h>
+}
```
msluszniak:
```suggestion
extern "C" {
#include <pfft/pfft.h>
}
```

packages/react-native-executorch/common/pfft/pfft.h
```diff
@@ -0,0 +1,185 @@
...
+void *pffft_aligned_malloc(size_t nb_bytes);
+void pffft_aligned_free(void *);
+
+/** return 4 or 1 wether support SSE/Altivec instructions was enable when
+ * building pffft.c */
+int pffft_simd_size(void);
+
+#ifdef __cplusplus
+}
+#endif
```
msluszniak:
```suggestion
#endif

```

packages/react-native-executorch/common/rnexecutorch/data_processing/FFT.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace rnexecutorch::dsp {
+
+class FFT {
+public:
+  explicit FFT(int size);
+  ~FFT();
+
+  void doFFT(float *in, std::vector<std::complex<float>> &out);
+  void doInverseFFT(std::vector<std::complex<float>> &in, float *out);
```
msluszniak: There is no such function in source file `FFT.cpp`, should it be deleted here?

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,62 @@
+#include <cstddef>
+#include <math.h>
+#include <rnexecutorch/data_processing/FFT.h>
+#include <rnexecutorch/data_processing/dsp.h>
+#include <vector>
```
msluszniak: Header file `dsp.h` has already this include, remove it from here.
```suggestion
```

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,62 @@
...
+  int numBins = fftWindowSize / 2;
+  auto hann = hannWindow(fftWindowSize);
+  auto inBuffer = std::vector<float>(fftWindowSize);
+  auto outBuffer = std::vector<std::complex<float>>(fftWindowSize);
+
+  // Output magnitudes in dB
+  std::vector<float> magnitudes;
+  magnitudes.reserve(numFrames * numBins);
+  const float magnitudeScale = 1.0f / fftWindowSize;
+  const float epsilon = 1e-10f;
```
msluszniak: And include `#include <limits>`. This is the better way to define epsilon. Especially, that epsilons in 32-bit floating point arithmetic are not so precise as 1e-10. Check out the value of  `std::numeric_limits<float>::epsilon();` to see, that is has much bigger value, order of magnitude 1e-7.
```suggestion
  constexpr auto epsilon = std::numeric_limits<float>::epsilon();
```

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,62 @@
...
+  size_t numFrames = 1 + (waveform.size() - fftWindowSize) / hopSize;
+  int numBins = fftWindowSize / 2;
+  auto hann = hannWindow(fftWindowSize);
+  auto inBuffer = std::vector<float>(fftWindowSize);
+  auto outBuffer = std::vector<std::complex<float>>(fftWindowSize);
+
+  // Output magnitudes in dB
+  std::vector<float> magnitudes;
+  magnitudes.reserve(numFrames * numBins);
+  const float magnitudeScale = 1.0f / fftWindowSize;
```
msluszniak:
```suggestion
  const auto magnitudeScale = 1.0f / static_cast<float>(fftWindowSize);
```

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,62 @@
...
+  return window;
+}
+
+std::vector<float> stftFromWaveform(std::span<float> waveform,
+                                    size_t fftWindowSize, size_t hopSize) {
+  // Initialize FFT
+  FFT fft(fftWindowSize);
+
+  size_t numFrames = 1 + (waveform.size() - fftWindowSize) / hopSize;
+  int numBins = fftWindowSize / 2;
```
msluszniak: I'm marking this only here, but it's general comment to all such places. Look that below you marked `magnitudeScale` with `const` qualifier, but here you don't. Make sure that you're consistent with marking variables with const. This way, it will be easy to detect variables that will really change in the function ;)
```suggestion
  const size_t numFrames = 1 + (waveform.size() - fftWindowSize) / hopSize;
  const int numBins = fftWindowSize / 2;
```

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,62 @@
...
+  std::vector<float> magnitudes;
+  magnitudes.reserve(numFrames * numBins);
+  const float magnitudeScale = 1.0f / fftWindowSize;
+  const float epsilon = 1e-10f;
+
+  for (size_t t = 0; t < numFrames; ++t) {
+    size_t offset = t * hopSize;
+
+    // Clear the input buffer first
+    std::fill(inBuffer.begin(), inBuffer.end(), 0.0f);
```
msluszniak: This version of fill comes from `<algorithm>`, please include this header also
```suggestion
    std::ranges::fill(inBuffer, 0.0f);
```

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,62 @@
...
+    for (size_t i = 0; i < samplesToRead; i++) {
+      inBuffer[i] = waveform[offset + i] * hann[i];
+    }
+    // Perform FFT using the FFT class
+    fft.doFFT(inBuffer.data(), outBuffer);
+
+    // Calculate magnitudes in dB (only positive frequencies)
+    for (int i = 0; i < numBins; i++) {
+      float magnitude = std::abs(outBuffer[i]) * magnitudeScale;
+      float magnitude_db = 20.0f * log10f(magnitude + epsilon);
```
msluszniak: 20.0f is a magic number, please move it to a named constant

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -49,6 +50,21 @@ getValue<std::shared_ptr<jsi::Function>>(const jsi::Value &val,
       val.asObject(runtime).asFunction(runtime));
 }
 
+template <>
```
msluszniak: Look that here you have identical implementation for both `int32_t` and `int64_t`. Please implement this as a template, and then explicitly instantiate variants for both `int32_t` and `int64_t`.

I guess that the same case is with `std::vector<JSTensorViewIn>` and `std::vector<JSTensorViewIn>`

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -170,6 +186,89 @@ getValue<std::set<std::string, std::less<>>>(const jsi::Value &val,
...
+  jsi::ArrayBuffer arrayBuffer =
+      bufferValue.asObject(runtime).getArrayBuffer(runtime);
+  size_t byteOffset =
+      getValue<size_t>(obj.getProperty(runtime, "byteOffset"), runtime);
+  size_t length = getValue<size_t>(obj.getProperty(runtime, "length"), runtime);
+
+  T *dataPtr = reinterpret_cast<T *>(
+      static_cast<uint8_t *>(arrayBuffer.data(runtime)) + byteOffset);
+
+  return std::span<T>(dataPtr, length);
```
msluszniak:
```suggestion
  return {dataPtr, length};
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -170,6 +186,89 @@ getValue<std::set<std::string, std::less<>>>(const jsi::Value &val,
...
+inline std::span<int8_t> getValue<std::span<int8_t>>(const jsi::Value &val,
+                                                     jsi::Runtime &runtime) {
+  return getTypedArrayAsSpan<int8_t>(val, runtime);
+}
+
+template <>
+inline std::span<uint8_t> getValue<std::span<uint8_t>>(const jsi::Value &val,
+                                                       jsi::Runtime &runtime) {
+  return getTypedArrayAsSpan<uint8_t>(val, runtime);
+}
```
msluszniak, line 285: Same comment as for std::vector implementations. please define template for `std::span<T>` and then explicitly instantiate instead of copying the same implementation all the time ;) Specialization is used to implement let's say "special" version which do not match the template implementation. Here it's not a case

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.cpp
```diff
@@ -83,6 +83,9 @@ BaseModel::getAllInputShapes(std::string methodName) {
   return output;
 }
 
+/// @brief This method is a forward wrapper that is created solely to be exposed
+/// to JS. It is not meant to be used within C++. If you want to call forward
+/// from C++ on a BaseModel, please use BaseModel::forward.
 std::vector<JSTensorViewOut>
 BaseModel::forwardJS(const std::vector<JSTensorViewIn> tensorViewVec) {
```
msluszniak: Top level const in declaration should be removed, maybe you meant `BaseModel::forwardJS(const std::vector<JSTensorViewIn> &tensorViewVec)` instead?
```suggestion
BaseModel::forwardJS(std::vector<JSTensorViewIn> tensorViewVec) {
```

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.h
```diff
@@ -0,0 +1,31 @@
...
+namespace rnexecutorch {
+using namespace facebook;
+using executorch::aten::Tensor;
+using executorch::runtime::EValue;
+
+class EncoderDecoderBase {
+public:
+  EncoderDecoderBase(const std::string &encoderPath,
+                     const std::string &decoderPath,
+                     std::shared_ptr<react::CallInvoker> callInvoker);
```
msluszniak:
```suggestion
  explicit EncoderDecoderBase(const std::string &encoderPath,
                     const std::string &decoderPath,
                     std::shared_ptr<react::CallInvoker> callInvoker) noexcept;
```

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.h
```diff
@@ -0,0 +1,31 @@
...
+  size_t getMemoryLowerBound() {
```
msluszniak:
```suggestion
  size_t getMemoryLowerBound() const noexcept {
```

msluszniak: And move implementation to cpp file, I prefer clear separation, even though implementation is a one-liner

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.cpp
```diff
@@ -0,0 +1,33 @@
+#include "MoonshineStrategy.h"
+#include <executorch/extension/tensor/tensor_ptr_maker.h>
+#include <executorch/runtime/core/portable_type/scalar_type.h>
+
+namespace rnexecutorch {
+
+using namespace ::executorch::extension;
+using namespace ::executorch::aten;
+
+std::pair<TensorPtr, std::vector<int32_t>>
```
msluszniak: `std::pair<TensorPtr, std::vector<int32_t>>`, sounds like a very specific type, maybe name it with something meaningful?

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.cpp
```diff
@@ -0,0 +1,33 @@
...
+  std::vector<int32_t> tensorSizes = {1,
+                                      static_cast<int32_t>(prevTokens.size())};
+  return make_tensor_ptr(tensorSizes, const_cast<int64_t *>(prevTokens.data()),
+                         ScalarType::Long);
+}
+
+int64_t MoonshineStrategy::extractOutputToken(const void *outputPtr,
+                                              std::vector<int32_t> &sizes) {
+  size_t innerDim = sizes.at(1);
+  const int64_t *data = static_cast<const int64_t *>(outputPtr);
```
msluszniak:
```suggestion
  const auto *data = static_cast<const int64_t *>(outputPtr);
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.cpp
```diff
@@ -0,0 +1,33 @@
...
+  auto tensorPtr =
+      make_tensor_ptr(inputShape, waveform.data(), ScalarType::Float);
+  return {tensorPtr, inputShape};
+}
+
+TensorPtr
+MoonshineStrategy::prepareTokenInput(const std::vector<int64_t> &prevTokens) {
+  std::vector<int32_t> tensorSizes = {1,
+                                      static_cast<int32_t>(prevTokens.size())};
+  return make_tensor_ptr(tensorSizes, const_cast<int64_t *>(prevTokens.data()),
```
msluszniak: Is this const_cast really needed?, please check this out.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.h
```diff
@@ -0,0 +1,20 @@
...
+  std::pair<TensorPtr, std::vector<int32_t>>
+  prepareAudioInput(std::span<float> waveform) override;
+
+  TensorPtr prepareTokenInput(const std::vector<int64_t> &prevTokens) override;
+
+  std::string getDecoderMethod() const override { return "forward_cached"; }
+
+  int64_t extractOutputToken(const void *outputPtr,
+                             std::vector<int32_t> &sizes) override;
+};
```
msluszniak: Make sure, that you marked everything you can with `const`, `noexcept`, and `static` qualifiers. Also you don't have constructor cleared here, so maybe consider removing all of them explicitly.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -0,0 +1,76 @@
...
+  if (!decoderResult.ok()) {
+    throw std::runtime_error(
+        "Forward pass failed during decoding, error code: " +
+        std::to_string(static_cast<int>(decoderResult.error())));
+  }
+
+  auto decoderOutputTensor = decoderResult.get().at(0).toTensor();
+  auto decoderOutputTensorSizes = decoderOutputTensor.sizes();
+  std::vector<int32_t> sizesVec(decoderOutputTensorSizes.begin(),
+                                decoderOutputTensorSizes.end());
```
msluszniak: Why you not use `decoderOutputTensorSizes` in `strategy->extractOutputToken` and you're copying content to `sizesVec`?

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -0,0 +1,76 @@
+#include "ReactCommon/CallInvoker.h"
+#include "rnexecutorch/models/EncoderDecoderBase.h"
+#include <memory>
+#include <rnexecutorch/models/speech_to_text/MoonshineStrategy.h>
+#include <rnexecutorch/models/speech_to_text/SpeechToText.h>
+#include <rnexecutorch/models/speech_to_text/WhisperStrategy.h>
+#include <stdexcept>
+#include <string>
+#include <vector>
```
msluszniak: Again vector is included in header of `SpeachToText.h`. Make sure that you do not repeat such imports

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
```diff
@@ -0,0 +1,28 @@
...
+class SpeechToText : public EncoderDecoderBase {
+public:
+  SpeechToText(const std::string &encoderPath, const std::string &decoderPath,
+               const std::string &modelName,
+               std::shared_ptr<react::CallInvoker> callInvoker);
+  void encode(std::span<float> waveform);
+  int64_t decode(std::vector<int64_t> prevTokens);
+
+private:
+  const std::string modelName;
```
msluszniak, line 24: There is a clang tidy check for such const in members
```suggestion
  std::string modelName;
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
```diff
@@ -0,0 +1,28 @@
...
+#include <vector>
+
+namespace rnexecutorch {
+class SpeechToText : public EncoderDecoderBase {
+public:
+  SpeechToText(const std::string &encoderPath, const std::string &decoderPath,
+               const std::string &modelName,
+               std::shared_ptr<react::CallInvoker> callInvoker);
+  void encode(std::span<float> waveform);
+  int64_t decode(std::vector<int64_t> prevTokens);
```
msluszniak: again, probably some members might be marked with some qualifiers

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.cpp
```diff
@@ -0,0 +1,40 @@
...
+std::pair<TensorPtr, std::vector<int32_t>>
+WhisperStrategy::prepareAudioInput(std::span<float> waveform) {
+  preprocessedData = dsp::stftFromWaveform(waveform, 512, 160);
+  auto numFrames = preprocessedData.size() / 256;
+  std::vector<int32_t> inputShape = {static_cast<int32_t>(numFrames), 256};
+
+  auto tensorPtr =
+      make_tensor_ptr(inputShape, preprocessedData.data(), ScalarType::Float);
+  return {tensorPtr, inputShape};
+}
```
msluszniak: Many magic numbers, move to constants

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.cpp
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  std::vector<int32_t> tensorSizes = {1, static_cast<int32_t>(tokens32.size())};
+  return make_tensor_ptr(tensorSizes, tokens32.data(), ScalarType::Int);
+}
+
+int64_t WhisperStrategy::extractOutputToken(const void *outputPtr,
+                                            std::vector<int32_t> &sizes) {
+  size_t innerDim = sizes.at(1);
+  const int32_t *data = static_cast<const int32_t *>(outputPtr);
```
msluszniak:
```suggestion
  const auto *data = static_cast<const int32_t *>(outputPtr);
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "SpeechToTextStrategy.h"
+#include <vector>
+
+namespace rnexecutorch {
+
+class WhisperStrategy : public SpeechToTextStrategy {
+private:
+  mutable std::vector<float> preprocessedData;
+  mutable std::vector<int32_t> tokens32;
```
msluszniak: Why explicitly marked with mutable?

msluszniak (changes requested):

packages/react-native-executorch/common/rnexecutorch/data_processing/FFT.h
msluszniak: Ok, you handled it already while I was creating this review. Resolving

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
msluszniak: Namely this check: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.cpp
msluszniak: https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <set>
+#include <span>
 #include <type_traits>
 #include <unordered_map>
```
msluszniak, line 7: Here, on the other hand, you should include `<vector>`, now it only works because other included headers already had `<vector>`. There is a clang tidy check to explicitly include all used headers. Check if you lack any.

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.h
```diff
@@ -0,0 +1,31 @@
...
+    return encoder_->getMemoryLowerBound() + decoder_->getMemoryLowerBound();
+  }
+  void unload();
```
msluszniak: Even though reset for pointer arguments is not explicitly mark as noexcept: https://en.cppreference.com/w/cpp/memory/shared_ptr/reset you can mark it because use pass `nullptr` in implementation.
```suggestion
  void unload() noexcept;
```

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.h
chmjkb: This shouldn't be noexcept given that it's creating a basemodel, which is not noexcept

msluszniak: Yeap, absolutely, I missed that BaseModel constructor throws. If I remember correctly I've already complained once that BaseModel can throw ;p

[EDIT] Yeap heh: https://github.com/software-mansion/react-native-executorch/pull/415#discussion_r2174501318

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.cpp
chmjkb: `make_tensor_ptr` accepts non-const pointers, therefore im doing a `const_cast` here

msluszniak: Yeah, makes sense. `std::vector<T,Allocator>::data` have two variants:
`T* data();`
`const T* data() const;`
and when vector is const (here it is since it is: `const std::vector<int64_t> &prevTokens`) then second one will be chosen, and yeah const_cast is needed.

Sooo, here maybe better approach will be passing vector by value in case `make_tensor_ptr` takes ownership on data or just use moving semantic if you don't need `prevTokens` anymore?

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
chmjkb: Not really since each method can throw and encode/decode uses a encoderOutput member, if you meant const noexcept, i made the constructor explicit.

msluszniak: I meant only these applicable so it's perfectly fine

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+std::vector<float> stftFromWaveform(std::span<float> waveform,
+                                    size_t fftWindowSize, size_t hopSize) {
+  // Initialize FFT
+  FFT fft(fftWindowSize);
+
+  const auto numFrames = 1 + (waveform.size() - fftWindowSize) / hopSize;
+  const auto numBins = fftWindowSize / 2;
+  auto hann = hannWindow(fftWindowSize);
```
msluszniak: I think you can mark it const al well
```suggestion
  const auto hann = hannWindow(fftWindowSize);
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.h
```diff
@@ -0,0 +1,24 @@
...
+#include <vector>
+
+namespace rnexecutorch {
+
+class WhisperStrategy : public SpeechToTextStrategy {
+private:
+  std::vector<float> preprocessedData;
+  std::vector<int32_t> tokens32;
+
+public:
```
msluszniak: I prefer to have public members first, then protected and and on the bottom the private ones. The reason for that is because people read from top to bottom and for them public members are most important since the will use it and might check something out. Check if this order is preserved in each new class.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "SpeechToTextStrategy.h"
+
+namespace rnexecutorch {
+
+class MoonshineStrategy : public SpeechToTextStrategy {
```
msluszniak: I guess that classes with dsp, fft, Moonshine, and Whisper might be marked final since there will be no inheritance (at least now) in these cases.

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
```diff
@@ -24,10 +24,14 @@ class BaseModel {
   getAllInputShapes(std::string methodName = "forward");
   std::vector<JSTensorViewOut>
   forwardJS(std::vector<JSTensorViewIn> tensorViewVec);
-
-protected:
   Result<std::vector<EValue>> forward(const EValue &input_value);
```
msluszniak, line 27: Just a suggestion, but you may mark functions that are exposed and have heavy returns or where you want to be clear that these results needs to be assigned with `[[no_discard("<reason for that>")]]`. I don't know if there are such functions so just a hint to maybe check this out

packages/react-native-executorch/common/rnexecutorch/data_processing/dsp.cpp
```diff
@@ -0,0 +1,65 @@
...
+  for (int i = 0; i < size; i++) {
+    window[i] = 0.5f * (1 - std::cos(2 * M_PI * i / size));
+  }
+  return window;
+}
+
+std::vector<float> stftFromWaveform(std::span<float> waveform,
+                                    size_t fftWindowSize, size_t hopSize) {
+  // Initialize FFT
+  FFT fft(fftWindowSize);
```
msluszniak, line 24: prevent implicit conversion from size_t to int since FFT constructor accepts int
```suggestion
  FFT fft(static_cast<int>(fftWindowSize));
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.cpp
```diff
@@ -0,0 +1,31 @@
...
+                                      static_cast<int32_t>(prevTokens.size())};
+  // prevTokens gets copied!!
+  return make_tensor_ptr(std::move(tensorSizes), prevTokens);
+}
+
+int64_t MoonshineStrategy::extractOutputToken(const void *outputPtr,
+                                              size_t innerDim) const {
+  const auto *data = static_cast<const int64_t *>(outputPtr);
+  return data[innerDim - 1];
+}
```
msluszniak, line 29: I guess that all three might be `static`

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.cpp
```diff
@@ -0,0 +1,38 @@
...
+  }
+  auto tensorSizes = {1, static_cast<int32_t>(tokens32.size())};
+  return make_tensor_ptr(std::move(tensorSizes), std::move(tokens32));
+}
+
+int64_t WhisperStrategy::extractOutputToken(const void *outputPtr,
+                                            size_t innerDim) const {
+  const auto *data = static_cast<const int32_t *>(outputPtr);
+  return static_cast<int64_t>(data[innerDim - 1]);
+}
```
msluszniak, line 36: Also ~all methods are~  `WhisperStrategy::extractOutputToken` is static

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.h
chmjkb: this was actually an accident, thanks:D

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "ReactCommon/CallInvoker.h"
+#include "executorch/runtime/core/evalue.h"
+#include <cstdint>
+#include <memory>
+#include <rnexecutorch/models/EncoderDecoderBase.h>
+#include <rnexecutorch/models/speech_to_text/SpeechToTextStrategy.h>
+#include <span>
+#include <vector>
```
msluszniak, line 11: Include `<string>` here as you use string e.g. in constructor

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -0,0 +1,73 @@
+#include "ReactCommon/CallInvoker.h"
+#include "rnexecutorch/models/EncoderDecoderBase.h"
+#include <memory>
+#include <rnexecutorch/models/speech_to_text/MoonshineStrategy.h>
+#include <rnexecutorch/models/speech_to_text/SpeechToText.h>
+#include <rnexecutorch/models/speech_to_text/WhisperStrategy.h>
+#include <stdexcept>
+#include <string>
```
msluszniak: Remove `<string>` from here, it will be included in header

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "SpeechToTextStrategy.h"
+#include <vector>
```
msluszniak, line 5: Add `<span>`

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
chmjkb: I'll create a separate issue for that, good idea!

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/WhisperStrategy.cpp
msluszniak: Correction, not all `TensorPtr WhisperStrategy::prepareAudioInput(std::span<float> waveform)` uses member `preprocessedData`, and `TensorPtr
WhisperStrategy::prepareTokenInput(const std::vector<int64_t> &prevTokens)` uses `tokens32`

chmjkb: Can't do that since `extractOutputToken` is virtual from base class

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -0,0 +1,72 @@
+#include "ReactCommon/CallInvoker.h"
+#include "rnexecutorch/models/EncoderDecoderBase.h"
+#include <memory>
```
msluszniak: Already included in header

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -1,12 +1,11 @@
 #pragma once
 
+#include <ReactCommon/CallInvoker.h>
 #include <cstdio>
 #include <string>
+#include <thread>
```
msluszniak, line 5: Where is used thread?

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -1,12 +1,11 @@
 #pragma once
 
+#include <ReactCommon/CallInvoker.h>
```
msluszniak, line 3: You should include memory since you use `std::shared_ptr` here

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
chmjkb: https://github.com/software-mansion/react-native-executorch/blob/b0290da95bfabd5f7ca95858d377b46a047e775d/packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h#L162

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -1,12 +1,11 @@
...
 #include <cstdio>
```
msluszniak: cstdio is also unused here I guess

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/MoonshineStrategy.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
```
msluszniak, line 2: Include `<vector>`

msluszniak: and `<span>`

packages/react-native-executorch/src/modules/natural_language_processing/SpeechToTextModule.ts
```diff
@@ -52,12 +52,12 @@ export class SpeechToText {
...
   }
 
-  static async encode(waveform: number[]) {
+  static async encode(waveform: Float32Array) {
     return await this.module.encode(waveform);
   }
 
-  static async decode(seq: number[], encodings: number[]) {
-    return await this.module.decode(seq, encodings);
+  static async decode(seq: number[]) {
```
mkopcins, line 59: This needs to be changed in docs too iirc

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
```diff
@@ -0,0 +1,30 @@
...
+  void encode(std::span<float> waveform);
+  int64_t decode(std::vector<int64_t> prevTokens);
+
+private:
+  const std::string modelName;
+  executorch::runtime::EValue encoderOutput;
+  std::unique_ptr<SpeechToTextStrategy> strategy;
+
+  void initializeStrategy();
+};
```
msluszniak, line 29:
```suggestion
};

```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -0,0 +1,69 @@
...
+    throw std::runtime_error(
+        "Forward pass failed during decoding, error code: " +
+        std::to_string(static_cast<int>(decoderResult.error())));
+  }
+
+  const auto decoderOutputTensor = decoderResult.get().at(0).toTensor();
+  const auto innerDim = decoderOutputTensor.size(1);
+  return strategy->extractOutputToken(decoderOutputTensor.const_data_ptr(),
+                                      innerDim);
+}
```
msluszniak, line 68:
```suggestion
}

```

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <ReactCommon/CallInvoker.h>
+#include <memory>
+#include <rnexecutorch/models/BaseModel.h>
+#include <string>
+
+namespace rnexecutorch {
```
msluszniak, line 8:
```suggestion
namespace rnexecutorch {

```

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.h
```diff
@@ -0,0 +1,29 @@
...
+  void unload() noexcept;
+
+protected:
+  std::shared_ptr<react::CallInvoker> callInvoker;
+  std::unique_ptr<BaseModel> encoder_;
+  std::unique_ptr<BaseModel> decoder_;
+
+private:
+  size_t memorySizeLowerBound;
+};
```
msluszniak, line 29:
```suggestion
};

```

packages/react-native-executorch/common/rnexecutorch/models/EncoderDecoderBase.cpp
```diff
@@ -0,0 +1,19 @@
...
+      encoder_(std::make_unique<BaseModel>(encoderPath, callInvoker)),
+      decoder_(std::make_unique<BaseModel>(decoderPath, callInvoker)) {};
+
+size_t EncoderDecoderBase::getMemoryLowerBound() const noexcept {
+  return encoder_->getMemoryLowerBound() + decoder_->getMemoryLowerBound();
+}
+void EncoderDecoderBase::unload() noexcept {
+  encoder_.reset(nullptr);
+  decoder_.reset(nullptr);
+}
```
msluszniak:
```suggestion
}

void EncoderDecoderBase::unload() noexcept {
  encoder_.reset(nullptr);
  decoder_.reset(nullptr);
}

```

packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
```diff
@@ -39,4 +43,4 @@ class BaseModel {
   std::size_t memorySizeLowerBound{0};
   std::vector<int32_t> getTensorShape(const executorch::aten::Tensor &tensor);
 };
-} // namespace rnexecutorch
\ No newline at end of file
+} // namespace rnexecutorch
```
msluszniak, line 47:
```suggestion

} // namespace rnexecutorch
```

packages/react-native-executorch/common/rnexecutorch/data_processing/FFT.cpp
```diff
@@ -0,0 +1,20 @@
...
+FFT::~FFT() {
+  pffft_destroy_setup(pffftSetup_);
+  pffft_aligned_free(work_);
+}
+
+void FFT::doFFT(float *in, std::vector<std::complex<float>> &out) {
+  pffft_transform_ordered(pffftSetup_, in,
+                          reinterpret_cast<float *>(out.data()), work_,
+                          PFFFT_FORWARD);
+}
```
msluszniak, line 19:
```suggestion
}

```

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -48,4 +48,4 @@ int32_t TokenizerModule::tokenToId(std::string token) const {
 std::size_t TokenizerModule::getMemoryLowerBound() const noexcept {
   return memorySizeLowerBound;
 }
-} // namespace rnexecutorch
\ No newline at end of file
+} // namespace rnexecutorch
```
msluszniak, line 52:
```suggestion

} // namespace rnexecutorch
```

msluszniak (commented):
Unification of spaces between namespace brackets.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
pweglik: Those kinds of nitpicks should be done with code formatter, not by manual review ;)

msluszniak: These changes will be not checked by the formatter because it uses default clang tidy setup which accepts both of this changes. To make these kind of things checked by the formatter we will need to remove current default clang format setup and create a specialized custom one. So for the time when there is no such thing I believe we have no other way than manual fixes.

docs/docs/typescript-api/SpeechToTextModule.md
```diff
@@ -42,8 +42,8 @@ const transcribedText = await SpeechToTextModule.transcribe(waveform);
...
+| `decode`              | `(tokens: number[]) => Promise<number>`                                                                                                                                                                                                                                                                  | Runs the decoder of the model. Returns a single token representing a next token in the output. It uses internal cached encodings from previous `encode` call, meaning that you have to call `encode` prior to decoding.                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
mkopcins:
```suggestion
| `decode`              | `(tokens: number[]) => Promise<number>`                                                                                                                                                                                                                                                                  | Runs the decoder of the model. Returns a single token representing a next token in the output. It uses internal cached encodings from the most recent `encode` call, meaning that you have to call `encode` prior to decoding.                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
pweglik: So we may let it be instead of nitpicking - if it's not enforced by formatter. And not produce overhead for dev checking their review and for yourself manually checking every curly bracket :D

msluszniak: >So we may let it be instead of nitpicking

I strongly disagree. This way I point out the correct styling for the future, increase self-awareness about such kind of styling. If someone will learn by the time the non-written project code styling then it won't be necessary to add such comments in the future ;) And I provided it given to Kuba, so it was only click and push. All cognitive effort was removed because Kuba didn't needed to search anything, so I guess the overhead from his side was near zero.

pweglik: It is increasing awareness about totally subjective newline spacing. If it is really important and part of some kind of C++ standard formatting then we should it it to clang formatter options.

> non-written project code styling

Such thing shouldn't exist or else we'll need someone for full time job to format our PRs.

> All cognitive effort was removed

Yours wasn't and we really value it and we wouldn't want to waste your time

msluszniak: >It is increasing awareness about totally subjective newline spacing

Before we establish some standard then yeah, they might look kinda random. But I'm very glad that this discussion started because this is a great occasion to actually create some standard like: only current LLVM formatter, everything that is compliant shouldn't be explicitly change. I guess that this should be decision of the whole team, and I will be ok with every version which will be chosen. From my side, I will suggest to changes to default LLVM which will force to change the default setup. 
1. Add one new line after opening bracket of namespace and and one new line before ending bracket of namespace. This makes the code separated in a pretty way.
2. Do not align preprocessor instructions maximally to the left, but aligned them to the content of line below.
3. Force to add ending newline to make all files POSIX compliant.

This way we will achieve both my suggested formatting changes and no manual interventions in the future :).

>Yours wasn't and we really value it and we wouldn't want to waste your time

I was quite important for me, so I don't feel I wasted my 2 minutes ;)

PR #382, opened 2025-06-09
Refactor model URL constants

msluszniak:
## Description

This PR refactors URLs for models. This also exposed great inconsistencies in naming across hugging face repository:

* Only for `llama3_2-3B_qat_lora.pte` there is `-` between name and number of parameters, while for the rest is `_`.
* For `WHISPER_TINY_DECODER` `en` is concatenated in the first part of URL with `.` and not with `-` as in the rest of examples.
* There is word `all` in `efficientnet_v2_s_coreml_all` and only for that model.
* For some of the models initial name is concatenated with `-` and the model name as well - `phi-4-mini` and `phi-4-mini_bf16.pte` while in others there is `-` and `_` e.g. `qwen-2.5` and `qwen2_5_3b_8da4w.pte`.
* Some model names adds backend info at the beginning: `xnnpack_whisper_encoder.pte` and some on the end or almost end: `style_transfer_udnie_xnnpack.pte`.
* `ssdlite320-mobilenet-v3-large` v3 here is concatenated with `-` but in model name not: `ssdlite320-mobilenetv3-large.pte`.
* For craft detector number means size of the intput tensor while for the other suffixed numbers means model sizes.
* Some models B and M meaning billion and million of parameters is capitalized in both first part of url and name: `smolLm-2-135M` and `smolLm2_135M_bf16.pte` and for some not: `qwen-2.5-1.5B` and `qwen2_5_0_5b_8da4w.pte`.

And there are even more, but you will spot them while correcting rest. I think that there is need to make this unified. And then I will be able to refactor this even more.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (improves or adds clarity to existing documentation)

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

No apply

### Screenshots

No apply

### Related issues

No apply

### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation accordingly
- [x] My changes generate no new warnings

PR #397, opened 2025-06-13
feat: Image embeddings

pweglik:
## Description

Introducing image embeddings feature.
TODO:
benchmarks
Everything else is good to review.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (improves or adds clarity to existing documentation)

### Tested on

- [x] iOS
- [x] Android

### Related issues

#353 

### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation accordingly
- [ ] My changes generate no new warnings

Issue #359, opened 2025-06-04
Make loading models easier


pweglik:
Right now we need to pass all 3 paths/urls, but they are already premade for most of models configurations, so instead:
```
 const llama = useLLM({
    modelSource: LLAMA3_2_1B,
    tokenizerSource: LLAMA3_2_TOKENIZER,
    tokenizerConfigSource: LLAMA3_2_TOKENIZER_CONFIG,
  });
```
we could have:
```
 const llama = useLLM({
    model: LLAMA3_2_1B,
  });
```

We should introduce it as a non-breaking change, make all attributes optional and add checks that either model or the previous trio needs to be passed

chmjkb:
We could just export preset objects, for example:
```typescript
export const LLAMA_3_2_1B_PRESET =  {
    modelSource: LLAMA3_2_1B,
    tokenizerSource: LLAMA3_2_TOKENIZER,
    tokenizerConfigSource: LLAMA3_2_TOKENIZER_CONFIG,
  }
```
Then the user would just do:
```typescript
import { LLAMA_3_2_1B_PRESET } from 'react-native-executorch';

const llm = useLLM(LLAMA_3_2_1B_PRESET)
```

PR #430, opened 2025-07-01
Refactor model URL constants

msluszniak:
## Description

Same as #382, but now I have no permission to push directly into separate branch on repo so changing to PR from fork.

### DIFF
--- a/.cspell-wordlist.txt
+++ b/.cspell-wordlist.txt
@@ -52,3 +52,5 @@ MPNET
 QINT
 FNUZ
 wordlist
+coreml
+mobilenetv
--- a/packages/react-native-executorch/src/constants/modelUrls.ts
+++ b/packages/react-native-executorch/src/constants/modelUrls.ts
@@ -1,196 +1,159 @@
 import { Platform } from 'react-native';
 
-// LLM's
+const BASE_URL_PREFIX =
+  'https://huggingface.co/software-mansion/react-native-executorch';
+const VERSION_TAG = 'resolve/v0.4.0';
+const NEXT_DEV_VERSION_TAG = 'resolve/v0.5.0';
 
-// LLAMA 3.2
-export const LLAMA3_2_3B =
-  'https://huggingface.co/software-mansion/react-native-executorch-llama-3.2/resolve/v0.4.0/llama-3.2-3B/original/llama3_2_3B_bf16.pte';
-export const LLAMA3_2_3B_QLORA =
-  'https://huggingface.co/software-mansion/react-native-executorch-llama-3.2/resolve/v0.4.0/llama-3.2-3B/QLoRA/llama3_2-3B_qat_lora.pte';
-export const LLAMA3_2_3B_SPINQUANT =
-  'https://huggingface.co/software-mansion/react-native-executorch-llama-3.2/resolve/v0.4.0/llama-3.2-3B/spinquant/llama3_2_3B_spinquant.pte';
-export const LLAMA3_2_1B =
-  'https://huggingface.co/software-mansion/react-native-executorch-llama-3.2/resolve/v0.4.0/llama-3.2-1B/original/llama3_2_bf16.pte';
-export const LLAMA3_2_1B_QLORA =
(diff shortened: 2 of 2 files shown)


msluszniak:
>Add info to change VERSION_TAG in model url constants. I created a https://github.com/software-mansion/react-native-executorch/pull/382which have introduced such tag. @pweglik in one https://github.com/software-mansion/react-native-executorch/pull/397 added models that are introduced with version tag for currently developed version aka 0.5.0. As a consequence, I probably introduce something like NEXT_DEV_VERSION_TAG in rebasing the PR. These will need to be changed once published a new version. I can add this one separately once rebase and push my PR.

This is what I meant to change the version tag. Right now, we have:
```ts
const VERSION_TAG = 'resolve/v0.4.0';
const NEXT_DEV_VERSION_TAG = 'resolve/v0.5.0';
```
And every time new version is release, we need to change the version of `VERSION_TAG` and `NEXT_DEV_VERSION_TAG` one up, and replace every `NEXT_DEV_VERSION_TAG` in the urls with `VERSION_TAG` since it is no longer next dev.
cc: @mkopcins @pweglik

pweglik:
This is good suggestion, we should probably complete it alongside with : https://github.com/software-mansion/react-native-executorch/issues/359

msluszniak:
@pweglik I don't know the list of models that require such presets from #359 . If you could send me some instructions I can implement this as e.g. present for LLM will have model, tokenizer, and config but text embeddings will have model, tokenizer, and mean pooling parameters. Otherwise, feel free to contribute to this PR.

mkopcins:
You can try to match it by names, that should be fairly straight forward, for speech-to-text it is already defined, so that's that.  You can add mean pooling falg alongside that. Regarding how to do that I think a simple json for all models on our huggingface as a starting point would be just fine. I am not sure how to generalize it beyond our models.

apps/llm/app/llm_structured_output/index.tsx
```diff
@@ -75,12 +73,8 @@ function LLMScreen() {
   const textInputRef = useRef<TextInput>(null);
   const { setGlobalGenerating } = useContext(GeneratingContext);
 
-  const llm = useLLM({
-    // try out 4B model it this one struggles with following structured output
-    modelSource: QWEN3_1_7B_QUANTIZED,
-    tokenizerSource: QWEN3_TOKENIZER,
-    tokenizerConfigSource: QWEN3_TOKENIZER_CONFIG,
-  });
+  // try out 4B model it this one struggles with following structured output
```
NorbertKlockiewicz:
```suggestion
  // try out 4B model if this one struggles with following structured output
```

NorbertKlockiewicz (approved):

msluszniak:
Currently, the whole codebase must be revised because there are situations where there is the following ussage:
```ts
await ImageSegmentationModule.load(DEEPLAB_V3_RESNET50);
```

and this one:

```ts
  const model = useImageSegmentation({
    modelSource: DEEPLAB_V3_RESNET50,
  });
 ```

Also onDownloadProgressCallback is used is a way that we currently do not support in this issue:
```ts
await LLMModule.load({
  modelSource: LLAMA3_2_1B_QLORA,
  tokenizerSource: LLAMA3_2_TOKENIZER,
  tokenizerConfigSource: LLAMA3_2_TOKENIZER_CONFIG,
  onDownloadProgressCallback: printDownloadProgress,
});
```

Ideally, we need to change every single usage in the codebase, to use presets, that are exposed to the end users.

msluszniak:
@NorbertKlockiewicz do you have any idea how to deal with this?

pweglik:
Regarding this:
>Also onDownloadProgressCallback is used is a way that we currently do not support in this issue:
 ```
await LLMModule.load({
  modelSource: LLAMA3_2_1B_QLORA,
  tokenizerSource: LLAMA3_2_TOKENIZER,
  tokenizerConfigSource: LLAMA3_2_TOKENIZER_CONFIG,
  onDownloadProgressCallback: printDownloadProgress,
});
```
I think we can simply change it to:
```
await LLMModule.load({
  ...LLAMA3_2_1B,
  onDownloadProgressCallback: printDownloadProgress,
});
```

Regarding the other issue - yeah, I think we need to go through every place in docs and example apps and change it

msluszniak:
Closing in favour of #477

PR #433, opened 2025-07-02
build: fix problems with crashes on iPads

NorbertKlockiewicz:
## Description

This PR updates static libraries(.a) files for iOS, so they support iOS in version 16 and newer, it also solves the problem with crashes on iPads. Also missing libkernels_portable was added. I suggest to wait with this PR until all features are rewritten to CPP and then also remove `s.ios.vendored_frameworks` line from podspec as it won't be needed anymore, and currently it overwrites linked static libraries

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -16,42 +16,42 @@ Pod::Spec.new do |s|
...
-      '-framework "MetalPerformanceShaders"', 
-      '-framework "MetalPerformanceShadersGraph"', 
       "-force_load \"#{et_binaries_path}\"/libbackend_coreml_simulator.a", 
       "-force_load \"#{et_binaries_path}\"/libbackend_mps_simulator.a", 
       "-force_load \"#{et_binaries_path}\"/libbackend_xnnpack_simulator.a", 
       "-force_load \"#{et_binaries_path}\"/libexecutorch_simulator.a", 
       "-force_load \"#{et_binaries_path}\"/libkernels_custom_simulator.a", 
       "-force_load \"#{et_binaries_path}\"/libkernels_optimized_simulator.a", 
       "-force_load \"#{et_binaries_path}\"/libkernels_quantized_simulator.a",
+      "\"#{et_binaries_path}\"/libkernels_portable_simulator.a",
```
chmjkb, line 54: I believe you can use vendored_libraries instead of OTHER_LDFLAGS, a bit cleaner though. The latter was used mostly due to the need of passing the --force-load flag. Same goes for tokenizers. You can try to make it work :D

msluszniak:
All cpp files are in, so I guess there is nothing blocking this PR anymore.

PR #435, opened 2025-07-06
Add clang-format and pre-commit check for clang-format

msluszniak:
### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

To check if pre-commit works add a new file that do not conform the clang format check and try to commit this file. Then, check if the file was formatted automatically.

### Checklist


.clang-format-ignore
```diff
@@ -0,0 +1,18 @@
...
+*.lock
+third-party/
+**/third-party/
+node_modules/
+**/node_modules/
+docs/
+apps/
+./packages/react-native-executorch/common/ada/
+./packages/react-native-executorch/ios/
+./packages/react-native-executorch/android/
```
msluszniak:
```suggestion
./packages/react-native-executorch/android/

```

msluszniak:
I have a problem with this one. After pushing changes with this commit and every commit that is children of this one, some random file of size 86.5MB is pushed, and I have no idea why is that.

pweglik:
Don't we already run clang on cpp files? Defined in `lefthook.yml`?

msluszniak:
Ok, you're right. Right now, it is used default clang format which is llvm one, the same as in the PR, so it is indeed redundant. However, this one cpp file is not formatted, so let's merge only this one formatted file then.

chmjkb (approved):

PR #445, opened 2025-07-14
refactor: ImageEmbedding and TextEmbedding, remove normalization, update URLs

jakmro:
## Description

Changes:
- Refactored CLIP URLs
- Updated textEmbeddings models URLs to v0.5.0
- Refactored textEmbeddings demo
- Removed normalization from ImageEmbeddings
- Refactored ImageEmbeddings c++ code
- Removed attentionMask from TextEmbeddings postProcessing

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/apps/text-embeddings/utils/math.ts
+++ b/apps/text-embeddings/utils/math.ts
@@ -1,0 +1,7 @@
+export const dotProduct = (a: Float32Array, b: Float32Array) => {
+  let sum = 0;
+  for (let i = 0; i < a.length; i++) {
+    sum += a[i] * b[i];
+  }
+  return sum;
+};
--- a/apps/text-embeddings/utils/math.ts
+++ b/apps/text-embeddings/utils/math.ts
@@ -1,4 +1,8 @@
 export const dotProduct = (a: Float32Array, b: Float32Array) => {
+  if (a.length !== b.length) {
+    throw new Error('Vectors must be of the same length');
+  }
+
   let sum = 0;
   for (let i = 0; i < a.length; i++) {
     sum += a[i] * b[i];
--- a/packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp
@@ -1,0 +1,27 @@
+#include "BaseEmbeddings.h"
+
+#include <span>
+
+namespace rnexecutorch {
+
+BaseEmbeddings::BaseEmbeddings(const std::string &modelSource,
+                               std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker) {}
+
+std::shared_ptr<OwningArrayBuffer>
+BaseEmbeddings::postprocess(const Result<std::vector<EValue>> &forwardResult) {
+  auto forwardResultTensor = forwardResult->at(0).toTensor();
+  auto dataPtr = forwardResultTensor.mutable_data_ptr();
+  auto outputNumel = forwardResultTensor.numel();
+
+  std::span<float> modelOutput(static_cast<float *>(dataPtr), outputNumel);
+
+  auto createBuffer = [](const auto &data, size_t size) {
+    auto buffer = std::make_shared<OwningArrayBuffer>(size);
+    std::memcpy(buffer->data(), data, size);
+    return buffer;
+  };
+  return createBuffer(modelOutput.data(), modelOutput.size_bytes());
+}
+
+} // namespace rnexecutorch
--- a/packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
@@ -9,7 +9,7 @@ namespace rnexecutorch {
 ImageEmbeddings::ImageEmbeddings(
     const std::string &modelSource,
     std::shared_ptr<react::CallInvoker> callInvoker)
-    : BaseModel(modelSource, callInvoker) {
+    : BaseEmbeddings(modelSource, callInvoker) {
   auto inputTensors = getAllInputShapes();
   if (inputTensors.size() == 0) {
     throw std::runtime_error("Model seems to not take any input tensors.");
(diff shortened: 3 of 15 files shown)


packages/react-native-executorch/common/rnexecutorch/models/image_embeddings/ImageEmbeddings.cpp
```diff
@@ -31,30 +32,31 @@ ImageEmbeddings::generate(std::string imageSource) {
...
-  return buffer;
+std::shared_ptr<OwningArrayBuffer>
+ImageEmbeddings::postprocess(std::span<float> modelOutput) {
+  auto createBuffer = [](const auto &data, size_t size) {
+    auto buffer = std::make_shared<OwningArrayBuffer>(size);
+    std::memcpy(buffer->data(), data, size);
+    return buffer;
+  };
+
+  return createBuffer(modelOutput.data(), modelOutput.size_bytes());
```
msluszniak: This code is 1-to-1 duplication of code from text embeddings. Please refactor it so you can use them without code duplication.

packages/react-native-executorch/common/rnexecutorch/models/image_embeddings/ImageEmbeddings.cpp
```diff
@@ -31,30 +32,31 @@ ImageEmbeddings::generate(std::string imageSource) {
...
-  }
-
-  auto &outputs = result.get();
-
-  if (outputs.size() > 1) {
-    throw std::runtime_error("It returned multiple outputs!");
+  auto forwardResult = BaseModel::forward(inputTensor);
+  if (!forwardResult.ok()) {
+    throw std::runtime_error(
+        "Failed to forward, error: " +
```
msluszniak: `"Fail to forward"` is not informative, I'd prefer something like: `"Function forward in ImageEmbeddings failed with error code: <and so on>"`

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
```diff
@@ -28,35 +28,18 @@ ImageEmbeddings::ImageEmbeddings(
 }
 
 std::shared_ptr<OwningArrayBuffer>
-ImageEmbeddings::generate(std::string imageSource) {
+ImageEmbeddings::generate(const std::string imageSource) {
```
msluszniak: Use const reference, and if not possible pass just `std::string`, top const qualifiers in arguments of function declaration are automatically dropped.

packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <span>
+
+namespace rnexecutorch {
+
+BaseEmbeddings::BaseEmbeddings(const std::string &modelSource,
+                               std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker) {}
+
+std::shared_ptr<OwningArrayBuffer>
+BaseEmbeddings::postprocess(const Result<std::vector<EValue>> &forwardResult) {
```
msluszniak, line 12: Is this post-processing for all types of embeddings? If so, mark this method `final` so it won't be overridden.

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.h
```diff
@@ -4,22 +4,19 @@
...
-#include <rnexecutorch/models/BaseModel.h>
-#include <span>
+#include <rnexecutorch/models/embeddings/BaseEmbeddings.h>
 
 namespace rnexecutorch {
 using executorch::extension::TensorPtr;
 using executorch::runtime::EValue;
 
-class ImageEmbeddings : public BaseModel {
+class ImageEmbeddings : public BaseEmbeddings {
```
msluszniak: mark `ImageEmbeddings` class final

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.h
```diff
@@ -4,22 +4,19 @@
...
 public:
   ImageEmbeddings(const std::string &modelSource,
                   std::shared_ptr<react::CallInvoker> callInvoker);
-  std::shared_ptr<OwningArrayBuffer> generate(std::string imageSource);
+  std::shared_ptr<OwningArrayBuffer> generate(const std::string imageSource);
```
msluszniak: ~Const ref~ or remove const

packages/react-native-executorch/common/rnexecutorch/models/embeddings/text/TextEmbeddings.h
```diff
@@ -11,7 +10,7 @@ struct TokenIdsWithAttentionMask {
   std::vector<int64_t> attentionMask;
 };
 
-class TextEmbeddings : public BaseModel {
+class TextEmbeddings : public BaseEmbeddings {
```
msluszniak: mark `TextEmbeddings` class final

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
msluszniak: https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

msluszniak:
@jakmro note that also code of constructors in `StyleTransfer`, `ObjectDetection`, and `ImageSegmentation` has duplications. It might be refactored similarly as here in a separate PR ;)

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
chmjkb: I don't think he should pass const ref here, given that this function will be the owner of that string:
```
      if constexpr (std::is_void_v<decltype(std::apply(
                        std::bind_front(FnPtr, model), argsConverted))>) {
        // For void functions, just call the function and return undefined
        std::apply(std::bind_front(FnPtr, model), std::move(argsConverted));
        return jsi::Value::undefined();
      } else {
        // For non-void functions, capture the result and convert it
        auto result =
            std::apply(std::bind_front(FnPtr, model), std::move(argsConverted));
        return jsiconversion::getJsiValue(std::move(result), runtime);
      }
 ```
 i believe we've had this discussion a while back in DMs

Comment (chmjkb) [reply packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp]: but I agree that it shoul;dn't be const though.

Comment (msluszniak) [reply packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp]: Oh, this is the case, so remove const then.

File: packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp (hunk 4 above)
Comment (jakmro) [reply packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp:12]: I think it's best to leave it as is. To use final, I'd need to declare this function as virtual, but since there's no need for polymorphism here at the moment, I don't think that's necessary. WDYT?

**Comment by jakmro** [comment]
> @jakmro note that also code of constructors in `StyleTransfer`, `ObjectDetection`, and `ImageSegmentation` has duplications. It might be refactored similarly as here in a separate PR ;)

Might look at that in the future, thanks ;)

File: packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp (hunk 4 above)
Comment (msluszniak) [reply packages/react-native-executorch/common/rnexecutorch/models/embeddings/BaseEmbeddings.cpp:12]: Yeah, if postprocess is not an integral part of BaseModel just leave it as it is.

**Review by msluszniak — APPROVED** [review]

File: apps/text-embeddings/utils/math.ts [hunk 8]
```diff
@@ -0,0 +1,7 @@
+export const dotProduct = (a: Float32Array, b: Float32Array) => {
+  let sum = 0;
+  for (let i = 0; i < a.length; i++) {
+    sum += a[i] * b[i];
+  }
+  return sum;
+};
```
mkopcins: check for both array lengths

PR #450, opened 2025-07-16
feat: enable 'no-console' lint rule for executorch src except Logger, add Logger.log method

jakmro:
## Description

Changes:
- Added a new override to enforce the no-console rule for files in `react-native-executorch/src`, excluding Logger.ts
- Added a static log method to Logger class

### Type of change

### Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,7 +38,6 @@ module.exports = {
   overrides: [
     {
       files: ['packages/react-native-executorch/src/**/*.{js,jsx,ts,tsx}'],
-      excludedFiles: ['packages/react-native-executorch/src/common/Logger.ts'],
       rules: {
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -36,6 +36,13 @@ module.exports = {
   },
   plugins: ['prettier', 'markdown'],
   overrides: [
+    {
+      files: ['packages/react-native-executorch/src/**/*.{js,jsx,ts,tsx}'],
+      excludedFiles: ['packages/react-native-executorch/src/common/Logger.ts'],
+      rules: {
+        'no-console': 'warn',
+      },
+    },
```
pweglik: I think it is better to put an eslint exception in file itself instead of littering in config file

Issue #499, opened 2025-08-14
Change `useMemo()` to `useRef()` In hooks


mlodyjesienin:
Currently in all hooks we use `useMemo()` with empty dependency array, but more fitting to our needs approach is to make use of `useRef()` instead.

PR #509, opened 2025-08-19
refactor: change useMemo to useState in hooks.

mlodyjesienin:
## Description

in React [`useMemo` docs](https://react.dev/reference/react/useMemo#caveats) they explicitly state, that `useMemo` is an performance optimization and it can throw away the cached value if it is beneficial. Therefore relying on `useRef` or `useState` is better in situations where you need to be sure that object is created exactly once. 

Although the `useRef` hook  is more "idiomatic",  it  lacks a built-in lazy initializer, requiring a more verbose checks. 
That is why, `useState` seems like the best option to make sure that objects are initialized exactly once in a clean way.

### Introducing Breaking Changes

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Related Issues 

#499 

### Checklist


packages/react-native-executorch/src/hooks/natural_language_processing/useTokenizer.ts
```diff
@@ -14,7 +14,7 @@ export const useTokenizer = ({
   const [isReady, setIsReady] = useState(false);
   const [isGenerating, setIsGenerating] = useState(false);
   const [downloadProgress, setDownloadProgress] = useState(0);
-  const _tokenizer = useMemo(() => new TokenizerModule(), []);
+  const [_tokenizer] = useState(() => new TokenizerModule());
```
jakmro: Lets rename this to tokenizerInstance

jakmro (changes requested):

jakmro (approved):

PR #534, opened 2025-08-25
feat: added configurable num of threads for xnnpack to fix android performance

mkopcins:
## Description

After migrating llms to cpp architecture we saw significant drop in performance (up to 10x slower). After debugging the main culprit turned out to be the number of threads spawned for xnnpack threadpool, which defaulted to the number of cores (the underlying reason is still unknown). 

- [x] Yes - migrating back to cpp architecture for llms
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
```diff
@@ -74,5 +83,19 @@ void RnExecutorchInstaller::injectJSIBindings(
...
+      *jsiRuntime, "loadSpeechToText",
+      RnExecutorchInstaller::loadModel<SpeechToText>(jsiRuntime, jsCallInvoker,
+                                                     "loadSpeechToText"));
+
+#ifdef defined(__ANDROID__) && defined(__aarch64__)
+  auto num_of_cores = std::min(
+      ::executorch::extension::cpuinfo::get_num_performant_cores(), 3u);
+  ::executorch::extension::threadpool::get_threadpool()
+      ->_unsafe_reset_threadpool(num_of_cores);
+  log(LOG_LEVEL::Info, "Configuring xnnpack for ", num_of_cores, "threads");
```
msluszniak:
```suggestion
  log(LOG_LEVEL::Info, "Configuring xnnpack for ", num_of_cores, " threads");
```

msluszniak: or use `std::format("Configuring xnnpack for {} threads", num_of_cores)`. I have a plan to change all bulky string compositions like in error messages to `std::format`.

packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
```diff
@@ -6,12 +6,19 @@
...
 #include <rnexecutorch/models/embeddings/text/TextEmbeddings.h>
 #include <rnexecutorch/models/image_segmentation/ImageSegmentation.h>
+#include <rnexecutorch/models/llm/LLM.h>
 #include <rnexecutorch/models/object_detection/ObjectDetection.h>
 #include <rnexecutorch/models/ocr/OCR.h>
 #include <rnexecutorch/models/speech_to_text/SpeechToText.h>
 #include <rnexecutorch/models/style_transfer/StyleTransfer.h>
 #include <rnexecutorch/models/vertical_ocr/VerticalOCR.h>
 
+#if defined(__ANDROID__) && defined(__aarch64__)
```
msluszniak, line 16: I guess you should add `#include <format>` in this block.

mkopcins: not needed, works just fine without include

msluszniak: Yeah, I know, but in general it is a good practise to include headers that contains functionality directly. There is clang tidy check for such includes: https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html, specifically `MissingIncludes` option.

This protects you from this kind of situations:
I have a file `z.cpp` that uses some functionality from `x.h` but I don't include `x.h`, because it works without it. I have plenty of other includes, and one, lets say `y.h` includes `x.h`, so it works. One day, I'd decide to change `y.h` and remove `x.h` from includes because it is not used anymore. But now my `z.cpp` stops to work.

There are plenty of such missing includes in our code, and I'm trying to add them when I spot some ;)

apps/llm/ios/llm/llm.entitlements
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-<dict/>
+<dict>
+	<key>com.apple.developer.kernel.increased-debugging-memory-limit</key>
+	<true/>
+	<key>com.apple.developer.kernel.increased-memory-limit</key>
+	<true/>
+</dict>
```
chmjkb, line 9: I'm wondering if we truly need this, the app works without this. And XCode won't let me run the app when this is specified.

packages/react-native-executorch/android/libs/arm64-v8a/libexecutorch.so
chmjkb, line 1: Do we need this file? shared/static libs are already stored in third-party/android/libs/

packages/react-native-executorch/android/libs/x86_64/libexecutorch.so
chmjkb, line 1: Do we need this file? shared/static libs are already stored in third-party/android/libs/

packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
chmjkb: I know that we kind of decided to stray away from supporting any other architectures than arm64. Is this a reason for this check here or maybe some other reason that I don't see? :D

packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
```diff
@@ -74,5 +82,23 @@ void RnExecutorchInstaller::injectJSIBindings(
...
+      *jsiRuntime, "loadSpeechToText",
+      RnExecutorchInstaller::loadModel<SpeechToText>(jsiRuntime, jsCallInvoker,
+                                                     "loadSpeechToText"));
+
+#if defined(__ANDROID__) && defined(__aarch64__)
+  auto num_of_perf_cores =
+      ::executorch::extension::cpuinfo::get_num_performant_cores();
+  log(LOG_LEVEL::Info,
+      std::format("Detected {} performant cores", num_of_perf_cores));
+  auto num_of_cores = static_cast<uint32_t>(num_of_perf_cores / 2) + 1;
```
chmjkb, line 103: Can you add a comment explaining the formula here? This might not be clear for someone without the context

packages/react-native-executorch/third-party/android/libs/executorch/executorch.aar
chmjkb, line 1: I don't think we need this

packages/react-native-executorch/android/src/main/cpp/CMakeLists.txt
```diff
@@ -33,13 +33,37 @@ set(RN_VERSION_LINK_LIBRARIES
...
 # ------- Executorch -------
 
 add_library(executorch SHARED IMPORTED)
 
 set_target_properties(executorch PROPERTIES
                       IMPORTED_LOCATION "${LIBS_DIR}/executorch/${ANDROID_ABI}/libexecutorch.so")
 
+                      
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+  target_compile_definitions(react-native-executorch PRIVATE ARCH_ARM64)
```
chmjkb: Why not use 
`if(ANDROID_ABI STREQUAL "arm64-v8a")` instead, as it is done in the following?

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
mkopcins: We are not dropping support for x86, just the fix right now is not built for x86. x86 is for chromebooks and emulators afaik, so I don't think this is a high priority task anyway

PR #535, opened 2025-08-25
RR

mlodyjesienin:

- [ ] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


packages/react-native-executorch/common/rnexecutorch/models/ocr/Constants.h
```diff
@@ -3,32 +3,32 @@
...
+inline constexpr int32_t kMaxWidth =
+    kLargeRecognizerWidth + (kLargeRecognizerWidth * 0.15);
+inline constexpr int32_t kSingleCharacterMinSize = 70;
+inline constexpr int32_t kRecognizerImageSize = 1280;
+inline constexpr int32_t kVerticalLineThreshold = 20;
 
-inline const cv::Scalar MEAN(0.485, 0.456, 0.406);
-inline const cv::Scalar VARIANCE(0.229, 0.224, 0.225);
+inline const cv::Scalar kMean(0.485, 0.456, 0.406);
+inline const cv::Scalar kVariance(0.229, 0.224, 0.225);
```
msluszniak: Could you rename these to something better? ;)

msluszniak:
>Sometimes the nested namespaces are very long. For example having to write models::object_detection::types::Detection in JsiConversions.h seems a bit redundant.

This is not a problem as well. If you feel that this is too long you can always use `namespace types = models::object_detection::types` at the very beginning of the file and then access via `types::Detection`

>TokenizerModule .h /.cpp is currently in the very high rnexecutorch directory. It is right next to the RnExecutorchInstaller.h file. Seems way off, it has to be moved somewhere, but not really into the models dir. Maybe the solution is to create new Tokenizer dir? It is the only thing apart from the models, that we export to js. Maybe it should be somehow visible in the structure (like directory with models and tokenizers directories inside?).

Different dir is way better than unseparated files next to installer.

>Maybe we could introduce more directory-structure-like namespaces? For example right now the namespaces in data_processing directory are not connected, we have rnexecutorch::image_processing, rnexecutorch::file_utils, rnexecutorch::dsp and rnexecutorch::numerical. Maybe it all could be somehow refactored? My lovely friend Gemini proposed something like media directory with namespaces than like media::audio, media::image (but then numerical and file_utils should be moved somewhere else), but I am not sure if I like this approach.

I'm okay with anything here that will be well documented and consistent.

> currently things that are related to jsi are mostly in the plain rnexecutorch:: namespace. The exception is JsiConversions.h that lives in the rnexecutroch::jsi_conversions namespace (but it is actually in host_object directory, while we have jsi directory elsewhere.) How to refactor this?

Maybe use for all jsi related things `rnexecutorch::jsi` namespace?

>I do not like the EncoderDecoderBase files being at the same height as BaseModel. For example BaseEmbeddings are at different directory depth, and the structure of embeddings directory feels better. How to handle this?

Hmm, maybe enclose `BaseModel` and `EncoderDecoderBase` in separate dir? On the other hand, it makes sense that `EncoderDecoderBase` is higher than any particular model, since I guess there are a few models with encoder-decoder architecture.

msluszniak (commented):
I think you should also use k prefix for all constexpr and const objects that are unchanged and independent of any runtime values in codebase (all constexpr and some const). See: https://google.github.io/styleguide/cppguide.html#Constant_Names

mkopcins:
I am not a fan of the k-preifx naming convention. Might be because I'm used to UPPER_CASE for consts. Also I still think that even tho we have te `constants` namespace the UPPER_CASE add glance value, sort of pops more. Anyhow another benefit is that we have the same naming convention between .ts and .cpp files with UPPER_CASE consts.

msluszniak:
>I am not a fan of the k-preifx naming convention. Might be because I'm used to UPPER_CASE for consts. Also I still think that even tho we have te constants namespace the UPPER_CASE add glance value, sort of pops more. Anyhow another benefit is that we have the same naming convention between .ts and .cpp files with UPPER_CASE consts.

I see. Generally reasons for the UPPER_CASE convention for constants is historical. Before C++11 there was no `constexpr` and compile time constants were created exactly the same way as macros, via preprocessor's instructions. So it was reasonable to use the same convention. Now, since there are better ways to use C++ as a language to describe some concepts, and as a consequence ways to avoid clash with macro definitions, I think that convention with k-prefix is beneficial in long term. 

There are many older codebases that are unable to change UPPER_CASE naming convention because it is so deeply embedded into the code, that even they want to they simply cannot. Our library is still at the early stages, so it is easy now to make such changes, and once we stick to the old convention, one day there might be no way around to move back to the newer one. I think that using newer and generally preferred conventions are "long-term investments".

I believe that having the same convention in both typescript and C++ might be convenient, but I consider it rather as a matter of time when these two will diverge. Google suggests to use UPPER_CASE in typescript constants, and k-prefix for c++, so they deliberately decided to not keep that consistency.

ExecuTorch uses convention with k-prefix so it might be easy in the future to integrate their code into our codebase without changing conventions. Also creators of C++ standard tend to migrate to newer convention:
`M_PI in <math.h> -> std::numbers::pi_v<T> in <numbers>`

Also C++ core guidelines created by the creator of C++ and main chairman of C++ committee explicitly suggest to use UPPER_CASE only for macros: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-all-caps
What is more, they don't suggest you convention for naming constants, so what they mean is: "Use convention you'd like, but we explicitly discourage you from using UPPER_CASE for anything other than macros" ;p

msluszniak:
>I think you should also use k prefix for all constexpr and const objects that are unchanged and independent of any runtime values in codebase (all constexpr and some const). See: https://google.github.io/styleguide/cppguide.html#Constant_Names

What about this one?

PR #562, opened 2025-08-26
Refactor jsi conversion & NITs

msluszniak:
## Description

As in the title, some refactor of jsi.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

Needs to be manually testes / unit test (TODO)

### Checklist


packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -160,7 +111,7 @@ getValue<std::set<std::string, std::less<>>>(const jsi::Value &val,
 
   for (size_t i = 0; i < length; ++i) {
     jsi::Value element = array.getValueAtIndex(runtime, i);
-    result.insert(getValue<std::string>(element, runtime));
+    result.emplace(getValue<std::string>(element, runtime));
```
msluszniak:
```suggestion
    result.insert(getValue<std::string>(element, runtime));
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -209,12 +160,31 @@ inline std::vector<T> getArrayAsVector(const jsi::Value &val,
 
   for (size_t i = 0; i < length; ++i) {
     const jsi::Value element = array.getValueAtIndex(runtime, i);
-    result.push_back(getValue<T>(element, runtime));
+    result.emplace_back(getValue<T>(element, runtime));
```
msluszniak:
```suggestion
    result.push_back(getValue<T>(element, runtime));
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -29,8 +29,6 @@ template <typename T> T getValue(const jsi::Value &val, jsi::Runtime &runtime);
 template <typename T>
   requires meta::IsNumeric<T>
 inline T getValue(const jsi::Value &val, jsi::Runtime &runtime) {
-  static_assert(std::is_integral<T>::value || std::is_floating_point<T>::value,
-                "Only integral and floating-point types are supported");
```
msluszniak, line 33: This assert is deleted because `requires meta::IsNumeric<T>` guarantees that the function will instantiate only if a value `T` already meets this condition (`meta::isNumeric<T>` is implemented as `std::is_arithmetic_v<T>`), so it is completely redundant.

msluszniak:
Awaiting for the test plan for the modified files

chmjkb (approved):

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -45,29 +43,13 @@ inline std::string getValue<std::string>(const jsi::Value &val,
   return val.getString(runtime).utf8(runtime);
 }
 
-template <>
 inline std::shared_ptr<jsi::Function>
```
msluszniak:
```suggestion
template <>
inline std::shared_ptr<jsi::Function>
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -83,8 +66,8 @@ inline JSTensorViewIn getValue<JSTensorViewIn>(const jsi::Value &val,
   tensorView.sizes.reserve(numShapeDims);
 
   for (size_t i = 0; i < numShapeDims; ++i) {
-    int dim = getValue<int>(shapeArray.getValueAtIndex(runtime, i), runtime);
-    tensorView.sizes.push_back(static_cast<int32_t>(dim));
+    int32_t dim = getValue<int32_t>(shapeArray.getValueAtIndex(runtime, i), runtime);
```
mkopcins, line 69: Just out of curiosity, why here you opted for `int32_t` instead of `auto`?

msluszniak: because function specialisation for `int32_t` might return arbitrary type in general. Maybe here a function name kinda suggest that type probably be the same, but there is no guarantee. Here e.g.,

```cpp

template <typename T>
auto someValue(T value,  T diff_value){
  return static_cast<float>(value / diff_value);
}
```
result value will be different for different specialization (floating point vs. integer types) but result is always float. But `auto` is also reasonable option here since we know the implementation of `getValue`.

On the other hand, when dealing with constructors, modern c++ casts, the resultant type is obvious so you can always safely use `auto`.

Also if the type of vector `tensorView.sizes` would be visible here, I'll probably opt to use `auto`, but here it's not visible at first glance

PR #594, opened 2025-09-11
feat: add s2t unload and make streamInsert and streamStop sync

jakmro:
## Description

Changes:
- add s2t unload
- make streamInsert and streamStop sync

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


packages/react-native-executorch/src/modules/natural_language_processing/SpeechToTextModule.ts
```diff
@@ -38,6 +38,11 @@ export class SpeechToTextModule {
     );
   }
 
+  public unload(): void {
+    this.nativeModule.unload();
+    this.nativeModule = null;
+  }
+
```
chmjkb: This will fail if called twice, also this seems inconsistent across other modules, guess we should fix that too. For example, in BaseModel which is inherited from by StyleTransfer for example, we don't specify unload, but delete :D

chmjkb: VerticalOCR / OCR modules also provide `delete`, instead of `unload`.

Issue #565, opened 2025-08-27
Implement threadpool for running async JSI functions


mkopcins:
Currently we are spawning a new thread for each async function, instead a better and more efficient option is to have a threadpool initialised at startup and then utilise that for the async function calls

PR #603, opened 2025-09-15
feat: Implemented GlobalThreadPool for async functions

mkopcins:
## Description

Added singleton class GlobalThreadPool for single threadpool management so that we don't have to spawn new threads for each async function and instead we can delegate functions to the threadpool.

Also added pthreadpool and cpuinfo binaries for iOS to allow for XNNPack threadpool configuration just like on Android

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist


packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
```diff
@@ -0,0 +1,77 @@
...
+#pragma once
+
+#include <executorch/extension/threadpool/cpuinfo_utils.h>
+#include <memory>
+#include <mutex>
+#include <rnexecutorch/Log.h>
+#include <rnexecutorch/threads/HighPerformanceThreadPool.h>
+
+namespace rnexecutorch {
+namespace threads {
```
msluszniak: These namespaces can be squashed

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
```diff
@@ -0,0 +1,77 @@
...
+
+class GlobalThreadPool {
+private:
+  inline static std::unique_ptr<HighPerformanceThreadPool> instance;
+  inline static std::once_flag initFlag;
+
+  GlobalThreadPool() = delete;
```
msluszniak, line 15: Automatically generated ctor c-assignment are deleted implicitly because `instance` does not support copying. But moving ctor and assignment will be generated. You probably want to delete them as well ;)

Also it is good practice to delete constructors in `public` (better error messages when trying to instantiate) and define public members first, then private

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+#ifdef __APPLE__
+#include <sys/syscall.h>
+#endif
+
+#ifdef __ANDROID__
+#include <sys/types.h>
+#endif
+
+namespace rnexecutorch {
+namespace threads {
```
msluszniak: again, squash namespaces

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+enum class Priority { LOW, NORMAL, HIGH, REALTIME };
+
+struct ThreadConfig {
+  bool pinToPerformanceCores{true};
+  std::string namePrefix{"RN_ET_Worker"};
+};
+
+class HighPerformanceThreadPool {
+public:
+private:
```
msluszniak, line 131: Access specifiers are off in the `HighPerformanceThreadPool` class. Prefer `public` members at the very top, and then `protected` and `private`.

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+
+    std::vector<CoreInfo> cores;
+    const auto numOfCores = std::thread::hardware_concurrency();
+
+    for (int i = 0; i < numOfCores; i++) {
+      std::string path = "/sys/devices/system/cpu/cpu" + std::to_string(i) +
+                         "/cpufreq/cpuinfo_max_freq";
+      std::ifstream file(path);
+      if (!file.good())
+        break;
```
msluszniak:
```suggestion
      if (!file.good()) {
        break;
      }
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+      cores.push_back(info);
+    }
+
+    if (cores.empty()) {
+      log(LOG_LEVEL::Debug, "Could not detect CPU topology");
+      return;
+    }
+
+    // Sort by frequency
+    std::sort(cores.begin(), cores.end(),
```
msluszniak: from `<algorithm>`
```suggestion
    std::ranges::sort(cores,
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+
+    // Classify cores
+    const auto numOfPerfCores =
+        ::executorch::extension::cpuinfo::get_num_performant_cores();
+
+    for (int i = 0; i < cores.size(); ++i) {
+      if (i < numOfPerfCores) {
+        performanceCores.push_back(core.id);
+        log(LOG_LEVEL::Debug, "Performance core:", core.id, "(",
+            core.maxFreq / 1000000.0, "GHz)");
```
msluszniak: dividing by 10^6 to get GHz is kinda unintuitive to me, is maxFreq in KHz really?

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+    }
+#endif
+  }
+
+  void setThreadPriority() {
+    // pthread_setschedparam doesn't work on android because permissions reasons
+    // and in general does not provide visible improvements on iOS
+
+    // Set nice value as fallback or additional priority boost
+    const int nice_value = 0;
```
msluszniak: I you want to make it const, use constexpr here, otherwise it is not const, maybe `constinit`, but probably not applicable here.

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+
+      {
+        std::unique_lock<std::mutex> lock(queueMutex);
+        condition.wait(lock, [this] { return !taskQueue.empty() || !running; });
+
+        if (!running && taskQueue.empty())
+          break;
+
+        if (!taskQueue.empty()) {
+          item = std::move(const_cast<WorkItem &>(taskQueue.top()));
```
msluszniak, line 350: Is this const cast necessary here?

EDIT: ok I see signature of top `now`, `const_reference top() const;`, please ignore this comment

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+    // Create a packaged task
+    auto boundFunc =
+        std::bind(std::forward<Func>(func), std::forward<Args>(args)...);
+    auto task = std::make_unique<Task<decltype(boundFunc), ReturnType>>(
+        std::move(boundFunc));
+    auto future = task->getFuture();
+
+    // Add to queue
+    {
+      std::lock_guard<std::mutex> lock(queueMutex);
```
msluszniak: Prefer always `std::scoped_lock` over `std::lock_guard` if c++ version >= 17, https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-scoped-lock.html

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+  // Fire and forget task
+  template <typename Func, typename... Args>
+  void submitDetached(Func &&func, Args &&...args) {
+    submit(std::forward<Func>(func), std::forward<Args>(args)...);
+    // Future is destroyed, task still runs
+  }
+
+  void shutdown() {
+    if (!running.exchange(false))
+      return;
```
msluszniak:
```suggestion
    if (!running.exchange(false)) {
      return;
    }
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+
+    condition.notify_all();
+
+    for (auto &worker : workers) {
+      if (worker.joinable()) {
+        worker.join();
+      }
```
msluszniak: You can probably avoid this by using `std::jthread` instead of `std::thread`

packages/react-native-executorch/common/rnexecutorch/threads/ThreadUtils.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <executorch/extension/threadpool/cpuinfo_utils.h>
+#include <executorch/extension/threadpool/threadpool.h>
+#include <rnexecutorch/Log.h>
+
+namespace rnexecutorch {
+namespace threads {
```
msluszniak: squash

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+#ifdef __ANDROID__
+    struct CoreInfo {
+      int id;
+      long maxFreq;
+    };
+
+    std::vector<CoreInfo> cores;
+    const auto numOfCores = std::thread::hardware_concurrency();
+
+    for (int i = 0; i < numOfCores; i++) {
```
msluszniak: return type of `std::thread::hardware_concurrency` is `unsigned int`, so to safely compare signed and usinged types use `std::cmp_less` here
```suggestion
    for (int i = 0; std::cmp_less(i, numOfCores); ++i) {
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+
+    try {
+      item.task->execute();
+    } catch (const std::exception &e) {
+      log(LOG_LEVEL::Error, "Task failed:", e.what());
+      throw;
+    }
+
+    auto endTime = std::chrono::steady_clock::now();
+    auto executionTime = std::chrono::duration_cast<std::chrono::milliseconds>(
```
msluszniak: It's unused (same as `waitTime`), you probably wanted to log time here, right?

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+      std::lock_guard<std::mutex> lock(queueMutex);
+
+      if (!running) {
+        throw std::runtime_error("Thread pool is shutting down");
+      }
+
+      WorkItem item;
+      item.task = std::move(task);
+      item.priority = priority;
+      item.enqueueTime = std::chrono::steady_clock::now();
```
msluszniak: This is completely up to you, but I would probably define constructor for this struct, it's not classic POD because you have a non-copyable `std::unique_ptr` here. What are the benefits? More concise initialization, `priority` and `enqueueTime` might be marked `private`.

packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
```diff
@@ -1,5 +1,6 @@
 #include "RnExecutorchInstaller.h"
 
+#include <rnexecutorch/Log.h>
```
chmjkb: I think we can get rid of that

jakmro: Unused include

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
```diff
@@ -0,0 +1,77 @@
+// GlobalThreadPool.h
+#pragma once
+
+#include <executorch/extension/threadpool/cpuinfo_utils.h>
+#include <memory>
+#include <mutex>
+#include <rnexecutorch/Log.h>
```
chmjkb, line 8: redundant include

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
chmjkb: I can't see where the core variable is defined

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <executorch/extension/threadpool/cpuinfo_utils.h>
+#include <fstream>
+#include <functional>
+#include <future>
+#include <memory>
+#include <mutex>
+#include <optional>
```
chmjkb: unused header

packages/react-native-executorch/common/rnexecutorch/threads/ThreadUtils.h
```diff
@@ -0,0 +1,35 @@
...
+
+class ThreadUtils {
+public:
+  static void unsafeSetupThreadPool(uint32_t num_of_cores = 0) {
```
jakmro: I think this can be just a function living in a namespace `threads::utlis`, no need for a class with a static function

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
```diff
@@ -0,0 +1,77 @@
...
+
+class GlobalThreadPool {
+private:
+  inline static std::unique_ptr<HighPerformanceThreadPool> instance;
+  inline static std::once_flag initFlag;
+
+  GlobalThreadPool() = delete;
+
+public:
+  static void initialize(uint32_t numThreads = 0, ThreadConfig config = {}) {
```
jakmro: Consider using `std::optional` for the numThreads

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
```diff
@@ -0,0 +1,77 @@
...
+#include <executorch/extension/threadpool/cpuinfo_utils.h>
+#include <memory>
+#include <mutex>
+#include <rnexecutorch/Log.h>
+#include <rnexecutorch/threads/HighPerformanceThreadPool.h>
+
+namespace rnexecutorch {
+namespace threads {
+
+class GlobalThreadPool {
```
jakmro, line 13: Consider removing this class and leaving all functions inside the `threads::global_thread_pool` namespace

msluszniak: `HighPerformanceThreadPool` should be probably implemented similarly to this one: 
```cpp
class Singleton {
public:
    Singleton(const Singleton&) = delete;
    Singleton& operator=(const Singleton&) = delete;
    Singleton(Singleton&&) = delete;
    Singleton& operator=(Singleton&&) = delete;

    static Singleton& getInstance() {
        static std::once_flag initFlag;
        if (!instance) { 
            throw std::logic_error("Instance not initialized.");
        }
        return *instance;
    }

    static void initialize(int arg) {
        static std::once_flag initFlag;
        std::call_once(initFlag, [arg]() {
            instance = std::make_unique<Singleton>(arg);
        });
    }

private:
    int value;
    Singleton(int val) : value(val) {}

    static std::unique_ptr<Singleton> instance;
};

std::unique_ptr<Singleton> Singleton::instance = nullptr;
```

msluszniak:
Add conditional closing of this issue: https://github.com/software-mansion/react-native-executorch/issues/565 in the PR

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -24,8 +24,11 @@ Pod::Spec.new do |s|
     'MetalPerformanceShadersGraph'
   ]
 
+  pthreadpool_binaries_path = File.expand_path('$(PODS_TARGET_SRCROOT)/ios/libs/pthreadpool', __dir__)
+  cpuinfo_binaries_path = File.expand_path('$(PODS_TARGET_SRCROOT)/ios/libs/cpuinfo', __dir__)
+
   s.user_target_xcconfig = {
-    "HEADER_SEARCH_PATHS" => "$(PODS_TARGET_SRCROOT)/third-party/include",
+    "HEADER_SEARCH_PATHS" => "$(PODS_TARGET_SRCROOT)/third-party/include $(PODS_TARGET_SRCROOT)/../../third-party/executorch/backends/xnnpack/third-party/pthreadpool/include $(PODS_TARGET_SRCROOT)/../../third-party/executorch/backends/xnnpack/third-party/cpuinfo/include",
```
chmjkb: I think we should stick for one "source of truth" for includes, so I suggest copying the needed include to a specific dir in third-party/include. Im also not sure how does this cpuinfo differ from the cpuinfo that is already under third-party/include/cpuinfo.

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+
+  // Configuration
+  ThreadConfig config;
+
+private:
+  void detectCPUTopology() {
+#ifdef __ANDROID__
+    struct CoreInfo {
+      int id;
+      long maxFreq;
```
chmjkb:
```suggestion
      int32_t id;
      int64_t maxFreq;
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+  std::mutex queueMutex;
+  std::condition_variable condition;
+  std::atomic<bool> running{true};
+  std::atomic<size_t> activeWorkers{0};
+  std::atomic<size_t> totalTasksProcessed{0};
+
+#ifdef __ANDROID__
+  // Performance cores
+  std::vector<int> performanceCores;
+  std::vector<int> efficiencyCores;
```
chmjkb:
```suggestion
  std::vector<int32_t> performanceCores;
  std::vector<int32_t> efficiencyCores;
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,367 @@
...
+    }
+  }
+
+  void processTask(const WorkItem &item) {
+    activeWorkers++;
+
+    auto startTime = std::chrono::steady_clock::now();
+    auto waitTime = std::chrono::duration_cast<std::chrono::milliseconds>(
+                        startTime - item.enqueueTime)
+                        .count();
```
chmjkb: unused variable

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
mkopcins: Since I am using `log` in this file I want to have all needed includes present

mkopcins: Since I am using `log` in this file I want to have all includes present

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
mkopcins: I think you meant `GlobalThreadPool`, right? `HighPerformanceThreadPool` was never supposed to be a singleton

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
mkopcins: Yes, in linux system it is usually expressed in kHz

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
msluszniak: Currently, `HighPerformanceThreadPool` is used only in the context of `GlobalThreadPool`. I guess Jakub meant that `GlobalThreadPool` is only a wrapper on a class that might be a singleton itself. If in the future there might be any other way to use `HighPerformanceThreadPool` the current design will make more sense to me then.

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
mkopcins: jthread is not compiling for iOS

msluszniak: Eh, indeed this feature is not implemented in Apple Clang:
Screenshot_20250916_165222_Chrome

packages/react-native-executorch/react-native-executorch.podspec
mkopcins: the files were already present there, so I just removed this change from here

packages/react-native-executorch/common/rnexecutorch/threads/GlobalThreadPool.h
mkopcins: That was my idea when creating this, `HighPerformanceThreadPool` is not meant to be a singleton, but `GlobalThreadPool` is. Also semantically it makes more sense to me to keep `GlobalThreadPool` as a class instead of a namespace even though it doesn't have any nonstatic members.

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,364 @@
...
+    std::future<Result> getFuture() { return promise.get_future(); }
+
+  private:
+    Func func;
+    std::promise<Result> promise;
+  };
+
+  class WorkItem {
+  public:
+    WorkItem() {}
```
msluszniak: https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-default.html
```suggestion
    WorkItem() = default;
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,364 @@
...
+#ifdef __ANDROID__
+    struct CoreInfo {
+      int32_t id;
+      int64_t maxFreq;
+    };
+
+    std::vector<CoreInfo> cores;
+    const auto numOfCores = std::thread::hardware_concurrency();
+
+    for (int i = 0; std::cmp_less(i, numOfCores); ++i) {
```
msluszniak:
```suggestion
    for (int32_t i = 0; std::cmp_less(i, numOfCores); ++i) {
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,364 @@
...
+    std::ranges::sort(cores, [](const CoreInfo &a, const CoreInfo &b) {
+      return a.maxFreq > b.maxFreq;
+    });
+
+    // Classify cores
+    const auto numOfPerfCores =
+        ::executorch::extension::cpuinfo::get_num_performant_cores();
+
+    constexpr float kKiloToGigaRatio = 1e6;
+    for (int i = 0; i < cores.size(); ++i) {
```
msluszniak:
```suggestion
    for (size_t i = 0; i < cores.size(); ++i) {
```

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,364 @@
...
+
+    setThreadPriority();
+
+    log(LOG_LEVEL::Debug, "Worker", workerIndex,
+        "configured:", threadName.c_str());
+  }
+
+  void setCPUAffinity() {
+    // AFAIK it is not possible on iOS
+#ifdef __ANDROID__
```
msluszniak, line 283: You should wrap the whole function implementation with `#ifdef __ANDROID__` to not introduce overhead of translating empty function body on iOS. Check other applicable places

packages/react-native-executorch/common/rnexecutorch/threads/HighPerformanceThreadPool.h
```diff
@@ -0,0 +1,364 @@
...
+    // and in general does not provide visible improvements on iOS
+
+    // Set nice value as fallback or additional priority boost
+    constexpr int nice_value = 0;
+    if (setpriority(PRIO_PROCESS, 0, nice_value) != 0) {
+      log(LOG_LEVEL::Debug, "Failed to set nice value");
+    } else {
+      log(LOG_LEVEL::Debug, "Set nice value", nice_value);
+    }
+  }
```
msluszniak, line 318: Some functions like here or `getCurrentThreadId` might be marked static or moved outside of the class

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -65,7 +72,9 @@ Pod::Spec.new do |s|
     "HEADER_SEARCH_PATHS" =>
       '"$(PODS_TARGET_SRCROOT)/ios" '+
       '"$(PODS_TARGET_SRCROOT)/third-party/include" '+
-      '"$(PODS_TARGET_SRCROOT)/common" ',
+      '"$(PODS_TARGET_SRCROOT)/common" '+
+      '"$(PODS_TARGET_SRCROOT)/third-party/executorch/backends/xnnpack/third-party/pthreadpool/include" '+
+      '"$(PODS_TARGET_SRCROOT)/third-party/executorch/backends/xnnpack/third-party/cpuinfo/include" ',
```
chmjkb: Shouldn't this be third-party/include/executorch/...? If this is the case then it is already included in the line above

chmjkb (commented):
besides the single comment lgtm

PR #625, opened 2025-09-25
feat:voice activity detection model

### LINKED ISSUE #547: Implement voice activity detection
Implement VAD for purposes of improving the speech to text API

mlodyjesienin:
## Description
This PR introduces voice activity detection (vad) feature into the react native executorch library.

**This PR is not ready for merge yet,**
however it is ready for the review of C++ code. 

Things that are missing:
- Documentation
- exported model on official swm huggingface 
- benchmarks 
- maybe example usage in one app? idk 

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Related issues

Closes #547 
### Checklist


packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace rnexecutorch::models::voice_activity_detection::constants {
+
+inline constexpr uint32_t nextPowerOfTwo(uint32_t n) noexcept {
```
msluszniak: Use `std::bit_ceil` instead of this function

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -0,0 +1,40 @@
...
+  n |= n >> 8;
+  n |= n >> 16;
+  return n + 1;
+}
+
+inline constexpr uint32_t kSampleRate = 16000;
+inline constexpr float kMstoSecond = 0.001;
+inline constexpr uint32_t kWindowSizeMs = 25;
+inline constexpr uint32_t kHopLengthMs = 10;
+inline constexpr uint32_t kWindowSize =
```
msluszniak: Everywhere if you use static cast and type name length is longer than 4 characters, use `auto` instead

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -0,0 +1,31 @@
...
+        0.46f * std::cos((2.0f * std::numbers::pi_v<float> * i) / (size - 1));
+  }
+  return window;
+}
+
+size_t getNonSpeechClassProbabilites(const executorch::aten::Tensor &tensor,
+                                     size_t numClass, size_t size,
+                                     std::vector<float> &resultVector,
+                                     size_t startIdx) {
+  auto rawData = static_cast<const float *>(tensor.const_data_ptr());
```
msluszniak: `const_data_ptr<float>` and you can remove static cast then

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <numeric>
+
+namespace rnexecutorch::models::voice_activity_detection::utils {
+using namespace constants;
+
+const std::array<float, kWindowSize> generateHammingWindow() noexcept {
+  constexpr size_t size = static_cast<size_t>(constants::kWindowSize);
+  std::array<float, size> window;
+  for (size_t i = 0; i < size; ++i) {
+    window[i] =
```
msluszniak: `i` and `size - 1` will upcast to double, use static cast to float on them to preventing unwanted upcast since you assign to float either way

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+  constexpr size_t totalPadding = kPaddedWindowSize - kWindowSize;
+  constexpr size_t leftPadding = totalPadding / 2;
+  constexpr size_t rightPadding = totalPadding - leftPadding;
+  for (size_t i = 0; i < numFrames; i++) {
+
+    auto windowView = waveform.subspan(i * kHopLength, kWindowSize);
+    std::ranges::copy(windowView, frameBuffer[i].begin() + leftPadding);
+    auto frameView =
+        std::span{frameBuffer[i].data() + leftPadding, kWindowSize};
+    const float sum = std::accumulate(frameView.begin(), frameView.end(), 0.0f);
```
msluszniak: You can use `std::reduce` here instead

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+    const float mean = sum / kWindowSize;
+    std::ranges::transform(frameView, frameView.begin(),
+                           [mean](float value) { return value - mean; });
+
+    // apply pre-emphasis filter
+    for (auto j = frameView.size() - 1; j > 0; --j) {
```
msluszniak, line 46: Use `std::partial_sum` here with reverse iterator and `std::minus<>` as bin op

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+  if (remainder < kModelInputMin) {
+    auto paddingSize = kModelInputMin - remainder;
+    lastChunkSize = kModelInputMin;
+    windowedInput.insert(windowedInput.end(), paddingSize,
+                         std::array<float, kPaddedWindowSize>{});
+  }
+  TensorPtr inputTensor;
+  size_t startIdx = 0;
+
+  for (size_t i = 0; i < chunksNumber; i += 1) {
```
msluszniak: `i++`

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+  if (triggered) {
+    endSegment = scores.size();
+    speechSegments.emplace_back(startSegment, endSegment);
+  }
+
+  for (auto &[start, end] : speechSegments) {
+    // std::max(start-kSpeedchPad, 0) might be underflow that is why we use ?
+    // operator.
+    start = (start > kSpeechPad ? start - kSpeechPad : 0) * kHopLength;
```
msluszniak, line 154: Use `std::max(start - kSpeechPad, 0)` instead of ternary op

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -439,3 +440,10 @@ export const BK_SDM_TINY_VPRED_256 = {
   unetSource: `${URL_PREFIX}-bk-sdm-tiny/${VERSION_TAG}/unet/model.256.pte`,
   decoderSource: `${URL_PREFIX}-bk-sdm-tiny/${VERSION_TAG}/vae/model.256.pte`,
 };
+
+// Voice Activity Detection
+const FSMN_VAD_MODEL = `${URL_PREFIX}-fsmn-vad/${NEXT_VERSION_TAG}/xnnpack/fsmn-vad_xnnpack.pte`;
```
msluszniak, line 445: Add `fsmn` and `FSMN` to cspell

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+  std::vector<std::array<float, kPaddedWindowSize>> frameBuffer(
+      numFrames, std::array<float, kPaddedWindowSize>{});
+
+  constexpr size_t totalPadding = kPaddedWindowSize - kWindowSize;
+  constexpr size_t leftPadding = totalPadding / 2;
+  constexpr size_t rightPadding = totalPadding - leftPadding;
+  for (size_t i = 0; i < numFrames; i++) {
+
+    auto windowView = waveform.subspan(i * kHopLength, kWindowSize);
+    std::ranges::copy(windowView, frameBuffer[i].begin() + leftPadding);
```
msluszniak: Use `namespace ranges = std::ranges;` at the top of the file and the use ranges::copy etc. elsewhere

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
mlodyjesienin: since `start` and `kSpeechPad` are `size_t` the difference might be underflow which results in unwanted behavior, as it is explained in the comment above.

msluszniak: Yeap, my fault

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -0,0 +1,40 @@
...
+  if (n <= 1)
+    return 1;
+  n--;
+  n |= n >> 1;
+  n |= n >> 2;
+  n |= n >> 4;
+  n |= n >> 8;
+  n |= n >> 16;
+  return n + 1;
+}
```
chmjkb: I think this is an util

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -0,0 +1,31 @@
+#include "Utils.h"
+
+#include <numbers>
+#include <numeric>
```
chmjkb: Unused header

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+const std::array<float, kWindowSize> generateHammingWindow() noexcept {
+  constexpr size_t size = static_cast<size_t>(constants::kWindowSize);
+  std::array<float, size> window;
+  for (size_t i = 0; i < size; ++i) {
+    window[i] =
+        0.54f -
+        0.46f * std::cos((2.0f * std::numbers::pi_v<float> * i) / (size - 1));
+  }
+  return window;
```
chmjkb: You can check if this will work with the hannWindow defined in dsp.cpp. If so, then you don't need to define it. Otherwise, move this function to dsp.cpp

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
chmjkb, line 1: utils.cpp often ends up as a bit of a bag for random functions. Might be worth thinking about whether some of these belong in more focused files.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+  auto kHammingWindowArray = utils::generateHammingWindow();
+
+  const size_t numFrames = (waveform.size() - kWindowSize) / kHopLength;
+
+  std::vector<std::array<float, kPaddedWindowSize>> frameBuffer(
+      numFrames, std::array<float, kPaddedWindowSize>{});
+
+  constexpr size_t totalPadding = kPaddedWindowSize - kWindowSize;
+  constexpr size_t leftPadding = totalPadding / 2;
+  constexpr size_t rightPadding = totalPadding - leftPadding;
```
chmjkb: unused variable

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,159 @@
...
+using executorch::extension::TensorPtr;
+
+VoiceActivityDetection::VoiceActivityDetection(
+    const std::string &modelSource,
+    std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker) {}
+
+std::vector<std::array<float, kPaddedWindowSize>>
+VoiceActivityDetection::preprocess(std::span<float> waveform) const {
+  auto kHammingWindowArray = utils::generateHammingWindow();
```
chmjkb: Since it's not actually depending on the size of the input, nor the input - calling this function on each preprocess call is redundant.

chmjkb (changes requested):

packages/react-native-executorch/src/modules/natural_language_processing/VADModule.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (paths === null || paths.length < 1) {
+      throw new Error('Download interrupted.');
+    }
+    this.nativeModule = global.loadVAD(paths[0] || '');
+  }
+
+  async forward(waveform: Float32Array | number[]): Promise<Segment[]> {
+    if (this.nativeModule == null)
+      throw new Error(getError(ETError.ModuleNotLoaded));
+    return await this.nativeModule.generate(waveform);
```
chmjkb: I believe this will crash when number[] is passed. Is there any reason why you want to pass number[] here?

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
mlodyjesienin: In the end I got rid of this function; (thanks to the @msluszniak comment about std::bit_ceil ), but to clarify, I agree that it should be inside `Utils.h`, and that was my initial approach, but due to circular dependencies I could not make it work. Obviously, one could solve this by introducing new file etc, but actually moving it to the `Constants.h` due to its simplicity seemed like the best solution to me. Writing this comment to ask, are there any better solutions I am not aware of?

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
mlodyjesienin: My approach is simple: In the model (`VoiceActivityDetection.h/.cpp` in this case ) files, I want only three functions, that is: **preprocess**, **generate**, **postprocess**. Basically everything else goes to `Utils.cpp`.
Often utils files are not very long so I think this works ok. Introducing some other files might be 
 less helpful, and more confusing but it is only my opinion. Also sometimes it is just hard to find the "common denominator" of more than one function.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
mlodyjesienin: I have checked it, it works fine. I deleted this function entirely, and I use `hannWindow` instead. 
The differences are very subtle (e.g. on the 60s audio clip with 16 speech segments, 14 of them are **exactly** the same, and two of them differ by circa 100ms, making it negligible.)

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
mlodyjesienin: As I mentioned in comment above, in the end I deleted this function, but for clarification I will say this:
Yes, this is redundant, but there is no easy way to bypass it. The problem is, `std::cos` is `constexpr` since `C++23` and we use `C++20`. At first I tried to simply write it as constexpr and calculate it in compile time, but as it turns out it is impossible. Obviously, what could be done is one of these:
1) we could write cosine function independently, and we could make it constexpr
2) hardcode the output in some different way. 

But I reckon it is just unnecessary complications (in the end it is 400 element array so it should be negligible)

Since I have decided to use the `hannWindow` instead, it does not matter (but obviously, the `hannWindow` has the same disadvantage of being calculated with each model run, and what is  even worse - it is `std::vector` and not `std::array`.... )

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
mlodyjesienin: What I am doing here is basically:
`frameView[j] = frameView[j] - kPreemphasisCoeff*frameView[j-1]`
Since the value of `j`th element relies on *original* `j-1`th element, we need the backwards loop.
I do not know, how to implement this using `std::partial_sum`. 
I think there is actually no simple way

mlodyjesienin:
I think it is ready to merge, but there is one thing that might be done before (or possibly after) merge:
I measured inference time for benchmarks for android only on OnePlus because Samsung Galaxy S24 was unavailable.
Since it is my last day in work, I won't be able to do it, so someone else should complete the benchmarking.

Obviously, review needs to be done, but I think we are in the clear.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
msluszniak: Right, this constant multiplier makes it cumbersome to handle this other way than loop,

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -0,0 +1,15 @@
+#include "Utils.h"
+
+namespace rnexecutorch::models::voice_activity_detection::utils {
+size_t getNonSpeechClassProbabilites(const executorch::aten::Tensor &tensor,
+                                     size_t numClass, size_t size,
+                                     std::vector<float> &resultVector,
+                                     size_t startIdx) {
+  auto rawData = tensor.const_data_ptr<float>();
```
msluszniak:
```suggestion
  const auto* rawData = tensor.const_data_ptr<float>();
```

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -0,0 +1,161 @@
...
+      numFrames, std::array<float, kPaddedWindowSize>{});
+
+  constexpr size_t totalPadding = kPaddedWindowSize - kWindowSize;
+  constexpr size_t leftPadding = totalPadding / 2;
+  for (size_t i = 0; i < numFrames; i++) {
+
+    auto windowView = waveform.subspan(i * kHopLength, kWindowSize);
+    ranges::copy(windowView, frameBuffer[i].begin() + leftPadding);
+    auto frameView =
+        std::span{frameBuffer[i].data() + leftPadding, kWindowSize};
```
msluszniak, line 39: `frameBuffer` is a `std::vector` of size `numFrames`
```suggestion
  for (auto frame : frameBuffer ) {

    auto windowView = waveform.subspan(i * kHopLength, kWindowSize);
    ranges::copy(windowView, frame.begin() + leftPadding);
    auto frameView =
        std::span{frame.data() + leftPadding, kWindowSize};
```

PR #631, opened 2025-10-01
Delete unused objective-c++ files

msluszniak:
## Description

As in the title, utils defined in the `ios/RnExecutorch/utils` are not longer used so they can be deleted.

- [ ] Yes
- [X] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


msluszniak:
@chmjkb Can we also delete objective-c++ installers, namely: https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/ios/RnExecutorch/ETInstaller.mm and header for it?

msluszniak:
From CI it seems that `ScalarType.mm` is still used in compilation of `LLM` demo app. I don't see where it is included though.

chmjkb:
> @chmjkb Can we also delete objective-c++ installers, namely: https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/ios/RnExecutorch/ETInstaller.mm and header for it?

This one we're actually using



> From CI it seems that `ScalarType.mm` is still used in compilation of `LLM` demo app. I don't see where it is included though.

I'll check it out

NorbertKlockiewicz:
@msluszniak can you try to remove the cocoapods caching part from build-ios workflow file. I believe it's causing the problems with CI fails.

NorbertKlockiewicz:
> @msluszniak can you try to remove the cocoapods caching part from build-ios workflow file. I believe it's causing the problems with CI fails.

image

The install cocoapods action still hit the cache as it has if that checks if the `Pods` directory exists, can you please also remove it and try again?

msluszniak:
@NorbertKlockiewicz should these cache things from ios workflow be deleted permanently?

NorbertKlockiewicz:
> @NorbertKlockiewicz should these cache things from ios workflow be deleted permanently?

Yes, we already cache native folders and installing the cocoapods dependencies doesn't take that much time.

Issue #651, opened 2025-10-15
Bad unicode in transcription result


michaldudek:
### Description

The transcription result I get from `model.transcribe()` seems to have some malformed unicode structure.

I have the following code:

```
const result = await model.transcribe(recording, {
  language: 'pl',
});

console.log('result', result);
console.log('object', { result });
console.log(JSON.stringify({ result }));
```

The output from this is:

In React DevTools (Chrome):

<img width="1397" height="239" alt="Image" src="https://github.com/user-attachments/assets/2881099c-790f-4cee-b270-59222c6c584a" />

In Terminal:

<img width="1068" height="214" alt="Image" src="https://github.com/user-attachments/assets/299f3b55-0ade-4fef-b747-41cec8d06b08" />

(interestingly, the 1st `console.log('result', result)` does not get logged in Terminal, but it does in devtools)

And this results in the following error:

```
[Error: InternalError Metro has encountered an error: Bad Unicode escape in JSON at position 63 (line 1 column 64): /Users/michaldudek/Developer/Lab/ExecuTorch/node_modules/metro/src/HmrServer.js (135:19)

  133 |     let data;
  134 |     try {
> 135 |       data = JSON.parse(String(message));
      |                   ^
  136 |     } catch (error) {
  137 |       send([sendFn], {
  138 |         type: "error",]
```

Removing the logs removes the issue, but when I POST that result to my nodejs api it also throws the same JSON encoding error (rightly so, the JSON is definitely malformed with this string).

When I switch the `language` to `en` and do speak English, the issue does not occur (but there are no special/accent chars in the transcription).

I tested with multilingual versions of `whisper_tiny` and `whisper_base`. I did not have this issue when testing with OpenAI's hosted whisper-1 model via their realtime api (but I suppose this is a completely different model there anyway?). It very well might be an issue with the model, but maybe something somewhere in the lib isn't getting set correctly?

### Steps to reproduce

1. Run `apps/speech-to-text` example app from this repo.
2. Apply the following patch: (change model to `WHISPER_TINY`, set `language: 'pl'`, console.log the result):

<img width="897" height="532" alt="Image" src="https://github.com/user-attachments/assets/eea642f3-1039-466b-ae1e-91f304c0cc53" />

3. If running in Simulator, you can use this recording I prepared: `https://cdn.studylm.app/assets/executorch-test.wav`
4. You should get the same error.

### React Native Executorch version

0.5.10

### React Native version

0.79.5

### Platforms

[...]

PR #652, opened 2025-10-17
bug: fix to corrupted utf-8 encoding for special characters in transcription

mkopcins:
## Description

Fix for issue #651. The problem was the serialization of cpp string into jsi::Value and then back to a string, causing invalid characters and bytes. Now cpp returns array of bytes and the decoding is done on the JS side.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
+++ b/packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
@@ -66,7 +66,8 @@ inline JSTensorViewIn getValue<JSTensorViewIn>(const jsi::Value &val,
   tensorView.sizes.reserve(numShapeDims);
 
   for (size_t i = 0; i < numShapeDims; ++i) {
-    int32_t dim = getValue<int32_t>(shapeArray.getValueAtIndex(runtime, i), runtime);
+    int32_t dim =
+        getValue<int32_t>(shapeArray.getValueAtIndex(runtime, i), runtime);
     tensorView.sizes.push_back(dim);
   }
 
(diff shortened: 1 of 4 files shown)


packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -60,7 +61,9 @@ std::string SpeechToText::transcribe(std::span<float> waveform,
       transcription += word.content;
     }
   }
-  return transcription;
+
+  std::vector<char> charVector(transcription.begin(), transcription.end());
+  return charVector;
```
msluszniak:
```suggestion
  return {transcription.begin(), transcription.end()};
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -280,6 +280,15 @@ inline jsi::Value getJsiValue(const std::vector<int32_t> &vec,
   return {runtime, array};
 }
 
+inline jsi::Value getJsiValue(const std::vector<char> &vec,
+                              jsi::Runtime &runtime) {
+  jsi::Array array(runtime, vec.size());
+  for (size_t i = 0; i < vec.size(); i++) {
+    array.setValueAtIndex(runtime, i, jsi::Value(static_cast<char>(vec[i])));
```
msluszniak: Is this static_cast to char really needed here?

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -1,6 +1,7 @@
 #include <thread>
 
 #include "SpeechToText.h"
+#include <vector>
```
msluszniak: This is not needed as `SpeechToText.h` already has this include
```suggestion
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "rnexecutorch/models/speech_to_text/stream/OnlineASRProcessor.h"
+#include <vector>
```
msluszniak, line 6: We can also add here `#include <span>` and `#include <string>`

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -94,12 +97,18 @@ void SpeechToText::stream(std::shared_ptr<jsi::Function> callback,
...
-    nativeCallback(res.committed, res.nonCommitted, false);
+
+    std::vector<char> committedVec(res.committed.begin(), res.committed.end());
+    std::vector<char> nonCommittedVec(res.nonCommitted.begin(), res.nonCommitted.end());
+
+    nativeCallback(committedVec, nonCommittedVec, false);
     this->readyToProcess = false;
   }
 
   std::string committed = this->processor->finish();
```
msluszniak, line 105: Maybe we change signature (so it will return `std::vector<char>` instead of `std::string`) of `this->processor->finish` and what is returned in `ProcessResult` if these functions are used only here. If used anywhere else as `std::string` we should probably keep it as it is

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
mkopcins: no, it is not

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
mkopcins: not sure, but I want this shipped today and as this is a mostly cosmetic change I don't want to spend time on deliberating :)

msluszniak: makes sense, I'll keep that in mind and check this

PR #658, opened 2025-10-23
feat: Remove stft calculation within the encoder

mkopcins:
## Description

The Whisper model export now takes in a plain waveform instead of pre-computed STFT. This PR aims to change the current API to accept waveforms instead. Before merging this, make sure to re-export all the existing Whisper models with the new export script. 

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/src/constants/modelUrls.ts
+++ b/packages/react-native-executorch/src/constants/modelUrls.ts
@@ -2,7 +2,7 @@ import { Platform } from 'react-native';
 
 const URL_PREFIX =
   'https://huggingface.co/software-mansion/react-native-executorch';
-const VERSION_TAG = 'resolve/v0.6.0';
+const VERSION_TAG = 'resolve/v0.5.0';
 const NEXT_VERSION_TAG = 'resolve/v0.6.0';
 
 // LLMs


packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/asr/ASR.cpp
```diff
@@ -246,19 +244,12 @@ std::vector<Segment> ASR::transcribe(std::span<const float> waveform,
   return results;
 }
 
-std::vector<float> ASR::encode(std::span<const float> waveform) const {
-  constexpr int32_t fftWindowSize = 512;
-  constexpr int32_t stftHopLength = 160;
-  constexpr int32_t innerDim = 256;
-
-  std::vector<float> preprocessedData =
-      dsp::stftFromWaveform(waveform, fftWindowSize, stftHopLength);
```
msluszniak, line 255: If `stftFromWaveform` will be removed, this function should be removed completely from codebase since it is not used anywhere else.

chmjkb:
Let's reupload all the Whisper models to HF and this is good to merge!

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -2,7 +2,7 @@ import { Platform } from 'react-native';
 
 const URL_PREFIX =
   'https://huggingface.co/software-mansion/react-native-executorch';
-const VERSION_TAG = 'resolve/v0.5.0';
+const VERSION_TAG = 'resolve/v0.6.0';
```
IgorSwat: I would wait with this until the final release, especially since you use NEXT_VERSION_TAG in model URLs.

IgorSwat:
```suggestion
const VERSION_TAG = 'resolve/v0.5.0';
```

PR #693, opened 2025-11-27
feat: build with executorch 1.0 (#638)

IgorSwat:

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/runner/sampler.h
+++ b/packages/react-native-executorch/common/runner/sampler.h
@@ -26,12 +26,12 @@ namespace extension {
 namespace llm {
 // A simple llama2 sampler.
 
-template <typename T> struct ET_EXPERIMENTAL ProbIndex {
+template <typename T> struct ProbIndex {
   T prob;
   int32_t index;
 }; // struct used when sorting probabilities during top-p sampling
 
-class ET_EXPERIMENTAL Sampler {
+class Sampler {
 public:
   Sampler(int32_t vocab_size, float temperature, float topp,
           unsigned long long rng_seed);
--- a/packages/react-native-executorch/common/runner/io_manager.h
+++ b/packages/react-native-executorch/common/runner/io_manager.h
@@ -23,7 +23,7 @@ namespace llm {
  * Derived classes must implement the virtual methods to provide specific IO
  * management functionality.
  */
-class ET_EXPERIMENTAL IOManager {
+class IOManager {
 public:
   /**
    * @brief Construct an IOManager bound to a Module.
--- a/packages/react-native-executorch/common/runner/irunner.h
+++ b/packages/react-native-executorch/common/runner/irunner.h
@@ -68,7 +68,7 @@ struct GenerationConfig {
 };
 
 // Base interface for LLM runners
-class ET_EXPERIMENTAL IRunner {
+class IRunner {
 public:
   virtual ~IRunner() = default;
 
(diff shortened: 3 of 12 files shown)


packages/react-native-executorch/common/runner/sampler.h
```diff
@@ -26,19 +26,19 @@ namespace extension {
 namespace llm {
 // A simple llama2 sampler.
 
-template <typename T> struct ProbIndex {
+template <typename T> struct ET_EXPERIMENTAL ProbIndex {
```
chmjkb: Can we get rid of the ET_EXPERIMENTAL macros (or find a way to avoid compiler warnings for this)?

packages/react-native-executorch/third-party/ios/libs/executorch/libkernels_portable_ios.a
chmjkb, line 1: Why are we getting rid of the portable kernels?

packages/react-native-executorch/third-party/ios/libs/executorch/libkernels_torchao_ios.a
chmjkb, line 1: Is this included in the Android build as well?

chmjkb (commented):
Code looks good, left some small questions, good job 👏🏻 
I'll test if things work today and we can merge

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.h
```diff
@@ -30,12 +30,6 @@ class LLM : public BaseModel {
 
 private:
   std::unique_ptr<example::Runner> runner;
-
-  // A typical input for parallel processing in exported LLM model consists of 2
-  // tensors of shapes [1, N] and [1], where N is the number of tokens. Hovewer,
-  // some exported models require inputs of shapes [1, N] and [N], which needs
-  // to be marked before using LLM runner.
-  bool extended_input_mode_ = false;
```
msluszniak, line 38: Why this is removed, is it not needed anymore for Gemma?

msluszniak (commented):
I guess that all the files from `runner` and `third-party` dirs are copied so no improvements wanted / needed. Are there any other cpp files than in `common/rnexecutorch` to check? If not, please make sure that someone will test this PR, unfortunately I don't have time to do this.

chmjkb:
I tested example apps and it looks good to me, are pytorch tokenizers included in this binary?

packages/react-native-executorch/third-party/ios/libs/executorch/libkernels_portable_ios.a
IgorSwat: It's no longer being produced by the ExecuTorch iOS build. I believe it's now replaced by *libkernels_llm_ios.a*.

packages/react-native-executorch/third-party/ios/libs/executorch/libkernels_torchao_ios.a
IgorSwat: Yes - Android libs are built with the same setup as iOS ones, so the resulting libs should contain the same content.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.h
IgorSwat: ExecuTorch 1.0 introduces its own implementation of this mechanism with the new TextDecoderRunner and specifically `populate_start_pos_or_cache_position()` function, so our extra code is no longer needed.

IgorSwat:
> I tested example apps and it looks good to me, are pytorch tokenizers included in this binary?

No, we use *tokenizers-cpp* as for now. I strongly believe that switching to pytorch tokenizers should be a separate task, since it's not a critical issue for now, and I suspect it might not be as straightforward as it seems to be.

Issue #546, opened 2025-08-26
Text-to-Speech


mkopcins:
Implement text-to-speech models (good starting point [kitten t2s](https://github.com/KittenML/KittenTTS), [kokoro](https://kokorotts.net/))

msluszniak:
@mkopcins cloud you close this one: https://github.com/software-mansion/react-native-executorch/issues/355 as it is a duplicate of this issue?

IgorSwat:
A new chance of completing the task appeared after fixing the issue related to Kokoro's Decoder export.

PR #710, opened 2026-01-09
feat: text to speech (#546)

IgorSwat:

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Checklist

### Additional notes

There are still a few side things to be done on this feature.


packages/react-native-executorch/common/rnexecutorch/data_processing/Sequential.h
```diff
@@ -0,0 +1,54 @@
...
+  if (data.size() != repetitions.size()) {
+    throw std::invalid_argument(
+        "repeatInterleave(): repetitions vector must be the same size as data,"
+        " expected " +
+        std::to_string(data.size()) + " but got " +
+        std::to_string(repetitions.size()));
+  }
+
+  IType totalReps = std::accumulate(repetitions.begin(), repetitions.end(),
+                                    static_cast<IType>(0));
```
msluszniak:
```suggestion
  IType totalReps = std::reduce(repetitions.begin(), repetitions.end());
```

packages/react-native-executorch/common/rnexecutorch/data_processing/Sequential.h
```diff
@@ -0,0 +1,54 @@
...
+  std::vector<T> result(totalReps);
+
+  IType filled = 0;
+  for (int i = 0; i < data.size(); ++i) {
```
msluszniak:
```suggestion
  for (size_t i = 0; i < data.size(); ++i) {
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Decoder.cpp
```diff
@@ -0,0 +1,77 @@
...
+                      2 * inputConfig.duration, " but got ", nPred.size());
+    throw std::runtime_error("[Kokoro::Decoder] Invalid input shape");
+  }
+  // ref_hs should be a half of a voice reference vector
+  if (ref_ls.size() != constants::kVoiceRefHalfSize) {
+    rnexecutorch::log(rnexecutorch::LOG_LEVEL::Error,
+                      "Unexpected voice ref length: expected ",
+                      constants::kVoiceRefHalfSize, " but got ", ref_ls.size());
+    throw std::runtime_error("[Kokoro::Decoder] Invalid input shape");
+  }
```
msluszniak: This should be a separate function

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Decoder.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <span>
+
```
msluszniak, line 7: Missing headers for std::string, std::vector, std::shared_ptr

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Constants.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include <cstdint>
+#include <unordered_map>
```
msluszniak, line 5: Missing header for std::string

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Decoder.h
```diff
@@ -0,0 +1,23 @@
...
+#include <executorch/extension/tensor/tensor.h>
+
+#include "Constants.h"
+#include <rnexecutorch/models/BaseModel.h>
+
+namespace rnexecutorch::models::text_to_speech::kokoro {
+
+class Decoder : public BaseModel {
+public:
+  Decoder(const std::string &modelSource,
```
msluszniak: If you always pass std::string and not e.g. char* you can make constructor explicit

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -0,0 +1,159 @@
...
+  int32_t originalLength =
+      std::distance(tokens.begin(),
+                    std::find(tokens.begin() + 1, tokens.end(), 0)) +
+      1;
+  int32_t effDuration = std::distance(
+      indices.begin(),
+      std::lower_bound(indices.begin(), indices.end(), originalLength));
+
+  if (effDuration < inputConfig.duration)
+    effDuration *= 0.95;
```
msluszniak, line 89:
```suggestion
  if (effDuration < inputConfig.duration) {
    effDuration *= 0.95;
  }
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -0,0 +1,159 @@
...
+    float scaled = scaleFactor * durationsPtr[i];
+    float remainder =
+        shrinking ? std::ceil(scaled) - scaled : scaled - std::floor(scaled);
+
+    durationsPtr[i] = static_cast<int64_t>(shrinking ? std::ceil(scaled)
+                                                     : std::floor(scaled));
+    scaledSum += durationsPtr[i];
+
+    // Keeps the entries sorted by the remainders
+    remainders.push({remainder, i});
```
msluszniak:
```suggestion
    remainders.emplace(remainder, i);
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -0,0 +1,184 @@
...
+  // Perform backtracking to obtain all the substrings.
+  // Note that because of backtracking, the order is reversed.
+  int32_t end = phonemes.size() - 1;
+  while (end != -1) {
+    int32_t begin = mem[end].second + 1;
+    result.push_back(phonemes.substr(begin, end - begin + 1));
+    end = mem[end].second;
+  }
+
+  std::reverse(result.begin(), result.end());
```
msluszniak:
```suggestion
  std::ranges::reverse(result);
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    if (count >= constants::kAudioCroppingSteps &&
+        sum / constants::kAudioCroppingSteps >=
+            constants::kAudioSilenceThreshold)
+      return i;
+
+    i = reverse ? i - 1 : i + 1;
+  }
+
+  return reverse ? 0 : length - 1;
```
msluszniak, line 47: if `reverse` is true then you always return 0, so you can skip all the computations earlier. So what is purpose of using reverse in while loop?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+} // namespace
+
+std::span<const float> stripAudio(std::span<const float> audio, size_t margin) {
+  auto lbound = findAudioBound<false>(audio);
+  auto rbound = findAudioBound<true>(audio);
+
+  lbound = lbound > margin ? lbound - margin : 0;
```
msluszniak:
```suggestion
  lbound = std::max(lbound - margin, 0);
```

msluszniak (commented):
Please apply changes to all applicable places, I didn't repeated these comments. Bot overall, it looks really cool these suggestions are just very small nits ;)

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
IgorSwat: The thing is, there is another return statement inside the while loop - so it could return something other than 0 even if `reverse` is true.

The `reverse` variable is just used to iterate in the opposite direction without duplicating the code.

msluszniak: Ok, I missed that inner return

msluszniak:
Since you added demo app in this PR, you should also update `README.md` in the following section:

image

apps/speech/screens/TextToSpeechScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+      // fixedModel: "large"
+    },
+  });
+
+  const [inputText, setInputText] = useState('');
+  const [isPlaying, setIsPlaying] = useState(false);
+  const [readyToGenerate, setReadyToGenerate] = useState(false);
+
+  const audioContextRef = useRef<AudioContext | null>(null);
+  const sourceRef = useRef<any>(null);
```
chmjkb: Can we type this?

apps/speech/screens/TextToSpeechScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+  useEffect(() => {
+    AudioManager.setAudioSessionOptions({
+      iosCategory: 'playAndRecord',
+      iosMode: 'spokenAudio',
+      iosOptions: ['defaultToSpeaker'],
+    });
+
+    // Initialize context once
```
chmjkb:
```suggestion
```

apps/speech/screens/TextToSpeechScreen.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+          source.onEnded = () => resolve();
+
+          source.start();
+        });
+      };
+
+      const onEnd = async () => {
+        setIsPlaying(false);
+        setReadyToGenerate(true);
```
chmjkb: doesnt the useEffect above handle this?

packages/react-native-executorch/scripts/create-package.sh
chmjkb, line 1: Why is this deleted?

packages/react-native-executorch/src/types/tts.ts
```diff
@@ -0,0 +1,44 @@
+import { ResourceSource } from './common';
+
+// List all the languages available in TTS models
+// The values should match the one used within the native side.
+export enum TextToSpeechLanguage {
+  EN_US = 0,
+  EN_GB = 1,
+}
```
chmjkb: S2T shared this as string literal union and I recommend using it instead, so we have a unified approach:
```typescript
export type SpeechToTextLanguage =
  | 'af'
  | 'sq'
  | 'ar'
// this continues
```

packages/react-native-executorch/src/types/tts.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Voice configuration
+// So far in Kokoro, each voice is directly associated with a language.
+// The 'data' field corresponds to (usually) binary file with voice tensor.
```
chmjkb: Can we use JSDoc for this kind of comments?
For example:
```
**
 * Voice configuration
 * * So far in Kokoro, each voice is directly associated with a language.
 * The 'data' field corresponds to (usually) binary file with voice tensor.
 */
```

Overall I think that this is a good approach for everything we share to the public API as it is easy for the user to figure out what the thing is doing without the docs.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1,9 +1,9 @@
 import { Platform } from 'react-native';
 
-const URL_PREFIX =
+export const URL_PREFIX =
   'https://huggingface.co/software-mansion/react-native-executorch';
-const VERSION_TAG = 'resolve/v0.6.0';
-// const NEXT_VERSION_TAG = 'resolve/v0.7.0';
+export const VERSION_TAG = 'resolve/v0.6.0';
+export const NEXT_VERSION_TAG = 'resolve/v0.7.0';
```
chmjkb: Does this need to be exported? This makes it possible for the users to import this, which is unnecessary

chmjkb (changes requested):
Left some nits for the moment, I feel like we need to add docs to have a solid understanding of the public API layer. However, this looks like really a lot of solid work, congrats!

Haven't finished reviewing tho :D

packages/react-native-executorch/src/types/tts.ts
```diff
@@ -0,0 +1,44 @@
...
+  f0nPredictorSource: ResourceSource;
+  textEncoderSource: ResourceSource;
+  textDecoderSource: ResourceSource;
+}
+
+// Model + voice configurations
+export interface TextToSpeechConfig {
+  model: KokoroConfig; // ... add other model types in the future
+  voice?: VoiceConfig;
+  options?: any; // A completely optional model-specific configuration
```
chmjkb: Would it be possible to find a way to not use any or unknown here? See the comment in `TextToSpeechModule.ts`

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,121 @@
...
+    // TODO: this check is pretty dubious and should be replaced with something better.
+    const uri = (config.model as any)[anySourceKey];
+    if (uri.includes('kokoro')) {
+      await this.loadKokoro(
+        config.model,
+        config.voice!,
+        onDownloadProgressCallback,
+        config.options
+      );
+    }
```
chmjkb: We shouldn't do it this way. What if someone changes the name of this file, or uses require()?

I would recommend doing something like this:

```typescript
// 1. Define specific option types first for reusability
interface KokoroOptions {
  temperature: number;
  contextSize: number;
}

interface OtherOptions {
  language: string;
  beamSize: number;
}

// 2. Define the Configs with a strict 'type' discriminator
interface KokoroConfig {
  type: 'kokoro';
  modelPath: string;
  options: KokoroOptions;
}

interface OtherModelConfig {
  type: 'other';
  modelPath: string;
  options: OtherOptions;
}

type ModelConfig = KokoroConfig | OtherModelConfig;

class GenericModel {
  public async load(config: ModelConfig): Promise<void> {
    switch (config.type) {
      case 'kokoro':
        // TypeScript implies config is KokoroConfig here
        await this.loadKokoro(config.modelPath, config.options);
        break;
      
      case 'other':
        // TypeScript implies config is OtherModelConfig here
        await this.loadOther(config.modelPath, config.options);
        break;
    }
  }

  private async loadKokoro(path: string, options: KokoroOptions) {
    console.log(`Loading Kokoro with temp: ${options.temperature}`);
    // implementation...
  }

  private async loadOther(path: string, options: OtherOptions) {
    console.log(`Loading Other with lang: ${options.language}`);
    // implementation...
  }
}
```

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,121 @@
...
+    model: KokoroConfig,
+    voice: VoiceConfig,
+    onDownloadProgressCallback: (progress: number) => void,
+    options?: any
+  ): Promise<void> {
+    if (!voice.extra || !voice.extra.tagger || !voice.extra.lexicon) {
+      throw new Error(
+        'Kokoro: voice config is missing required extra fields: tagger and/or lexicon.'
+      );
+    }
```
chmjkb: https://github.com/software-mansion/react-native-executorch/pull/710/changes#r2681356511

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const paths = await ResourceFetcher.fetch(
+      onDownloadProgressCallback,
+      ...Object.values(model),
+      voice.data,
+      voice.extra!.tagger,
+      voice.extra!.lexicon
```
chmjkb: Let's not use non-null assertion operators

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  public async forward(text: string, speed: number = 1.0) {
+    if (this.nativeModule == null)
+      throw new Error(getError(ETError.ModuleNotLoaded));
+    return await this.nativeModule.generate(text, speed);
+  }
+
+  public async stream({
```
chmjkb: Maybe we could make this a generator? See the STT implementation

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,245 @@
+#include "Kokoro.h"
+#include "Utils.h"
+
+#include <algorithm>
+#include <fstream>
+#include <numeric>
+#include <rnexecutorch/data_processing/Sequential.h>
+#include <stdexcept>
+
+#include <chrono>
```
chmjkb: numeric and chrono are unused

packages/react-native-executorch/common/rnexecutorch/RnExecutorchInstaller.cpp
```diff
@@ -100,6 +101,10 @@ void RnExecutorchInstaller::injectJSIBindings(
       *jsiRuntime, "loadSpeechToText",
       RnExecutorchInstaller::loadModel<models::speech_to_text::SpeechToText>(
           jsiRuntime, jsCallInvoker, "loadSpeechToText"));
+  jsiRuntime->global().setProperty(
+      *jsiRuntime, "loadTextToSpeechKokoro",
+      RnExecutorchInstaller::loadModel<models::text_to_speech::kokoro::Kokoro>(
+          jsiRuntime, jsCallInvoker, "loadTextToSpeechKokoro"));
```
chmjkb, line 107: Maybe we should do loadTextToSpeech and dispatch to specific model implementations based on a string passed from JS instead of doing it this way? If we introduce other models, this file will get even more bloated than it already is

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Divide the phonemes string intro substrings.
+  // Affects the further calculations only in case of string size
+  // exceeding the biggest model's input.
+  auto subsentences =
+      partitioner_.divide<Partitioner::Strategy::TOTAL_TIME>(phonemes);
+
+  std::vector<float> audio = {};
+  for (const auto &subsentence : subsentences) {
+    size_t inputSize = subsentence.size() + 2;
```
chmjkb, line 82: Why the +2? A comment would be handy

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,245 @@
...
+  // Use specialized implementation to minimize the latency between the
+  // sentences.
+  auto subsentences =
+      partitioner_.divide<Partitioner::Strategy::LATENCY>(phonemes);
+
+  // We follow the implementation of generate() method, but
+  // instead of accumulating results in a vector, we push them
+  // back to the JS side with the callback.
+  for (const auto &subsentence : subsentences) {
+    size_t inputSize = subsentence.size() + 2;
```
chmjkb, line 135: Same here

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,245 @@
...
+      method, config, std::span(tokens),
+      std::span(reinterpret_cast<bool *>(textMask.data()), textMask.size()),
+      std::span<float>(pred_aln_trg.data_ptr<float>(), pred_aln_trg.numel()));
+  auto asr = encoding->at(0).toTensor();
+
+  // Inference 4 - Decoder
+  auto decoding = decoder_.generate(
+      method, config, std::span<float>(asr.data_ptr<float>(), asr.numel()),
+      std::span<float>(F0_pred.data_ptr<float>(), F0_pred.numel()),
+      std::span<float>(N_pred.data_ptr<float>(), N_pred.numel()), ref_ls);
```
chmjkb: `data_ptr()` is deprecated. Use `const_data_ptr()` or `mutable_data_ptr()` instead.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -0,0 +1,144 @@
...
+using ::executorch::aten::ScalarType;
+using ::executorch::extension::make_tensor_ptr;
+using ::executorch::extension::TensorPtr;
+
+DurationPredictor::DurationPredictor(
+    const std::string &modelSource,
+    std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker) {
+  std::string testMethod =
+      "forward_" + std::to_string(constants::kInputSmall.noTokens);
```
chmjkb: This should be a const I think

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+std::vector<float> Kokoro::generateForConfig(const std::u32string &phonemes,
+                                             const Configuration &config,
+                                             float speed) {
+  // Determine the appropriate method for given input configuration
+  std::string method = "forward_" + std::to_string(config.noTokens);
+
+  // Map phonemes to tokens
+  auto tokens = utils::tokenize(phonemes, {config.noTokens});
```
chmjkb: I think both should be const

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  auto shape = durations.sizes();
+  if (shape.size() != 1) {
+    throw std::runtime_error(
+        "[Kokoro::DurationPredictor] Attempted to scale an ill-shaped tensor");
+  }
+
+  int32_t nTokens = shape[0];
+  int64_t *durationsPtr = durations.data_ptr<int64_t>();
```
chmjkb: use `const_data_ptr()` or `mutable_data_ptr()`

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.h
```diff
@@ -0,0 +1,43 @@
...
+namespace rnexecutorch::models::text_to_speech::kokoro {
+
+using executorch::aten::Tensor;
+
+class DurationPredictor : public BaseModel {
+public:
+  explicit DurationPredictor(const std::string &modelSource,
+                             std::shared_ptr<react::CallInvoker> callInvoker);
+
+  // Returns a tuple (d, indices, effectiveDuration)
```
chmjkb: I think it would be nice if there was some explaination what exactly the returned stuff is

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Encoder.h
```diff
@@ -0,0 +1,27 @@
...
+#include <rnexecutorch/models/BaseModel.h>
+
+namespace rnexecutorch::models::text_to_speech::kokoro {
+
+class Encoder : public BaseModel {
+public:
+  explicit Encoder(const std::string &modelSource,
+                   std::shared_ptr<react::CallInvoker> callInvoker);
+
+  Result<std::vector<EValue>> generate(const std::string &method,
```
chmjkb, line 34: same here, a comment would be nice

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Encoder.h
```diff
@@ -0,0 +1,27 @@
...
+                                       const Configuration &inputConfig,
+                                       std::span<Token> tokens,
+                                       std::span<bool> textMask,
+                                       std::span<float> pred_aln_trg);
+};
```
chmjkb, line 39: Same here, high-level comments would be nice

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,245 @@
...
+  auto audioTensor = decoding->at(0).toTensor();
+
+  // Cut the resulting audio vector according to the effective duration
+  int32_t effLength = constants::kTicksPerDuration * effectiveDuration;
+  auto audio =
+      std::span<const float>(audioTensor.const_data_ptr<float>(), effLength);
+  auto croppedAudio =
+      utils::stripAudio(audio, constants::kSamplesPerMilisecond * 50);
+
+  std::vector<float> result(croppedAudio.begin(), croppedAudio.end());
```
chmjkb, line 229: Maybe we can return span instead of vector to avoid copying?

chmjkb (changes requested):

packages/react-native-executorch/src/constants/modelUrls.ts
IgorSwat: It doesn't, but not exporting this would mean that TTS constants files (models.ts and voices.ts) would require it's own copies of the constants. Then after the version changes we would have to change those constants in all of those files.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
msluszniak: @IgorSwat I think this change is not applied

msluszniak:
Please also fix all the alerts that are right now. 
execu

IgorSwat:
> Please also fix all the alerts that are right now. execu

image
No alerts on my screen 🤷

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (paths === null || paths.length !== 7 || paths.some((p) => p == null)) {
+      throw new Error('Download interrupted or missing resource.');
+    }
+
+    const modelPaths = paths.slice(0, 4) as [string, string, string, string];
+    const voiceDataPath = paths[4] as string;
+    const phonemizerPaths = paths.slice(5, 7) as [string, string];
+
+    console.log('[rnExecutorch] model paths: ', modelPaths);
```
msluszniak: Console log shoudn't be used. We have a special logging via Logger ;)

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,135 @@
...
+    onDownloadProgressCallback: (progress: number) => void,
+    options?: any
+  ): Promise<void> {
+    if (!voice.extra || !voice.extra.tagger || !voice.extra.lexicon) {
+      throw new Error(
+        'Kokoro: voice config is missing required extra fields: tagger and/or lexicon.'
+      );
+    }
+
+    console.log('[rnExecutorch] Pobierane zasoby:', {
```
msluszniak: remove console.log and use logger instead, also switch polish to english ;p

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -0,0 +1,153 @@
...
+  float scaleFactor = static_cast<float>(targetDuration) / totalDur;
+  bool shrinking = scaleFactor < 1.F;
+
+  // We need to scale partial durations (integers) corresponding to each token
+  // in a way that they all sum up to target duration, while keeping the balance
+  // between the values.
+  std::priority_queue<std::pair<float, int>>
+      remainders; // Sorted by the first value
+  int64_t scaledSum = 0;
+  for (int i = 0; i < nTokens; i++) {
```
msluszniak: Change int to uint32_t

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  return result;
+}
+
+const Configuration &Kokoro::selectConfig(size_t inputSize) const {
+  std::string modelLabel =
+      fixedModel_.has_value()                         ? fixedModel_.value()
+      : inputSize <= constants::kInputSmall.noTokens  ? "small"
+      : inputSize <= constants::kInputMedium.noTokens ? "medium"
+                                                      : "large";
```
benITo47, line 239: Heavy ternary operator usage makes it unreadable. Please try to use if statements or switch/case if possible  😁

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -0,0 +1,187 @@
...
+// Helper function - cost estimation (by string size)
+Partitioner::Cost Partitioner::cost(size_t stringSize) {
+  size_t effSize = stringSize + 2;
+
+  // If fixed model is set, we are limited to using only one of the models.
+  std::string activeModel =
+      fixedModel_.has_value()                       ? fixedModel_.value()
+      : effSize <= constants::kInputSmall.noTokens  ? "small"
+      : effSize <= constants::kInputMedium.noTokens ? "medium"
+                                                    : "large";
```
benITo47, line 145: Another difficult to read ternary operator

packages/react-native-executorch/scripts/create-package.sh
chmjkb, line 1: You changed the file permissions so it won't be able to execute on CI now.

msluszniak:
When you navigate to example with STT (and TTS) you cannot return to starting page, please make sure to implement some simple navigation for these examples ;)

Also in the quiz, when someone had chosen the answer before lector finished, error:
```
 ERROR  [Error: ModelGenerating]
 ```
 is generated

File: packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts [hunk 40]
```diff
@@ -0,0 +1,136 @@
...
+import {
+  KokoroConfig,
+  KokoroOptions,
+  TextToSpeechConfig,
+  TextToSpeechStreamingInput,
+  VoiceConfig,
+} from '../../types/tts';
+
+export class TextToSpeechModule extends BaseModule {
+  public async load(
```
chmjkb, line 14:
```suggestion
  public override async load(
```

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -0,0 +1,136 @@
...
+import { BaseModule } from '../BaseModule';
+import {
+  KokoroConfig,
+  KokoroOptions,
+  TextToSpeechConfig,
+  TextToSpeechStreamingInput,
+  VoiceConfig,
+} from '../../types/tts';
+
+export class TextToSpeechModule extends BaseModule {
```
chmjkb: I'm not sure it should inherit from BaseModule. What happens if I call getInputShape or forwardET?

packages/react-native-executorch/src/types/tts.ts
```diff
@@ -0,0 +1,91 @@
...
+  | 'en-gb'; // British English
+
+/**
+ * Voice configuration
+ *
+ * So far in Kokoro, each voice is directly associated with a language.
+ *
+ * @property {TextToSpeechLanguage} lang - speaker's language
+ * @property {ResourceSource} voiceSource - a source to a binary file with voice embedding
+ * @property {Record<string, unknown>} [extra] - an optional extra sources or properties related to specific voice
```
chmjkb: I think it should be KokoroVoiceExtras

IgorSwat:
> When you navigate to example with STT (and TTS) you cannot return to starting page, please make sure to implement some simple navigation for these examples ;)
> 
> Also in the quiz, when someone had chosen the answer before lector finished, error:
> 
> ```
>  ERROR  [Error: ModelGenerating]
> ```
> 
> is generated

Fixed.

packages/react-native-executorch/scripts/create-package.sh
IgorSwat: I made it executable back with chmod +x. I hope it is enough.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
IgorSwat: Sorry, I like ternary a lot (:

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
IgorSwat: Sorry, I like ternary a lot (:

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
benITo47: But I insist :-)

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
benITo47: But I insist :-)

chmjkb (approved):

chmjkb:
great work!

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
IgorSwat: But I insist harder :-)

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
IgorSwat: But I insist harder :-)

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
benITo47: Overall, that's some solid chunk of code. Great job, so I won't haggle over semantics.  Approved! 👍

danieloi:
Stoked to try this out! Curious how it performs compared to the onnx version that often blocks the main thread.

Issue #692, opened 2025-11-27
Re-export OCR to use single weights


chmjkb:
### Problem description

Currently our OCR pipeline needs using 3 different input shapes, so there is 3 different models exported, each with a different input shape, but with the same weights. This creates unnecessary overhead, since the users need to download the same weights 3 times. 

### Proposed solution

see https://docs.pytorch.org/executorch/stable/using-executorch-export.html#multi-method-models

### Benefits to React Native ExecuTorch

Less overhead for the users

msluszniak:
Don't dynamic shapes work here still?

chmjkb:
> Don't dynamic shapes work here still?

@msluszniak If you're suggesting to export the model with dynamic shapes then unfortunately the export fails when we do so, only static shape export passes AFAIK

msluszniak:
Yeap, that's what I meant

msluszniak:
Maybe we should share the export errors with dynamic shapes with executorch team then?

PR #716, opened 2026-01-14
feat: Change OCR to multi-method model 

benITo47:
## Description
This PR refactors the OCR and VerticalOCR to improve efficiency. 

Previously, our implementation relied on multiple instances of the same detector and recognizer models, each handling a different input size (3x Detector, 4x Recognizer instances). This approach was resource-intensive.

This update introduces a more streamlined approach by using a single detector and a single recognizer model, each with multiple forward_ methods (e.g., forward_800, forward_320). These methods handle different input widths within the same model instance, significantly reducing the number of loaded models and simplifying the API.

This change is a breaking change as it modifies the arguments for useOCR, useVerticalOCR, OCRModule, and VerticalOCRModule

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Manual sanity checks. 

### Related issues

[#692](https://github.com/software-mansion/react-native-executorch/issues/692)

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
@@ -1,13 +1,28 @@
 #include "Detector.h"
+#include "Constants.h"
+#include <cstdint>
 #include <rnexecutorch/data_processing/ImageProcessing.h>
 #include <rnexecutorch/models/ocr/Constants.h>
 #include <rnexecutorch/models/ocr/utils/DetectorUtils.h>
+#include <stdexcept>
 #include <string>
-
 namespace rnexecutorch::models::ocr {
 Detector::Detector(const std::string &modelSource,
                    std::shared_ptr<react::CallInvoker> callInvoker)
-    : BaseModel(modelSource, callInvoker) {}
+    : BaseModel(modelSource, callInvoker) {
+
+  for (auto input_size : constants::kDetectorInputWidths) {
+    std::string methodName = "forward_" + std::to_string(input_size);
+    auto inputShapes = getAllInputShapes(methodName);
+    if (inputShapes[0].size() < 2) {
+      throw std::runtime_error(
+          "Unexpected detector model input size for method:" + methodName +
+          ", expected "
+          "at least 2 dimensions but got: " +
+          std::to_string(inputShapes[0].size()) + ".");
+    }
+  }
+}
 
 std::vector<types::DetectorBBox> Detector::generate(const cv::Mat &inputImage,
                                                     int32_t inputWidth) {
@@ -19,24 +34,13 @@ std::vector<types::DetectorBBox> Detector::generate(const cv::Mat &inputImage,
    original aspect ratio and the missing parts are filled with padding.
    */
 
+  utils::validateInputWidth(inputWidth, constants::kDetectorInputWidths,
+                            "Detector");
+
   std::string methodName = "forward_" + std::to_string(inputWidth);
-
   auto inputShapes = getAllInputShapes(methodName);
-  if (inputShapes.empty()) {
-    throw std::runtime_error("Detector model: invalid method name " +
-                             methodName);
-  }
 
-  std::vector<int32_t> modelInputShape = inputShapes[0];
-
-  if (modelInputShape.size() < 2) {
-    throw std::runtime_error("Detector model: invalid method name: " +
-                             methodName);
-  }
-
-  cv::Size modelInputSize =
-      cv::Size(modelInputShape[modelInputShape.size() - 1],
-               modelInputShape[modelInputShape.size() - 2]);
+  cv::Size modelInputSize = calculateModelImageSize(inputWidth);
 
   cv::Mat resizedInputImage =
       image_processing::resizePadded(inputImage, modelInputSize);
@@ -44,18 +48,32 @@ std::vector<types::DetectorBBox> Detector::generate(const cv::Mat &inputImage,
       inputShapes[0], resizedInputImage, constants::kNormalizationMean,
       constants::kNormalizationVariance);
   auto forwardResult = BaseModel::execute(methodName, {inputTensor});
+
   if (!forwardResult.ok()) {
     throw std::runtime_error(
-        "Failed to forward, error: " +
+        "Failed to " + methodName + " error: " +
         std::to_string(static_cast<uint32_t>(forwardResult.error())));
   }
 
   return postprocess(forwardResult->at(0).toTensor(), modelInputSize);
 }
 
+cv::Size Detector::calculateModelImageSize(int32_t methodInputWidth) {
+
+  utils::validateInputWidth(methodInputWidth, constants::kDetectorInputWidths,
+                            "Detector");
+  std::string methodName = "forward_" + std::to_string(methodInputWidth);
+
+  auto inputShapes = getAllInputShapes(methodName);
+  std::vector<int32_t> modelInputShape = inputShapes[0];
+  cv::Size modelInputSize =
+      cv::Size(modelInputShape[modelInputShape.size() - 1],
+               modelInputShape[modelInputShape.size() - 2]);
+  return modelInputSize;
+}
+
 std::vector<types::DetectorBBox>
-Detector::postprocess(const Tensor &tensor,
-                      const cv::Size &modelInputSize) const {
+Detector::postprocess(const Tensor &tensor, const cv::Size &modelInputSize) {
   /*
    The output of the model consists of two matrices (heat maps):
    1. ScoreText(Score map) - The probability of a region containing character.
--- a/packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.h
+++ b/packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.h
@@ -17,15 +17,20 @@ namespace rnexecutorch::models::ocr {
 using executorch::aten::Tensor;
 using executorch::extension::TensorPtr;
 
-class Detector final : public BaseModel {
+class Detector : public BaseModel {
 public:
   explicit Detector(const std::string &modelSource,
                     std::shared_ptr<react::CallInvoker> callInvoker);
-  std::vector<types::DetectorBBox> generate(const cv::Mat &inputImage,
-                                            int32_t inputWidth);
+  virtual std::vector<types::DetectorBBox> generate(const cv::Mat &inputImage,
+                                                    int32_t inputWidth);
 
-private:
-  std::vector<types::DetectorBBox>
-  postprocess(const Tensor &tensor, const cv::Size &modelInputSize) const;
+  cv::Size calculateModelImageSize(int32_t methodInputWidth);
+
+protected:
+  TensorPtr runInference(const cv::Mat &inputImage, int32_t inputWidth,
+                         const std::string &detectorName);
+
+  std::vector<types::DetectorBBox> postprocess(const Tensor &tensor,
+                                               const cv::Size &modelInputSize);
 };
 } // namespace rnexecutorch::models::ocr
--- a/packages/react-native-executorch/common/rnexecutorch/models/ocr/OCR.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/ocr/OCR.cpp
@@ -5,7 +5,8 @@
 
 namespace rnexecutorch::models::ocr {
 OCR::OCR(const std::string &detectorSource, const std::string &recognizerSource,
-         std::string symbols, std::shared_ptr<react::CallInvoker> callInvoker)
+         const std::string &symbols,
+         std::shared_ptr<react::CallInvoker> callInvoker)
     : detector(detectorSource, callInvoker),
       recognitionHandler(recognizerSource, symbols, callInvoker) {}
 
--- a/packages/react-native-executorch/common/rnexecutorch/models/ocr/Recognizer.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/ocr/Recognizer.cpp
@@ -1,9 +1,11 @@
 #include "Recognizer.h"
+#include "Constants.h"
 #include <numeric>
 #include <rnexecutorch/data_processing/ImageProcessing.h>
 #include <rnexecutorch/data_processing/Numerical.h>
 #include <rnexecutorch/models/ocr/Constants.h>
 #include <rnexecutorch/models/ocr/Types.h>
+#include <rnexecutorch/models/ocr/utils/DetectorUtils.h>
 #include <rnexecutorch/models/ocr/utils/RecognizerUtils.h>
 #include <string>
 
@@ -23,11 +25,14 @@ Recognizer::generate(const cv::Mat &grayImage, int32_t inputWidth) {
    The `generate` function as an argument accepts an image in grayscale
    already resized to the expected size.
   */
+  utils::validateInputWidth(inputWidth, constants::kRecognizerInputWidths,
+                            "Recognizer");
+
   std::string method_name = "forward_" + std::to_string(inputWidth);
   auto shapes = getAllInputShapes(method_name);
   if (shapes.empty()) {
-    throw std::runtime_error("Recognizer model: invalid method name " +
-                             method_name);
+    throw std::runtime_error("Recognizer model: Input shapes for  " +
+                             method_name " not found");
   }
   std::vector<int32_t> tensorDims = shapes[0];
   TensorPtr inputTensor =
--- a/packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
@@ -11,36 +11,37 @@ namespace rnexecutorch::models::ocr {
 VerticalDetector::VerticalDetector(
     const std::string &modelSource,
     std::shared_ptr<react::CallInvoker> callInvoker)
-    : BaseModel(modelSource, callInvoker) {
-  this->modelSmallImageSize =
-      calculateImageSizeForWidth(constants::kSmallDetectorWidth);
-  this->modelMediumImageSize =
-      calculateImageSizeForWidth(constants::kMediumDetectorWidth);
-  this->modelLargeImageSize =
-      calculateImageSizeForWidth(constants::kLargeDetectorWidth);
-}
+    : Detector(modelSource, callInvoker) {}
 
 std::vector<types::DetectorBBox>
-VerticalDetector::generate(const cv::Mat &inputImage, int32_t inputWidth,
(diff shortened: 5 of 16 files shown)


packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.h
```diff
@@ -20,11 +20,11 @@ class Detector final : public BaseModel {
 public:
   explicit Detector(const std::string &modelSource,
                     std::shared_ptr<react::CallInvoker> callInvoker);
-  std::vector<types::DetectorBBox> generate(const cv::Mat &inputImage);
-  cv::Size getModelImageSize() const noexcept;
+  std::vector<types::DetectorBBox> generate(const cv::Mat &inputImage,
+                                            const int inputWidth);
```
IgorSwat: Unnecessary const.
Also use int32_t (or any other explicit int type you want) instead of an int.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -2,54 +2,60 @@
...
    original aspect ratio and the missing parts are filled with padding.
    */
-  auto inputShapes = getAllInputShapes();
+
+  std::string methodName = "forward_" + std::to_string(inputWidth);
+
+  auto inputShapes = getAllInputShapes(methodName);
+  if (inputShapes.empty()) {
+    throw std::runtime_error("Detector model has no input shape for method: " +
+                             methodName);
```
IgorSwat: I would change the error message to something like "Detector model: invalid method name XYZ"

msluszniak:
The cpp code and docs look good, someone needs to review the rest and test it

packages/react-native-executorch/common/rnexecutorch/models/ocr/Recognizer.cpp
```diff
@@ -5,25 +5,15 @@
...
-    throw std::runtime_error("Unexpected Recognizer model input shape.");
-  }
-  modelImageSize = cv::Size(modelInputShape[modelInputShape.size() - 1],
-                            modelInputShape[modelInputShape.size() - 2]);
-}
+    : BaseModel(modelSource, callInvoker) {}
 
 std::pair<std::vector<int32_t>, float>
-Recognizer::generate(const cv::Mat &grayImage) {
+Recognizer::generate(const cv::Mat &grayImage, int inputWidth) {
```
IgorSwat: int32_t please :)

packages/react-native-executorch/common/rnexecutorch/models/ocr/Recognizer.cpp
```diff
@@ -33,10 +23,16 @@ Recognizer::generate(const cv::Mat &grayImage) {
    The `generate` function as an argument accepts an image in grayscale
    already resized to the expected size.
   */
-  std::vector<int32_t> tensorDims = getAllInputShapes()[0];
+  std::string method_name = "forward_" + std::to_string(inputWidth);
+  auto shapes = getAllInputShapes(method_name);
+  if (shapes.empty()) {
+    throw std::runtime_error(
+        "Recognizer model has no input tensors for method " + method_name);
+  }
```
IgorSwat, line 36: I would change the error message to something like "Recognizer model: invalid method name XYZ"

packages/react-native-executorch/common/rnexecutorch/models/ocr/Recognizer.h
```diff
@@ -25,12 +25,11 @@ class Recognizer final : public BaseModel {
 public:
   explicit Recognizer(const std::string &modelSource,
                       std::shared_ptr<react::CallInvoker> callInvoker);
-  std::pair<std::vector<int32_t>, float> generate(const cv::Mat &grayImage);
+  std::pair<std::vector<int32_t>, float> generate(const cv::Mat &grayImage,
+                                                  int inputWidth);
```
IgorSwat: int32_t please :)

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
```diff
@@ -5,51 +5,49 @@
...
-}
-
-cv::Size VerticalDetector::getModelImageSize() const noexcept {
-  return modelImageSize;
+  modelSmallImageSize =
+      calculateImageSizeForWidth(constants::kSmallDetectorWidth);
+  modelMediumImageSize =
+      calculateImageSizeForWidth(constants::kMediumDetectorWidth);
+  modelLargeImageSize =
+      calculateImageSizeForWidth(constants::kLargeDetectorWidth);
```
IgorSwat: I recommend using either this-> (as in the line above, ex. this->modelSmallImageSize), or marking the class members with _ at the end (ex. modelSmallImageSize_).

By doing so it's clear which fields are class members, and which one are not.

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
```diff
@@ -67,20 +65,20 @@ VerticalDetector::postprocess(const Tensor &tensor) const {
    */
   auto [scoreTextMat, scoreAffinityMat] = utils::interleavedArrayToMats(
       tensorData,
-      cv::Size(modelImageSize.width / 2, modelImageSize.height / 2));
-  float txtThreshold = this->detectSingleCharacters
+      cv::Size(modelInputSize.width / 2, modelInputSize.height / 2));
+  float txtThreshold = detectSingleCharacters
```
IgorSwat, line 69: Why is this-> removed?

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
```diff
@@ -67,20 +65,20 @@ VerticalDetector::postprocess(const Tensor &tensor) const {
...
           scoreTextMat, scoreAffinityMat, txtThreshold,
-          constants::kLinkThreshold, this->detectSingleCharacters);
+          constants::kLinkThreshold, detectSingleCharacters);
   const float restoreRatio = utils::calculateRestoreRatio(
       scoreTextMat.rows, constants::kRecognizerImageSize);
   utils::restoreBboxRatio(bBoxesList, restoreRatio);
 
   // if this is Narrow Detector, do not group boxes.
-  if (!this->detectSingleCharacters) {
+  if (!detectSingleCharacters) {
```
IgorSwat, line 81: Why is this-> removed?

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
benITo47: Before detectSingleCharacters was member variable of Detector class.

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
benITo47: It's no longer member variable

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -2,54 +2,60 @@
...
-    throw std::runtime_error("Unexpected detector model input size, expected "
-                             "at least 2 dimensions but got: " +
-                             std::to_string(modelInputShape.size()) + ".");
-  }
-  modelImageSize = cv::Size(modelInputShape[modelInputShape.size() - 1],
-                            modelInputShape[modelInputShape.size() - 2]);
-}
-
-cv::Size Detector::getModelImageSize() const noexcept { return modelImageSize; }
+    : BaseModel(modelSource, callInvoker) {}
```
chmjkb: I think we can still do the check that was done previously, but for each exported method.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -2,54 +2,60 @@
...
+    throw std::runtime_error("Detector model: invalid method name " +
+                             methodName);
+  }
+
+  std::vector<int32_t> modelInputShape = inputShapes[0];
+
+  if (modelInputShape.size() < 2) {
+    throw std::runtime_error("Detector model: invalid method name: " +
+                             methodName);
+  }
```
chmjkb: I think the error message is wrong here, this can be checked in constructor too.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -2,54 +2,60 @@
...
    original aspect ratio and the missing parts are filled with padding.
    */
-  auto inputShapes = getAllInputShapes();
+
+  std::string methodName = "forward_" + std::to_string(inputWidth);
+
+  auto inputShapes = getAllInputShapes(methodName);
+  if (inputShapes.empty()) {
+    throw std::runtime_error("Detector model: invalid method name " +
+                             methodName);
```
chmjkb: This is never going to happen, see `getAllInputShapes` source:
```
  auto method_meta = module_->method_meta(methodName);
  if (!method_meta.ok()) {
    throw std::runtime_error(
        "Failed to get metadata for method '" + methodName + "': Error " +
        std::to_string(static_cast<int>(method_meta.error())));
  }
```

packages/react-native-executorch/common/rnexecutorch/models/ocr/OCR.cpp
```diff
@@ -1,16 +1,13 @@
...
 #include <rnexecutorch/models/ocr/Constants.h>
 
 namespace rnexecutorch::models::ocr {
-OCR::OCR(const std::string &detectorSource,
-         const std::string &recognizerSourceLarge,
-         const std::string &recognizerSourceMedium,
-         const std::string &recognizerSourceSmall, std::string symbols,
-         std::shared_ptr<react::CallInvoker> callInvoker)
+OCR::OCR(const std::string &detectorSource, const std::string &recognizerSource,
+         std::string symbols, std::shared_ptr<react::CallInvoker> callInvoker)
```
chmjkb: Why std::string symbols?

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
```diff
@@ -90,4 +87,45 @@ VerticalDetector::postprocess(const Tensor &tensor) const {
...
+VerticalDetector::calculateImageSizeForWidth(const int methoInputWidth) {
+
+  std::string methodName = "forward_" + std::to_string(methoInputWidth);
+
+  auto inputShapes = getAllInputShapes(methodName);
+
+  if (inputShapes.empty()) {
+    throw std::runtime_error("Detector model has no input shape for method: " +
+                             methodName);
+  }
```
chmjkb: Same as above, getAllInputshapes will throw when the method doesn't exist.

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.h
```diff
@@ -15,16 +15,17 @@ namespace rnexecutorch::models::ocr {
...
+ obtain detected boxes from heeat maps it utilizes
+ `getDetBoxesFromTextMapVertical()` function rather than
  'getDetBoxesFromTextMap()`. Other than that, refer to the standard OCR
  Detector.
 
-  2. Narrow Detector -- it is designed to detect a single characters bounding
- boxes. `getDetBoxesFromTextMapVertical()` function acts differently for Narrow
- Detector and different textThreshold Value is passed. Additionally, the
- grouping of detected boxes is completely omited.
+  2. Narrow Detector throguh forward_320 method -- it is designed to detect a
```
chmjkb:
```suggestion
  2. Narrow Detector through forward_320 method -- it is designed to detect a
```

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1,9 +1,9 @@
 import { Platform } from 'react-native';
 
-const URL_PREFIX =
+export const URL_PREFIX =
   'https://huggingface.co/software-mansion/react-native-executorch';
-const VERSION_TAG = 'resolve/v0.6.0';
-// const NEXT_VERSION_TAG = 'resolve/v0.7.0';
+export const VERSION_TAG = 'resolve/v0.7.0';
+// const NEXT_VERSION_TAG = 'resolve/v0.8.0';
```
chmjkb: u sure?

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
msluszniak: Couldn't be the case (hypothetically) when method worked and the returned tensor is empty?

benITo47: ???

chmjkb: > Couldn't be the case (hypothetically) when method worked and the returned tensor is empty?

Can you elaborate? We're not sure what you mean

packages/react-native-executorch/src/constants/modelUrls.ts
benITo47: Nah, not really

packages/react-native-executorch/common/rnexecutorch/models/vertical_ocr/VerticalDetector.cpp
msluszniak: getAllInputShapes didn't throw and inputShapes is an empty container. Indeed, there is a case when method takes no arguments, then output from `inputShapes` will be an empty vector. I don't know if it's a case, but hypothetically this control block catches cases of such empty vectors which are not raised in getAllInputShapes

packages/react-native-executorch/common/rnexecutorch/models/ocr/OCR.cpp
benITo47: Good catch! Symbols were consumed downstream by CTCLavelConverter class which accepted const ref in its constructor.  Compiler probably took care of it  and omitted unnecessary copies, but I changed signatures along the way to const refs for clarity.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -1,13 +1,34 @@
...
+    std::string methodName = "forward_" + std::to_string(input_size);
+    auto inputShapes = getAllInputShapes(methodName);
+    if (inputShapes[0].size() < 2) {
+      throw std::runtime_error(
+          "Unexpected detector model input size for method:" + methodName +
+          ", expected "
+          "at least 2 dimensions but got: " +
+          std::to_string(inputShapes[0].size()) + ".");
+    }
+  }
```
IgorSwat, line 24: I would not store this as class members. Those are just 2 integers, so recalculating this is cheap and could simplify the code logic.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -19,40 +40,46 @@ std::vector<types::DetectorBBox> Detector::generate(const cv::Mat &inputImage,
...
-        "Failed to forward, error: " +
+        "Failed to " + methodName + " error: " +
         std::to_string(static_cast<uint32_t>(forwardResult.error())));
   }
 
   return postprocess(forwardResult->at(0).toTensor(), modelInputSize);
 }
 
+cv::Size Detector::getModelImageSize(int inputWidth) const noexcept {
+  switch (inputWidth) {
```
IgorSwat: I would remove this function and use calculateImageSizeForWidth instead to keep the code simpler - but this is just my subjective opinion.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -103,4 +130,28 @@ Detector::postprocess(const Tensor &tensor,
...
+  if (inputShapes.empty()) {
+    throw std::runtime_error("Detector model has no input shape for method: " +
+                             methodName);
+  }
+  std::vector<int32_t> modelInputShape = inputShapes[0];
+
+  if (modelInputShape.size() < 2) {
+    throw std::runtime_error("Unexpected detector model input size, expected "
+                             "at least 2 dimensions but got: " +
+                             std::to_string(modelInputShape.size()) + ".");
```
IgorSwat: I think you can refactor this part and replace most of it with a call to validateInputWidth().

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.h
```diff
@@ -17,14 +17,25 @@ namespace rnexecutorch::models::ocr {
...
+                                                    int32_t inputWidth);
+
+  cv::Size getModelImageSize(int inputWidth) const noexcept;
+
+protected:
+  cv::Size calculateImageSizeForWidth(const int methoInputWidth);
+
+  cv::Size modelSmallImageSize;
+  cv::Size modelMediumImageSize;
+  cv::Size modelLargeImageSize;
```
IgorSwat: Same as above - I would personally not store those values as class members.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.cpp
```diff
@@ -103,4 +130,28 @@ Detector::postprocess(const Tensor &tensor,
   return bBoxesList;
 }
 
+cv::Size Detector::calculateImageSizeForWidth(const int methoInputWidth) {
```
IgorSwat: Use int32_t without const.

packages/react-native-executorch/common/rnexecutorch/models/ocr/Detector.h
benITo47: Agreed. Calculation is cheap and rare  - we can get rid of it in favour of readability

msluszniak:
I tested OCR on demo apps on your branch, and I don't know if the models are just off or something is wrong:
image
image

benITo47:
image
@msluszniak 
Checked your example against main branch. It's not caused by my changes

msluszniak:
Ok, these fields with 0.000 were especially concerning. I don't think it should show any box that recognises no signs 😅 Maybe we should fix this in demo app to not show them?

msluszniak:
I guess we released code which does not compile:

image
 It was not visible because CI failed earlier because of cache bloating. Please fix this @benITo47

benITo47:
Rebasing error. On it

PR #730, opened 2026-01-15
Remove redundant packages in CI 

msluszniak:
## Description

Currently, CI for Android fails because machine is bloated with not needed code. After adding step in CI that unbloats machine the ci works: 
https://github.com/software-mansion/react-native-executorch/actions/runs/21036801074

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist

### DIFF
--- a/.github/workflows/build-android-llm-example.yml
+++ b/.github/workflows/build-android-llm-example.yml
@@ -23,19 +23,13 @@ jobs:
       group: android-${{ github.ref }}
       cancel-in-progress: true
     steps:
-      - name: Free Disk Space (Ubuntu)
-        uses: jlumbroso/free-disk-space-action@v1.3.1
-        with:
-            # ⚠️ IMPORTANT: If your job builds the Android app, set 'android: false'
-            # to prevent deleting the Android SDK/NDK.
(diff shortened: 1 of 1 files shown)


.github/workflows/build-android-llm-example.yml
```diff
@@ -23,6 +23,13 @@ jobs:
       group: android-${{ github.ref }}
       cancel-in-progress: true
     steps:
+      - name: Free Disk Space (Manual)
+        run: |
+              sudo rm -rf /usr/share/dotnet
```
chmjkb, line 28: do we need this? we're not using dotnet

msluszniak: Do you have a list of packages that are preinstalled on machine used for CI? Then, I will be able to remove packages precisely. And machines used for CIs usually have dotnet preinstalled.

chmjkb: I don't really have such list, but if this works then It's good enough. Thanks!

chmjkb (approved):

PR #759, opened 2026-01-27
feat!: add modular resource fetcher adapters for Expo and bare React Native

### LINKED ISSUE #549: Explore removing expo dependency
Explore how we could remove expo dependency by separating Resource Fetcher outside of our library (this it the only feature depending on expo)

rizalibnu:
## Description

This PR introduces modular resource fetcher adapters to support both Expo and bare React Native environments, replacing the previous monolithic approach with a flexible, platform-specific architecture.

**Key Changes**
New Adapter Packages:

- @rn-executorch/expo-adapter: Resource fetcher for Expo projects using expo-file-system and expo-asset
- @rn-executorch/bare-adapter: Resource fetcher for bare React Native projects using @dr.pogodin/react-native-fs and @kesha-antonov/react-native-background-downloader

**Initialization Changes:**

- Added initExecutorch() function that requires explicit adapter selection
- Users must now choose and configure the appropriate adapter for their project type
- Provides better separation of concerns and platform-specific optimizations

**Documentation Updates:**
- Created individual README.md files for each adapter package

- [x] Yes
- [ ] No

**Migration Required:**
Users must now explicitly initialize the library with a resource fetcher adapter:

```typescript
// Before (no initialization needed)
import { useLLM } from 'react-native-executorch';

// After (required initialization)
import { initExecutorch, useLLM } from 'react-native-executorch';
import { ExpoResourceFetcher } from '@rn-executorch/expo-adapter'; // or BareResourceFetcher

initExecutorch({
  resourceFetcher: ExpoResourceFetcher,
});
```

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

**For Expo projects:**

- Install dependencies: `yarn add @rn-executorch/expo-adapter expo-file-system expo-asset`
- Initialize: `initExecutorch({ resourceFetcher: ExpoResourceFetcher })`
- Run existing LLM example app to verify model downloads work correctly

**For bare React Native projects:**

- Install dependencies: `yarn add @rn-executorch/bare-adapter @dr.pogodin/react-native-fs @kesha-antonov/react-native-background-downloader`
- Initialize: `initExecutorch({ resourceFetcher: BareResourceFetcher })`
- Run the bare React Native example app in PR #763

> Note: A separate PR will add a dedicated bare React Native example app to make this PR easier to review. The Expo example apps can be used to verify the Expo adapter functionality.

### Related issues

Closes #549 

### Checklist

### Additional notes

**Why This Change:**

- Different React Native environments have different filesystem APIs and c
- apabilities
- Expo projects benefit from using Expo's managed filesystem APIs
- Bare React Native projects can leverage native libraries with background download support
- Modular architecture allows for better platform-specific optimizations
- Enables future extensibility for other environments (e.g., React Native Windows, macOS)

**Split Into Multiple PRs:**
To make review easier, this work has been split:

- This PR: Core adapter infrastructure and Expo adapter implementation
- Follow-up PR: Bare React Native example app demonstrating the bare adapter usage

BREAKING CHANGE: `initExecutorch()` with explicit adapter selection is now required before using any react-native-executorch hooks. Users must install and configure either `@rn-executorch/expo-adapter` or `@rn-executorch/bare-adapter` depending on their project type.

### DIFF
--- a/packages/react-native-executorch/src/utils/ResourceFetcher.ts
+++ b/packages/react-native-executorch/src/utils/ResourceFetcher.ts
@@ -1,6 +1,7 @@
 import { ResourceSource } from '../types/common';
 import { RnExecutorchError } from '../errors/errorUtils';
 import { RnExecutorchErrorCode } from '../errors/ErrorCodes';
+import { ResourceFetcherUtils } from './ResourceFetcherUtils';
 
 /**
  * Adapter interface for resource fetching operations.
@@ -105,6 +106,15 @@ export class ResourceFetcher {
     callback: (downloadProgress: number) => void = () => {},
     ...sources: ResourceSource[]
   ) {
+    for (const source of sources) {
+      if (typeof source === 'string') {
+        try {
+          ResourceFetcherUtils.triggerHuggingFaceDownloadCounter(source);
+        } catch (error) {
+          throw error;
+        }
+      }
+    }
     return this.getAdapter().fetch(callback, ...sources);
   }
 


msluszniak:
Lint CI fails (you don't need to worry about the second failing CI ;)). Please could you fix the errors from this CI?

rizalibnu:
@msluszniak I’m not able to reproduce the lint CI failure locally — everything passes on my side.

I’ll take a closer look and investigate further to see what might be causing the discrepancy (environment, cache, or config differences). I’ll follow up with a fix or more details as soon as I find the root cause.

Screenshot 2026-01-27 at 17 19 22

msluszniak:
@rizalibnu Sure thing, maybe the configuration of the CI itself does not work with as it should. We'll also look at this, don't worry :)

rizalibnu:
@msluszniak Found the issue 👍
CI was failing because `react-native-executorch` types come from `lib/typescript`, which isn’t built on a fresh run. It worked locally since I already had the build artifacts.

Fixed by adding a build step before adapter type checks and bumped Node to 22 in `.nvmrc` due to an ESM-only dependency (`arktype` via `react-native-builder-bob`).

package.json
```diff
@@ -5,7 +5,12 @@
   "workspaces": {
     "packages": [
       "packages/react-native-executorch",
-      "apps/*"
+      "packages/bare-adapter",
+      "packages/expo-adapter",
+      "apps/computer-vision",
+      "apps/llm",
+      "apps/speech",
+      "apps/text-embeddings"
```
msluszniak: Why we need to expanded `apps/*`

packages/expo-adapter/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@rn-executorch/expo-adapter",
+  "version": "0.1.0",
```
msluszniak: Do we want to have version of adapters to be divergent from version of React Native ExecuTorch cc: @chmjkb @mkopcins ?

chmjkb: I think that's fine. We just need to keep in mind to add a compatibility table, if some version is not compatible at some point. We can also make sure that RNET is a peer dependency of the adapters so we can narrow down on the versions there.

msluszniak: I just remember from one project, that we wanted at some point unify the versions of packages in monorepo, but it was to late since we started with the approach as here. Same thing as Apple did with OS versions on multiple devices ;p

package.json
rizalibnu: @msluszniak We need to expand `apps/*` because `apps/bare_rn` in PR #763 must be excluded from the Yarn workspace. An alternative is explicitly excluding it in package.json as shown below.

```json
"workspaces": [
  "apps/*",
  "!apps/bare_rn"
]
```

Which approach do you prefer?

msluszniak: Personally, I prefer the alternative, because if we add another apps in the future, we don't need to remember to update this file. But I would like to know others opinion, cc: @chmjkb @mkopcins

rizalibnu: I agree with the alternative 👍
Using `apps/*` with an explicit exclusion is more future-proof.
Should we apply the same glob + exclusion pattern to `packages/` as well for consistency?

```json
"workspaces": {
  "packages": [
    "packages/*",
    "apps/*",
    "!apps/bare_rn"
  ]
}
```

packages/expo-adapter/package.json
rizalibnu: That’s a fair concern.

While independent adapter releases are useful early on, IMO I think a good compromise might be to align major/minor versions between adapters and RNET, while allowing independent patch releases. This keeps compatibility clear for users without blocking adapter-only fixes.

Happy to follow whatever versioning direction the maintainers prefer.

apps/llm/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "customConditions": ["react-native"],
     "noEmit": true,
     "paths": {
-      "react-native-executorch": ["../../packages/react-native-executorch/src"]
+      "react-native-executorch": ["../../packages/react-native-executorch/src"],
+      "@rn-executorch/expo-adapter": ["../../packages/expo-adapter/src"]
```
chmjkb: nit: we're likely going to call the package `react-native-executorch/expo-resource-fetcher` or `react-native-executorch/bare-resource-fetcher`, so let's stick to that naming 👍🏻

package.json
chmjkb: I think I'm on the alternative side as we likely won't add much bare RN example apps in the future.

packages/react-native-executorch/src/index.ts
```diff
@@ -1,4 +1,20 @@
...
+export interface ExecutorchConfig {
+  resourceFetcher: ResourceFetcherAdapter;
+}
+
+export function initExecutorch(config: ExecutorchConfig) {
+  ResourceFetcher.setAdapter(config.resourceFetcher);
+}
+
+export function cleanupExecutorch() {
+  ResourceFetcher.setAdapter(null as unknown as ResourceFetcherAdapter);
```
chmjkb: Maybe it would be a tiny bit cleaner if we had a resetAdapter or something like this exposed from ResourceFetcher

packages/bare-adapter/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+## When to Use
+
+Use this adapter if you're working with:
+- Bare React Native projects (created with `npx @react-native-community/cli@latest init`)
+- Projects that need true background downloads
+- Projects requiring direct native filesystem access
+
+This adapter uses `@dr.pogodin/react-native-fs` and `@kesha-antonov/react-native-background-downloader` for enhanced file operations and background download support.
```
chmjkb: Maybe it would be cool to link here to their installation steps

packages/bare-adapter/src/ResourceFetcherUtils.ts
```diff
@@ -0,0 +1,92 @@
...
+      } finally {
+        results.push({ source, type, length, previousFilesTotalLength });
+      }
+    }
+    return { results, totalLength };
+  }
+
+  export async function createDirectoryIfNoExists() {
+    if (!(await checkFileExists(RNEDirectory))) {
+      await RNFS.mkdir(RNEDirectory);
```
chmjkb: Can we try catch here and re-throw rnexecutorch erorr?

packages/expo-resource-fetcher/src/ResourceFetcherUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export { HTTP_CODE, DownloadStatus, SourceType, ResourceSourceExtended };
+
+export interface DownloadResource {
+  downloadResumable: DownloadResumable;
+  status: DownloadStatus;
+  extendedInfo: ResourceSourceExtended;
+}
+
+export namespace ResourceFetcherUtils {
```
chmjkb, line 37: Is there a reason for exporting this as a namespace?

packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -28,427 +28,80 @@
...
-    sourceExtended.results.push(result);
+  ): Promise<string[] | null>;
+  pauseFetching(...sources: ResourceSource[]): Promise<void>;
+  resumeFetching(...sources: ResourceSource[]): Promise<void>;
+  cancelFetching(...sources: ResourceSource[]): Promise<void>;
+  listDownloadedFiles(): Promise<string[]>;
+  listDownloadedModels(): Promise<string[]>;
+  deleteResources(...sources: ResourceSource[]): Promise<void>;
+  getFilesTotalSize(...sources: ResourceSource[]): Promise<number>;
+  readAsString(path: string): Promise<string>;
```
chmjkb: I'm wondering if these should be enforced, in the library internals, we only use `fetch()`, meaning that in case where the user wants to implement his own resource fetcher, he only has to implement fetch() for the library internals to work. The other methods are just for convenience. We should either add a JSDoc here saying that or we should remove them from this interface completely. cc @mkopcins, wdyt?

mkopcins: yes, we should require the bare minimum for the fetcher to work and mark other method as optional to make implementing custom fetcher as simple as possible

packages/expo-resource-fetcher/src/ResourceFetcherUtils.ts
rizalibnu: @chmjkb It follows the original implementation in the existing codebase. I kept the same pattern to stay consistent and avoid unnecessary refactors.
https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts#L50

packages/react-native-executorch/src/utils/ResourceFetcher.ts
rizalibnu: I’ve simplified the interface to include only the methods that are actually used. Please re-review. @mkopcins @chmjkb

packages/bare-adapter/src/ResourceFetcherUtils.ts
rizalibnu: Updated. Please re-review. @chmjkb

packages/bare-adapter/README.md
rizalibnu: Updated. Please re-review. @chmjkb

packages/react-native-executorch/src/index.ts
rizalibnu: I’ve added resetAdapter. Please re-review. @chmjkb

apps/llm/tsconfig.json
rizalibnu: Updated. Please re-review. @chmjkb

package.json
rizalibnu: I’ve applied the glob pattern to `apps/` and `packages/`. Please re-review. @msluszniak @chmjkb

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -0,0 +1,473 @@
...
+      const nextSource = sourceExtended.next;
+      nextSource.results.push(...sourceExtended.results);
+      return this.singleFetch(nextSource);
+    }
+    sourceExtended.callback!(1);
+    return sourceExtended.results;
+  },
+
+  async pause(source: ResourceSource) {
+    const resource = this.downloads.get(source)!;
```
msluszniak:
```suggestion
    const resource = this.downloads.get(source);
    if (!resource) return;
```

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -0,0 +1,473 @@
...
+            progress.bytesDownloaded / progress.bytesTotal
+          );
+        })
+        .done(async () => {
+          if (
+            !this.downloads.has(source) ||
+            this.downloads.get(source)!.status === DownloadStatus.PAUSED
+          ) {
+            resolve(null);
+            return;
```
msluszniak: This is repeated across this file at least two times, please refactor this one, to some `onDownloadSuccess` or something like that ;)

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -0,0 +1,473 @@
...
+              info[0]!.length,
+              callback
+            )
+          : () => {},
+      results: [],
+    };
+
+    let node = head;
+    for (let idx = 1; idx < sources.length; idx++) {
+      node.next = {
```
msluszniak, line 143:
```suggestion
      const node.next = {
```

msluszniak:
Also I think that adding these changes will make section `Fundamentals` in documentation out of date. Could look at these three subsections, and make sure they will be up to date? Also, a docs for `ResourceFetcher` itself would need an update.

packages/bare-resource-fetcher/src/ResourceFetcher.ts
rizalibnu: @msluszniak Could you clarify what the actual suggested change was?

The suggested change `const node.next =` is invalid syntax. You cannot use `const` with property assignments. `const` is only for variable declarations.

msluszniak: I don't know what I had here in mind, ignore it 😅

NorbertKlockiewicz:
Hi, I worked a bit with this PR by building bare react native app. From what I've tested I didn't get a single issue with the bare resource fetcher and the integration was also very smooth. To sum up my experience with it was good and I don't have any issues. Thank you @rizalibnu for this amazing piece of code :D

packages/expo-adapter/package.json
mkopcins: I think separating versions between main lib and adapters is a better approach, as it allows us to develop and change adapters at their own rate independent of the lib

packages/bare-resource-fetcher/src/ResourceFetcherUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+    return { results, totalLength };
+  }
+
+  export async function createDirectoryIfNoExists() {
+    if (!(await checkFileExists(RNEDirectory))) {
+      try {
+        await RNFS.mkdir(RNEDirectory);
+      } catch (error) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.AccessFailed,
```
chmjkb: This is an ExecuTorch runtime error code, please use FileWriteFailed instead.

msluszniak:
The last thing from my side is to add `API Reference` to this PR generated from typedoc comments & resolve warning / errors that comes from pre-commit hook.

packages/react-native-executorch/src/modules/computer_vision/OCRModule.ts
```diff
@@ -30,12 +30,17 @@ export class OCRModule {
...
-    );
+    try {
+      await this.controller.load(
+        model.detectorSource,
+        model.recognizerSource,
+        model.language,
+        onDownloadProgressCallback
+      );
+    } catch (error) {
+      console.error('Load Failed:', error);
```
msluszniak:
```suggestion
      Logger.error('Load failed:', error);
```

packages/react-native-executorch/src/modules/computer_vision/OCRModule.ts
msluszniak, line 6:
```suggestion
import { Logger } from '../../common/Logger';

```

packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts
```diff
@@ -42,49 +42,54 @@ export class TextToImageModule extends BaseModule {
...
+        encoderPath,
+        unetPath,
+        decoderPath,
+        schedulerConfig.beta_start,
+        schedulerConfig.beta_end,
+        schedulerConfig.num_train_timesteps,
+        schedulerConfig.steps_offset
+      );
+    } catch (error) {
+      console.error('Load Failed:', error);
```
msluszniak:
```suggestion
      Logger.error('Load failed:', error);
```

packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts
msluszniak:
```suggestion
import { RnExecutorchError } from '../../errors/errorUtils';
import { Logger } from '../../common/Logger';
```

msluszniak:
Todo: 
* Remove two `console.errors` and replace them with `Logger.error`
* Fix offending commit be28882 (use accurate model, not classification in every `load`)

packages/react-native-executorch/src/controllers/BaseOCRController.ts
```diff
@@ -71,7 +72,14 @@ export abstract class BaseOCRController {
       this.isReady = true;
       this.isReadyCallback(this.isReady);
     } catch (e) {
-      if (this.errorCallback) {
+      if (
+        e &&
+        typeof e === 'object' &&
+        'code' in e &&
+        e.code == RnExecutorchErrorCode.NotImplemented
```
msluszniak:
```suggestion
        e.code === RnExecutorchErrorCode.NotImplemented
```

docs/docs/01-fundamentals/01-getting-started.md
```diff
@@ -62,7 +62,7 @@ Installation is pretty straightforward, just use your favorite package manager.
   </TabItem>
 </Tabs>
 
-If you're using bare React Native (instead of a managed Expo project), you also need to install Expo Modules because the underlying implementation relies on expo-file-system. Since expo-file-system is an Expo package, bare React Native projects need **Expo Modules** to properly integrate and use it. The link provided (https://docs.expo.dev/bare/installing-expo-modules/) offers guidance on setting up Expo Modules in a bare React Native environment.
+Our library offers support for both bare React Native and Expo projects. We encourage you to use Expo project if possible. If you are planning to migrate from bare React Native to Expo project, the link (https://docs.expo.dev/bare/installing-expo-modules/) offers a guidance on setting up Expo Modules in a bare React Native environment.
```
chmjkb: I think the getting started page should mention the additional installation step you most likely will have to do (our at least point to the loading models section where things are explained in detail)

docs/docs/01-fundamentals/02-loading-models.md
```diff
@@ -4,6 +4,46 @@ title: Loading Models
 
 There are three different methods available for loading model files, depending on their size and location.
 
+## Prerequisites
+
+In our library, you can use two different resource fetching mechanisms. One is implemented using Expo module, the other one uses external file-system library. We encourage you to use implementation utilizing Expo if possible. But sometimes you cannot add Expo module to your project, then you should use bare React Native implementation.
```
chmjkb:
```suggestion
In our library, you can use two different resource fetching mechanisms. One is implemented using Expo FileSystem, the other one uses external library. We encourage you to use implementation utilizing Expo if possible. But sometimes you cannot add Expo module to your project, then you should use bare React Native implementation.
```

chmjkb: Also, the last sentence seems redundant to me

docs/docs/01-fundamentals/02-loading-models.md
```diff
@@ -4,6 +4,46 @@ title: Loading Models
...
+
+To use initialize Expo adapter, please add these libraries:
```
chmjkb:
```suggestion
To use the Expo adapter, please add these libraries:
```

docs/docs/01-fundamentals/02-loading-models.md
```diff
@@ -4,6 +4,46 @@ title: Loading Models
...
+
+```bash
+yarn add @rn-executorch/expo-adapter
```
chmjkb: I thought we agreed on a different name? cc @mkopcins

docs/docs/06-api-reference/react-native-executorch/namespaces/ResourceFetcherUtils/functions/getFilenameFromUri.md
chmjkb, line 1: Should this be here? Is this even exported to the public API?

docs/docs/06-api-reference/react-native-executorch/namespaces/ResourceFetcherUtils/functions/calculateDownloadProgress.md
chmjkb, line 1: Should this be here? Is this even exported to the public API?

docs/docs/06-api-reference/interfaces/ResourceSourceExtended.md
chmjkb, line 1: Should this be here? Is this even exported to the public API?

docs/docs/06-api-reference/react-native-executorch/namespaces/ResourceFetcherUtils/functions/hashObject.md
chmjkb, line 1: Should this be here? Is this even exported to the public API?

docs/docs/06-api-reference/react-native-executorch/namespaces/ResourceFetcherUtils/functions/removeFilePrefix.md
chmjkb, line 1: Should this be here? Is this even exported to the public API?

packages/react-native-executorch/src/controllers/BaseOCRController.ts
```diff
@@ -71,7 +72,14 @@ export abstract class BaseOCRController {
       this.isReady = true;
       this.isReadyCallback(this.isReady);
     } catch (e) {
-      if (this.errorCallback) {
+      if (
+        e &&
+        typeof e === 'object' &&
+        'code' in e &&
+        e.code === RnExecutorchErrorCode.NotImplemented
```
chmjkb: Isn't NotImplemented an error code coming from ET runtime? If it's thrown by us then it likely shouldn't be

packages/react-native-executorch/src/modules/computer_vision/ImageEmbeddingsModule.ts
```diff
@@ -20,17 +21,24 @@ export class ImageEmbeddingsModule extends BaseModule {
...
       );
+
+      if (paths === null || paths.length < 1) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.DownloadInterrupted,
+          'The download has been interrupted. As a result, not every file was downloaded. Please retry the download.'
+        );
+      }
+
+      this.nativeModule = global.loadImageEmbeddings(paths[0] || '');
```
chmjkb: I know that this is done all over the codebase but we shouldn't default to an empty string. We should check for paths[0], and then load.

packages/react-native-executorch/src/modules/computer_vision/ImageSegmentationModule.ts
```diff
@@ -22,17 +23,24 @@ export class ImageSegmentationModule extends BaseModule {
...
       );
+
+      if (paths === null || paths.length < 1) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.DownloadInterrupted,
+          'The download has been interrupted. As a result, not every file was downloaded. Please retry the download.'
+        );
+      }
+
+      this.nativeModule = global.loadImageSegmentation(paths[0] || '');
```
chmjkb: dont default to empty stirng

packages/react-native-executorch/src/modules/computer_vision/ObjectDetectionModule.ts
```diff
@@ -22,17 +23,24 @@ export class ObjectDetectionModule extends BaseModule {
...
       );
+
+      if (paths === null || paths.length < 1) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.DownloadInterrupted,
+          'The download has been interrupted. As a result, not every file was downloaded. Please retry the download.'
+        );
+      }
+
+      this.nativeModule = global.loadObjectDetection(paths[0] || '');
```
chmjkb: dont default to empty string

packages/react-native-executorch/src/modules/computer_vision/StyleTransferModule.ts
```diff
@@ -21,17 +22,24 @@ export class StyleTransferModule extends BaseModule {
...
       );
+
+      if (paths === null || paths.length < 1) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.DownloadInterrupted,
+          'The download has been interrupted. As a result, not every file was downloaded. Please retry the download.'
+        );
+      }
+
+      this.nativeModule = global.loadStyleTransfer(paths[0] || '');
```
chmjkb: dont default to empty string

packages/react-native-executorch/src/modules/general/ExecutorchModule.ts
```diff
@@ -22,17 +23,22 @@ export class ExecutorchModule extends BaseModule {
...
+        onDownloadProgressCallback,
+        modelSource
       );
+      if (paths === null || paths.length < 1) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.DownloadInterrupted,
+          'The download has been interrupted. As a result, not every file was downloaded. Please retry the download.'
+        );
+      }
+      this.nativeModule = global.loadExecutorchModule(paths[0] || '');
```
chmjkb: dont default to empty stirng

packages/react-native-executorch/src/modules/natural_language_processing/VADModule.ts
```diff
@@ -22,17 +23,22 @@ export class VADModule extends BaseModule {
...
+        onDownloadProgressCallback,
+        model.modelSource
       );
+      if (paths === null || paths.length < 1) {
+        throw new RnExecutorchError(
+          RnExecutorchErrorCode.DownloadInterrupted,
+          'The download has been interrupted. As a result, not every file was downloaded. Please retry the download.'
+        );
+      }
+      this.nativeModule = global.loadVAD(paths[0] || '');
```
chmjkb: dont default do empty string

docs/docs/01-fundamentals/02-loading-models.md
msluszniak: This is correct in the code, need to update it here

msluszniak:
```suggestion
yarn add @react-native-executorch/expo-adapter
```

docs/docs/01-fundamentals/02-loading-models.md
```diff
@@ -4,6 +4,46 @@ title: Loading Models
...
+
+initExecutorch({
+  resourceFetcher: ExpoResourceFetcher,
+});
+```
+
+If you cannot use Expo in your project, proceed with the following steps:
+
+```bash
+yarn add @rn-executorch/bare-adapter
```
msluszniak:
```suggestion
yarn add @react-native-executorch/bare-adapter
```

docs/docs/01-fundamentals/02-loading-models.md
```diff
@@ -4,6 +4,46 @@ title: Loading Models
...
+yarn add @dr.pogodin/react-native-fs @kesha-antonov/react-native-background-downloader
+```
+
+and
+
+```typescript
+import { initExecutorch } from 'react-native-executorch';
+import { BareResourceFetcher } from '@rn-executorch/bare-adapter';
```
msluszniak:
```suggestion
import { BareResourceFetcher } from '@react-native-executorch/bare-adapter';
```

docs/docs/01-fundamentals/02-loading-models.md
```diff
@@ -4,6 +4,46 @@ title: Loading Models
 
 There are three different methods available for loading model files, depending on their size and location.
 
+## Prerequisites
+
+In our library, you can use two different resource fetching mechanisms. One is implemented using Expo FileSystem, the other one uses external library. We encourage you to use implementation utilizing Expo if possible. But sometimes you cannot add Expo module to your project, then you should use bare React Native implementation.
```
msluszniak:
```suggestion
In our library, you can use two different resource fetching mechanisms. One is implemented using Expo FileSystem, the other one uses external library. We encourage you to use implementation utilizing Expo if possible.
```

docs/docs/01-fundamentals/01-getting-started.md
msluszniak:
```suggestion
Our library offers support for both bare React Native and Expo projects. Please follow the instructions from [Loading models section](./02-loading-models.md) to make sure you setup your project correctly. We encourage you to use Expo project if possible. If you are planning to migrate from bare React Native to Expo project, the link (https://docs.expo.dev/bare/installing-expo-modules/) offers a guidance on setting up Expo Modules in a bare React Native environment.
```

packages/react-native-executorch/src/controllers/BaseOCRController.ts
msluszniak: This comes from here (I didn't write this code): https://github.com/software-mansion/react-native-executorch/pull/759/changes#diff-6db8988f6b270a4ca2f4a34a53a32ec515d01b9e2986bd1516eb3b3c95a4d432R89

What error code should be used here?

PR #766, opened 2026-01-28
fix: throw on invalid user configs in LLM, OCR, Object detection, distinct crane(bird) and crane(machine) in imagenet labels

chmjkb:
## Description

Due to failing tests, this PR adds throwing errors when invalid configs happen in OCR, LLM, Object detection. 

- [ ] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist


msluszniak:
There is one thing that might be added in this PR:
in classification one of the labels in `kImagenet1kV1Labels` namely `crane` occurs two times which cause that our code does not work correctly. Names for these two should be change to indicate that the first crane is about a bird, and the second one is a machine.

please update it here: https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/common/rnexecutorch/models/classification/Constants.h

because currently the following code:
```cpp
std::unordered_map<std::string_view, float> probs;
for (std::size_t cl = 0; cl < resultVec.size(); ++cl) {
  probs[constants::kImagenet1kV1Labels[cl]] = resultVec[cl];
}
```

generates the code with only 999 classes, not 1000

PR #781, opened 2026-02-03
Migrate to PyTorch tokenizers

benITo47:
## Description

This PR migrates us from tokenisers-cpp to PyTorch tokenisers bundled with executorch 

- [ ] Yes
- [x] No - User faces no changes 

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

This changes need to be tested manually. Try running all our apps that consume tokenizers and see  whether the output is ok. 
By "OK" I mean 
- no unwanted special tokens 
- response that is related to the question (if model start answering about aquariums after being asked about ketchup, then it's not ok) 
- Output has proper punctuation, no missing or added spaces. 

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
@@ -19,8 +19,8 @@ TokenizerModule::TokenizerModule(
   auto status = tokenizer->load(source);
 
   if (status != tokenizers::Error::Ok) {
-    throw RnExecutorchError(RnExecutorchErrorCode::ModuleNotLoaded,
-                            "Unexpected issue when loading tokenizer");
+    throw RnExecutorchError(RnExecutorchErrorCode::TokenizerError,
+                            "Unexpected issue occured while loading tokenizer");
   };
 }
 
--- a/packages/react-native-executorch/common/rnexecutorch/tests/CMakeLists.txt
+++ b/packages/react-native-executorch/common/rnexecutorch/tests/CMakeLists.txt
@@ -21,6 +21,7 @@ set(MONOREPO_ROOT "${PACKAGE_ROOT}/../..")
 set(THIRD_PARTY_DIR "${MONOREPO_ROOT}/third-party")
 set(REACT_NATIVE_DIR "${MONOREPO_ROOT}/node_modules/react-native")
 set(ANDROID_THIRD_PARTY "${PACKAGE_ROOT}/third-party/android/libs/")
+set(TOKENIZERS_DIR "${PACKAGE_ROOT}/third-party/include/executorch/extension/llm/tokenizers/include")
 
 # Add Gtest as a subdirectory
 add_subdirectory(${THIRD_PARTY_DIR}/googletest ${PROJECT_BINARY_DIR}/googletest)
@@ -80,14 +81,8 @@ target_link_libraries(opencv_deps INTERFACE
 )
 target_link_options(opencv_deps INTERFACE -fopenmp -static-openmp)
 
-# Tokenizers (Interface Library)
-set(TOKENIZERS_LIBS_DIR "${ANDROID_THIRD_PARTY}/tokenizers-cpp/${ANDROID_ABI}")
 add_library(tokenizers_deps INTERFACE)
-target_link_libraries(tokenizers_deps INTERFACE
-    ${TOKENIZERS_LIBS_DIR}/libtokenizers_cpp.a
-    ${TOKENIZERS_LIBS_DIR}/libtokenizers_c.a
-    ${TOKENIZERS_LIBS_DIR}/libsentencepiece.a
-)
+target_include_directories(tokenizers_deps INTERFACE "${TOKENIZERS_DIR}")
 
 # Source Definitions
 set(CORE_SOURCES
@@ -110,6 +105,7 @@ add_library(rntests_core STATIC ${CORE_SOURCES})
 
 target_include_directories(rntests_core PUBLIC
     ${RNEXECUTORCH_DIR}/data_processing
+    ${TOKENIZERS_DIR}
     ${RNEXECUTORCH_DIR}
     ${COMMON_DIR}
     ${PACKAGE_ROOT}/third-party/include
--- a/packages/react-native-executorch/common/rnexecutorch/tests/run_tests.sh
+++ b/packages/react-native-executorch/common/rnexecutorch/tests/run_tests.sh
@@ -80,7 +80,7 @@ REQUIRED_LIBS=(
 # Get libc++_shared.so directly from NDK
 MONOREPO_ROOT="$PACKAGE_ROOT/../../.."
 
-LIBFBJNI_PATH="$SCRIPT_DIR/../integration/libs/libfbjni.so"
+LIBFBJNI_PATH="$SCRIPT_DIR/integration/libs/libfbjni.so"
 
 if [ -z "$LIBFBJNI_PATH" ]; then
   echo "Error: libfbjni.so not found."
--- a/packages/react-native-executorch/src/controllers/LLMController.ts
+++ b/packages/react-native-executorch/src/controllers/LLMController.ts
@@ -347,7 +347,7 @@ export class LLMController {
   ): string {
     if (!tokenizerConfig.chat_template) {
       throw new RnExecutorchError(
-        RnExecutorchErrorCode.InvalidConfig,
+        RnExecutorchErrorCode.TokenizerError,
         "Tokenizer config doesn't include chat_template"
       );
     }


packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -1,18 +1,27 @@
 #include "TokenizerModule.h"
 #include "Error.h"
 #include "ErrorCodes.h"
+#include <cstdint>
 #include <executorch/extension/module/module.h>
 #include <filesystem>
+#include <pytorch/tokenizers/error.h>
 #include <rnexecutorch/data_processing/FileUtils.h>
```
msluszniak: Is this header used here if you removed loading file using file utils?

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -23,31 +32,49 @@ void TokenizerModule::ensureTokenizerLoaded(
...
-  return tokenizer->GetVocabSize();
+  return static_cast<size_t>(tokenizer->vocab_size());
 }
 
-std::string TokenizerModule::idToToken(int32_t tokenId) const {
+std::string TokenizerModule::idToToken(uint64_t tokenId) const {
   ensureTokenizerLoaded("idToToken");
-  return tokenizer->IdToToken(tokenId);
+  auto result = tokenizer->id_to_piece(
+      static_cast<uint64_t>(tokenId)); // TODO: Change accepted type to uint64_t
```
msluszniak: Is this resolved?

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -23,31 +32,49 @@ void TokenizerModule::ensureTokenizerLoaded(
...
+  return "";
 }
 
-int32_t TokenizerModule::tokenToId(std::string token) const {
+uint64_t TokenizerModule::tokenToId(std::string token) const {
   ensureTokenizerLoaded("tokenToId");
-  return tokenizer->TokenToId(token);
+  auto result = tokenizer->piece_to_id(token);
+  if (result.ok()) {
+    return static_cast<uint64_t>(result.get()); // TODO: CHANGE RETURN TYPE
```
msluszniak: Same here, is this resolved?

packages/react-native-executorch/common/runner/runner.cpp
```diff
@@ -35,35 +36,39 @@ static constexpr auto kUseSDPAWithKVCache = "use_sdpa_with_kv_cache";
...
-      metadata_({
-          {kEnableDynamicShape, false},
-          {kMaxSeqLen, 128},
-          {kMaxContextLen, 128},
-          {kUseKVCache, true},
-          {kUseSDPAWithKVCache, false},
-      }) {}
+    : config_(config), module_(module),
+      tokenizer_(std::make_unique<tokenizers::HFTokenizer>()),
+      tokenizer_path_(tokenizer_path), metadata_({
```
msluszniak: Order of initialization here is differeng than order of definition in header, fix this. It should be the same as in header.

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -315,6 +346,17 @@ inline jsi::Value getJsiValue(int val, jsi::Runtime &runtime) {
   return {runtime, val};
 }
 
+template <typename T>
+  requires(meta::IsNumeric<T> && !meta::SameAs<T, int> &&
```
chmjkb: https://en.cppreference.com/w/cpp/types/is_floating_point.html

packages/react-native-executorch/common/runner/runner.cpp
```diff
@@ -35,35 +36,39 @@ static constexpr auto kUseSDPAWithKVCache = "use_sdpa_with_kv_cache";
...
 
-  // Load tokenizer.
-  auto blob = rnexecutorch::file_utils::loadBytesFromFile(tokenizer_path_);
-  tokenizer_ = tokenizers::Tokenizer::FromBlobJSON(blob);
+  if (status != tokenizers::Error::Ok) {
+    throw rnexecutorch::RnExecutorchError(
+        rnexecutorch::RnExecutorchErrorCode::ModuleNotLoaded,
+        "Unexpected issue when loading tokenizer");
+  };
+  ET_CHECK_OK_OR_RETURN_ERROR(module_->load_method("forward"));
```
chmjkb, line 57: not sure if this is a correct error code though, modulenotloaded should be thrown when the user tries to run a model that is not yet loaded, which is not the case here

chmjkb (changes requested):
Please fix tests b4 merging 👍🏻

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
chmjkb, line 1: Please remove the Error.get() calls

chmjkb (changes requested):
Please remove the Error.get() calls

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
benITo47: Yes it's used. You can see usage e.g line 20

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
benITo47: Yeap, good catch. Thanks!

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
benITo47: Yeap, good catch. Thanks!

packages/react-native-executorch/common/runner/runner.cpp
benITo47: Yeap, another good catch. Thanks! 🥇

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -315,6 +346,17 @@ inline jsi::Value getJsiValue(int val, jsi::Runtime &runtime) {
...
+           !meta::SameAs<T, bool>)
+inline jsi::Value getJsiValue(T val, jsi::Runtime &runtime) {
+  return jsi::Value(static_cast<double>(val));
+}
+
+inline jsi::Value getJsiValue(bool val, jsi::Runtime &runtime) {
```
mkopcins: I think this should be above the template function

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
msluszniak: I think we talk about two different headers. I mean FileUtils.h

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
msluszniak: There are not equivalent, so author needs to precise if floating point numbers were his intentions ;)

mkopcins: yes, but as it stands now we cast int64 to double, does not seem to be the correct way

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -315,6 +346,17 @@ inline jsi::Value getJsiValue(int val, jsi::Runtime &runtime) {
...
+  return jsi::Value(val);
+}
```
mkopcins, line 365: I think this should be above the template and below

```
inline jsi::Value getJsiValue(int val, jsi::Runtime &runtime) {
  return {runtime, val};
}
```

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -1,18 +1,27 @@
...
-          file_utils::loadBytesFromFile(source))),
-      memorySizeLowerBound(std::filesystem::file_size(source)) {}
+    : tokenizer(std::make_unique<tokenizers::HFTokenizer>()),
+      memorySizeLowerBound(std::filesystem::file_size(source)) {
+
+  auto status = tokenizer->load(source);
+
+  if (status != tokenizers::Error::Ok) {
+    throw RnExecutorchError(RnExecutorchErrorCode::ModuleNotLoaded,
+                            "Unexpected issue when loading tokenizer");
```
mkopcins: Do we have access to more verbose error message here?

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -23,31 +32,49 @@ void TokenizerModule::ensureTokenizerLoaded(
   }
 }
 
-std::vector<int32_t> TokenizerModule::encode(std::string s) const {
+std::vector<uint64_t> TokenizerModule::encode(std::string s) const {
   ensureTokenizerLoaded("encode");
-  return tokenizer->Encode(s);
+  // Two last arguments represent number of bos and eos tokens added to the
+  // encoded string
```
mkopcins: Instead of this comment change the 0s to constexpr variables
```
constexpr numOfAddedBoSTokens = 0;
constexpr numOfAddedEoSTokens = 0;
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
benITo47: Changed to specification for size_t and uint64_t

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -23,31 +32,49 @@ void TokenizerModule::ensureTokenizerLoaded(
...
-std::string TokenizerModule::idToToken(int32_t tokenId) const {
+std::string TokenizerModule::idToToken(uint64_t tokenId) const {
   ensureTokenizerLoaded("idToToken");
-  return tokenizer->IdToToken(tokenId);
+  auto result = tokenizer->id_to_piece(
+      static_cast<uint64_t>(tokenId)); // TODO: Change accepted type to uint64_t
+  if (result.ok()) {
+    return result.get();
+  }
+  return "";
```
mkopcins: So when we cant resolve the token we return empty string and fail silently?

mkopcins: Perhaps we should reject the promise on the js side

chmjkb: Likewise in the tokentoid function

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
```diff
@@ -23,31 +32,49 @@ void TokenizerModule::ensureTokenizerLoaded(
...
 
-int32_t TokenizerModule::tokenToId(std::string token) const {
+uint64_t TokenizerModule::tokenToId(std::string token) const {
   ensureTokenizerLoaded("tokenToId");
-  return tokenizer->TokenToId(token);
+  auto result = tokenizer->piece_to_id(token);
+  if (result.ok()) {
+    return static_cast<uint64_t>(result.get()); // TODO: CHANGE RETURN TYPE
+  }
+  return -1;
```
mkopcins: like above, maybe reject promise on js side

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
mkopcins: @chmjkb wdyt

chmjkb: yeah i guess we should throw here
@benITo47, when is the result not `.ok()`? Do you know any cases? What happens if i use a token thats not in the vocab?

packages/react-native-executorch/common/runner/runner.cpp
mkopcins: this is legacy behaviour, I think we can leave it for now and then change i when we migrate to vlm runner

packages/react-native-executorch/common/runner/runner.cpp
```diff
@@ -188,7 +193,12 @@ Error Runner::generate(const std::string &prompt,
 
   int64_t context_len_left = static_cast<int64_t>(max_context_length) - pos_;
 
-  std::vector<int32_t> prompt_tokens = tokenizer_->Encode(prompt);
+  // Two last arguments represent number of bos and eos tokens added to the
+  // encoded string
+  // If the used tokenizer.json has defined post_processor field,
+  // setting any of those flags to value other than 0 will result in running the
+  // post_processor with 'add_special_token' flag
```
mkopcins: like in the same case before, avoid magic numbers with comments describing them, change it to named variables and then leave comment if necesary

packages/react-native-executorch/common/runner/runner.cpp
chmjkb: Is it? I think the current runner doesnt have such check, and I don't see any other model throwing this error in load / constructor:
```
Error Runner::load() {
  if (is_loaded()) {
    return Error::Ok;
  }

  ET_CHECK_OK_OR_RETURN_ERROR(module_->load_method("forward"));

  // Load tokenizer.
  auto blob = rnexecutorch::file_utils::loadBytesFromFile(tokenizer_path_);
  tokenizer_ = tokenizers::Tokenizer::FromBlobJSON(blob);
```

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
msluszniak: Could you tell the reason for that, aesthetic one?

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -293,6 +305,15 @@ inline jsi::Value getJsiValue(const std::vector<int32_t> &vec,
   return {runtime, array};
 }
 
+inline jsi::Value getJsiValue(const std::vector<uint64_t> &vec,
```
msluszniak, line 308: This one is for a different issue, but maybe it will be valuable to write some generic function similar to `getTypedArrayAsSpan` but in case of translating C++ to JSI? If so, I'll create an issue for that.

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
benITo47: Not really. Tokenizer's error doesn't provide any messages. Best we could do is specified std::to_string that shows enum name instead of error code.

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
benITo47: There are several cases where result is not ok, and it will resolve to Error type: 
1. Tokenizer is not initialised and the function is called - likely won't happen in our case (handled by the ensureTokenierLoaded call) 
2. Tokenizer has corrupted token maps - with no values present, this might happen if the passed .json is malformed. 
3. Token is not in the vocab - might be due to .json but might not.

packages/react-native-executorch/common/rnexecutorch/TokenizerModule.cpp
benITo47: Agreed. But the comment will stay as inner workings of PyTorch tokenizers are sometimes surprising

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
benITo47: +1 for msluszniak. 
Could you elaborate?

packages/react-native-executorch/common/runner/runner.cpp
mkopcins: Right, I was looking at wrong line, thought you meant `ET_CHECK_OK_OR_RETURN_ERROR`, we should return error here, not throw one

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
mkopcins: aesthetic one

msluszniak:
@chmjkb I guess you wanted to approve, not add another review ;)

Issue #783, opened 2026-02-04
Problems with Tool Calling


msluszniak:
### Description

When prompted sequentially about creating a meeting in calendar, two times I got empty response and then strange error:

```
[React] { [Error: [object Object]] code: 1, cause: undefined }
 ERROR  [Error: [object Object]] 

Code: errorUtils.ts
   6 |
   7 |   constructor(code: number, message: string, cause?: unknown) {
>  8 |     super(message);
     |     ^
   9 |     this.code = code;
  10 |     this.message = message;
  11 |     this.cause = cause;
Call Stack
  RnExecutorchError#constructor (packages/react-native-executorch/src/errors/errorUtils.ts:8:5)
  parseUnknownError (packages/react-native-executorch/src/errors/errorUtils.ts:28:31)
  forward (packages/react-native-executorch/src/controllers/LLMController.ts:232:30)
```

Reference:

<img width="3342" height="1686" alt="Image" src="https://github.com/user-attachments/assets/5ce27945-af22-4dc0-a12f-b44cfefceec2" />

### Steps to reproduce

1. Open llm > Tool Calling app in demo apps.
2. Ask about creating a meeting as in the reference image above.
3. Observe strange behaviour.

### Snack or a link to a repository

demo apps

### React Native Executorch version

main

### React Native version

81.5

### Platforms

Platform agnostic

### JavaScript runtime

Hermes

### Workflow

Expo Go

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### AI model

Tool calling model from llm demo app

### Acknowledgements

Yes

msluszniak:
Update, after applying fix for creating correctly error I get:

```
[React] { [Error: Failed to generate text] code: 18, cause: undefined }
 ERROR  [Error: Failed to generate text] 

Code: errorUtils.ts
  19 |
  20 |   constructor(code: number, message: string, cause?: unknown) {
> 21 |     super(message);
     |     ^
  22 |     this.code = code;
  23 |     this.message = message;
  24 |     this.cause = cause;
Call Stack
  RnExecutorchError#constructor (packages/react-native-executorch/src/errors/errorUtils.ts:21:5)
  parseUnknownError (packages/react-native-executorch/src/errors/errorUtils.ts:30:33)
  forward (packages/react-native-executorch/src/controllers/LLMController.ts:232:30)
```

[...]

PR #784, opened 2026-02-04
Fix creating error out of object

msluszniak:
## Description

Partially fixes problems of #783 with uninformative error

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

Run code as in #783 described. Then apply fix and check if the error message is now created correctly.

### Checklist

### DIFF
--- a/packages/react-native-executorch/src/errors/errorUtils.ts
+++ b/packages/react-native-executorch/src/errors/errorUtils.ts
@@ -1,5 +1,18 @@
 import { RnExecutorchErrorCode } from './ErrorCodes';
 
+function isRnExecutorchError(
+  e: unknown
+): e is { code: number; message: string } {
+  return (
+    typeof e === 'object' &&
+    e !== null &&
+    'code' in e &&
+    'message' in e &&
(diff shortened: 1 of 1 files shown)


packages/react-native-executorch/src/errors/errorUtils.ts
```diff
@@ -13,6 +13,18 @@ export class RnExecutorchError extends Error {
...
+    e !== null &&
+    'code' in e &&
+    'message' in e &&
+    typeof (e as any).code === 'number' &&
+    typeof (e as any).message === 'string'
+  ) {
+    const raw = e as { code: number; message: string };
+    return new RnExecutorchError(raw.code, raw.message);
+  }
+
```
chmjkb: We can use type guards here instead, i.e.
```typescript
function isRnExecutorchError(e: unknown): e is RawError {
  return (
    typeof e === 'object' &&
    e !== null &&
    'code' in e &&
    'message' in e &&
    typeof (e as RawError).code === 'number' &&
    typeof (e as RawError).message === 'string'
  );
}
```

PR #814, opened 2026-02-17
feat!: make image segmentation generic, general refactor

chmjkb:
## Description

Refactors image segmentation into a generic, multi-model architecture. Previously the module was hardcoded to DeepLab V3 — now it supports multiple built-in models (DeepLab V3, selfie segmentation, RF-DETR) and custom user-provided models with type-safe label maps.

**Key changes:**

- **C++ base class**: Extracted `BaseImageSegmentation` with virtual `preprocess()`/`postprocess()` methods. `ImageSegmentation` is now a thin subclass. This allows future models to override preprocessing (e.g. different normalization) or postprocessing without duplicating the pipeline.
- **Optional normalization in C++**: `readImageToTensor` now accepts optional `normMean`/`normStd` params, eliminating duplicated normalization logic. **also, imo it would be a good idea to do such factories for the entire API**
- **Generic TypeScript module**: `ImageSegmentationModule<T>` is generic over model name or custom `LabelEnum`. Two static factories: `fromModelName()` (built-in models with auto label resolution) and `fromCustomConfig()` (custom models with user-provided labels).
- **Generic hook**: `useImageSegmentation` infers the model's label types from the config — no explicit generic parameter needed. `forward()` return type narrows based on `classesOfInterest` passed in.
- **Correct return types**: `forward()` now returns `Record<'ARGMAX', Int32Array> & Record<K, Float32Array>` matching what the native side actually produces (was incorrectly typed as `number[]`).
- **ARGMAX always returned**: Removed `'ARGMAX'` from `classesOfInterest` — it's always in the output regardless, and the return type reflects this.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. Build and run the `computer-vision` demo app
2. Navigate to Image Segmentation screen
3. Pick an image and run segmentation — verify the ARGMAX overlay renders correctly
4. Verify the hook API works as expected:
```ts
const { isReady, forward } = useImageSegmentation({
  model: { modelName: 'deeplab-v3', modelSource: DEEPLAB_V3_RESNET50 },
});

// Returns Record<'ARGMAX', Int32Array> — no generic needed
const result = await forward(imageUri);

// Narrows return type to include 'PERSON' key as Float32Array
const result2 = await forward(imageUri, ['PERSON']);
```
5. Verify TypeScript autocompletion: `classesOfInterest` should only suggest valid label keys for the chosen model (e.g. `'PERSON'`, `'CAR'` for DeepLab, `'SELFIE'`/`'BACKGROUND'` for selfie segmentation)
6. You can also try changing the parameters, to say selfie segmentation and see how the return types react. Please contact me for weights for selfie segmentation as I'm not pushing them to HF yet 

### Checklist

### Additional notes

The `ImageSegmentationModule.fromCustomConfig()` API allows users to bring their own segmentation model with a custom label map:
```ts
const MyLabels = { BACKGROUND: 0, FOREGROUND: 1 } as const;
const seg = await ImageSegmentationModule.fromCustomConfig(
  'https://example.com/model.pte',
  { labelMap: MyLabels },
);
```

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.h
+++ b/packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.h
@@ -19,7 +19,6 @@ using executorch::extension::TensorPtr;
 
 class BaseImageSegmentation : public BaseModel {
 public:
-  ~BaseImageSegmentation() override = default;
   BaseImageSegmentation(const std::string &modelSource,
                         std::shared_ptr<react::CallInvoker> callInvoker);
 
--- a/packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
+++ b/packages/react-native-executorch/common/rnexecutorch/models/BaseModel.h
@@ -20,6 +20,8 @@ using executorch::runtime::Result;
 class BaseModel {
 public:
   virtual ~BaseModel() = default;
+  BaseModel(BaseModel &&) = default;
+  BaseModel &operator=(BaseModel &&) = default;
   BaseModel(
       const std::string &modelSource,
       std::shared_ptr<react::CallInvoker> callInvoker,


packages/react-native-executorch/src/modules/computer_vision/ImageSegmentationModule.ts
```diff
@@ -1,74 +1,204 @@
...
+    preprocessorConfig: undefined,
+  },
+  'selfie-segmentation': {
+    labelMap: SelfieSegmentationLabel,
+    preprocessorConfig: undefined,
+  },
+  'rfdetr': {
+    labelMap: CocoLabel,
+    preprocessorConfig: { normMean: IMAGENET_MEAN, normStd: IMAGENET_STD },
+  },
```
chmjkb: fyi this doesnt work as the rfdetr is an instance segmentation model and needs a separate implementation, but leaving this so you can get an idea on what the configuration looks like

msluszniak:
Please suppress these warnings:

image

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.cpp
```diff
@@ -0,0 +1,216 @@
...
+    normMean_ = cv::Scalar(normMean[0], normMean[1], normMean[2]);
+  }
+  if (normStd.size() >= 3) {
+    normStd_ = cv::Scalar(normStd[0], normStd[1], normStd[2]);
+  }
+}
+
+void BaseImageSegmentation::initModelImageSize() {
+  auto inputShapes = getAllInputShapes();
+  if (inputShapes.size() == 0) {
```
msluszniak: use empty instead

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.cpp
```diff
@@ -0,0 +1,216 @@
...
+  return postprocess(forwardResult->at(0).toTensor(), originalSize, allClasses,
+                     classesOfInterest, resize);
+}
+
+std::shared_ptr<jsi::Object> BaseImageSegmentation::postprocess(
+    const Tensor &tensor, cv::Size originalSize,
+    std::vector<std::string> &allClasses,
+    std::set<std::string, std::less<>> &classesOfInterest, bool resize) {
+
+  auto dataPtr = static_cast<const float *>(tensor.const_data_ptr());
```
msluszniak:
```suggestion
  const auto* dataPtr = (tensor.const_data_ptr<float>();
```

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.cpp
```diff
@@ -0,0 +1,216 @@
...
+      for (std::size_t cl = 0; cl < numChannels; ++cl) {
+        classBuffers[cl][pixel] = values[cl];
+        if (values[cl] > maxVal) {
+          maxVal = values[cl];
+          maxInd = static_cast<int>(cl);
+        }
+      }
+      argmaxData[pixel] = maxInd;
+    }
+  }
```
msluszniak: go over classes and then over pixels, not in this loop order. Because you are now jumping between classes which are not next to each other in memory, you have slow accesses. Instead use something like that:

```cpp
std::vector<float> maxLogits(outputPixels, -std::numeric_limits<float>::infinity());
std::vector<float> sumExp(outputPixels, 0.0f);

for (std::size_t cl = 0; cl < numChannels; ++cl) {
    for (std::size_t pixel = 0; pixel < outputPixels; ++pixel) {
        float val = classBuffers[cl][pixel];
        
        if (val > maxLogits[pixel]) {
            maxLogits[pixel] = val;
            argmaxData[pixel] = static_cast<int32_t>(cl);
        }
    }
}

for (std::size_t cl = 0; cl < numChannels; ++cl) {
    for (std::size_t pixel = 0; pixel < outputPixels; ++pixel) {
        float val = std::exp(classBuffers[cl][pixel] - maxLogits[pixel]);
        
        classBuffers[cl][pixel] = val; 
        sumExp[pixel] += val;
    }
}

for (std::size_t cl = 0; cl < numChannels; ++cl) {
    for (std::size_t pixel = 0; pixel < outputPixels; ++pixel) {
        classBuffers[cl][pixel] /= sumExp[pixel];
    }
}
```

Even though you uses 3 passes over data, this should be faster due to linear accesses to data. You can benchmark this part of code.

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.cpp
```diff
@@ -0,0 +1,216 @@
...
+  auto dataPtr = static_cast<const float *>(tensor.const_data_ptr());
+  auto resultData = std::span(dataPtr, tensor.numel());
+
+  // Read output dimensions directly from tensor shape
+  std::size_t numChannels =
+      (tensor.dim() >= 3) ? tensor.size(tensor.dim() - 3) : 1;
+  std::size_t outputH = tensor.size(tensor.dim() - 2);
+  std::size_t outputW = tensor.size(tensor.dim() - 1);
+  std::size_t outputPixels = outputH * outputW;
+  cv::Size outputSize(static_cast<int>(outputW), static_cast<int>(outputH));
```
msluszniak: Maybe int32_t for convention

msluszniak:
Also please fill out the PR template and write how to test this PR

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.cpp
chmjkb: I guess it would make sense due to cache locality, but I benchmarked it and it looks almost identical in terms of performance, but I think this is a tiny bit cleaner so I'll change it

msluszniak:
In a meantime, you can resolve conflicts and I'll test the PR ;)

docs/docs/03-hooks/02-computer-vision/useImageSegmentation.md
```diff
@@ -47,45 +54,70 @@ You need more details? Check the following resources:
...
-- The [`resizeToInput`](../../06-api-reference/interfaces/ImageSegmentationType.md#resizeToInput) flag specifies whether the output will be rescaled back to the size of the input image. The default is `true`. The model runs inference on a scaled (probably smaller) version of your image (224x224 for `DEEPLAB_V3_RESNET50`). If you choose to resize, the output will be `number[]` of size `width * height` of your original image.
+- `imageSource` (required) - The image to segment. Can be a remote URL, a local file URI, or a base64-encoded image.
+- `classesOfInterest` (optional) - An array of label keys indicating which per-class probability masks to include in the output. Defaults to `[]` (no class masks). The `ARGMAX` map is always returned regardless of this parameter.
+- `resizeToInput` (optional) - Whether to resize the output masks to the original input image dimensions. Defaults to `true`. If `false`, returns the raw model output dimensions (e.g. 224x224 for `DEEPLAB_V3_RESNET50`).
```
msluszniak: Why we removed redirecting links from there?

docs/docs/03-hooks/02-computer-vision/useImageSegmentation.md
```diff
@@ -47,45 +54,70 @@ You need more details? Check the following resources:
...
 
 :::warning
-Setting `resize` to true will make `forward` slower.
+Setting `resizeToInput` to `true` will make `forward` slower.
```
msluszniak: I think that the current version on `main` is better: `"Setting resizeToInput to false will make forward faster."` since `true` is default, so you actively need to change `resizeToInput` to false to gain the speed.

docs/docs/03-hooks/02-computer-vision/useImageSegmentation.md
```diff
@@ -47,45 +54,70 @@ You need more details? Check the following resources:
...
+`forward` returns a promise resolving to an object containing:
+
+- `ARGMAX` - An `Int32Array` where each element is the class index with the highest probability for that pixel.
+- For each label included in `classesOfInterest`, a `Float32Array` of per-pixel probabilities for that class.
 
-- For the key [`DeeplabLabel.ARGMAX`](../../06-api-reference/enumerations/DeeplabLabel.md#argmax) the array contains for each pixel an integer corresponding to the class with the highest probability.
-- For every other key from [`DeeplabLabel`](../../06-api-reference/enumerations/DeeplabLabel.md), if the label was included in [`classesOfInterest`](../../06-api-reference/interfaces/ImageSegmentationType.md#classesofinterest) the dictionary will contain an array of floats corresponding to the probability of this class for every pixel.
+The return type is fully typed — TypeScript narrows it based on the labels you pass in `classesOfInterest`.
```
msluszniak, line 82: I will keep links if applicable

docs/docs/04-typescript-api/02-computer-vision/ImageSegmentationModule.md
```diff
@@ -35,34 +36,75 @@ All methods of `ImageSegmentationModule` are explained in details here: [`ImageS
...
-- The [`resizeToInput`](../../06-api-reference/classes/ImageSegmentationModule.md#resizeToInput) flag specifies whether the output will be rescaled back to the size of the input image. The default is `true`. The model runs inference on a scaled (probably smaller) version of your image (224x224 for the `DEEPLAB_V3_RESNET50`). If you choose to resize, the output will be `number[]` of size `width * height` of your original image.
+- `imageSource` (required) - The image to segment. Can be a remote URL, a local file URI, or a base64-encoded image.
+- `classesOfInterest` (optional) - An array of label keys indicating which per-class probability masks to include in the output. Defaults to `[]`. The `ARGMAX` map is always returned regardless.
+- `resizeToInput` (optional) - Whether to resize the output masks to the original input image dimensions. Defaults to `true`. If `false`, returns the raw model output dimensions.
 
 :::warning
-Setting `resize` to true will make `forward` slower.
+Setting `resizeToInput` to `true` will make `forward` slower.
```
msluszniak: Same as above

docs/docs/03-hooks/02-computer-vision/useImageSegmentation.md
chmjkb: ARGMAX is no longer a key of Deeplab label, so the link is removed. Also, it is not just the DeeplabLabel now, but other labels too.

msluszniak: Ok

msluszniak (approved):

msluszniak:
There are some conflicts after PR with adapters.

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/ImageSegmentation.cpp
```diff
@@ -1,170 +1 @@
 #include "ImageSegmentation.h"
```
mkopcins: remove file

packages/react-native-executorch/src/constants/commonVision.ts
```diff
@@ -0,0 +1,4 @@
+import { Triple } from '../types/common';
+
+export const IMAGENET_MEAN: Triple<number> = [0.485, 0.456, 0.406];
+export const IMAGENET_STD: Triple<number> = [0.229, 0.224, 0.225];
```
mkopcins: they are not the same for all imagenet variants
```suggestion
export const IMAGENET1K_MEAN: Triple<number> = [0.485, 0.456, 0.406];
export const IMAGENET1K_STD: Triple<number> = [0.229, 0.224, 0.225];
```

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
```
mkopcins, line 1: We are missing virtual destructor in model classes

packages/react-native-executorch/src/hooks/computer_vision/useImageSegmentation.ts
```diff
@@ -1,23 +1,109 @@
...
+  const [isGenerating, setIsGenerating] = useState(false);
+  const [downloadProgress, setDownloadProgress] = useState(0);
+  const [instance, setInstance] = useState<ImageSegmentationModule<
+    ModelNameOf<C>
+  > | null>(null);
+
+  useEffect(() => {
+    if (preventLoad) return;
+
+    let currentInstance: ImageSegmentationModule<ModelNameOf<C>> | null = null;
```
mkopcins, line 49: add `isMounted` guard here (reference [here](https://github.com/software-mansion/react-native-executorch/blob/2d584100e49a94164dbeeed76ce0f9bf2b4bcafb/packages/react-native-executorch/src/hooks/natural_language_processing/useSpeechToText.ts#L32-L54))

packages/react-native-executorch/src/modules/computer_vision/ImageSegmentationModule.ts
```diff
@@ -1,74 +1,201 @@
...
+   * );
+   * ```
+   */
+  static async fromCustomConfig<L extends LabelEnum>(
+    modelSource: ResourceSource,
+    config: SegmentationConfig<L>,
+    onDownloadProgress: (progress: number) => void = () => {}
+  ): Promise<ImageSegmentationModule<L>> {
+    const paths = await ResourceFetcher.fetch(onDownloadProgress, modelSource);
+    if (!paths?.[0]) {
```
mkopcins, line 148: I guess we should be also checking for `paths.length == =1`

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/BaseImageSegmentation.cpp
```diff
@@ -0,0 +1,231 @@
...
+    const std::string &modelSource, std::vector<float> normMean,
+    std::vector<float> normStd, std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker) {
+  initModelImageSize();
+  if (normMean.size() >= 3) {
+    normMean_ = cv::Scalar(normMean[0], normMean[1], normMean[2]);
+  }
+  if (normStd.size() >= 3) {
+    normStd_ = cv::Scalar(normStd[0], normStd[1], normStd[2]);
+  }
```
benITo47: What if norm vector has less than 3  elements? Maybe we should log here?

packages/react-native-executorch/src/constants/commonVision.ts
msluszniak: Make sure you change it across all the code

packages/react-native-executorch/common/rnexecutorch/models/image_segmentation/ImageSegmentation.cpp
benITo47: Do we need a .cpp file that just includes header?

packages/react-native-executorch/src/modules/computer_vision/ImageSegmentationModule.ts
msluszniak: If so, we need to change this part in the rest of the models as well.

packages/react-native-executorch/common/rnexecutorch/models/object_detection/ObjectDetection.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <rnexecutorch/Error.h>
 #include <rnexecutorch/ErrorCodes.h>
+#include <rnexecutorch/Log.h>
```
msluszniak: Why you added this header here?

msluszniak (changes requested):
There are some errors in compilation of C++ code that spread across multiple models, not only segmentation. So I believe there are some errors maybe in `BaseModel` or similar location

Issue #776, opened 2026-02-03
LLM message history management not working + exceeding context window ([Error: Failed to generate text, error code: 18])


mateuszlampert:
### Problem description

Right now, the model from `useLLM` hook does not manage context window automatically and it results in errors when it is exceeded. Ideally, the model should remove the oldest tokens/messages automatically so the app does not crash. The problem occurs in both functional and managed approach.

After exceeding the window size,  an error is thrown: `[Error: Failed to generate text, error code: 18]`.

[Discord thread](https://discord.com/channels/464786597288738816/1466201459547701353/1468007304569028836)

### Steps to reproduce
- run few longer prompts (or many shorter prompts) using `.generate()` or `.sendMessage()` method

### What should be done

- in the managed approach, the oldest messages/tokens should be removed automatically if the context were to be exceeded
- in the functional approach, the state of messages should not be stored at all, the history of messages should be reset after each generation
- we should probably throw some more descriptive error

### Benefits to React Native ExecuTorch
- users won't have to manage the messages history by their own

PR #819, opened 2026-02-19
fix/feat!: LLMs context management

mateuszlampert:
## Description

This PR fixes few bugs related to the LLMs, caused by mixing two approaches - functional (as we pass whole messages history each time) and stateful (as we keep `pos_` in the runner, representing at which position the KV cache is), which resulted in 3 bugs:

- broken KV cache for reasoning models - in the runner, we counted tokens generated for the reasoning and included these in KV cache (`pos_ += num_generated_tokens`), but in next turns, `jinja template` removed these reasoning tokens from the messages history - as a result, KV-cache was incoherent
- duplicated tokens in KV cache - we were passing whole messages history to the runner (functional approach), but we were also appending all tokens (prompt and generated) to the KV cache (which position is represented by `pos_`) - as a result tokens were "duplicated" in the KV cache and we were running out of available tokens very fast (exceeding `context_window_length`)
-  stateful TS functional API - even though our `generate()` method is called functional, it kept internal state in the runner (e. g. `pos_`)

These bugs were fixed by resetting the runner before each generation, which makes it truly functional - old messages are prefilled and the KV cache can be still used during generation phase.

Additionally, this PR adds `ContextStrategy` to `ChatConfig` interface, so now it's possible to define (or use one of already implemented) strategy for managing context (e. g. naive, message count based, sliding window) - it gives us more flexibility and user can decide what's best for their use case. From now on, `SlidingWindowContextStrategy` is also configured as the default one.

- [x] Yes
- [ ] No

These changes will not break anything until max number of messages is not modified (I removed `contextWindowLength` from `ChatConfig` and replaced it with `contextStrategy`)

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Run example llm app, open executorch logs (`adb logcat | grep -i "executorch"` for example) and see if numbers of tokens are properly aligned and if `pos_` is correct.

To test different context management strategies, change `contextStrategy` in llm app and modify model configuration.

### Related issues

#776

### Checklist

### Additional notes

Position in KV cache, number of prompt tokens and number of generated tokens for both non-reasoning and reasoning models BEFORE changes.

LLAMA 3.2 1B SPINQUANT (without reasoning)

| pos_             | Prompt tokens | Generated tokens |
|------------------|---------------|------------------|
| 0                | 335           | 269              |
| 604=269+335      | 872           | 372              |
| 1848=604+872+372 | 1513          | CRASH            |

QWEN 3.0 0.6B QUANTIZED (with reasoning)
| pos_             | Prompt tokens | Generated tokens |
|------------------|---------------|------------------|
| 0                | 309           | 457              |
| 766=309+457      | 617 (<766!)   | 192              |
| 1575=766+617+192 | 925 (<1575!)  | CRASH            |


mateuszlampert:
I will update docs after the code gets approved

packages/react-native-executorch/common/runner/runner.cpp
```diff
@@ -368,4 +368,17 @@ int32_t Runner::resolve_max_new_tokens(int32_t num_prompt_tokens,
...
+
+  if (!encodeResult.ok()) {
+    throw rnexecutorch::RnExecutorchError(
+        rnexecutorch::RnExecutorchErrorCode::TokenizerError,
+        "Encoding failed during token count check.");
+  }
+
+  return static_cast<int32_t>(encodeResult.get().size());
+}
+
```
msluszniak: I'm wondering if calling encoder specifically to get only the size of encoded text is the most efficient way to get the size. Can we compute this during encoding phase?

mateuszlampert: If I understand this correctly that's not possible for 2 reasons:
- I need to calculate all tokens in the whole messages history  **+ new message** (so I don't know number of tokens for new message)
- I can't really use token counts from the runner reliably, because it counts reasoning tokens as well for reasoning models (and later on, this reasoning tokens are not included in jinja template) - it could lead to some discrepancies

mateuszlampert: Alternatively, maybe we could implement these context strategies in the c++ code, but as a result we give no flexibility to the user (?)

Also:
- the context management strategy is configurable, so it can be switched on/off oraz replaced with other strategy - by default, we do not use sliding window, but the strategy based on number of messages (like it was before)
- I think that encoding phase should not take that long

mkopcins: We don't need to bother ourselves with computational complexity here, this is just tokenizer encoding which is very cheap and only performed once for the given prompt and so only impacts Time To First Token

packages/react-native-executorch/common/runner/runner.cpp
```diff
@@ -368,4 +368,17 @@ int32_t Runner::resolve_max_new_tokens(int32_t num_prompt_tokens,
...
+  auto encodeResult =
+      tokenizer_->encode(text, numOfAddedBoSTokens, numOfAddedEoSTokens);
+
+  if (!encodeResult.ok()) {
+    throw rnexecutorch::RnExecutorchError(
+        rnexecutorch::RnExecutorchErrorCode::TokenizerError,
+        "Encoding failed during token count check.");
+  }
+
+  return static_cast<int32_t>(encodeResult.get().size());
```
msluszniak: Why you explicitly cast here to `int32_t` to use it as a `size_t` and implicitly cast back to `size_t` in `countTextTokens`?

mateuszlampert: I unified this to `size_t`

mateuszlampert:
I changed only docs for `useLLM` and `LLMModule` - the rest of changes is related to autogenerated api reference changes

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.h
```diff
@@ -21,14 +21,17 @@ class LLM : public BaseModel {
   std::string generate(std::string input,
                        std::shared_ptr<jsi::Function> callback);
   void interrupt();
+  void reset();
   void unload() noexcept;
   size_t getGeneratedTokenCount() const noexcept;
   size_t getPromptTokenCount() const noexcept;
+  size_t countTextTokens(std::string text) const;
```
msluszniak: This is passed by copy because otherwise JSI fails, right?

mateuszlampert: yes

msluszniak:
I run the demo llm app on my android device I got:
``` 
ERROR  [TypeError: Cannot convert 5 to BigInt] 

Code: SlidingWindowContextStrategy.ts
  39 |   ): Message[] {
  40 |     let localHistory = [...history];
> 41 |     const tokenBudget = maxContextLength - this.bufferTokens;
     |                         ^
  42 |
  43 |     while (localHistory.length > 1) {
  44 |       const candidateContext: Message[] = [
Call Stack
  buildContext (packages/react-native-executorch/src/utils/llms/context_strategy/SlidingWindowContextStrategy.ts:41:25)
  sendMessage (packages/react-native-executorch/src/controllers/LLMController.ts:326:51)
  sendMessage (packages/react-native-executorch/src/controllers/LLMController.ts:307:27)
  sendMessage (packages/react-native-executorch/src/hooks/natural_language_processing/useLLM.ts:97:44)
  sendMessage (apps/llm/app/llm/index.tsx:49:28)
  LLMScreen (apps/llm/app/llm/index.tsx:45:20)
  TouchableOpacity.props.onPress (apps/llm/app/llm/index.tsx:109:77)
  ```

File: packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp [hunk 4]
```diff
@@ -64,6 +64,14 @@ void LLM::interrupt() {
   runner->stop();
 }
 
+void LLM::reset() {
+  if (!runner || !runner->is_loaded()) {
+    throw RnExecutorchError(RnExecutorchErrorCode::ModuleNotLoaded,
+                            "Can't interrupt a model that's not loaded");
```
mkopcins:
```suggestion
                            "Can't reset a model that's not loaded");
```

packages/react-native-executorch/src/utils/llms/context_strategy/NaiveContextStrategy.ts
```diff
@@ -0,0 +1,28 @@
+import { ContextStrategy, Message } from '../../../types/llm';
+
+/**
+ * A context strategy that performs no filtering or trimming of the message history.
+ * * This strategy is ideal when the developer wants to manually manage the conversation
+ * context.
+ *
+ * @category Utils
+ */
+export class NaiveContextStrategy implements ContextStrategy {
```
mkopcins: imo sth like NoOpContextStrategy is better, Naive suggest that there is some simple logic underneath

PR #880, opened 2026-02-25
feat!: integrate all vision models with vision camera

NorbertKlockiewicz:
## Description

- Added `PixelData` input support to all vision model hooks and modules — `forward()` now accepts both string URIs and raw pixel buffers
- Added `runOnFrame` worklet to all vision model hooks for real-time VisionCamera v5 frame processing (`useClassification`, `useImageEmbeddings`, `useOCR`, `useVerticalOCR`, `useObjectDetection`, `useSemanticSegmentation`, `useStyleTransfer`)
- Refactored `StyleTransfer` C++ to unified `generateFromString` / `generateFromPixels` with a `saveToFile` flag
- Added new VisionCamera integration docs page with setup guide, full example, Module API pattern, and common issues section
- Updated all vision model hook and module docs to document `PixelData` input and `runOnFrame`
- Added integration tests: `VisionModelTest` (unit tests for `extractFromPixels` and `preprocess`), `FrameProcessorTest` (unit tests for `pixelsToMat`), `StyleTransferTest` (full coverage of both output modes), and `generateFromPixels` smoke tests across all vision models

#### `StyleTransferType.forward` return type changed

Before:
```typescript
  forward(imageSource: string): Promise<string>
```

After:
```typescript
  forward(input: string | PixelData, output?: 'pixelData' | 'url'): Promise<PixelData | string>
```

### Known Issues

The orientation handling is static right now, it will be fixed in the next PR.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [x] Run computer vision app -> vision camera screen -> test all models
- [ ] Also test rest of the models in this app.
- [ ] Run test suite

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
+++ b/packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
@@ -575,7 +575,7 @@ getJsiValue(const models::style_transfer::PixelDataResult &result,
   auto sizesArray = jsi::Array(runtime, 3);
   sizesArray.setValueAtIndex(runtime, 0, jsi::Value(result.height));
   sizesArray.setValueAtIndex(runtime, 1, jsi::Value(result.width));
-  sizesArray.setValueAtIndex(runtime, 2, jsi::Value(4));
+  sizesArray.setValueAtIndex(runtime, 2, jsi::Value(result.channels));
   obj.setProperty(runtime, "sizes", sizesArray);
 
   obj.setProperty(runtime, "scalarType",
--- a/packages/react-native-executorch/common/rnexecutorch/models/style_transfer/StyleTransfer.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/style_transfer/StyleTransfer.cpp
@@ -61,7 +61,7 @@ PixelDataResult toPixelDataResult(const cv::Mat &bgrMat) {
   cv::cvtColor(bgrMat, rgba, cv::COLOR_BGR2RGBA);
   std::size_t dataSize = static_cast<std::size_t>(size.width) * size.height * 4;
   auto pixelBuffer = std::make_shared<OwningArrayBuffer>(rgba.data, dataSize);
-  return PixelDataResult{pixelBuffer, size.width, size.height};
+  return PixelDataResult{pixelBuffer, size.width, size.height, rgba.channels()};
 }
 
 StyleTransferResult StyleTransfer::generateFromString(std::string imageSource,
--- a/packages/react-native-executorch/common/rnexecutorch/models/style_transfer/Types.h
+++ b/packages/react-native-executorch/common/rnexecutorch/models/style_transfer/Types.h
@@ -11,6 +11,7 @@ struct PixelDataResult {
   std::shared_ptr<OwningArrayBuffer> dataPtr;
   int width;
   int height;
+  int channels;
 };
 
 using StyleTransferResult = std::variant<PixelDataResult, std::string>;


barhanc:
On Android the output for front camera is upside down and mirrored. Is this an expected behaviour that will be fixed with orientation handling?

NorbertKlockiewicz:
> On Android the output for front camera is upside down and mirrored. Is this an expected behaviour that will be fixed with orientation handling?

yes, I am working on it now but this PR already has too much changes so it will be added in the next one.

apps/computer-vision/components/vision_camera/tasks/ClassificationTask.tsx
```diff
@@ -0,0 +1,120 @@
...
+              const [label, score] = entries[i]!;
+              if ((score as number) > bestScore) {
+                bestScore = score as number;
+                bestLabel = label;
+              }
+            }
+            scheduleOnRN(updateClass, { label: bestLabel, score: bestScore });
+          }
+        } catch {
+          // ignore
```
msluszniak, line 74: Why we are ignoring errors here?

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
```diff
@@ -1,40 +1,43 @@
 #include "ImageEmbeddings.h"
 
+#include <span>
```
msluszniak: Where is span used here?

packages/react-native-executorch/common/rnexecutorch/models/semantic_segmentation/BaseSemanticSegmentation.cpp
```diff
@@ -37,46 +36,71 @@ void BaseSemanticSegmentation::initModelImageSize() {
...
-  return postprocess(forwardResult->at(0).toTensor(), originalSize, allClasses_,
-                     classesOfInterest, resize);
+  return computeResult(forwardResult->at(0).toTensor(), originalSize,
+                       allClasses_, classesOfInterest, resize);
+}
+
+semantic_segmentation::SegmentationResult
+BaseSemanticSegmentation::generateFromString(
+    std::string imageSource,
+    std::set<std::string, std::less<>> classesOfInterest, bool resize) {
```
msluszniak, line 77: isn't `std::less` default comparator in `std::set`?

packages/react-native-executorch/common/rnexecutorch/models/semantic_segmentation/BaseSemanticSegmentation.cpp
```diff
@@ -161,8 +185,8 @@ std::shared_ptr<jsi::Object> BaseSemanticSegmentation::postprocess(
   }
 
   // Filter classes of interest
-  auto buffersToReturn = std::make_shared<std::unordered_map<
-      std::string_view, std::shared_ptr<OwningArrayBuffer>>>();
+  auto buffersToReturn = std::make_shared<
```
msluszniak, line 188: Why we changed `std::string_view` to `std::string` here?

packages/react-native-executorch/common/rnexecutorch/tests/integration/StyleTransferTest.cpp
```diff
@@ -40,57 +50,187 @@ INSTANTIATE_TYPED_TEST_SUITE_P(StyleTransfer, CommonModelTest,
...
-  auto result = model.generate(kValidTestImagePath);
-  test_utils::trimFilePrefix(result);
-  EXPECT_TRUE(std::filesystem::exists(result));
+  EXPECT_NO_THROW((void)model.generateFromString(kValidTestImagePath, false));
+  auto result1 = model.generateFromString(kValidTestImagePath, false);
+  auto result2 = model.generateFromString(kValidTestImagePath, false);
+  ASSERT_TRUE(std::holds_alternative<PixelDataResult>(result1));
+  ASSERT_TRUE(std::holds_alternative<PixelDataResult>(result2));
+  EXPECT_NE(std::get<PixelDataResult>(result1).dataPtr, nullptr);
+  EXPECT_NE(std::get<PixelDataResult>(result2).dataPtr, nullptr);
```
msluszniak: Why these are repeated?

packages/react-native-executorch/common/rnexecutorch/tests/integration/StyleTransferTest.cpp
```diff
@@ -40,57 +50,187 @@ INSTANTIATE_TYPED_TEST_SUITE_P(StyleTransfer, CommonModelTest,
...
+}
+
+TEST(StyleTransferPixelTests, ValidPixelsSaveToFileTrueHasFileScheme) {
+  StyleTransfer model(kValidStyleTransferModelPath, nullptr);
+  std::vector<uint8_t> buf;
+  auto view = makeRgbView(buf, 64, 64);
+  auto result = model.generateFromPixels(view, true);
+  ASSERT_TRUE(std::holds_alternative<std::string>(result));
+  EXPECT_TRUE(std::get<std::string>(result).starts_with("file://"));
+}
```
msluszniak: The first test is a subset of the second one

packages/react-native-executorch/common/rnexecutorch/tests/integration/StyleTransferTest.cpp
```diff
@@ -40,57 +50,187 @@ INSTANTIATE_TYPED_TEST_SUITE_P(StyleTransfer, CommonModelTest,
...
+  EXPECT_EQ(pr.height, 64);
+}
+
+// ============================================================================
+// Thread safety tests
+// ============================================================================
+TEST(StyleTransferThreadSafetyTests, TwoConcurrentGeneratesDoNotCrash) {
+  StyleTransfer model(kValidStyleTransferModelPath, nullptr);
+  std::atomic<int> successCount{0};
+  std::atomic<int> exceptionCount{0};
```
msluszniak:
```suggestion
  std::atomic<int32_t> successCount{0};
  std::atomic<int32_t> exceptionCount{0};
```

packages/react-native-executorch/common/rnexecutorch/tests/unit/FrameProcessorTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::vector<uint8_t> buf;
+  auto view = makeValidView(buf, 48, 64);
+  EXPECT_EQ(pixelsToMat(view).rows, 48);
+}
+
+TEST(PixelsToMatValidInput, ProducesCorrectCols) {
+  std::vector<uint8_t> buf;
+  auto view = makeValidView(buf, 48, 64);
+  EXPECT_EQ(pixelsToMat(view).cols, 64);
+}
```
msluszniak: Merge this two into one test

packages/react-native-executorch/common/rnexecutorch/tests/unit/FrameProcessorTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::vector<uint8_t> buf;
+  auto view = makeValidView(buf, 4, 4);
+  EXPECT_EQ(pixelsToMat(view).channels(), 3);
+}
+
+TEST(PixelsToMatValidInput, MatTypeIsCV_8UC3) {
+  std::vector<uint8_t> buf;
+  auto view = makeValidView(buf, 4, 4);
+  EXPECT_EQ(pixelsToMat(view).type(), CV_8UC3);
+}
```
msluszniak: Again, merge and make two check in one test

apps/computer-vision/components/vision_camera/tasks/ClassificationTask.tsx
NorbertKlockiewicz: It will just make a spam of errors on every frame, not sure if we want to handle it that way.

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
```diff
@@ -1,40 +1,43 @@
...
-  modelImageSize = cv::Size(modelInputShape[modelInputShape.size() - 1],
-                            modelInputShape[modelInputShape.size() - 2]);
 }
 
 std::shared_ptr<OwningArrayBuffer>
-ImageEmbeddings::generate(std::string imageSource) {
-  auto [inputTensor, originalSize] =
-      image_processing::readImageToTensor(imageSource, getAllInputShapes()[0]);
+ImageEmbeddings::runInference(cv::Mat image) {
+  std::scoped_lock lock(inference_mutex_);
```
msluszniak, line 32: You are locking scoped_locks on whole functions in this entire PR, are you sure that all the function call must be protected? I guess that pre/post-processing might be done concurrently and only forward needs to be protected. Can you verify this?

packages/react-native-executorch/common/rnexecutorch/models/semantic_segmentation/BaseSemanticSegmentation.cpp
chmjkb: I think the following is true:
```
It’s a classic C++ "gotcha." Even though std::less<T> is the default, writing out std::less< > (with the empty brackets) actually changes the type of the comparator to something more powerful.

If you just write std::set<std::string>, you get the "old school" default. People explicitly write std::less< > to unlock Heterogeneous Lookup.
```

msluszniak: Ok right, there will be no temporal copies if e.g. char* passed

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
NorbertKlockiewicz: Pre/postprocessing are cheap and they access shared member state, so unlocking around them would create a TOCTOU window with unload(). Only the image loading/extraction (which is the slow part) is kept outside the lock.

packages/react-native-executorch/common/rnexecutorch/models/semantic_segmentation/BaseSemanticSegmentation.cpp
NorbertKlockiewicz: Changed from std::string_view to std::string for the map keys in buffersToReturn to fix a potential dangling reference.

string_view is non-owning — the keys were pointing into allClasses_ memory. With the old synchronous populateDictionary this was safe since everything happened on the same thread within the same call. But now SegmentationResult is a value type that gets returned and can outlive the inference call (and potentially the model itself in concurrent scenarios with VisionCamera).

msluszniak: Okok now I see, thanks

packages/react-native-executorch/common/rnexecutorch/models/embeddings/image/ImageEmbeddings.cpp
msluszniak: If resizing image (in preprocess) is really that fast, then we can indeed lock for safety and simplicity.

docs/docs/03-hooks/02-computer-vision/useObjectDetection.md
```diff
@@ -107,6 +107,12 @@ function App() {
 }
 ```
 
+## VisionCamera integration
+
+For real-time object detection on camera frames, use `runOnFrame`. It runs synchronously on the JS worklet thread and returns `Detection[]`.
+
+See the full guide: [VisionCamera Integration](./visioncamera-integration.md).
```
chmjkb, line 112: Honestly I think just a link to the vision camera integration doc would be enough. This way we don't have to remember to update these docs when changing return types. It's a general comment to all the changes in docs.

apps/computer-vision/app/vision_camera/index.tsx
chmjkb, line 1: When downloading the model i got the maximum state depth update error

msluszniak:
That might look better:

Screenshot_20260313_160720_computer-vision

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -557,4 +560,62 @@ inline jsi::Value getJsiValue(const TranscriptionResult &result,
...
+      uint8ArrayCtor.callAsConstructor(runtime, arrayBuffer).getObject(runtime);
+  obj.setProperty(runtime, "dataPtr", uint8Array);
+
+  auto sizesArray = jsi::Array(runtime, 3);
+  sizesArray.setValueAtIndex(runtime, 0, jsi::Value(result.height));
+  sizesArray.setValueAtIndex(runtime, 1, jsi::Value(result.width));
+  sizesArray.setValueAtIndex(runtime, 2, jsi::Value(4));
+  obj.setProperty(runtime, "sizes", sizesArray);
+
+  obj.setProperty(runtime, "scalarType", jsi::Value(0));
```
chmjkb: use ScalarType::Byte instead of plain magic 0

msluszniak:
Also get these warnings when running demo apps:
```
 WARN  Route "./vision_camera/tasks/types.ts" is missing the required default export. Ensure a React component is exported as default.
 WARN  Route "./vision_camera/utils/colors.ts" is missing the required default export. Ensure a React component is exported as default.
```

msluszniak:
Trying to run regular style transfer freezes the app on my android phone.

This is probably related to: https://github.com/software-mansion/react-native-executorch/issues/962 so we can ignore it for now.

.eslintrc.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
         customWordListFile: path.resolve(__dirname, '.cspell-wordlist.txt'),
       },
     ],
-    'camelcase': 'error',
+    'camelcase': ['error', { properties: 'never' }],
```
chmjkb: Why are we ignoring this rule for properties?

packages/react-native-executorch/src/types/styleTransfer.ts
```diff
@@ -59,9 +59,34 @@ export interface StyleTransferType {
...
+   *
+   * @param input - Image source (string or PixelData object)
+   * @param output - Output format: `'pixelData'` (default) returns raw RGBA pixel data; `'url'` saves the result to a temp file and returns its `file://` path.
+   * @returns A Promise resolving to `PixelData` when `output` is `'pixelData'` (default), or a `file://` URL string when `output` is `'url'`.
    * @throws {RnExecutorchError} If the model is not loaded or is currently processing another image.
    */
-  forward: (imageSource: string) => Promise<string>;
+  forward<O extends 'pixelData' | 'url' = 'pixelData'>(
+    input: string | PixelData,
+    output?: O
```
chmjkb: maybe outputKind / outputType would be a bit better

apps/computer-vision/app/vision_camera/index.tsx
NorbertKlockiewicz: oh, on android or ios, also when downloading which model?

chmjkb: iOS and I believe it was the selfie segmentation one

packages/react-native-executorch/src/types/imageEmbeddings.ts
```diff
@@ -53,9 +53,30 @@ export interface ImageEmbeddingsType {
...
    * Executes the model's forward pass to generate embeddings (a feature vector) for the provided image.
-   * @param imageSource - A string representing the image source (e.g., a file path, URI, or base64 string) to be processed.
+   *
+   * Supports two input types:
+   * 1. **String path/URI**: File path, URL, or Base64-encoded string
+   * 2. **PixelData**: Raw pixel data from image libraries (e.g., NitroImage)
+   *
+   * **Note**: For VisionCamera frame processing, use `runOnFrame` instead.
+   *
+   * @param input - Image source (string or PixelData object)
```
chmjkb: I think you can link to pixeldata here

packages/react-native-executorch/src/types/classification.ts
```diff
@@ -53,9 +53,46 @@ export interface ClassificationType {
...
+  /**
+   * Synchronous worklet function for real-time VisionCamera frame processing.
+   * Automatically handles native buffer extraction and cleanup.
+   *
+   * **Use this for VisionCamera frame processing in worklets.**
+   * For async processing, use `forward()` instead.
+   *
+   * Available after model is loaded (`isReady: true`).
+   *
+   * @example
```
chmjkb: Other models dont have such example. I think we should follow a single approach

packages/react-native-executorch/src/modules/computer_vision/VisionLabeledModule.ts
```diff
@@ -1,6 +1,45 @@
...
+ */
+export type ResolveLabels<
+  T,
+  Configs extends Record<string, { labelMap: LabelEnum }>,
+> = T extends keyof Configs
+  ? Configs[T]['labelMap']
+  : T extends LabelEnum
+    ? T
+    : never;
+
```
chmjkb: i think to keep types seperate this should be moved to some other file

packages/react-native-executorch/src/modules/computer_vision/StyleTransferModule.ts
```diff
@@ -1,22 +1,17 @@
...
 /**
  * Module for style transfer tasks.
  *
  * @category Typescript API
  */
-export class StyleTransferModule extends BaseModule {
-  private constructor(nativeModule: unknown) {
-    super();
-    this.nativeModule = nativeModule;
-  }
```
chmjkb, line 18: Is there a reason why we're ditching this? I feel like this is a cleaner way of creating models, where the factory is responsible for creating the native object, and the module just receives a ready to use thing.

packages/react-native-executorch/src/modules/computer_vision/ImageEmbeddingsModule.ts
```diff
@@ -44,9 +39,9 @@ export class ImageEmbeddingsModule extends BaseModule {
         );
       }
 
-      return new ImageEmbeddingsModule(
-        await global.loadImageEmbeddings(paths[0])
-      );
+      const instance = new ImageEmbeddingsModule();
+      instance.nativeModule = await global.loadImageEmbeddings(paths[0]);
+      return instance;
```
chmjkb: same here regarding the construction method

packages/react-native-executorch/src/modules/computer_vision/ClassificationModule.ts
```diff
@@ -44,9 +41,9 @@ export class ClassificationModule extends BaseModule {
         );
       }
 
-      return new ClassificationModule(
-        await global.loadClassification(paths[0])
-      );
+      const instance = new ClassificationModule();
+      instance.nativeModule = await global.loadClassification(paths[0]);
+      return instance;
```
chmjkb: same here

packages/react-native-executorch/src/modules/computer_vision/ClassificationModule.ts
```diff
@@ -1,22 +1,19 @@
...
  * @category Typescript API
  */
-export class ClassificationModule extends BaseModule {
-  private constructor(nativeModule: unknown) {
-    super();
-    this.nativeModule = nativeModule;
-  }
-
+export class ClassificationModule extends VisionModule<{
+  [category: string]: number;
```
chmjkb, line 15: Not a scope of this PR, but I think we should make this one a VisionLabeledModule

packages/react-native-executorch/src/hooks/computer_vision/useStyleTransfer.ts
```diff
@@ -25,8 +26,14 @@ export const useStyleTransfer = ({
...
     runForward((inst) => inst.forward(imageSource));
 
-  return { error, isReady, isGenerating, downloadProgress, forward };
+  return {
+    error,
+    isReady,
+    isGenerating,
+    downloadProgress,
+    forward,
+  } as StyleTransferType;
```
chmjkb, line 47: If possible let's not use type assertions. Also I'm not sure but my code editor shows some type errors in this file, in line 25 and 30.

chmjkb: maybe its an issue with my LSP but let's double check

packages/react-native-executorch/src/hooks/computer_vision/useSemanticSegmentation.ts
chmjkb, line 1: same here, I get multiple type errors 🤨

packages/react-native-executorch/src/hooks/computer_vision/useObjectDetection.ts
chmjkb, line 1: i get type errors here too

packages/react-native-executorch/src/modules/computer_vision/VisionModule.ts
chmjkb, line 1: When can this happen? Maybe we should throw if it does?
```typescript
    if (!this.nativeModule?.generateFromFrame) {
      return null;
    }

```

packages/react-native-executorch/src/controllers/BaseOCRController.ts
```diff
@@ -87,7 +88,34 @@ export abstract class BaseOCRController {
     }
   };
 
-  public forward = async (imageSource: string): Promise<OCRDetection[]> => {
+  get runOnFrame(): ((frame: Frame) => OCRDetection[]) | null {
+    if (!this.nativeModule?.generateFromFrame) {
+      return null;
+    }
```
chmjkb: Same concern as in VisionModule

apps/computer-vision/app/vision_camera/index.tsx
NorbertKlockiewicz: I've tried to reproduce it in iPhone 16 Pro and SE 3 and wasn't able to, as it's just the example app I suggest to "jeździć obserwować"

packages/react-native-executorch/common/rnexecutorch/utils/FrameProcessor.cpp
```diff
@@ -25,4 +25,52 @@ cv::Mat extractFrame(jsi::Runtime &runtime, const jsi::Object &frameData) {
...
+                            errorMessage);
+  }
+
+  if (pixelData.scalarType != executorch::aten::ScalarType::Byte) {
+    throw RnExecutorchError(
+        RnExecutorchErrorCode::InvalidUserInput,
+        "Invalid pixel data: scalarType must be BYTE (Uint8Array)");
+  }
+
+  uint8_t *dataPtr = static_cast<uint8_t *>(pixelData.dataPtr);
```
chmjkb:
```suggestion
  auto *dataPtr = static_cast<uint8_t *>(pixelData.dataPtr);
```

packages/react-native-executorch/common/rnexecutorch/tests/integration/StyleTransferTest.cpp
```diff
@@ -40,57 +50,174 @@ INSTANTIATE_TYPED_TEST_SUITE_P(StyleTransfer, CommonModelTest,
...
+
+  std::thread a(task);
+  std::thread b(task);
+  a.join();
+  b.join();
+
+  EXPECT_EQ(successCount + exceptionCount, 2);
+}
+
+TEST(StyleTransferThreadSafetyTests,
```
chmjkb: I feel like this could be a typed test for all the models

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -326,8 +326,24 @@ template <typename Model> class ModelHostObject : public JsiHostObject {
...
+                            jsi::String::createFromUtf8(runtime, e.what()));
+      throw jsi::JSError(runtime, jsi::Value(runtime, std::move(errorData)));
+    } catch (const std::runtime_error &e) {
+      // This catch should be merged with the next one
+      // (std::runtime_error inherits from std::exception) HOWEVER react
+      // native has broken RTTI which breaks proper exception type
+      // checking. Remove when the following change is present in our
+      // version:
+      // https://github.com/facebook/react-native/commit/3132cc88dd46f95898a756456bebeeb6c248f20e
+      throw jsi::JSError(runtime, e.what());
```
mkopcins, line 342: I think this is already present in react native v0.81.5 https://github.com/facebook/react-native/blob/v0.81.5/packages/react-native/ReactCommon/yoga/cmake/project-defaults.cmake

NorbertKlockiewicz: https://github.com/software-mansion/react-native-executorch/issues/964?reload=1

packages/react-native-executorch/src/modules/computer_vision/ClassificationModule.ts
NorbertKlockiewicz: Agree, I was thinking about it today :D, I will create an issue.

NorbertKlockiewicz: https://github.com/software-mansion/react-native-executorch/issues/965

packages/react-native-executorch/src/hooks/computer_vision/useStyleTransfer.ts
NorbertKlockiewicz: It's here because the StyleTransfer is an edge case as it's forward function can return both PixelData and string which will be an url to saved image. `runOnFrame` however returns only PixelData. That inconsistency breaks the contract stated by VisionModule which uses the same output type for both forward and runOnFrame. I've tried to make it work without this type assertion but it required quite complex changes to types in VisionModule and I decided that it doesn't really make sense for just one model.

NorbertKlockiewicz: About the typescript error, doesn't show in my IDE, also the precommit doesn't catch anything so I suppose it should be alright.

packages/react-native-executorch/src/modules/computer_vision/VisionModule.ts
NorbertKlockiewicz: It would skip inference if the model isn't loaded, however I think we can just check `if (this.nativeModule == null)`

chmjkb: My point is that skipping inference silently here might be unexpected for the user.

packages/react-native-executorch/src/modules/computer_vision/StyleTransferModule.ts
NorbertKlockiewicz: I've selected wrong version in rebase :/

packages/react-native-executorch/common/rnexecutorch/models/VisionModel.h
```diff
@@ -53,6 +53,20 @@ class VisionModel : public BaseModel {
...
+   *
+   * Overrides BaseModel::unload() to acquire inference_mutex_ before
+   * resetting the module. This prevents a crash where BaseModel::unload()
+   * destroys module_ while generateFromFrame() is still executing on the
+   * VisionCamera worklet thread.
+   */
+  void unload() noexcept {
+    std::scoped_lock lock(inference_mutex_);
+    BaseModel::unload();
+  }
```
mkopcins: Why dont we do it inside base model?

mkopcins: we can move the `inference_mutex_` to BaseModel class

packages/react-native-executorch/common/rnexecutorch/models/VisionModel.h
```diff
@@ -53,6 +53,20 @@ class VisionModel : public BaseModel {
...
+
 protected:
   /**
```
mkopcins, line 65: field/method order https://google.github.io/styleguide/cppguide.html#Declaration_Order

packages/react-native-executorch/common/rnexecutorch/utils/FrameProcessor.cpp
```diff
@@ -25,4 +25,52 @@ cv::Mat extractFrame(jsi::Runtime &runtime, const jsi::Object &frameData) {
...
+  cv::Mat frame = extractFrame(runtime, frameObj);
+
+  // Camera sensors deliver landscape frames; rotate to portrait orientation.
+  if (frame.cols > frame.rows) {
+    cv::Mat upright;
+    cv::rotate(frame, upright, cv::ROTATE_90_CLOCKWISE);
+    return upright;
+  }
+  return frame;
+}
```
mkopcins, line 40: iirc we discussed today handling of landscape input, this won't work for landscape, right?

NorbertKlockiewicz: https://github.com/software-mansion/react-native-executorch/pull/880#issuecomment-4054511454

packages/react-native-executorch/common/rnexecutorch/models/VisionModel.h
NorbertKlockiewicz: I think it should stay as for the vision models we can run the inference from two threads (js and worklet) and for the most of the models we have, an inference is run only from the js thread so locking there isn't really needed.

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -557,4 +560,63 @@ inline jsi::Value getJsiValue(const TranscriptionResult &result,
...
+  auto uint8ArrayCtor =
+      runtime.global().getPropertyAsFunction(runtime, "Uint8Array");
+  auto uint8Array =
+      uint8ArrayCtor.callAsConstructor(runtime, arrayBuffer).getObject(runtime);
+  obj.setProperty(runtime, "dataPtr", uint8Array);
+
+  auto sizesArray = jsi::Array(runtime, 3);
+  sizesArray.setValueAtIndex(runtime, 0, jsi::Value(result.height));
+  sizesArray.setValueAtIndex(runtime, 1, jsi::Value(result.width));
+  sizesArray.setValueAtIndex(runtime, 2, jsi::Value(4));
```
msluszniak: What is 4 here?

NorbertKlockiewicz: It's a number of channels but it should be passed the same way as the height and width, on it now!

Issue #831, opened 2026-02-24
Configure eslint-plugin-jsdoc and eslint-plugin-tsdoc


mateuszlampert:
Add config for `eslint-plugin-jsdoc` and `eslint-plugin-tsdoc` to make sure that TypeDocs match actual function's signature and to define possible `@Category` tag values.

PR #883, opened 2026-02-26
chore: configure eslint-jsdoc-plugin

mateuszlampert:
## Description

This PR configures eslint-jsdoc-plugin so:
- JSDocs are consistent throughout codebase and jsdocs are consistent with method/interface signature
- there is defined set of `@category` tag values to prevent typos and generating additional sections in docs
- fixed few jsdocs manually

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

- try to break some jsdocs (e. g. remove `@returns` tag, set `@category` to some value outside of values defined in `eslintrc.js`, change `@param` tag so it does not match signature) and see if eslint catches these errors. 
- build and open docs locally and check out if api reference is valid

### Related issues

#831 

### Checklist

### Additional notes

**NOTE:** for now I switched off required destructured params as that's the convention we are mostly using, but we could move to destructured params eventually.

### DIFF
--- a/packages/react-native-executorch/src/modules/computer_vision/InstanceSegmentationModule.ts
+++ b/packages/react-native-executorch/src/modules/computer_vision/InstanceSegmentationModule.ts
@@ -99,7 +99,11 @@ export type InstanceSegmentationLabels<
   M extends InstanceSegmentationModelName,
 > = ResolveLabels<M>;
 
-/** @internal */
+/**
+ * Resolves the label type: if `T` is a {@link InstanceSegmentationModelName}, looks up its labels
+ * from the built-in config; otherwise uses `T` directly as a {@link LabelEnum}.
+ * @internal
+ */
 type ResolveLabels<T extends InstanceSegmentationModelName | LabelEnum> =
   ResolveLabelsFor<T, ModelConfigsType>;
 


packages/react-native-executorch/src/utils/llms/context_strategy/MessageCountContextStrategy.ts
```diff
@@ -3,24 +3,22 @@ import { ContextStrategy, Message } from '../../../types/llm';
...
   constructor(private readonly windowLength: number = 5) {}
 
   /**
    * Builds the context by slicing the history to retain only the most recent `windowLength` messages.
-   *
-   * @param {string} systemPrompt - The top-level instructions for the model.
-   * @param {Message[]} history - The complete conversation history.
-   * @param {(messages: Message[]) => number} _getTokenCount - Unused in this strategy.
-   * @param {number} _maxContextLength - Unused in this strategy.
-   * @returns {Message[]} The truncated message history with the system prompt at the beginning.
```
msluszniak, line 23: @mateuszlampert why this configuration deletes these types here and in other places?

mateuszlampert: For typescript projects it's recommended to get rid of these types as they are redundant with method signatures/interfaces etc.

API reference in our docs still has these types (they are now infered from method signatures, so it's guaranteed that api reference is consistent with an actual implementation)

we can also turn this rule off, but IMO it's better. WDYT?

msluszniak: >we can also turn this rule off, but IMO it's better. WDYT?

I think there is no need for that, I was just curious. Thank you for the explanation.

msluszniak:
When you'll return to this PR @mateuszlampert please also make sure that incorrect anchors in links will raise an error instead of warning.

packages/react-native-executorch/src/modules/computer_vision/InstanceSegmentationModule.ts
```diff
@@ -90,20 +92,14 @@ type ModelConfigsType = typeof ModelConfigs;
...
  */
 export type InstanceSegmentationLabels<
   M extends InstanceSegmentationModelName,
 > = ResolveLabels<M>;
 
-/**
- * @internal
- * Resolves the label type: if `T` is a {@link InstanceSegmentationModelName}, looks up its labels
- * from the built-in config; otherwise uses `T` directly as a {@link LabelEnum}.
- */
```
msluszniak, line 106: This was mark as an error and I was forced to change it.

mateuszlampert: you can move `@internal` to the bottom of the jsdoc and it should work

packages/react-native-executorch/src/modules/computer_vision/InstanceSegmentationModule.ts
```diff
@@ -90,20 +92,14 @@ type ModelConfigsType = typeof ModelConfigs;
...
+/** @internal */
```
msluszniak:
```suggestion
/**
 * Resolves the label type: if `T` is a {@link InstanceSegmentationModelName}, looks up its labels
 * from the built-in config; otherwise uses `T` directly as a {@link LabelEnum}.
 * @internal
 */
```

PR #886, opened 2026-02-27
fix: Make package.json accept newer versions of libs, e.g. Expo

### LINKED ISSUE #882: Installation expo 55
### Description

hello,

I've tried to migrate to expo 55 but i'm getting some trouble with the latest version.

```
D:\github\nola>npm install
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: react-native-executorch@0.7.1
npm error Found: expo-asset@55.0.7
npm error node_modules/expo-asset
npm error   expo-asset@"~55.0.7" from the root project
npm error   expo-asset@"~55.0.7" from expo@55.0.2
npm error   node_modules/expo
npm error     expo@"^55.0.2" from the root project
npm error     peer expo@"*" from @expo/dom-webview@55.0.3
npm error     node_modules/@expo/dom-webview
npm error       @expo/dom-webview@"^55.0.3" from @expo/log-box@55.0.7
npm error       node_modules/@expo/log-box
npm error         @expo/log-box@"55.0.7" from expo@55.0.2
npm error         1 more (@expo/cli)
npm error       1 more (expo)
npm error     40 more (@expo/log-box, @expo/metro-config, ...)
npm error   1 more (expo-audio)
npm error
npm error Could not resolve dependency:
npm error peer expo-asset@"^12.0.0" from react-native-executorch@0.7.1
npm error node_modules/react-native-executorch
npm error   react-native-executorch@"^0.7.1" from the root project
npm error
npm error Conflicting peer dependency: expo-asset@12.0.12
npm error node_modules/expo-asset
npm error   peer expo-asset@"^12.0.0" from react-native-executorch@0.7.1
npm error   node_modules/react-native-executorch
npm error     react-native-executorch@"^0.7.1" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
npm error
npm error
```
They changed libaries version to match the same version as the sdk

### Steps to reproduce

npm install

### Snack or a link to a repository

_No response_

### React Native Executorch version

0.7.1

### React Native version

0.83

### Platforms

Platform agnostic

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### AI model

_No response_

### Performance logs

_No response_

### Acknowledgements

Yes

msluszniak:
## Description

This PR fixes package.json in both expo and bare resource fetchers.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

* Run minimal app that utilizes this version of our library with expo 55 (pack react native executorch and expo resource fetchers into tgz and install in template app) and check if everything works correctly.
* Run text to image example as this PR introduces changes in this package and check if everything works ok.

### Related issues

Closes #882 

### Checklist

### DIFF
--- a/packages/bare-resource-fetcher/package.json
+++ b/packages/bare-resource-fetcher/package.json
@@ -26,8 +26,8 @@
     "clean": "del-cli lib"
   },
   "peerDependencies": {
-    "@dr.pogodin/react-native-fs": "^2.0.0",
-    "@kesha-antonov/react-native-background-downloader": "^4.0.0",
+    "@dr.pogodin/react-native-fs": ">=2.0.0",
+    "@kesha-antonov/react-native-background-downloader": ">=4.0.0",
     "react-native": "*",
     "react-native-executorch": "*"
   },
--- a/packages/bare-resource-fetcher/package.json
+++ b/packages/bare-resource-fetcher/package.json
@@ -26,8 +26,8 @@
     "clean": "del-cli lib"
   },
   "peerDependencies": {
-    "@dr.pogodin/react-native-fs": ">=2.0.0",
-    "@kesha-antonov/react-native-background-downloader": ">=4.0.0",
+    "@dr.pogodin/react-native-fs": "^2.0.0",
+    "@kesha-antonov/react-native-background-downloader": "^4.0.0",
     "react-native": "*",
     "react-native-executorch": "*"
   },
--- a/packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts
+++ b/packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts
@@ -1,7 +1,7 @@
 import { ResourceFetcher } from '../../utils/ResourceFetcher';
 import { ResourceSource } from '../../types/common';
 import { BaseModule } from '../BaseModule';
-import { Buffer } from 'buffer';
+
 import { PNG } from 'pngjs/browser';
 import { RnExecutorchErrorCode } from '../../errors/ErrorCodes';
 import { parseUnknownError, RnExecutorchError } from '../../errors/errorUtils';
@@ -121,10 +121,15 @@ export class TextToImageModule extends BaseModule {
       return '';
     }
     const png = new PNG({ width: imageSize, height: imageSize });
-    png.data = Buffer.from(outputArray);
+    png.data = outputArray as unknown as Buffer;
     const pngBuffer = PNG.sync.write(png, { colorType: 6 });
-    const pngString = pngBuffer.toString('base64');
-    return pngString;
+    const pngArray = new Uint8Array(pngBuffer as unknown as ArrayBufferLike);
+    let binary = '';
+    const chunkSize = 8192;
+    for (let i = 0; i < pngArray.length; i += chunkSize) {
+      binary += String.fromCharCode(...pngArray.subarray(i, i + chunkSize));
+    }
+    return btoa(binary);
   }
 
   /**
--- a/docs/docs/06-api-reference/classes/TextToImageModule.md
+++ b/docs/docs/06-api-reference/classes/TextToImageModule.md
@@ -178,7 +178,7 @@ The input shape as an array of numbers.
 
 > **interrupt**(): `void`
 
-Defined in: [modules/computer_vision/TextToImageModule.ts:133](https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts#L133)
+Defined in: [modules/computer_vision/TextToImageModule.ts:138](https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts#L138)
 
 Interrupts model generation. The model is stopped in the nearest step.
 
--- a/packages/expo-resource-fetcher/package.json
+++ b/packages/expo-resource-fetcher/package.json
@@ -27,8 +27,8 @@
   },
   "peerDependencies": {
     "expo": ">=54.0.0",
-    "expo-asset": "^12.0.0",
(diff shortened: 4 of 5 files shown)


packages/bare-resource-fetcher/package.json
```diff
@@ -26,8 +26,8 @@
     "clean": "del-cli lib"
   },
   "peerDependencies": {
-    "@dr.pogodin/react-native-fs": "^2.0.0",
-    "@kesha-antonov/react-native-background-downloader": "^4.0.0",
```
chmjkb, line 30: Should we do this for those packages as well? I feel like we can easily manage new Expo versions, but any major change here will be less expected

packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts
```diff
@@ -121,10 +121,15 @@ export class TextToImageModule extends BaseModule {
...
     const pngBuffer = PNG.sync.write(png, { colorType: 6 });
-    const pngString = pngBuffer.toString('base64');
-    return pngString;
+    const pngArray = new Uint8Array(pngBuffer as unknown as ArrayBufferLike);
+    let binary = '';
+    const chunkSize = 8192;
+    for (let i = 0; i < pngArray.length; i += chunkSize) {
+      binary += String.fromCharCode(...pngArray.subarray(i, i + chunkSize));
+    }
+    return btoa(binary);
```
chmjkb, line 132: Why do we need this change?

packages/bare-resource-fetcher/package.json
msluszniak: Ok, fair point. I'll revert these changes.

packages/react-native-executorch/src/modules/computer_vision/TextToImageModule.ts
msluszniak: Because otherwise there was an error caused by the `Buffer` package. I rewrite this to remove Buffer entirely. Maybe you have a better idea how to fix this?

msluszniak: For more context:

```
The original error (before commit 4017db56) came from `TextToImageModule.ts` having:


import { Buffer } from 'buffer';
// ...
png.data = Buffer.from(outputArray);
const pngString = pngBuffer.toString('base64');
Root cause — two conflicting Buffer implementations:

pngjs/browser.js (v3.4.0) is a browserified bundle that ships its own internal buffer polyfill (v5.7.1, found in node_modules/bl/node_modules/buffer)
import { Buffer } from 'buffer' resolves to v6.0.3 at the root node_modules/buffer
In Expo 55 with Metro, these are treated as different classes. So:

pngBuffer returned by PNG.sync.write() is an instance of the v5 Buffer
Calling .toString('base64') on it fails or returns garbled output because the v6 Buffer class (what you imported) doesn't recognize it
instanceof Buffer checks across the two versions also fail
```

msluszniak: But I also found this: https://github.com/expo/expo/issues/42828

which suggest that if with bump svg the problem will be resolved as well.
[EDIT] ~We already have SVG 15.3.3, so I don't think this this is it.~ The alternative fix is to add `buffer` package as an explicit dependency.

msluszniak:
Unfortunately, it seems that Expo 55.0.3 does not included fix with newer RNScreens version, which blocks us from bumping expo in our example apps look here: https://github.com/software-mansion/react-native-screens/issues/3682. So we need to wait for this until this will be merged.

chmjkb (approved):
lgtm, but we need to do the same thing on the 0.7 release branch and release a patched version so 0.7 supports Expo 55

chmjkb:
also TBH i don't understand why we're not just saving the image to FS on the native side, and returning the URI but that's not the point of this PR, am i missing something here?

msluszniak:
>also TBH i don't understand why we're not just saving the image to FS on the native side, and returning the URI but that's not the point of this PR, am i missing something here?

You mean specifically TextToImage TS part which was modified in this PR?
[EDIT] Same thing is with Style transfer. Fairly honest, I don't see any reason at first so I'll create an issue for that.

chmjkb:
> > also TBH i don't understand why we're not just saving the image to FS on the native side, and returning the URI but that's not the point of this PR, am i missing something here?
> 
> You mean specifically TextToImage TS part which was modified in this PR? [EDIT] Same thing is with Style transfer. Fairly honest, I don't see any reason at first so I'll create an issue for that.

Yeah, style transfer does this:
```
std::string StyleTransfer::postprocess(const Tensor &tensor,
                                       cv::Size originalSize) {
  cv::Mat mat = image_processing::getMatrixFromTensor(modelImageSize, tensor);
  cv::resize(mat, mat, originalSize);

  return image_processing::saveToTempFile(mat);
}
```
It would make sense to me that TTI does this too (and we wouldnt have to do the png stuff). This isn't really related to changes made in this PR, but  a thought of mine

msluszniak:
Ok, right style transfer is ok, but typedocs are incorrect:

```typescript
  /**
   * Executes the model's forward pass, where `imageSource` can be a fetchable resource or a Base64-encoded string.
   *
   * @param imageSource - The image source to be processed.
   * @returns The stylized image as a Base64-encoded string. THIS IS INCORRECT
   */
  async forward(imageSource: string): Promise<string> {
    if (this.nativeModule == null)
      throw new RnExecutorchError(
        RnExecutorchErrorCode.ModuleNotLoaded,
        'The model is currently not loaded. Please load the model before calling forward().'
      );
    return await this.nativeModule.generate(imageSource);
  }
  ```

PR #940, opened 2026-03-06
chore: add support for quantized versions of CV models CLIP, Style Transfer, EfficientNetV2, SSDLite

### LINKED ISSUE #719: Quantize CV models
Quantize CV models to improve performance, especially for real time usecases

barhanc:
## Description

Adds support for quantized versions of CV models CLIP, Style Transfer, EfficientNetV2, SSDLite and updates paths to non-quantized models exported with ExecuTorch v1.1.0.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. Run the Computer Vision example app:
   - Object detection with model set to:
      - `SSDLITE_320_MOBILENET_V3_LARGE`  
   - Classification with model set to:
      - `EFFICIENTNET_V2_S`,
      - `EFFICIENTNET_V2_S_QUANTIZED` 
   - Style transfer with model set to: 
      - `STYLE_TRANSFER_CANDY`,
      - `STYLE_TRANSFER_MOSAIC`,
      - `STYLE_TRANSFER_UDNIE`,
      - `STYLE_TRANSFER_RAIN_PRINCESS`,
      - `STYLE_TRANSFER_CANDY_QUANTIZED`,
      - `STYLE_TRANSFER_MOSAIC_QUANTIZED`,
      - `STYLE_TRANSFER_UDNIE_QUANTIZED`,
      - `STYLE_TRANSFER_RAIN_PRINCESS_QUANTIZED`,
2. Run the Text Embeddings example app:
    - CLIP embeddings with image model set to:
       - `CLIP_VIT_BASE_PATCH32_IMAGE`,
       - `CLIP_VIT_BASE_PATCH32_IMAGE_QUANTIZED`
3. Check HF pages for updated models:
   - https://huggingface.co/software-mansion/react-native-executorch-style-transfer-candy
   - https://huggingface.co/software-mansion/react-native-executorch-style-transfer-mosaic
   - https://huggingface.co/software-mansion/react-native-executorch-style-transfer-rain-princess
   - https://huggingface.co/software-mansion/react-native-executorch-style-transfer-udnie
   - https://huggingface.co/software-mansion/react-native-executorch-efficientnet-v2-s
   - https://huggingface.co/software-mansion/react-native-executorch-ssdlite320-mobilenet-v3-large
   - https://huggingface.co/software-mansion/react-native-executorch-clip-vit-base-patch32

### Related issues

Closes #719 

### Checklist

### DIFF
--- a/docs/docs/02-benchmarks/memory-usage.md
+++ b/docs/docs/02-benchmarks/memory-usage.md
@@ -2,32 +2,68 @@
 title: Memory Usage
 ---
 
-:::info
-All the below benchmarks were performed on iPhone 17 Pro (iOS) and OnePlus 12 (Android).
-:::
-
 ## Classification
 
-| Model             | Android (XNNPACK) [MB] | iOS (Core ML) [MB] |
-| ----------------- | :--------------------: | :----------------: |
-| EFFICIENTNET_V2_S |          230           |         87         |
+:::info
+Memory usage values represent the peak memory increase observed while the model was
+loaded and actively running inference, relative to the baseline app memory
+before model initialization.
+:::
+
+| Model / Device                   | iPhone 17 Pro [MB] | Google Pixel 10 [MB] |
+| -------------------------------- | :----------------: | :------------------: |
+| EFFICIENTNET_V2_S (XNNPACK FP32) |        101         |         122          |
+| EFFICIENTNET_V2_S (XNNPACK INT8) |         62         |          78          |
+| EFFICIENTNET_V2_S (Core ML FP32) |        101         |          -           |
+| EFFICIENTNET_V2_S (Core ML FP16) |         87         |          -           |
 
 ## Object Detection
 
-| Model                          | Android (XNNPACK) [MB] | iOS (XNNPACK) [MB] |
-| ------------------------------ | :--------------------: | :----------------: |
-| SSDLITE_320_MOBILENET_V3_LARGE |          164           |        132         |
+:::info
+Memory usage values represent the peak memory increase observed while the model was
+loaded and actively running inference, relative to the baseline app memory
+before model initialization.
+:::
+
+| Model / Device                                | iPhone 17 Pro [MB] | Google Pixel 10 [MB] |
+| --------------------------------------------- | :----------------: | :------------------: |
+| SSDLITE_320_MOBILENET_V3_LARGE (XNNPACK FP32) |         94         |         104          |
+| SSDLITE_320_MOBILENET_V3_LARGE (Core ML FP32) |         83         |          -           |
+| SSDLITE_320_MOBILENET_V3_LARGE (Core ML FP16) |         62         |          -           |
 
 ## Style Transfer
 
-| Model                        | Android (XNNPACK) [MB] | iOS (Core ML) [MB] |
-| ---------------------------- | :--------------------: | :----------------: |
-| STYLE_TRANSFER_CANDY         |          1200          |        380         |
-| STYLE_TRANSFER_MOSAIC        |          1200          |        380         |
-| STYLE_TRANSFER_UDNIE         |          1200          |        380         |
-| STYLE_TRANSFER_RAIN_PRINCESS |          1200          |        380         |
+:::info
+Memory usage values represent the peak memory increase observed while the model was
+loaded and actively running inference, relative to the baseline app memory
+before model initialization.
+:::
+
+| Model / Device                              | iPhone 17 Pro [MB] | Google Pixel 10 [MB] |
+| ------------------------------------------- | :----------------: | :------------------: |
+| STYLE_TRANSFER_CANDY (XNNPACK FP32)         |        1200        |         1200         |
+| STYLE_TRANSFER_CANDY (XNNPACK INT8)         |        800         |         800          |
+| STYLE_TRANSFER_CANDY (Core ML FP32)         |        400         |          -           |
+| STYLE_TRANSFER_CANDY (Core ML FP16)         |        380         |          -           |
+| STYLE_TRANSFER_MOSAIC (XNNPACK FP32)        |        1200        |         1200         |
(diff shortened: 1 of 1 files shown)


NorbertKlockiewicz:
I will run the new models later today to see if they work. I think you should also benchmark them and add the results to our docs.

You can ask @IgorSwat for the tips about benchmarking ;D

IgorSwat:
@barhanc Did you profile the added XNNPACK models following [these](https://gl.swmansion.com/swm-ai/ai_knowledge_base/-/wikis/INTERNAL-DOCS/HOW-TO/PROFILE_EXPORTED_MODELS?redirected_from=INTERNAL-DOCS/HOW-TO/{new_page_title}) instructions? 

I guess they should be fine considering that those are only a quantized versions of already profiled models, but it's always nice to check if everything is alright with export.

barhanc:
I've added the profiling results to the corresponding READMEs in the internal exports gitlab. They all look fine to me (>80% delegated ops), but you can also take a look to make sure everything is correct.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -384,8 +388,18 @@ export const EFFICIENTNET_V2_S = {
   modelSource: EFFICIENTNET_V2_S_MODEL,
 };
 
+/**
+ * @category Models - Classification
+ */
+export const EFFICIENTNET_V2_S_QUANTIZED = {
+  modelSource: EFFICIENTNET_V2_S_QUANTIZED_MODEL,
+};
```
NorbertKlockiewicz, line 398:
```suggestion
export const EFFICIENTNET_V2_S_QUANTIZED = {
  modelName: "efficientnet-v2-s-quantized"
  modelSource: EFFICIENTNET_V2_S_QUANTIZED_MODEL,
};
```

Please add modelName to each new model

NorbertKlockiewicz (changes requested):
Please change examples in demo apps to use quantized models by default.

packages/react-native-executorch/src/constants/modelUrls.ts
barhanc: Done

msluszniak:
Please add in benchmark section how memory usage was measured

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -374,18 +374,34 @@ export const LFM2_5_1_2B_INSTRUCT_QUANTIZED = {
...
 const EFFICIENTNET_V2_S_MODEL =
   Platform.OS === `ios`
-    ? `${URL_PREFIX}-efficientnet-v2-s/${VERSION_TAG}/coreml/efficientnet_v2_s_coreml_all.pte`
-    : `${URL_PREFIX}-efficientnet-v2-s/${VERSION_TAG}/xnnpack/efficientnet_v2_s_xnnpack.pte`;
+    ? `${URL_PREFIX}-efficientnet-v2-s/${NEXT_VERSION_TAG}/coreml/efficientnet_v2_s_coreml_fp32.pte`
+    : `${URL_PREFIX}-efficientnet-v2-s/${NEXT_VERSION_TAG}/xnnpack/efficientnet_v2_s_xnnpack_fp32.pte`;
+const EFFICIENTNET_V2_S_QUANTIZED_MODEL =
+  Platform.OS === `ios`
+    ? `${URL_PREFIX}-efficientnet-v2-s/${NEXT_VERSION_TAG}/coreml/efficientnet_v2_s_coreml_fp16.pte`
+    : `${URL_PREFIX}-efficientnet-v2-s/${NEXT_VERSION_TAG}/xnnpack/efficientnet_v2_s_xnnpack_int8.pte`;
```
chmjkb, line 382: I'm thinking if we should silently run the coreml, or let the user handle it? Some users might prefer consistency across platforms, rather than performance. cc @NorbertKlockiewicz

NorbertKlockiewicz: I think we should discuss this and plan accordingly. The simplest way would be to have an constant for both xnnpack and coreml model, but with quantized models that's already 4 constants. Maybe we can figure out a better way so the users will be able to easily switch between models delegated to different backends. Answering your question, we should definitely let users select it.

barhanc: I guess we can add constants like `EFFICIENTNET_V2_S_XNNPACK_FP32` that signal unambiguously both the backend and the precision for "power users" who want control; and additionally have simple constants like just `EFFICIENTNET_V2_S` which will just be the fastest variant of the model on the given platform.

NorbertKlockiewicz: Let's keep it as it is now and we we will do it properly in other PR.

msluszniak:
Just a question, why for [react-native-executorch-ssdlite320-mobilenet-v3-large](https://huggingface.co/software-mansion/react-native-executorch-ssdlite320-mobilenet-v3-large) for xnnpack there is only f32 version? And is there any reason why there is no CoreML for [react-native-executorch-clip-vit-base-patch32](https://huggingface.co/software-mansion/react-native-executorch-clip-vit-base-patch32). If there were some problems with exports, I think we should create issues on ExecuTorch repo for that.

barhanc:
@msluszniak I didn't export the CLIP Vision model to CoreML backend because the xnnpack variant is already extremely fast on iOS (<20ms `forward()` on iPhone 17 Pro) and I don't think we would see any benefit from CoreML in that case, but I can add it of course. I had some problems with quantizing SSDLite and since xnnpack fp32 is already fast (<20ms `forward()` on Pixel 10) I didn't think it was worth it.

msluszniak:
The question is if they are super-fast also on low-tier devices. I guess that the fact that models are super-fast on top devices, doesn't mean that we can speed them up for slower devices ;) but of course you are in the better position right now to decide if this is worth trying.

PR #936, opened 2026-03-04
feat!: text-to-speech x LLM integration

IgorSwat:
## Description

This pull request introduces a few changes to the Text-to-Speech module:
- Improved streaming mode by allowing an incrementally expanded text input. This change focuses on integrating T2S with text generation models (e.g. Llama 3.2).
- Added simple test cases for T2S module.

### Introduces a breaking change?

- [x] Yes
- [ ] No

### Type of change

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Documentation update (improves or adds clarity to existing documentation)
- [x] Other (chores, tests, code style improvements etc.)

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

To test the Text-to-Speech module, run the set of tests for this module.
To test the new streaming mode and it's integration with text generation models, one can use 'text-to-speech-llm' demo app.

### Related issues

#773 
#897 

### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation accordingly
- [x] My changes generate no new warnings

PR #943, opened 2026-03-06
fix(kokoro): voice loading, method selection, padding, and audio safety fixes

yocontra:
Fixes for Kokoro TTS native code. Addresses voice data truncation, missing Synthesizer method selection, progressive speed-up on longer inputs, phoneme token reordering, and several additional safety fixes.

### Voice loading reads only 128 of 510 rows

`voice_` was a fixed `std::array<..., kMaxInputTokens>` (128 elements), but `hexgrad/Kokoro-82M` voice files contain 510 rows. The remaining 382 rows were silently dropped.

Changed `voice_` to `std::vector`, sized dynamically from the file. Also fixed an OOB in `voiceID` — upstream used `std::min(phonemes.size() - 1, noTokens)` where `noTokens` could equal 128, indexing past the end of a 128-element array. Now uses a three-way `std::min({phonemes.size() - 1, noTokens - 1, voice_.size() - 1})`.

### Synthesizer doesn't do method selection

`DurationPredictor` discovers and selects from `forward_8`/`forward_32`/`forward_64`/`forward_128` based on input size, but `Synthesizer` only knew about `forward`. Added the same discovery and selection logic. Falls back to `"forward"` if no `forward_N` methods exist, so older models still work.

### Audio progressively speeds up on longer inputs

The Synthesizer's attention mechanism drifts on longer input sequences (60+ tokens), causing later phonemes to be spoken progressively faster than the Duration Predictor intended. The DP's timing predictions are correct, but the Synthesizer compresses later phonemes into fewer samples.

Fixed by capping `inputTokensLimit` to 60, which forces the Partitioner to split text into shorter chunks that the Synthesizer can render faithfully. Each chunk is roughly one sentence (~15-20 words).

### `tokenize()` scrambles phoneme order on invalid tokens

`std::partition` was used to filter out invalid (unrecognized) phoneme tokens, but `partition` does not preserve relative order. When any phonemes fall outside the vocabulary, the remaining valid tokens could be reordered, producing garbled audio.

Changed to `std::stable_partition` which preserves relative order.

### `stripAudio` unsigned integer underflow

`lbound - margin` wraps `size_t` to ~2^64 when the audio's first non-silent sample is near the start of the buffer (i.e., `lbound < margin`). `std::max(huge_value, 0)` returns the huge value, and `audio.subspan()` reads out-of-bounds. This is especially triggered in streaming mode where `paddingMs=15` (margin = 360 samples) on short chunks.

Fixed by guarding the subtraction: `lbound > margin ? lbound - margin : 0`. Also guarded `audio.size() - 1` against empty spans.

### `isStreaming_` data race

`isStreaming_` is a plain `bool` read by `stream()` on a background thread and written by `streamStop()` from the JS thread. Non-atomic access is undefined behavior — the compiler may optimize away the read, making `streamStop()` ineffective.

Changed to `std::atomic<bool>`.

### `scaleDurations` drops phonemes

When aggressively shrinking durations (many tokens, few total ticks), individual token durations can be driven to 0 by the correction loop. A zero-duration token is skipped by `repeatInterleave`, effectively deleting that phoneme from the output.

Fixed by clamping each scaled duration to a minimum of 1, and guarding the correction loop so it never drives a duration below 1. Without the correction loop guard, the clamp is immediately undone — the priority queue picks clamped entries (they have high remainders) and subtracts 1, defeating the purpose.

### Misc perf

- Replace temporary pause zero-vectors with `resize()` directly on the output
- Move-capture audio in the streaming callback instead of copying

## Changes

- `Kokoro.h` — `voice_` from fixed array to vector, `isStreaming_` to `std::atomic<bool>`
- `Kokoro.cpp` — `loadVoice()`, `synthesize()`, `generate()`, `stream()`, constructor token limit cap
- `DurationPredictor.cpp` — `scaleDurations()` min-1 clamp with correction loop guard
- `Synthesizer.h` — `forwardMethods_` member
- `Synthesizer.cpp` — method discovery and selection
- `Utils.cpp` — `stable_partition` in `tokenize()`, `stripAudio` underflow guard

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
@@ -71,7 +71,7 @@ Result<std::vector<EValue>> Synthesizer::generate(std::span<const Token> tokens,
                                         ref_s.data(), ScalarType::Float);
 
   // Select appropriate forward method based on token count
-  auto it = std::find_if(forwardMethods_.begin(), forwardMethods_.end(),
+  auto it = std::ranges::find_if(forwardMethods_,
       [noTokens](const auto &entry) { return static_cast<int32_t>(entry.second) >= noTokens; });
   std::string selectedMethod = (it != forwardMethods_.end()) ? it->first : forwardMethods_.back().first;
 


chmjkb:
Thanks for the contribution, @yocontra! We'll be taking a closer look at this soon. In the meantime, if you're open to sharing your experience using this or any other APIs in the library, we'd love to hear your feedback!

yocontra:
Right now I'm trying to get TTS working well - I was using an onnx based approach and got the quality pretty high (after writing my own phonemizer, because nobody else had theirs quite right). Switched to this library as I need multiple types of things now, and been fixing issues to get the quality to where that other library was.

I'm also sending some PRs to the phonemizer this library uses as there are some words it pronounces wrong and that should get it to perfect again :)

yocontra:
I'm very eager for https://github.com/software-mansion/react-native-executorch/pull/936 as I'm trying to get a fully local replacement for gemini live/chatgpt realtime by doing Speech to Text -> Llama 3.2 -> Kokoro TTS

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -174,8 +174,9 @@ void DurationPredictor::scaleDurations(Tensor &durations, size_t nTokens,
     float remainder =
         shrinking ? std::ceil(scaled) - scaled : scaled - std::floor(scaled);
 
-    durationsPtr[i] = static_cast<int64_t>(shrinking ? std::ceil(scaled)
-                                                     : std::floor(scaled));
+    durationsPtr[i] = std::max(1LL,
```
msluszniak: Do not use long long here since it clashes with `int64_t` (aka long int on 64-bit Android). Even if they both take 64 bits, compiler doesn't have an overload for `std::max` to match these types. Insead, please use `static_cast<int64_t>(1)` or constructor directly.

yocontra: @msluszniak Fixed

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/DurationPredictor.cpp
```diff
@@ -174,8 +174,9 @@ void DurationPredictor::scaleDurations(Tensor &durations, size_t nTokens,
     float remainder =
         shrinking ? std::ceil(scaled) - scaled : scaled - std::floor(scaled);
 
-    durationsPtr[i] = static_cast<int64_t>(shrinking ? std::ceil(scaled)
-                                                     : std::floor(scaled));
+    durationsPtr[i] = std::max(static_cast<int64_t>(1),
+        static_cast<int64_t>(shrinking ? std::ceil(scaled)
+                                       : std::floor(scaled)));
```
IgorSwat: I think it's better not to do that. The reason is not that simple, but I will try to explain it briefly:

The way Kokoro is being exported to ExecuTorch format, forces us to set some upper bound for total duration. The current bound is 296 duration ticks, and increasing this will increase the inference time as well as memory usage. The reason for that is a well-known ExecuTorch issue with exporting LSTMs with dynamic shapes. So there is no way to omit it without sacrificing some performance.

This is very important, because if we get a total duration higher than 296 and not scale it down to the 296 cap, the model will throw a hard exception of invalid input shape and the app using it will surely crash. If we keep this implementation with the condition that scaling cannot reduce duration to 0, we can accidentally end up with duration scaled to something like 297 or 298, which will cause a crash.

Now, the scalling applied in `scaleDurations()` method is used very rarely - only if the sum of predicted durations is below 16 or exceeds 296. From my experience with the model, I can tell that the first case practically never happens, and the second case happens only for very long sentences around maximum amount of 128 tokens - but those are also very rarely processed due to the way the partitioning algorithm works (it tries to avoid processing long sequences). So the `scaleDurations()` call is rare in itself, and missing a phoneme by calling it is even rarer.

So I strongly believe that allowing to miss single phonemes in very, very rare cases is better than allowing some accidental crash.

I hope the explanation wasn't too boring :)

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
```diff
@@ -54,14 +70,19 @@ Result<std::vector<EValue>> Synthesizer::generate(std::span<const Token> tokens,
   auto voiceRefTensor = make_tensor_ptr({1, constants::kVoiceRefSize},
                                         ref_s.data(), ScalarType::Float);
 
-  // Execute the appropriate "forward_xyz" method, based on given method name
-  auto results = forward(
+  // Select appropriate forward method based on token count
+  auto it = std::find_if(forwardMethods_.begin(), forwardMethods_.end(),
```
msluszniak: we can use `std::ranges::find_if` from `algorithm` here

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -35,41 +35,50 @@ Kokoro::Kokoro(const std::string &lang, const std::string &taggerDataSource,
...
   context_.inputTokensLimit = durationPredictor_.getTokensLimit();
   context_.inputDurationLimit = synthesizer_.getDurationLimit();
+
+  // Cap effective token limit to prevent the Synthesizer's attention from
+  // drifting on longer sequences, which manifests as progressive speed-up
+  // in the generated audio.  Shorter chunks keep timing faithful to the
+  // Duration Predictor's output.
+  static constexpr size_t kSafeTokensLimit = 60;
+  context_.inputTokensLimit =
+      std::min(context_.inputTokensLimit, kSafeTokensLimit);
```
IgorSwat: I would rather not do that. The most destructive thing for the output is cutting the text in the middle of a sentence, which creates a very unnatural-sounding pause. We try to avoid that by cutting the text on end-of-sentence or pause marks (such as ',', '.', '?' or '!'), but reducing the input tokens limit makes it harder to do so.

When implementing the module, I also felt that longer inputs produce better, more natural speech, and the short inputs are the ones which can result in a too quick of a speech. 

We would be  grateful if you could provide a sample input which motivated you to make this change, cause it would make it easier to understand the issue :)

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -35,41 +35,50 @@ Kokoro::Kokoro(const std::string &lang, const std::string &taggerDataSource,
...
+  // Resize voice vector to hold all rows from the file
+  voice_.resize(rows);
+
+  if (!in.read(reinterpret_cast<char *>(voice_.data()->data()), readBytes)) {
     throw RnExecutorchError(
         RnExecutorchErrorCode::FileReadFailed,
-        "[Kokoro::loadSingleVoice]: failed to read voice weights");
+        "[Kokoro::loadVoice]: failed to read voice weights");
   }
 }
```
IgorSwat, line 76: We never really loaded the entire voice array, cause we don't process inputs longer than 128 tokens - and the rest of the voice data which we omit is used specifically for longer inputs.

I guess reading the entire file is also OK, although it's a little bit of memory waste since we don't use the remainder of the voice data.

yocontra:
@IgorSwat Thanks for the feedback - I'm new to this TTS stuff and was mainly just kind of hack and slashing my way through getting this to work on my app. The duration stuff is a bit sketchy, I'll try to revert those and find another way to fix the issues I was seeing.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
yocontra: @IgorSwat This change fixed a bunch of the voice files I was trying to use (the kokoro 82m voice files in their repo, not the .bin files y'all provided). I don't think the input length impacts how many rows of the voice weights are needed but I can look into it more here.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
yocontra: Updated

IgorSwat:
> @IgorSwat Thanks for the feedback - I'm new to this TTS stuff and was mainly just kind of hack and slashing my way through getting this to work on my app. The duration stuff is a bit sketchy, I'll try to revert those and find another way to fix the issues I was seeing.

Yeah, that's absolutely fine. There is a lot of hidden nuances here, mostly because of ExecuTorch limitations, which force us to do some things in a way they normally should not be done, or at least complicating the stuff.

Hovewer, I saw that the [LSTM issue](https://github.com/pytorch/executorch/issues/5176) has been resolved recently, so there is a possibility we will be able to remove the entire duration scalling part very soon.

msluszniak:
@yocontra all of your changes are now on main, thank you once again for the contribution, really appreciate :))

PR #944, opened 2026-03-07
fix: move model construction off the JS thread

yocontra:
## Problem

The `loadModel` template in `RnExecutorchInstaller.h` constructs model objects **synchronously on the JS thread**. For models like Kokoro TTS, the constructor loads `.pte` files, initializes the phonemizer, and reads voice data — blocking the JS thread for several seconds.

This prevents React from rendering any pending state updates (loading spinners, progress indicators, etc.) until construction completes. For example, calling `setIsSynthesizing(true)` followed by `tts.load()` will not show the loading UI for 3-4 seconds because the JS thread is blocked by the synchronous native constructor.

## Solution

Make `loadModel` return a **Promise** and dispatch the heavy model construction to `GlobalThreadPool::detach` — matching the pattern already used by `promiseHostFunction` for inference calls like `generate()`.

### C++ (`RnExecutorchInstaller.h`)

- JSI arguments are still parsed on the JS thread (required for `jsi::Value` access)
- Model construction is dispatched to `GlobalThreadPool::detach`
- The Promise resolves on the JS thread via `callInvoker->invokeAsync` with the host object
- Error handling follows the same pattern as `promiseHostFunction` in `ModelHostObject.h`

### TypeScript

- All `global.load*()` call sites updated to `await` the now-async result
- Global type declarations updated to return `Promise<any>`
- `SpeechToText` already used `await`, so no change needed there

## Breaking Change

This is a breaking change for any consumers calling `global.load*()` directly and expecting a synchronous return value. However:

- The public module APIs (`load()` methods) are already `async` — no user-facing API changes needed
- The `global.load*()` functions are internal/undocumented

## Test Plan

- Verify model loading still works for all model types (TTS, LLM, OCR, etc.)
- Verify that React state updates (loading spinners) render immediately when `load()` is called, rather than being delayed until construction completes
- Verify error handling still propagates correctly from native construction failures


msluszniak:
There is an error in CI build, specifically in VerticalOCR, I guess we just need to mark it as promise like for the other models.

yocontra:
@msluszniak Fixed! I can't see the CI but pretty sure I made the correct changes.

msluszniak:
Cool now all important checks work. Now, we just need to fix lint, I will fix this for you later today and then proceed with the actual review ;)

yocontra:
@msluszniak Fixed the lint errors - my bad

msluszniak:
FYI @yocontra, I will push changes to the API reference so CI will pass. Other then that, I tested changes and everything was correct. :))

chmjkb:
Thanks for another contribution @yocontra 🫶, I will perform some quality checks tomorrow and then I think we're ready to merge :)

PR #946, opened 2026-03-08
Allow passing pre-computed phonemes to Kokoro TTS

yocontra:
Right now if you want to use Kokoro TTS, you have to go through the built-in phonemis G2P pipeline. There's no way around it. This PR adds `generateFromPhonemes` / `streamFromPhonemes` methods that let you skip phonemis and pass your own IPA phoneme strings directly to the synthesis engine.

Why would you want this? A few reasons we've run into:

- phonemis doesn't handle every word well. Libraries like [phonemizer](https://github.com/bootphon/phonemizer) (espeak-ng backend) do better on edge cases, foreign words, etc.
- Custom lexicons. If you have domain-specific pronunciation (game character names, medical terms), you probably want control over the G2P step.
- Server-side G2P. Pre-compute phonemes on a server with a proper NLP pipeline, send them to the device.
- Languages phonemis doesn't cover yet.

## What changed

The existing `generate()` / `stream()` methods now delegate to shared internal helpers (`generateFromPhonemesImpl` / `streamFromPhonemesImpl`). The new public methods call the same helpers but skip the `phonemizer_.process()` step. No behavior change for existing callers.

Changes across layers:
- C++ `Kokoro`: `generateFromPhonemes`, `streamFromPhonemes` + input validation (empty string, invalid UTF-8)
- JSI `ModelHostObject`: exposes new methods
- `TextToSpeechModule`: `forwardFromPhonemes()`, `streamFromPhonemes()` (shared `streamImpl` helper, no copy-paste)
- `useTextToSpeech` hook: same, with shared guard + streaming orchestration
- Types: `TextToSpeechPhonemeInput`, `TextToSpeechStreamingPhonemeInput`, `TextToSpeechStreamingCallbacks`

## Usage

```typescript
const tts = new TextToSpeechModule();
await tts.load(config);

// text path (unchanged -- goes through phonemis)
const audio = await tts.forward("Hello world");

// phoneme path (bypasses phonemis)
const audio = await tts.forwardFromPhonemes("həloʊ wɝːld");

// streaming
for await (const chunk of tts.streamFromPhonemes({ phonemes: "həloʊ wɝːld", speed: 1.0 })) {
  playAudio(chunk);
}
```

## Test plan

- [ ] Existing `generate()` and `stream()` still work (refactor is internal)
- [ ] `generateFromPhonemes()` with known Kokoro IPA strings
- [ ] `streamFromPhonemes()` produces same audio as `stream()` for identical phonemes
- [ ] Multi-byte UTF-8 phoneme characters (ʊ, ɪ, ŋ, etc.)
- [ ] Empty string and invalid UTF-8 rejected with proper error

### DIFF
--- a/.cspell-wordlist.txt
+++ b/.cspell-wordlist.txt
@@ -127,3 +127,4 @@ detr
 metaprogramming
 ktlint
 lefthook
+espeak
--- a/docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
+++ b/docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
@@ -82,16 +82,23 @@ You need more details? Check the following resources:
 
 ## Running the model
 
-The module provides two ways to generate speech:
+The module provides two ways to generate speech using either raw text or pre-generated phonemes:
 
-1.  [**`forward(text, speed)`**](../../06-api-reference/interfaces/TextToSpeechType.md#forward): Generates the complete audio waveform at once. Returns a promise resolving to a `Float32Array`.
+### Using Text
+
+1.  [**`forward({ text, speed })`**](../../06-api-reference/interfaces/TextToSpeechType.md#forward): Generates the complete audio waveform at once. Returns a promise resolving to a `Float32Array`.
+2.  [**`stream({ text, speed, onNext, ... })`**](../../06-api-reference/interfaces/TextToSpeechType.md#stream): An async generator that yields chunks of audio as they are computed. This is ideal for reducing the "time to first audio" for long sentences.
+
+### Using Phonemes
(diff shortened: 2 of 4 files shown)


IgorSwat:
Hey @yocontra !

Thank you very much for your contribution. I reviewed the code and it looks pretty good - and the new API methods indeed make a lot of sense.
I just added some changes to the docs to match the new API.

I can also reveal that we have started to work on big update to both Phonemis phonemizer as well as Kokoro-based core, which will add new languages support and improve the performance of the model significantly. So if you are interested in this module, stay tuned for that :)

IgorSwat (approved):

Issue #948, opened 2026-03-09
Allocated memory is not cleanup after `llm.delete`


lukachi:
### Description

Hi, is there any guidance about how to free allocated memory? i'm using LLMModule and already tried the interrupt and delete methods. Although it says that the model is unloaded (assume it's unloaded by hitting interrupt twice and it throws error that model is not loaded) the memory never drops down. (checking memory with expo dev tool - performance monitor). Once i assumed that monitor is just cached, so i tried to load model one more time, and the app is crashed. (This is kinda reproducible bug, with just snippets from the docs, and few simple buttons and states to simulate chat.)

device: iphone 16 pro
expo: v55
"react-native-executorch": "^0.7.1",
"react-native-worklets": "0.7.2",

### Steps to reproduce

Simple snippet to use LLMModule instead of useLLM to access to `.delete` method.

Add below components to generate and button to call llm.interrupt and llm.delete.

Replace GiftedChat.append in onSend with plain state `prev => [...prev, ...new]`

```
  const [messages, setMessages] = useState<IMessage[]>([])

  const llmId = useMemo(() => v4(), [])

  const { data: llm } = useQuery({
    queryKey: ['llm', 'HAMMER2_1_1_5B_QUANTIZED'],
    queryFn: async () => {
      const llm = new LLMModule({
        tokenCallback: token => console.log(token),
        messageHistoryCallback: messages => console.log(messages),
      })

      llm.configure({
        chatConfig: {
          systemPrompt: `You are helpful assistant. Current time and date: ${new Date().toString()}`,
        },
        toolsConfig: {
          tools: [
            {
              name: 'get_weather',
              description: 'Get/check weather in given location.',
              parameters: {
                type: 'dict',
                properties: {
                  location: {
                    type: 'string',
                    description: 'Location where user wants to check weather',
                  },
                },
                required: ['location'],
              },
            },
          ],
          executeToolCallback: async call => {
            if (call.toolName === 'get_weather') {
              console.log('Checking weather!')
              // perform call to weather API
              // ...
              const mockResults = 'Weather is chinazes!'
              return mockResults
            }
            return null
          },
          displayToolCalls: true,
        },
      })

      await llm.load(HAMMER2_1_1_5B_QUANTIZED, progress => console.log(progress))

      return llm
    },
  })

  const handleGenerate = useCallback(
    async (chat: Message[]) => {
      // Chat completion - returns the generated response
      const response = await llm?.generate(chat)
      return response
    },
[...]

PR #892, opened 2026-03-02
feat: implementation of multimodal runner

NorbertKlockiewicz:
## Description

Adds vision/multimodal support to useLLM: load a VLM by passing capabilities: ['vision'], then use sendMessage(text, { imagePath }) to send messages with images. Under the hood this introduces a pluggable encoder architecture (IEncoder / VisionEncoder), a dedicated MultimodalRunner, and a refactored BaseLLMRunner with cleaner ownership and shared state. Also exposes getVisualTokenCount() JSI method for accurate token counting with images. No changes to the text-only path.

### Introduces a breaking change?

- [ ] Yes
- [x] No

### Type of change

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [x] Documentation update (improves or adds clarity to existing documentation)
- [x] Other (chores, tests, code style improvements etc.)

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Run the `llm` example app, select `multimodal llm` screen. Select an image and prompt the model.

### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation accordingly
- [x] My changes generate no new warnings

PR #954, opened 2026-03-11
fix: fix unload in LLM and ModelHostObject to properly free LLM memory

barhanc:
## Description

Fixes two related memory management bugs in the LLM delete flow:
- **LLM.cpp**. `LLM::unload()` was destroying the runner object but never calling `BaseModel::unload()` and thus not releasing the module memory. `LLM::unload()` now calls `BaseModel::unload()` after resetting the runner.
- **LLMController.ts**.  When `load()` was called on an already-loaded `LLMController`, the previous native module instance was simply overwritten without calling `unload()` first. This caused a memory leak. It now unloads the existing native module before loading a new one.
- **ModelHostObject.h**. Reset JSI external memory pressure to `0` after `model->unload()` so the JS GC is correctly informed that native memory has been freed.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Use the provided screen (you can simply replace the current LLM app screen) to reproduce the bug and verify the fix. It lets you load and unload an LLM and observe memory behavior with `vmmap` / `adb`.

- [x] Run the LLM example app
- [x] Prepare memory monitors
       - **iOS**: `xcrun simctl spawn booted launchctl list | grep llm` and `watch -n 0.1 "vmmap <pid> | tail -12"`
       - **Android**: `watch -n 0.1 "adb shell dumpsys meminfo com.anonymous.llm"`
- [x] Press Load - wait for ready
- [x] Note baseline native memory in profiler
- [x] Press Unload - verify native memory returns to baseline
- [x] Press Load again without restarting - verify no accumulation
- [x] Repeat load/unload 5+ times - confirm no upward drift in native memory and no crash
- [x] Press Load a few times without pressing Unload -- verify no memory accumulation

```typescript
import { useEffect, useRef, useState } from 'react';
import { StyleSheet, Text, TouchableOpacity, View } from 'react-native';
import { LLMModule, HAMMER2_1_1_5B_QUANTIZED } from 'react-native-executorch';

export default function LLMScreen() {
  const [status, setStatus] = useState('idle');
  const llmRef = useRef<LLMModule | null>(null);

  useEffect(() => {
    llmRef.current = new LLMModule();
    return () => {
      try {
        llmRef.current?.interrupt();
        llmRef.current?.delete();
      } catch {}
    };
  }, []);

  const handleLoad = async () => {
    setStatus('loading...');
    try {
      await llmRef.current!.load(HAMMER2_1_1_5B_QUANTIZED, (p) =>
        setStatus(`loading ${(p * 100).toFixed(0)}%`)
      );
      setStatus('ready');
    } catch (e: any) {
      setStatus(`load error: ${e?.message}`);
    }
  };

  const handleUnload = () => {
    try {
      llmRef.current?.interrupt();
      llmRef.current?.delete();
      setStatus('unloaded');
    } catch (e: any) {
      setStatus(`unload error: ${e?.message}`);
    }
  };

  const isLoading = status.startsWith('loading');
  const canLoad = !isLoading;
  const canUnload = status === 'ready';

  return (
    <View style={styles.container}>
      <Text style={styles.status}>{status}</Text>
      <TouchableOpacity
        style={[styles.button, !canLoad && styles.disabled]}
        onPress={handleLoad}
        disabled={!canLoad}
      >
        <Text style={styles.buttonText}>Load</Text>
      </TouchableOpacity>
      <TouchableOpacity
        style={[styles.button, styles.unload, !canUnload && styles.disabled]}
        onPress={handleUnload}
        disabled={!canUnload}
      >
        <Text style={styles.buttonText}>Unload</Text>
      </TouchableOpacity>
    </View>
  );
}

const styles = StyleSheet.create({
  container: { flex: 1, alignItems: 'center', justifyContent: 'center', gap: 12 },
  status: { fontSize: 16, marginBottom: 8 },
  button: { backgroundColor: '#2563eb', paddingHorizontal: 32, paddingVertical: 14, borderRadius: 8 },
  unload: { backgroundColor: '#dc2626' },
  buttonText: { color: '#fff', fontSize: 16, fontWeight: '600' },
  disabled: { opacity: 0.4 },
});
```

### Related issues

#948 

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
@@ -256,7 +256,7 @@ int32_t LLM::getMaxContextLength() const {
 }
 
 void LLM::unload() noexcept {
-  runner.reset(nullptr);
+  runner_.reset(nullptr);
   BaseModel::unload();
 }
 


msluszniak:
Do you believe that similar problems might be in other modules?

NorbertKlockiewicz:
We should wait with this PR until #892 is merged and check if the problem occurs also in this version.

barhanc:
@msluszniak I don't think the problem with LLM not freeing the module memory occurs in other models as they don't override the `BaseModel::unload()` method and thus do `module_.reset(nullptr);`. The JS GC `setExternalMemoryPressure` fix is an ad hoc one though, as I'm not an expert and was guided mostly by [this](https://github.com/facebook/hermes/issues/982) discussion. This one is also more tricky because after the fix with `LLM::unload()` the memory profiler shows that the used memory is being freed properly, but doing the cycle load -> unload multiple times results in sudden crash of the whole app suggesting it is a problem with the GC. The last problem with calling `load()` method multiple times causing memory accumulation, I think, will pretty much be mitigated when we move to constructing models using factory pattern.

msluszniak:
Ok, now we can check if VLMs are also affected.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
```diff
@@ -255,6 +255,9 @@ int32_t LLM::getMaxContextLength() const {
   return runner_->get_max_context_length();
 }
 
-void LLM::unload() noexcept { runner_.reset(nullptr); }
+void LLM::unload() noexcept {
+  runner.reset(nullptr);
```
msluszniak: Norbert's PR changed naming a little bit.
```suggestion
  runner_.reset(nullptr);
```

msluszniak (approved):
I tested both iOS and Android and works correctly, if the VLMs are good, we can push it. Otherwise, re-request review and I will test changes for VLMs as well.

barhanc:
@msluszniak I went through the changes that VLMs introduced and basically the first fix with `BaseModel::unload()` is redundant as now the runner takes unique pointer to the `module` but I guess we can leave the fix if in the future someone decides to change something in the runner again. The other two problems still persist without the proposed fixes applied (especially the GC crash is very easy to trigger: VLM screen -> LLM screen -> VLM screen -> should crash on iOS with `hermes  0x106aaa064 hermes::vm::GCBase::oom(std::__1::error_code)`). I think we can merge, unless you want me to delete the `BaseModel::unload()` fix.

msluszniak:
We can leave it as you mentioned if it does not generates double frees etc. I ok with that :))

PR #982, opened 2026-03-18
chore: Fix bounding box annotation

### LINKED ISSUE #975: Make bounding boxes coherent across demo apps
https://github.com/software-mansion/react-native-executorch/pull/893#issuecomment-4076242516

msluszniak:
## Description

This PR fixes the problem in demo apps where bounding boxes were utilized. For small objects annotation width was set to width of the bounding box since they were in the same view. Fix is to make these two be in relation child - parent components.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

- [ ] Compare result of annotation for apps that utilizes bounding boxes for small detected objects. 

### Screenshots

Now bounding boxes on small objects looks correct (in the #975 was the original result for the same image):
image

### Related issues

Closes #975

### DIFF
--- a/apps/computer-vision/components/BoundingBoxes.tsx
+++ b/apps/computer-vision/components/BoundingBoxes.tsx
@@ -1,25 +1,27 @@
 import React from 'react';
 import { StyleSheet, Text, View } from 'react-native';
-import { Detection } from 'react-native-executorch';
+import { Detection, LabelEnum } from 'react-native-executorch';
 import { labelColor, labelColorBg } from './utils/colors';
 
-interface Props {
-  detections: Detection[];
+interface Props<L extends LabelEnum> {
+  detections: Detection<L>[];
   scaleX: number;
   scaleY: number;
   offsetX: number;
   offsetY: number;
   mirrorLabels?: boolean;
+  containerWidth?: number;
 }
 
-export default function BoundingBoxes({
+export default function BoundingBoxes<L extends LabelEnum>({
   detections,
   scaleX,
   scaleY,
   offsetX,
   offsetY,
   mirrorLabels = false,
-}: Props) {
+  containerWidth,
+}: Props<L>) {
   return (
     <>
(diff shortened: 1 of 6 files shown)


msluszniak:
~TODO: Unify this with with bounding boxes from vision camera so all bounding boxes will be in different colors plus apply these changes to examples from vision camera.~
DONE

apps/computer-vision/components/vision_camera/utils/colors.ts
```diff
@@ -1,41 +1 @@
...
-export function labelColor(label: string): string {
-  const color = CLASS_COLORS[hashLabel(label)]!;
-  return `rgba(${color[0]},${color[1]},${color[2]},1)`;
-}
-
-export function labelColorBg(label: string): string {
-  const color = CLASS_COLORS[hashLabel(label)]!;
-  return `rgba(${color[0]},${color[1]},${color[2]},0.75)`;
-}
+export * from '../../utils/colors';
```
NorbertKlockiewicz: I think we can completely skip this file and import directly from the utils/colors.ts

apps/computer-vision/components/BoundingBoxes.tsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import { StyleSheet, Text, View } from 'react-native';
+import { Detection } from 'react-native-executorch';
+import { labelColor, labelColorBg } from './utils/colors';
+
+interface Props {
+  detections: Detection[];
+  scaleX: number;
```
chmjkb: It's no biggie but since Detection is generic maybe we can make this generic as well and pass labels from the demo apps

msluszniak:
> Can we do something about the case where an object is at the edge of a picture, and the label text is clipped? Image

I'll add ellipsis to make it looking like this:
image

Issue #351, opened 2025-06-02
Add minimal documentation in README.md in each example in `apps` directory


msluszniak:
### Problem description

It was not clear at first glance for me how to navigate through each example, what does formats are expected etc.

### Proposed solution

I think it will be beneficial to add very short description of every example in `README.md`

### Alternative solutions

No alternative solutions

### Benefits to React Native ExecuTorch

Clearer examples, easier navigation through example apps, etc.

PR #985, opened 2026-03-19
chore: Add basic statistics to demo apps

### LINKED ISSUE #959: Add basic statistics to demo app
This should introduce inference times, maybe memory consumption etc.

msluszniak:
## Description

Added inference times, plus for some apps TTFT, tokens/second.
Fixed T2S streaming for unfinished sentences.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

- [ ] Run all apps and check if basic statistics are correctly displayed
- [x] Check that text without end of sequence sign is processed correctly in text to speech example

### Related issues

Closes #959 

### Checklist

### DIFF
--- a/apps/computer-vision/components/ModelPicker.tsx
+++ b/apps/computer-vision/components/ModelPicker.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   View,
   StyleSheet,
@@ -29,6 +29,10 @@ export function ModelPicker<T>({
 }: Props<T>) {
   const [open, setOpen] = useState(false);
   const selected = models.find((m) => m.value === selectedModel);
+
+  useEffect(() => {
+    if (disabled) setOpen(false);
+  }, [disabled]);
 
   return (
     <View style={styles.container}>


apps/computer-vision/app/text_to_image/index.tsx
```diff
@@ -124,6 +132,9 @@ export default function TextToImageScreen() {
           </TouchableOpacity>
         </View>
 
+        {/* Added StatsBar here, just above the input row */}
```
barhanc:
```suggestion
```

barhanc:
Instance segmentation app is missing the inference timing.

msluszniak:
> Instance segmentation app is missing the inference timing.

And model listing as well. It's because instance segmentation was merged after the PR with demo app fixes. I'll add these to instance segmentation. Good catch.

DONE

apps/text-embeddings/app/clip-embeddings/index.tsx
barhanc, line 1: I think we should somehow differentiate reporting embedding times for images and text.

apps/text-embeddings/app/clip-embeddings/index.tsx
```diff
@@ -123,9 +128,12 @@ function ClipEmbeddingsScreen() {
       return;
 
     try {
+      const start = Date.now();
+      // Array.from to get numbers[]
```
barhanc:
```suggestion
```
I don't think this comment applies anymore.

apps/llm/hooks/useLLMStats.ts
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    if (
+      !isGenerating &&
+      sendTimeRef.current !== null &&
+      firstTokenTimeRef.current !== null
+    ) {
+      const endTime = Date.now();
+      const ttft = firstTokenTimeRef.current - sendTimeRef.current;
+      const totalTime = (endTime - firstTokenTimeRef.current) / 1000;
+      const totalTokens = Math.round(lastResponseRef.current.length / 4);
```
barhanc: I think this heuristic to compute the number tokens requires some comment (perhaps even some info in the app screen).

barhanc:
Additionally there are no inference timings for other speech models, but I understand it would be difficult to properly measure the inference time there as they operate in streaming mode. Maybe @IgorSwat has some idea how this could be done.

msluszniak:
@barhanc about streaming tasks, yeah I eventually decided that I don't have good idea how to present these times

IgorSwat:
> Additionally there are no inference timings for other speech models, but I understand it would be difficult to properly measure the inference time there as they operate in streaming mode. Maybe @IgorSwat has some idea how this could be done.

We could measure the average time of executing transcribe() inside the streaming loop - the question is whether we want to make a breaking change just for that.

Same goes for Text to Speech - measuring it's inference time is possible, but would require digging into TypeScript API at very least.

apps/llm/hooks/useLLMStats.ts
msluszniak: I will add information about this together with this solving this issue: https://github.com/software-mansion/react-native-executorch/issues/351

NorbertKlockiewicz: We already include information about number of generated tokens in return type so let's just use it.

apps/computer-vision/app/instance_segmentation/index.tsx
```diff
@@ -60,12 +84,14 @@ export default function InstanceSegmentationScreen() {
       height: image.height ?? 0,
     });
     setInstances([]);
+    setInferenceTime(null);
```
NorbertKlockiewicz, line 87: This is missing in the rest of the computer vision screens, so once we switch the model in those the stale inference time still will be visible.

apps/computer-vision/app/text_to_image/index.tsx
barhanc, line 1: The app can still crash if one starts generating without closing the model picker and then tries to switch the model. Additionally it is possible to trigger `Error: mutex lock failed: Invalid argument` if in the above, one interrupts the model and then quickly tries to switch the model.

msluszniak: I cannot handle this correctly. If you have an idea of to fix this in demo app, feel free to propose changes :)

Issue #612, opened 2025-09-18
Model constants grouped by type


a-szymanska:
### Problem description

Currently, the model constants don’t give much intuition about what each model is for, and so users need to reference documentation to know which models are available and what they’re used for. This problem also extends to Hugging Face, where models are listed ungrouped.

### Proposed solution

- Prefix names with model categories - maybe similar to how it is done in style transfer.
- On Hugging Face, create collections that group models by the hook/module they are designed for.
Then in the hook docs, we could replace links to specific models with links to the related collection, so users can easily browse all supported models for that hook.

### Alternative solutions

Add JSDoc to model constants.

### Benefits to React Native ExecuTorch

Users would be able to quickly identify what a model does without looking up the docs. This would reduce confusion and improve developer experience.

msluszniak:
I don't think it is possible to create such code that will so intuitive here, so user could refer to a model without even looking once into documentation. The PR #748 adds organised section of models for each task, links it clearly in documentation. This way it solves the mentioned problem in a different way. That's why I will mark it for resolving by the PR.

msluszniak:
The is an idea to not only suggest list of models from documentation, but also from the level of language server (e.g. tabbing in VSC). The proposed solution might look like this:

```typescript
import { MODELS_REGISTRY } from "react-native-executorch;"

const llm = useLLM({ MODELS_REGISTRY.LLM.SOME_MODEL })
```

mkopcins:
The final approach would be as follows:
1. internally we maintain model map will following structure:
```
const MODEL_MAP = {
  LLM: {
    LLAMA3_2_1B: {
      'base': {
        'xnnpack': LLAMA3_2_1B
      },
      'qlora': {
        'xnnpack': LLAMA3_2_1B_QLORA
      },
      'spinquant': LLAMA3_2_1B_SPINQUANT
    },
    QWEN2_5_0_5B: {
      'base': {
        'xnnpack': QWEN2_5_0_5B
      },
      '8dw4a': {
        'xnnpack': QWEN2_5_0_5B_QUANTIZED
      }
    },
    ...
  }
};
```
2.  we export public model registry
```
export const MODEL_REGISTRY = {
  LLM: {
    LLAMA3_2_1B: modelAccessor(
      LLAMA3_2_1B,
      ({
        quant = 'base',
        backend = 'xnnpack',
      }: {
        quant?: 'base' | 'qlora' | 'spinquant';
        backend?: 'xnnpack';
[...]

PR #990, opened 2026-03-23
feat: added downloads tracking separate from huggingface

mkopcins:
## Description

Added separate request on model download to track downloads on our managed service. This would allow us to track things like geolocation and have insights on the entire timeline without the need to pay for huggingface premium.

Set up so that it fails silently.

Addtionally added model registry named `MODEL_REGISTRY` for better model discoverability.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Checklist


packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts
```diff
@@ -193,6 +194,49 @@ export namespace ResourceFetcherUtils {
...
+      fetch(DOWNLOAD_EVENT_ENDPOINT, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          modelName: getModelNameFromUri(uri),
+          countryCode: getCountryCode(),
+        }),
+      });
+    } catch (e) {
+      console.log(e);
```
msluszniak: Console logs are explicitly prohibited by eslint rules, we should replace it with throwing an error or sth.

msluszniak:
Does this PR closes https://github.com/software-mansion/react-native-executorch/issues/612? If so, add `Closes #612` in PR template.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -976,3 +976,240 @@ export const FSMN_VAD = {
...
+ *
+ * // List all model names
+ * const names = Object.values(MODEL_REGISTRY.ALL_MODELS).map(m => m.modelName);
+ *
+ * // List all speech-to-text models
+ * const sttModels = Object.values(MODEL_REGISTRY.SPEECH_TO_TEXT);
+ * ```
+ * @category Utils
+ */
+export const MODEL_REGISTRY = {
```
chmjkb, line 997: Do you think an additional backend layer here would be useful? For now if a model is exported to CoreML, we silently make the user download it. And the user may wish to run the XNNPack model on both platforms for whatever reason.

chmjkb: I'm fine with this structure to keep it simple for now, but I wonder if it's something we want to add in the future relases.

msluszniak: I think we should rather have models named `LLAMA3_2_3B_XNNPACK` etc. than adding another layer of nesting here.

mkopcins: Just like you said @chmjkb, this is simple for now, I can even remove the type classification (the main reason for this basic registry is that I need it right now to send proper model name to the database) and that might actually be the correct way to go as that would allow us to extend it later without any braking changes/backward compatibility worries.
As for the structure itself, we had a discussion today with Norbert and we didn't reach common ground on the structure so this is still open question and one we should discuss, I think tomorrow in the office is the best option.

mkopcins:
>Does this PR closes https://github.com/software-mansion/react-native-executorch/issues/612? If so, add Closes #612 in PR template. 

Not quite, still open question

packages/react-native-executorch/src/constants/modelUrls.ts
msluszniak: Agreed.

chmjkb: can we jot down the conclusion from our discussion today somewhere? cc @mkopcins 
I guess a good place would be #612.

docs/docs/05-utilities/model-registry.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+### Accessing a model directly
+
+Every model config is exported as a standalone constant:
+
+```typescript
+import { LLAMA3_2_1B } from 'react-native-executorch';
+
+const model = useExecutorchModule(LLAMA3_2_1B);
```
chmjkb: I don't think its a valid example

docs/docs/05-utilities/model-registry.md
```diff
@@ -0,0 +1,222 @@
...
+| Constant                | Model Name            |
+| ----------------------- | --------------------- |
+| `BK_SDM_TINY_VPRED_512` | bk-sdm-tiny-vpred-512 |
+| `BK_SDM_TINY_VPRED_256` | bk-sdm-tiny-vpred-256 |
+
+### Voice Activity Detection
+
+| Constant   | Model Name |
+| ---------- | ---------- |
+| `FSMN_VAD` | fsmn-vad   |
```
chmjkb: I'm not sure we should add this here. Each doc contains its own supported models section, so we're duplicating.

docs/docs/05-utilities/model-registry.md
chmjkb, line 1: Also i think we should link to API reference for the model registry variable

docs/docs/05-utilities/model-registry.md
mkopcins: Yeah, removing this it is also in API Reference anyway

PR #1023, opened 2026-03-27
refactor: resource fetcher

### LINKED ISSUE #807: Remove linked list from resourcefetcher implementation
Remove the linked list implementation inside of ResourceFetcher and replace it with simple array

### LINKED ISSUE #815: Resource fetcher should throw on error instead of returning empty array
In some cases ResourceFetcher can return empty array. Instead it should throw an error

chmjkb:
## Description

 - Extracted shared base class (BaseResourceFetcherClass<TDownload>) into the core package containing the fetch loop, source dispatching, and pause/resume/cancel plumbing — eliminating duplication between expo and bare
  implementations
  - Extracted platform-specific handlers (handlers.ts) for both packages, separating download mechanics from the class structure
  - Rewrote both fetchers as classes extending the base, implementing only platform-specific file system and download operations
  - Replaced the linked-list download pattern with a simple sequential for loop
  - Cancel now throws DownloadInterrupted instead of returning null, giving callers a consistent error-based API

**Note**: this PR still uses the legacy FS as the support for background / resumable downloads is WIP.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. Create a demo app trying to run all the functionalities exposed in Resource Fetchers
2. Fetch function with single and/or multiple resources
3. Cancel long fetching file, also verify fetch throws an error when its canceled
4. Check if a model with downloaded file works
5. Make sure deleting resources work
6. Try running a background download

### Checklist


msluszniak:
Please run `yarn lint --fix` in the root of the package and push changes

packages/bare-resource-fetcher/src/handlers.ts
```diff
@@ -0,0 +1,234 @@
...
+        try {
+          await RNFS.moveFile(cacheFileUri, fileUri);
+        } catch (error) {
+          downloads.delete(source);
+          reject(error);
+          return;
+        }
+
+        downloads.delete(source);
+        ResourceFetcherUtils.triggerHuggingFaceDownloadCounter(uri);
```
mkopcins: remove from here, i want those calls to be strictly inside core package so that they are mandatory.
Also there is a bug when we fire those calls even when file is already downloaded (see `packages/react-native-executorch/src/utils/ResourceFetcher.ts:128`

packages/bare-resource-fetcher/src/handlers.ts
```diff
@@ -0,0 +1,234 @@
...
+export async function handleRemote(
+  uri: string,
+  source: ResourceSource,
+  progressCallback: (progress: number) => void,
+  downloads: Map<ResourceSource, ActiveDownload>
+): Promise<string> {
+  if (downloads.has(source)) {
+    throw new RnExecutorchError(
+      RnExecutorchErrorCode.ResourceFetcherDownloadInProgress,
+      'Already downloading this file'
```
mkopcins, line 244: just a thought, do we need a message? the error seems pretty self-explanaotry

packages/bare-resource-fetcher/src/handlers.ts
```diff
@@ -0,0 +1,234 @@
...
+    return ResourceFetcherUtils.removeFilePrefix(fileUri);
+  }
+
+  await ResourceFetcherUtils.createDirectoryIfNoExists();
+
+  // We need a Promise whose resolution can be triggered from outside this function —
+  // by cancel() or resume() in the fetcher class. A plain async function can't do that,
+  // so we create the Promise manually and store settle/reject in the downloads map.
+  let settle: (path: string) => void = () => {};
+  let reject: (error: unknown) => void = () => {};
```
mkopcins: why this naming? resolve -> settle, but reject->reject

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -18,352 +18,128 @@
...
-    ...sources: ResourceSource[]
-  ) {
-    if (sources.length === 0) {
+    progressCallback: (progress: number) => void
+  ): Promise<string> {
+    return handleRemote(uri, source, progressCallback, this.downloads);
+  }
+
+  protected async pause(source: ResourceSource): Promise<void> {
+    const dl = this.downloads.get(source)!;
```
mkopcins: please not `dl`, use proper name
```suggestion
    const donwloadHandle = this.downloads.get(source)!;
```

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -18,352 +18,128 @@
...
+    if (dl.status === DownloadStatus.PAUSED) {
       throw new RnExecutorchError(
-        RnExecutorchErrorCode.InvalidUserInput,
-        'Empty list given as an argument'
+        RnExecutorchErrorCode.ResourceFetcherAlreadyPaused,
```
mkopcins, line 78: I am not a fan of this structure, the error type is already quite long and descriptive

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -18,352 +18,128 @@
...
-      return this.singleFetch(nextSource);
-    }
-    sourceExtended.callback!(1);
-    return sourceExtended.results;
-  },
-
-  async pause(source: ResourceSource) {
-    const resource = this.downloads.get(source);
-    if (!resource) {
+    if (!dl.task) {
```
mkopcins: wouldnt a better check be actually `Platform.OS === 'android'`

packages/bare-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -18,352 +18,128 @@
...
+    dl.task.resume();
+  }
+
+  protected async cancel(source: ResourceSource): Promise<void> {
+    const dl = this.downloads.get(source)!;
+
+    if (dl.task) {
+      // iOS: background downloader cancel
+      dl.task.stop();
+    } else if (dl.jobId !== undefined) {
```
mkopcins: Two things here
1. why are we checking with (dl.task) and the other time with (dl.jobId !==...) instead of (!dl.jobId)
2. why are we using those fields (task/jobId) as a proxy for checking what platform we are using, especially since you are then writing comments explaining the platforms

mkopcins (changes requested):

packages/bare-resource-fetcher/src/ResourceFetcher.ts
chmjkb: regarding the first point, jobId can be 0 which is falsy, so it would incorrectly tread a valid job with id 0 as missing. Task is an object ref, so !downloadHandle.task is safe.

chmjkb: and regarding the second point, I could do something like this instead, but I'm not sure if its much cleaner though:
```
    if (Platform.OS === 'ios') {
      if (!downloadHandle.task) {
        throw new RnExecutorchError(
          RnExecutorchErrorCode.ResourceFetcherDownloadFailed,
          'Download task is missing. This should not happen on iOS.'
        );
      }
      downloadHandle.task.stop();
    } else if (Platform.OS === 'android') {
      if (downloadHandle.jobId === undefined) {
        throw new RnExecutorchError(
          RnExecutorchErrorCode.ResourceFetcherDownloadFailed,
          'Download job ID is missing. This should not happen on Android.'
        );
      }
      RNFS.stopDownload(downloadHandle.jobId);
      if (await ResourceFetcherUtils.checkFileExists(downloadHandle.cacheFileUri)) {
        await RNFS.unlink(downloadHandle.cacheFileUri);
      }
    }
```

packages/bare-resource-fetcher/src/handlers.ts
```diff
@@ -0,0 +1,236 @@
...
+    await RNFS.copyFile(uri, fileUri);
+  }
+  return ResourceFetcherUtils.removeFilePrefix(fileUri);
+}
+
+// uri and source are separate parameters because for asset sources (dev mode),
+// source is the require() number the user holds (used as the downloads map key),
+// while uri is the resolved HTTP URL needed for the actual download.
+// For plain remote strings they are the same value.
+export async function handleRemote(
```
msluszniak, line 235: This function is huge and should be split into smaller subparts.

mkopcins (approved):

msluszniak (approved):
I performed detailed testing, code available on branch: https://github.com/msluszniak/react-native-executorch/tree/test/pr-1023-resource-fetcher

Everything worked correctly.

PR #1056, opened 2026-04-06
fix: add mutex to VoiceActivityDetection to prevent race between generate() and unload()

### LINKED ISSUE #1055: Android SIGILL crash in VAD unload path
### Description

We are seeing production Android crashes when using the built-in VAD model from `react-native-executorch`.

The dominant Android crash family is:

- `NDK crash detected with signal: SIGILL`

From symbolicated stacks, the relevant frames are:

```text
executorch::extension::module::ModuleD2
executorch::extension::module::ModuleD0
rnexecutorch::ModelHostObject<voice_activity_detection::VoiceActivityDetection>::unload

Suspected cause:
`VoiceActivityDetection` appears to use `BaseModel`, where `unload()` resets the module without synchronization, while model work is dispatched asynchronously from `ModelHostObject`. `VisionModel` already has unload/inference locking, but VAD does not. This suggests a race between in-flight VAD inference and unload, and possibly overlapping VAD `generate()` calls as well.

Suggested direction:
Please check whether VAD should get the same unload/inference synchronization pattern used by `VisionModel`, or otherwise guard `VoiceActivityDetection::generate()` and `unload()` against concurrent access.

### Steps to reproduce

I do not have a minimal public repro yet, but the production usage pattern is:

1. Mount useVAD({ model: FSMN_VAD })
2. Keep the model alive for the app/session lifetime
3. Continuously feed audio chunks into forward() from microphone callbacks during a study flow
4. After some time in session, native crashes occur either: during VAD unload on Android (SIGILL) or during VAD generate (SIGSEGV)

This is not tied to a single handset model. We have seen it on multiple physical Android devices.

### Snack or a link to a repository

_No response_

### React Native Executorch version

0.7.2

### React Native version

0.83.4

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Observed on multiple devices, for example:  Motorola motorola one action Infinix X626 TECNO KM4

### AI model

Built-in VAD model: FSMN_VAD via useVAD

### Performance logs

_No response_

### Acknowledgements

Yes

### PR DESCRIPTION
**Pull request by radko93**
## Description

Add thread-safety to `VoiceActivityDetection` using the same mutex pattern already established by `VisionModel`.

Without this, `BaseModel::unload()` can destroy `module_` on the JS thread while `generate()` is still calling `forward()` on a worker thread, causing `SIGILL` / `SIGSEGV` crashes.

- [x] No

### Type of change

### Tested on

- [x] Android

### Testing instructions

Use `useVAD({ model: FSMN_VAD })` with continuous audio input and trigger rapid mount/unmount cycles. Confirm no crashes or deadlocks.

### Related issues

Fixes #1055

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.h
+++ b/packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.h
@@ -24,14 +24,6 @@ public:
   [[nodiscard("Registered non-void function")]] std::vector<types::Segment>
   generate(std::span<float> waveform) const;
 
-  /**
-   * @brief Thread-safe unload that waits for any in-flight inference to
-   * complete.
-   *
-   * Mirrors VisionModel::unload(). Without this, BaseModel::unload() can
-   * destroy module_ while generate() is still calling forward() on a worker
-   * thread, causing SIGILL / SIGSEGV crashes.
-   */
   void unload() noexcept;
 
 private:

### CODE REVIEW

**Comment by barhanc** [comment]
The problem with race condition occurs also in other models that do not inherit from `VisionModel` and thus we probably want to fix it on the `BaseModel.cpp` implementation level. Could you check if fixes introduced in branch `@bh/issue1055` solve your problem?

**Comment by radko93** [comment]
@barhanc unfortunately I would need to deploy that to production to really find out. It is possible if it would be safe but it's complicated as you can understand

File: packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.h [hunk 1]
```diff
@@ -23,7 +24,19 @@ class VoiceActivityDetection : public BaseModel {
...
   generate(std::span<float> waveform) const;
 
+  /**
+   * @brief Thread-safe unload that waits for any in-flight inference to
+   * complete.
+   *
+   * Mirrors VisionModel::unload(). Without this, BaseModel::unload() can
+   * destroy module_ while generate() is still calling forward() on a worker
+   * thread, causing SIGILL / SIGSEGV crashes.
+   */
```
barhanc:
```suggestion
```

PR #1067, opened 2026-04-09
fix: graceful degradation when native libraries are unavailable (Android)

radko93:
## Description

On 32-bit Android devices (or devices with 64-bit CPUs running 32-bit userspace), the `arm64-v8a` native `.so` files are not found. `ETInstaller.init` throws `RuntimeException` wrapping `UnsatisfiedLinkError`, crashing the app at startup with no consumer-side escape hatch.

This PR adds a fallback TurboModule (`ETInstallerUnavailable`) whose `install()` returns `false`. `RnExecutorchPackage` returns it only when the native library fails to load, so JS sees a real linked module but JSI bindings are never injected. This preserves the existing linking-error Proxy for genuinely mislinked installs on supported devices.

**Unsupported ABI**: real module exists → `install()` returns `false` → globals not set → `isAvailable` is `false`
**Supported but mislinked**: module is `null` → existing Proxy throws linking error (unchanged)

No public API break; preserves existing mislink failure behavior. `isAvailable` is a new additive export.

- [x] No

### Type of change

### Tested on

- [x] Android

### Testing instructions

1. Build and run on a 32-bit Android device (e.g. Galaxy A13, Moto G Play 2023)
2. App should start without crashing
3. `isAvailable` should be `false`
4. Verify on a supported 64-bit device that everything works as before

### Related issues

Fixes #1065

### DIFF
--- a/packages/react-native-executorch/android/src/main/java/com/swmansion/rnexecutorch/ETInstallerUnavailable.kt
+++ b/packages/react-native-executorch/android/src/main/java/com/swmansion/rnexecutorch/ETInstallerUnavailable.kt
@@ -1,0 +1,27 @@
+package com.swmansion.rnexecutorch
+
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactMethod
+import com.facebook.react.common.annotations.FrameworkAPI
+import com.facebook.react.module.annotations.ReactModule
+
+/**
+ * Fallback TurboModule returned when native ExecuTorch libraries cannot be
+ * loaded (e.g. 32-bit Android devices where only arm64-v8a binaries are
+ * shipped). Extends the same spec as ETInstaller so JS sees a real linked
+ * module, but install() returns false to signal unavailability.
+ */
+@OptIn(FrameworkAPI::class)
+@ReactModule(name = ETInstallerUnavailable.NAME)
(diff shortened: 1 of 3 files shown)


msluszniak:
Ok, it took us some time to find a suitable device to test these changes, but we will be testing in a while :))

msluszniak:
Hmm, I turned out that it will be a challenge for us. The motorola that we had is from 2016 with android 7 😅. For now, we don't have any device that has 32-bit arch to test it :/

radko93:
@msluszniak we don't have a device like that either. But a similar solution made crashes disappear at scale. There are also some devices that are 64bit but apparently run or on 32 bit. Try to Google, especially Samsung A13

msluszniak:
Ok, I think the same applies to Motorola G04 that we have. We will test this on ~Monday~ Tuesday at the latest :))

msluszniak:
Ok, I can confirm that without fix I get:
* Fatal signal 6 (SIGABRT) — the app crashes
* UnsatisfiedLinkError: dlopen failed: library "libexecutorch.so" not found
* RuntimeException: Could not load native module ETInstaller
* And eventually app crashed

with fix the `UnsatisfiedLinkError` is triggered aka libexecutorch.so not found. The app didn't crash.

msluszniak (approved):

PR #1080, opened 2026-04-17
feat: add Fishjam background blur integration

### LINKED ISSUE #949: React Native WebRTC integration
## Summary

Add support for running ExecuTorch vision models (e.g. object detection) on local WebRTC camera frames in real-time, using `react-native-webrtc`'s existing `ProcessorProvider` plugin system. This allows us to integrate with tools such as [Fishjam](https://fishjam-dev.github.io/fishjam-docs/next/tutorials/react-native).

## Motivation

Users building WebRTC video call apps often want to run on-device ML on the local camera feed — object detection, pose estimation, segmentation — before or alongside sending the stream to a peer. There is currently no way to do this with `react-native-executorch`.

## Approach

`react-native-webrtc` ships a plugin system for intercepting captured frames on both platforms:

**iOS** — implement `VideoFrameProcessorDelegate`:
```objc
- (RTCVideoFrame *)capturer:(RTCVideoCapturer *)capturer
       didCaptureVideoFrame:(RTCVideoFrame *)frame;
```

**Android** — implement `VideoFrameProcessor`:
```java
VideoFrame process(VideoFrame frame, SurfaceTextureHelper textureHelper);
```

On iOS, `RTCVideoFrame` wraps a `CVPixelBufferRef` — the exact same type our existing `FrameExtractor.cpp` already handles (same zero-copy path as VisionCamera). On Android, call `toI420()` and convert to RGB for the existing `generateFromPixels` path.

## References

- [`VideoFrameProcessor.h`](https://github.com/react-native-webrtc/react-native-webrtc/blob/72f9dfd19dfcf5ebc3d4bc9fe3b72ad6560ac7af/ios/RCTWebRTC/videoEffects/VideoFrameProcessor.h)
- [`VideoFrameProcessor.java`](https://github.com/react-native-webrtc/react-native-webrtc/blob/72f9dfd19dfcf5ebc3d4bc9fe3b72ad6560ac7af/android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoFrameProcessor.java)
- [`ProcessorProvider.h`](https://github.com/react-native-webrtc/react-native-webrtc/blob/72f9dfd19dfcf5ebc3d4bc9fe3b72ad6560ac7af/ios/RCTWebRTC/videoEffects/ProcessorProvider.h)

chmjkb:
## Description

 Adds react-native-executorch-webrtc package for real-time background blur in Fishjam WebRTC video calls using on-device ExecuTorch segmentation models.

  Key features:
  - useBackgroundBlur hook providing blurMiddleware for Fishjam's useCamera
  - blur compositing (OpenGL ES on Android, Core Image on iOS)
  - Morphological mask cleaning + EMA temporal smoothing (C++/OpenCV)

  Architecture:
  - Reuses BaseSemanticSegmentation from react-native-executorch for inference
  - Registers custom VideoFrameProcessor with Fishjam's WebRTC pipeline
  - All heavy processing in native (C++/Objective-C++) for performance

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

You'll need to setup your fishjam account, and verify this example works properly:

```import { useEffect, useState } from 'react';
import { StatusBar } from 'expo-status-bar';
import { StyleSheet, Text, View, TouchableOpacity } from 'react-native';
import {
  FishjamProvider,
  useConnection,
  useCamera,
  useInitializeDevices,
  useSandbox,
  RTCView,
} from '@fishjam-cloud/react-native-client';
import { useBackgroundBlur } from 'react-native-executorch-webrtc';
import { ResourceFetcher, SELFIE_SEGMENTATION, initExecutorch } from 'react-native-executorch';
import { ExpoResourceFetcher } from 'react-native-executorch-expo-resource-fetcher';

initExecutorch({ resourceFetcher: ExpoResourceFetcher });

const FISHJAM_ID = 'your-id';

function CameraScreen() {
  const { initializeDevices } = useInitializeDevices();
  const { cameraStream, cameraDevices, currentCamera, selectCamera, setCameraTrackMiddleware } = useCamera();
  const { joinRoom, leaveRoom, peerStatus } = useConnection();
  const { getSandboxPeerToken } = useSandbox();
  const [isJoining, setIsJoining] = useState(false);
  const [modelPath, setModelPath] = useState<string | null>(null);
  const [blurEnabled, setBlurEnabled] = useState(false);

  const { blurMiddleware } = useBackgroundBlur({
    modelUri: modelPath || '',
    blurRadius: 12,
  });

  // Download the selfie segmentation model
  useEffect(() => {
    const downloadModel = async () => {
      try {
        const paths = await ResourceFetcher.fetch(
          (progress) => setDownloadProgress(progress),
          SELFIE_SEGMENTATION.modelSource
        );
        if (paths?.[0]) {
          setModelPath(paths[0]);
        }
      } catch (error) {
        console.error('Failed to download model:', error);
      }
    };
    downloadModel();
  }, []);

  const handleFlipCamera = async () => {
    if (cameraDevices.length < 2) return;
    const currentIndex = cameraDevices.findIndex(
      (device) => device.deviceId === currentCamera?.deviceId
    );
    const nextIndex = (currentIndex + 1) % cameraDevices.length;
    await selectCamera(cameraDevices[nextIndex].deviceId);
  };

  const handleToggleBlur = async () => {
    if (!modelPath) return;
    if (blurEnabled) {
      await setCameraTrackMiddleware(null);
      setBlurEnabled(false);
    } else {
      await setCameraTrackMiddleware(blurMiddleware);
      setBlurEnabled(true);
    }
  };

  useEffect(() => {
    initializeDevices();
  }, []);

  const handleJoinRoom = async () => {
    setIsJoining(true);
    try {
      const roomName = 'demo-room';
      const peerName = `user_${Date.now()}`;
      const peerToken = await getSandboxPeerToken(roomName, peerName);
      await joinRoom({ peerToken });
    } catch (error) {
      console.error('Failed to join room:', error);
    } finally {
      setIsJoining(false);
    }
  };

  return (
    <View style={styles.container}>
      <StatusBar style="light" />

      <View style={styles.videoContainer}>
        {cameraStream ? (
          <RTCView
            mediaStream={cameraStream}
            style={styles.video}
            objectFit="cover"
            mirror={true}
          />
        ) : (
          <View style={styles.placeholder}>
            <Text style={styles.placeholderText}>Starting camera...</Text>
          </View>
        )}
      </View>

      <View style={styles.controls}>
        <Text style={styles.status}>Status: {peerStatus}</Text>
        {downloadProgress > 0 && downloadProgress < 1 && (
          <Text style={styles.status}>
            Downloading model: {(downloadProgress * 100).toFixed(0)}%
          </Text>
        )}
        <View style={styles.buttons}>
          <TouchableOpacity style={styles.flipButton} onPress={handleFlipCamera}>
            <Text style={styles.buttonText}>Flip</Text>
          </TouchableOpacity>
          <TouchableOpacity
            style={[styles.blurButton, blurEnabled && styles.blurButtonActive, !modelPath && styles.buttonDisabled]}
            onPress={handleToggleBlur}
            disabled={!modelPath}
          >
            <Text style={styles.buttonText}>{blurEnabled ? 'Blur On' : 'Blur'}</Text>
          </TouchableOpacity>
          {peerStatus === 'connected' ? (
            <TouchableOpacity style={styles.leaveButton} onPress={leaveRoom}>
              <Text style={styles.buttonText}>Leave Room</Text>
            </TouchableOpacity>
          ) : (
            <TouchableOpacity
              style={[styles.button, isJoining && styles.buttonDisabled]}
              onPress={handleJoinRoom}
              disabled={isJoining}
            >
              <Text style={styles.buttonText}>
                {isJoining ? 'Joining...' : 'Join Room'}
              </Text>
            </TouchableOpacity>
          )}
        </View>
      </View>
    </View>
  );
}

export default function App() {
  return (
    <FishjamProvider fishjamId={FISHJAM_ID}>
      <CameraScreen />
    </FishjamProvider>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: '#000',
  },
  videoContainer: {
    flex: 1,
  },
  video: {
    flex: 1,
  },
  placeholder: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: '#1a1a1a',
  },
  placeholderText: {
    color: '#666',
    fontSize: 18,
  },
  controls: {
    padding: 20,
    paddingBottom: 40,
    alignItems: 'center',
    gap: 12,
  },
  status: {
    color: '#888',
    fontSize: 14,
  },
  buttons: {
    flexDirection: 'row',
    gap: 12,
  },
  button: {
    backgroundColor: '#007AFF',
    paddingHorizontal: 32,
    paddingVertical: 14,
    borderRadius: 12,
  },
  flipButton: {
    backgroundColor: '#333',
    paddingHorizontal: 24,
    paddingVertical: 14,
    borderRadius: 12,
  },
  leaveButton: {
    backgroundColor: '#FF3B30',
    paddingHorizontal: 32,
    paddingVertical: 14,
    borderRadius: 12,
  },
  blurButton: {
    backgroundColor: '#5856D6',
    paddingHorizontal: 24,
    paddingVertical: 14,
    borderRadius: 12,
  },
  blurButtonActive: {
    backgroundColor: '#34C759',
  },
  buttonDisabled: {
    backgroundColor: '#444',
  },
  buttonText: {
    color: '#fff',
    fontSize: 16,
    fontWeight: '600',
  },
});
```

### Checklist


msluszniak (commented):
Image

apps/computer-vision/app.json
```diff
@@ -26,7 +27,12 @@
         "backgroundColor": "#ffffff"
       },
       "package": "com.anonymous.computervision",
-      "permissions": ["android.permission.CAMERA"]
+      "permissions": [
+        "android.permission.CAMERA",
+        "android.permission.INTERNET",
+        "android.permission.RECORD_AUDIO",
+        "android.permission.ACCESS_NETWORK_STATE"
+      ]
```
msluszniak: Why we have these changes?

packages/react-native-executorch-webrtc/android/src/main/cpp/FrameProcessorBridge.cpp
```diff
@@ -0,0 +1,216 @@
...
+      g_modelHeight = inputShapes[0][inputShapes[0].size() - 2];
+      g_modelWidth = inputShapes[0][inputShapes[0].size() - 1];
+    }
+
+    LOGD("Model input size: %dx%d", g_modelWidth, g_modelHeight);
+
+    // Pre-allocate buffers
+    g_resizedRgb = cv::Mat(g_modelHeight, g_modelWidth, CV_8UC3);
+
+    g_modelLoaded = true;
```
msluszniak: **Thread safety on the model lifecycle**

Three issues stacked on this line:

1. **Publication ordering** — plain `bool g_modelLoaded` is written here without ordering vs the prior writes to `g_segmentation` (L69) and dims (L75–76). The frame thread reading at L105 can observe `g_modelLoaded == true` with `g_segmentation == nullptr` → null deref. Data race per the C++ memory model regardless.

2. **TOCTOU vs `unloadModel`** — L105 checks the flag, L143 dereferences `g_segmentation` for tens of ms. `unloadModel` runs from `release()` on a different thread and `reset()`s the `unique_ptr` → UAF.

3. **Process-global state** — these are file-scope globals; two `ExecutorchFrameProcessor` instances clobber each other. The per-instance Kotlin `modelLoaded` flag misleadingly suggests isolation.

Suggested fix: hold a `std::shared_ptr<BaseSemanticSegmentation>` and snapshot it locally under a mutex at the start of `runSegmentation`, so unload just drops a refcount.

The iOS side (`ExecutorchFrameProcessor.mm`) has the same bugs and they're more reachable: load is explicitly `dispatch_async`'d to a background queue (L80), `unloadModel` is on the public API (L112), and the whole class is a `+sharedInstance` singleton. Worth converging both platforms on the same synchronization model.

Drive-by in this file: `g_resizedRgb` (L33) is allocated but never used; `g_lastDebugLogTime` (L36) is unused.

packages/react-native-executorch-webrtc/android/src/main/cpp/FrameProcessorBridge.cpp
```diff
@@ -0,0 +1,216 @@
...
+    cv::Mat rgbRotated;
+    if (rotation == 90) {
+      cv::rotate(rgb, rgbRotated, cv::ROTATE_90_CLOCKWISE);
+    } else if (rotation == 180) {
+      cv::rotate(rgb, rgbRotated, cv::ROTATE_180);
+    } else if (rotation == 270) {
+      cv::rotate(rgb, rgbRotated, cv::ROTATE_90_COUNTERCLOCKWISE);
+    } else {
+      rgbRotated = rgb;
+    }
```
msluszniak: Maybe we can refactor this into separate function?

packages/react-native-executorch-webrtc/android/src/main/cpp/FrameProcessorBridge.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+// Model input dimensions (dynamically read from model)
+static int g_modelHeight = 256;
+static int g_modelWidth = 256;
+
+// Pre-allocated buffers
+static cv::Mat g_resizedRgb;
+
+// Debug logging rate limiter
+static long long g_lastDebugLogTime = 0;
```
msluszniak: Plain `int` and `long long` types, may use `int32_t` and `int64_t` instead.

msluszniak (commented):
The functionality worked on both iOS and android. However the level of blur is not consistent across platforms:

iOS:
Image

Android:
Image

apps/computer-vision/app.json
chmjkb: Can't remember if all of those are needed, but I had some issues with webrtc without those.

msluszniak: Wowow, wait a minute, what computer vision app.json file has to do with the webrtc, completely separate package?

packages/react-native-executorch-webrtc/android/src/main/cpp/FrameProcessorBridge.cpp
```diff
@@ -0,0 +1,222 @@
...
+std::shared_ptr<LoadedModel> g_model; // null when unloaded
+
+std::shared_ptr<LoadedModel> snapshotModel() {
+  std::lock_guard<std::mutex> lock(g_modelMutex);
+  return g_model;
+}
+
+// Mask post-processing state (EMA temporal smoothing). Touched only on the
+// capture thread; unload resets it.
+cv::Mat g_previousMask;
```
msluszniak: The `LoadedModel`+mutex+`shared_ptr` snapshot doesn't cover `g_previousMask` / `g_hasHistory`. Capture thread reads/writes them at L158–165; `unloadModel` resets them at L216–217 from a different thread — `cv::Mat` isn't thread-safe for concurrent release/assignment, so an in-flight frame races with `release()`. The L41–42 "Touched only on the capture thread" comment contradicts itself.

Fix: hoist to a per-instance native handle (`loadModel` returns a `jlong`, others take it back) with the EMA state inside `LoadedModel` so the existing snapshot covers it.

packages/react-native-executorch-webrtc/ios/ExecutorchFrameProcessor.mm
```diff
@@ -0,0 +1,565 @@
...
+      }
+      NSLog(@"[ExecutorchFrameProcessor] Model loaded!");
+    } @catch (NSException *exception) {
+      NSLog(@"[ExecutorchFrameProcessor] Failed to load model: %@",
+            exception.reason);
+    }
+  });
+}
+
+- (void)setBlurRadius:(float)blurRadius {
```
msluszniak, line 98: Several cross-thread accesses without synchronization, all rooted in `+sharedInstance`:
- `_blurRadius` (L122 vs L242): plain ivar across JS/capture threads. `std::atomic<float>` is the one-liner. Android handles this via `@Volatile pendingBlurRadius`.
- `_isProcessing` (L165–177): plain BOOL, safe only because WebRTC happens to deliver frames serially.
- `_lastProcessedFrame`: raw ivar — concurrent ARC stores from capture queue (L181) vs unload nil-out (L136) can over-release.
- `_outputPool`: released in unload (L139) while capture queue may be reading it. UAF window.
- `_previousMask` — see Android comment.

All collapse when state moves off `+sharedInstance` onto per-instance ivars with a serial queue (or `os_unfair_lock`). The singleton also blocks multi-track / camera-switch.

packages/react-native-executorch-webrtc/react-native-executorch-webrtc.podspec
```diff
@@ -0,0 +1,35 @@
...
+
+  s.platforms    = { :ios => "13.0" }
+  s.source       = { :git => package["repository"]["url"], :tag => "#{s.version}" }
+
+  s.source_files = "ios/**/*.{h,m,mm}"
+
+  # react-native-executorch exposes rnexecutorch/* headers via its header_dir.
+  # However, executorch SDK headers and internal headers don't propagate to
+  # dependent pods, so we need to add them here.
+  rne_path = '${PODS_ROOT}/../../node_modules/react-native-executorch'
```
msluszniak: `${PODS_ROOT}/../../node_modules/react-native-executorch` assumes a flat layout. In yarn/pnpm workspaces `node_modules` hoists to the workspace root, so the path resolves to nowhere and the executorch SDK headers aren't found — breaks any workspace consumer, including this repo's own apps.

Two options:
```ruby
rne_path = ['../node_modules/react-native-executorch',
            '../../node_modules/react-native-executorch',
            '../../../node_modules/react-native-executorch']
  .map { |p| File.expand_path(p, __dir__) }
  .find { |p| File.exist?(p) }
```
Or better, expose `third-party/include/` and `common/` from `react-native-executorch.podspec` as `public_header_files`, removing the need for any `HEADER_SEARCH_PATHS` workaround here. The L18–20 comment already admits this is the design issue.

msluszniak (changes requested):
Remaining items:
- EMA state race + process-global model state — line on `FrameProcessorBridge.cpp`.
- iOS sibling races + singleton design — line on `ExecutorchFrameProcessor.mm`.
- Monorepo header path — line on the podspec.

Also — should we add a section about this package to the documentation?

msluszniak:
Also please resolve conflicts. I tested this PR on Android and it worked well. I will be glad if someone test it on iOS as well.

packages/react-native-executorch-webrtc/ios/ExecutorchFrameProcessor.mm
```diff
@@ -0,0 +1,466 @@
...
+    CVPixelBufferPoolRelease(poolToRelease);
+  }
+  NSLog(@"[ExecutorchFrameProcessor] Model unloaded");
+}
+
+#pragma mark - VideoFrameProcessorDelegate
+
+- (RTCVideoFrame *)capturer:(RTCVideoCapturer *)capturer
+       didCaptureVideoFrame:(RTCVideoFrame *)frame {
+  std::lock_guard<std::mutex> lock(_mutex);
```
msluszniak:
```suggestion
  std::scoped_lock lock(_mutex);
```

msluszniak (approved):
If CI will pass clearly then it's ok from my side. Maybe someone else should test it before merging.

Issue #1086, opened 2026-04-21
Integration findings - aggregated improvements


msluszniak:
While integrating LFM2-VL-1.6B into a multilingual Expo app on Android (Samsung Galaxy S26 Ultra) against `react-native-executorch@0.8.3` + `react-native-executorch-expo-resource-fetcher@0.8.0`, a number of papercuts and one real sampler-side bug surfaced. Filing them here as a single aggregated tracking issue. Each item is independently actionable and will be addressed in its own PR.

## 1. `LLMModule.generate()` does not auto-shape multimodal `mediaPath` messages

Native throws `"More images paths provided than '<image>' placeholders in prompt"`.

`LLMController.sendMessage` transforms user messages with a `mediaPath` into `content: [{type:'image'}, {type:'text', text}]` so the chat template emits the image placeholder. `LLMController.generate` does not — it only collects `imagePaths` and renders the template with the original string `content`. Callers using the lower-level `generate()` API must pre-shape the structured array themselves.

**Proposed fix:** apply the same `historyForTemplate` transformation inside `generate()` for any message with a `mediaPath`.

SOLVED by #1089

## 2. `Message.content` is typed `string` but the chat template accepts an array of content parts

Forces users of `generate()` (after fix for point 1 lands, or as a workaround today) to write `as unknown as string` casts. Type and runtime contract are out of sync.

**File:** `src/types/llm.ts` — `content: string`.

**Proposed fix:** make `content` a discriminated union `string | Array<{type:'image'} | {type:'text', text:string}>`. Likely bundles with 1 into one PR.

SOLVED by #1089

## 3. Image-path format requirement for multimodal `generate()` is undocumented and inconsistent

- `LLMModule.generate()` with vision capability throws `"Read image error: invalid argument"` if the path lacks a `file://` prefix.
- `ResourceFetcher.fetch()` docstring says it returns paths *without* `file://`.
- Vision module examples in source comments show `'file:///path/to/image.jpg'`.

A user threading the resource fetcher result into a vision call will hit this.

**Proposed fix:** normalize internally on the multimodal call (accept either form) or document the requirement loudly on `LLMModule.generate`/`sendMessage` and on every vision module.

SOLVED by #1090

## 4. `ResourceFetcher` class docstring promises methods that don't exist on the class

[...]

PR #1087, opened 2026-04-22
docs: clarify ResourceFetcher class docstring

msluszniak:
## Summary

Both the file-level docstring and the `ResourceFetcher` class docstring listed `pauseFetching`, `resumeFetching`, `cancelFetching`, `getFilesTotalSize`, `listDownloadedFiles`, `listDownloadedModels`, and `deleteResources` as if they were available on the class. They aren't — those methods only exist on the concrete adapter implementations (`ExpoResourceFetcher`, `BareResourceFetcher`).

This PR rewrites both docstrings to:
- Describe `ResourceFetcher` as a static-only entry point that delegates to a configured `ResourceFetcherAdapter`.
- Direct readers to the adapter packages when they want storage-management utilities.

Pure docs change — no behavior change.

## Issue

Addresses item 4 of #1086.

## Test plan

- [ ] `yarn typecheck` passes
- [ ] `npx eslint packages/react-native-executorch/src/utils/ResourceFetcher.ts` clean
- [ ] `npx prettier --check packages/react-native-executorch/src/utils/ResourceFetcher.ts` clean
- [ ] No source changes — only the file-level and class-level JSDoc blocks were modified

### DIFF
--- a/packages/react-native-executorch/src/utils/ResourceFetcher.ts
+++ b/packages/react-native-executorch/src/utils/ResourceFetcher.ts
@@ -1,31 +1,27 @@
 /**
  * Resource Fetcher
  *
- * Provides an interface for downloading files (via `ResourceFetcher.fetch()`)
+ * Static entry point that delegates resource operations to a configured
+ * `ResourceFetcherAdapter`. The adapter is registered for you when you call
+ * `initExecutorch({ resourceFetcher })` from one of the platform packages:
  *
- * Key functionality:
- * - Download control: pause, resume, and cancel operations through:
- *   - Single file: `.pauseFetching()`, `.resumeFetching()`, `.cancelFetching()`
- * - Downloaded file management:
- *   -  `.getFilesTotalSize()`, `.listDownloadedFiles()`, `.listDownloadedModels()`, `.deleteResources()`
+ * - `react-native-executorch-expo-resource-fetcher` (Expo apps)
+ * - `react-native-executorch-bare-resource-fetcher` (bare React Native)
  *
- * Remark: The pausing/resuming/canceling works only for fetching remote resources.
+ * Methods on this class:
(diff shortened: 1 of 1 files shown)


packages/react-native-executorch/src/utils/ResourceFetcher.ts
```diff
@@ -1,31 +1,27 @@
...
- * - Uses the `ResourceSourceExtended` interface to enable pause/resume functionality:
- *   - Wraps user-provided `ResourceSource` elements
- *   - Implements linked list behavior via the `.next` attribute
- *   - Automatically processes subsequent downloads when `.next` contains a valid resource
+ * Storage-management utilities such as `pauseFetching`, `resumeFetching`,
+ * `cancelFetching`, `getFilesTotalSize`, `listDownloadedFiles`,
+ * `listDownloadedModels`, and `deleteResources` are NOT methods on this
+ * class. They live on the concrete adapter implementations
+ * (`ExpoResourceFetcher`, `BareResourceFetcher`); use the adapter directly
+ * if you need them.
```
chmjkb: I'm ok with this change but I wonder if this comment is even needed here 😅 I guess it's kind of duplicating what the resource fetcher JSDoc should be saying, wdyt?

chmjkb (approved):

packages/react-native-executorch/src/utils/ResourceFetcher.ts
msluszniak: Yeah, we can limit to just one place.

msluszniak: Dropped the file-level block in 7915e5e2d — kept only the class docstring. Agreed that the duplication was exactly what caused this issue to begin with.

PR #1104, opened 2026-04-27
feat: add support openai/privacy-filter model

### LINKED ISSUE #1103: Adding support for OpenAI/Privacy-Filter

NorbertKlockiewicz:
## Description

Adds a new **Privacy Filter** native model for on-device PII detection, with a TypeScript hook (`usePrivacyFilter`), module (`PrivacyFilterModule`), types, and a demo screen in the `apps/llm` example. Two models are wired up out-of-the-box:

- **`PRIVACY_FILTER_OPENAI`** — [openai/privacy-filter](https://huggingface.co/openai/privacy-filter), 8 categories (person, email, phone, address, DOB, URL, account number, secret).
- **`PRIVACY_FILTER_NEMOTRON`** — [OpenMed/privacy-filter-nemotron-base](https://huggingface.co/OpenMed/privacy-filter-nemotron-base), 55+ categories adding medical, financial, demographic, technical, and government-document spans.

Both models share the same single-method `forward(input_ids, attention_mask) → logits` graph, the same o200k tokenizer (pad/eos id 199999), and 256-token banded attention. They differ only in the BIOES label space, which is supplied as `labelNames` at load time.

**Highlights:**
- **Constrained Viterbi decoding** with six tunable BIOES transition biases, matching OpenAI's `viterbi_calibration.json` schema. Defaults to neutral validity-only Viterbi.
- **Sliding-window inference**: 256-token windows, 50% overlap, no truncation — arbitrary-length input is supported.
- **`seq_len` is read from the model's `forward` input shape** at load time, so a 512-token export would Just Work without code changes.
- **Typed JSI return**: `generate()` returns `PiiEntity[]` directly via a `getJsiValue` overload — no JSON bridge.
- **Viterbi decoder is split** into its own `Viterbi.{h,cpp}` module under `rnexecutorch::models::privacy_filter::viterbi`, mirroring the `Utils.{h,cpp}` convention used by VAD.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. `cd apps/llm && yarn install`
2. iOS: `yarn pods && yarn ios` · Android: `yarn android`
3. From the home screen, open **"Privacy Filter"**.
4. Pick a model in the picker (OpenAI base or Nemotron fine-tune) — the model downloads on first run.
5. Tap **"Run"** to scan the bundled sample text.
6. Verify detected entity spans are highlighted inline with their label and color, and inference time is shown on the button.
7. Switch models and re-run — entity coverage should differ between the two (Nemotron picks up medical/financial/demographic categories the base model doesn't have).

### Checklist

### Additional notes

- The `tokenizer.json` is bit-identical between the OpenAI base and the Nemotron fine-tune; both use the GPT-4o `o200k` BPE.
- All PTE files are exported as a **single `forward` method** rather than the 18-method multimethod export — this brings RAM down from ~3.5 GB to ~1.2 GB on iPhone 16 Pro and inference from ~8s to ~1.5s, since planner buffers are reused across the full graph.
- Docs are added to `docs/docs/` (next-version only). The `06-api-reference/` typedoc output regenerates automatically from the `@category Models - Privacy Filter` tags on the exported constants and the public hook/module/types — no manual entries were added there.
- Benchmarks (`02-benchmarks/`) are intentionally *not* updated in this PR; we'll add measured numbers in a follow-up once we have proper multi-device timings.


packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -181,6 +182,14 @@ template <typename Model> class ModelHostObject : public JsiHostObject {
           JSI_EXPORT_FUNCTION(ModelHostObject<Model>, unload, "unload"));
     }
 
+    if constexpr (meta::SameAs<Model, models::privacy_filter::PrivacyFilter>) {
+      addFunctions(
+          JSI_EXPORT_FUNCTION(ModelHostObject<Model>, unload, "unload"));
+      addFunctions(JSI_EXPORT_FUNCTION(ModelHostObject<Model>,
+                                       promiseHostFunction<&Model::detect>,
+                                       "detect"));
+    }
```
chmjkb: I know it probably doesn't make a lot of sense semantically, but maybe lets use generate so its aligned with other models.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,355 @@
...
+// mismatch in the diagnostic at the preprocess step.
+constexpr int32_t kSeqLen = 256;
+constexpr int32_t kHiddenDim = 640;
+constexpr int32_t kNumLayers = 8;
+constexpr int32_t kNumLabels = 33;
+// RoBERTa pad token id used by openai/privacy-filter.
+constexpr int64_t kPadTokenId = 1;
+
+// BIOES scheme: 33 labels = 1 outside + 4 tags x 8 entity types.
+constexpr const char *kLabelNames[kNumLabels] = {
```
chmjkb: maybe use std array here?

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+  // 1) Tokenize. The HF tokenizer post-processor attached to
+  //    openai/privacy-filter adds <s>/</s> (CLS/SEP) automatically.
+  auto rawIds = tokenizer_->encode(text);
+
+  const int32_t usedLen = static_cast<int32_t>(
+      std::min<size_t>(rawIds.size(), static_cast<size_t>(kSeqLen)));
+
+  std::vector<int64_t> inputIds(kSeqLen, kPadTokenId);
+  std::vector<int64_t> attentionMask(kSeqLen, 0);
```
chmjkb: I think this could be an array too?

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,355 @@
...
+  for (int32_t i = 0; i < usedLen; ++i) {
+    inputIds[static_cast<size_t>(i)] = static_cast<int64_t>(rawIds[i]);
+    attentionMask[static_cast<size_t>(i)] = 1;
+  }
+
+  // 2) Preprocess
```
chmjkb: Either change those comments so they're meaningful or delete them

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,355 @@
...
+  std::vector<int32_t> idsShape = {1, kSeqLen};
+  auto inputIdsTensor =
+      make_tensor_ptr(idsShape, inputIds.data(), ScalarType::Long);
+  auto attentionMaskTensor =
+      make_tensor_ptr(idsShape, attentionMask.data(), ScalarType::Long);
+
+  auto preprocessResult =
+      BaseModel::execute("preprocess", {*inputIdsTensor, *attentionMaskTensor});
```
chmjkb: Wondering what this is actually doing

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,355 @@
...
+        }
+        diag += ", scalarType=" +
+                std::to_string(static_cast<int32_t>(t->scalar_type())) + ")";
+      }
+    } else {
+      diag += " (no preprocess method? error=" +
+              std::to_string(static_cast<int32_t>(metaResult.error())) + ")";
+    }
+    throw RnExecutorchError(preprocessResult.error(), diag);
+  }
```
chmjkb: damn this is a big boy, IF you're sure we need this, can we move this to some other file? but tbh it seems like an overkill.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,355 @@
...
+  if (preprocessOut.size() < 4) {
+    throw RnExecutorchError(
+        RnExecutorchErrorCode::UnknownError,
+        "PrivacyFilter: preprocess returned unexpected number of outputs");
+  }
+
+  std::vector<float> hidden = copyFloatTensor(preprocessOut[0]);
+  std::vector<float> cos = copyFloatTensor(preprocessOut[1]);
+  std::vector<float> sin = copyFloatTensor(preprocessOut[2]);
+  std::vector<float> attnMask4d = copyFloatTensor(preprocessOut[3]);
```
chmjkb: do we need to copy? or this memory gets somehow freed from ET once we unload

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.h
```diff
@@ -0,0 +1,41 @@
...
+
+class PrivacyFilter final : public BaseModel {
+public:
+  PrivacyFilter(const std::string &modelSource,
+                const std::string &tokenizerSource,
+                std::shared_ptr<react::CallInvoker> callInvoker);
+
+  // Returns a JSON string describing detected PII entities. Schema:
+  //   [{ "label": string, "text": string,
+  //      "startToken": number, "endToken": number }]
```
chmjkb: Maybe we could return a list of structs instead of JSON?

chmjkb (commented):
im not sure itf its good time to review this as this is a draft 😅 
but I only noticed after starting the review so heres a couple of omments

apps/llm/app.json
```diff
@@ -63,7 +63,8 @@
       },
       "entitlements": {
         "com.apple.developer.kernel.increased-memory-limit": true
-      }
+      },
+      "appleTeamId": "J5FM626PE2"
```
msluszniak: Hmm? ;)

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    throw RnExecutorchError(
+        RnExecutorchErrorCode::WrongDimensions,
+        "PrivacyFilter: expected forward input shape [1, seq_len].");
+  }
+  seqLen_ = inputShapes[0][1];
+  grammar_ = viterbi::buildGrammar(labelNames_, biases_);
+}
+
+std::string PrivacyFilter::labelEntityType(int32_t labelId) const {
+  if (labelId <= 0 || labelId >= static_cast<int32_t>(labelNames_.size())) {
```
msluszniak:
```suggestion
  if (labelId <= 0 || std::cmp_greater_equal(labelId, labelNames_.size()) {
```

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+void PrivacyFilter::runWindow(const std::vector<int64_t> &paddedInputIds,
+                              const std::vector<int64_t> &paddedAttentionMask,
+                              int32_t absStart, int32_t validLen,
+                              int32_t writeFromOffset, int32_t writeToOffset,
+                              std::vector<int32_t> &outLabels) {
+  std::vector<int32_t> idsShape = {1, seqLen_};
+  auto inputIdsTensor = make_tensor_ptr(
+      idsShape, const_cast<int64_t *>(paddedInputIds.data()), ScalarType::Long);
+  auto attentionMaskTensor = make_tensor_ptr(
+      idsShape, const_cast<int64_t *>(paddedAttentionMask.data()),
```
msluszniak: Do we really want to pass const reference to `paddedInputIds` and `paddedAttentionMask` to just then strip consts via `const_cast`?

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    throw RnExecutorchError(forwardResult.error(),
+                            "The model's forward function did not succeed. "
+                            "Ensure the model input is correct.");
+  }
+  auto &out = forwardResult.get();
+  if (out.empty()) {
+    throw RnExecutorchError(RnExecutorchErrorCode::UnknownError,
+                            "PrivacyFilter: forward returned no outputs");
+  }
+  if (validLen <= 0) {
```
msluszniak: We can move this higher to avoid inference

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+  const float *logits = logitsTensor.const_data_ptr<float>();
+
+  auto path = viterbi::decode(logits, validLen, grammar_);
+
+  for (int32_t t = writeFromOffset; t < writeToOffset; ++t) {
+    if (t >= validLen) {
+      break;
+    }
+    outLabels[static_cast<size_t>(absStart + t)] = path[static_cast<size_t>(t)];
+  }
```
msluszniak:
```suggestion
const int32_t end = std::min(writeToOffset, validLen);
std::copy(path.begin() + writeFromOffset,
          path.begin() + end,
          outLabels.begin() + absStart + writeFromOffset);
```

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::vector<int64_t> paddedInputIds(static_cast<size_t>(seqLen_),
+                                        kPadTokenId);
+    std::vector<int64_t> paddedAttentionMask(static_cast<size_t>(seqLen_), 0);
+    for (int32_t i = 0; i < validLen; ++i) {
+      paddedInputIds[static_cast<size_t>(i)] =
+          static_cast<int64_t>(rawIds[static_cast<size_t>(windowStart + i)]);
+      paddedAttentionMask[static_cast<size_t>(i)] = 1;
+    }
+
+    const bool isFirst = (windowStart == 0);
```
msluszniak:
```suggestion
    const bool isFirst = windowStart == 0;
```

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    const bool isLast = (windowStart + seqLen_ >= totalTokens);
```
msluszniak:
```suggestion
    const bool isLast = windowStart + seqLen_ >= totalTokens;
```

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (entity.empty()) {
+      ++i;
+      continue;
+    }
+    int32_t j = i + 1;
+    while (j < totalTokens &&
+           labelEntityType(predictedLabels[static_cast<size_t>(j)]) == entity) {
+      ++j;
+    }
+    spans.push_back({i, j, entity});
```
msluszniak: emplace back

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::vector<uint64_t> slice;
+    slice.reserve(static_cast<size_t>(span.end - span.start));
+    for (int32_t k = span.start; k < span.end; ++k) {
+      slice.emplace_back(rawIds[static_cast<size_t>(k)]);
+    }
+    std::string decoded;
+    try {
+      decoded = tokenizer_->decode(slice, /*skipSpecialTokens=*/true);
+    } catch (...) {
+      decoded = "";
```
msluszniak: This one is not needed I guess?

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    auto left = std::find_if(decoded.begin(), decoded.end(), notSpace);
+    auto right =
+        std::find_if(decoded.rbegin(), decoded.rend(), notSpace).base();
+    if (left < right) {
+      decoded = std::string(left, right);
+    } else {
+      decoded.clear();
+    }
+
+    entities.push_back({span.entity, std::move(decoded), span.start, span.end});
```
msluszniak: Add this one but correctly formatted
```suggestion
constexpr auto notSpace = [](unsigned char c) { return !std::isspace(c); };
auto left  = std::ranges::find_if(decoded, notSpace);
auto right = std::ranges::find_if(decoded | std::views::reverse, notSpace).base();
if (left < right) {
  decoded.assign(left, right);
} else {
  decoded.clear();
}
entities.emplace_back(span.entity, std::move(decoded), span.start, span.end);
```

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/Viterbi.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  size_t bestEnd = 0;
+  float bestScore = kNegInf;
+  for (size_t j = 0; j < N; ++j) {
+    if (dp[j] > bestScore) {
+      bestScore = dp[j];
+      bestEnd = j;
+    }
+  }
```
msluszniak: Replace with this, correctly formatted.
```suggestion
auto it = std::ranges::max_element(dp);
size_t bestEnd = std::distance(dp.begin(), it);
float bestScore = *it;
```

msluszniak:
FYI, I tested demo apps and they worked correctly :))

msluszniak (commented):
Second pass — a few correctness issues + polish.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,210 @@
...
+                             std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker),
+      tokenizer_(
+          std::make_unique<TokenizerModule>(tokenizerSource, callInvoker)),
+      labelNames_(std::move(labelNames)), seqLen_(0) {
+  if (labelNames_.empty()) {
+    throw RnExecutorchError(
+        RnExecutorchErrorCode::UnknownError,
+        "PrivacyFilter requires a non-empty labelNames vector "
+        "(must include 'O' at index 0).");
```
msluszniak, line 42: Wrong-sized `viterbiBiases` silently falls through to all-zero defaults with no diagnostic — should throw on `size() != 0 && size() != 6`. Currently latent because `packViterbiBiases` on the TS side always emits 6, but anyone calling the JSI binding from a different layer (or a future fork) gets a confusing "biases have no effect" debugging session.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,210 @@
...
+} // namespace
+
+PrivacyFilter::PrivacyFilter(const std::string &modelSource,
+                             const std::string &tokenizerSource,
+                             std::vector<std::string> labelNames,
+                             std::vector<float> viterbiBiases,
+                             std::shared_ptr<react::CallInvoker> callInvoker)
+    : BaseModel(modelSource, callInvoker),
+      tokenizer_(
+          std::make_unique<TokenizerModule>(tokenizerSource, callInvoker)),
```
msluszniak, line 36: Message claims "must include 'O' at index 0" but only `empty()` is checked. If `labelNames_[0] != "O"`, any prediction at index 0 is silently dropped by the `labelId <= 0` branch in `labelEntityType()`. Add the missing assertion to match the message.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.cpp
```diff
@@ -0,0 +1,210 @@
...
+                            "PrivacyFilter is not loaded");
+  }
+
+  auto rawIds = tokenizer_->encode(text);
+  const int32_t totalTokens = static_cast<int32_t>(rawIds.size());
+
+  std::vector<int32_t> predictedLabels(static_cast<size_t>(totalTokens), 0);
+
+  const int32_t stride = seqLen_ / 2;
+  const int32_t edgeMargin = seqLen_ / 4;
```
msluszniak, line 137: `stride = seqLen_ / 2` is 0 when `seqLen_ < 2`, making this window loop infinite. The constructor validates that the input shape exists but not its magnitude — worth a `seqLen_ < 2` check up there.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/Viterbi.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return grammar;
+}
+
+std::vector<int32_t> decode(const float *logits, int32_t validLen,
+                            const Grammar &grammar) {
+  if (validLen <= 0) {
+    return {};
+  }
+
```
msluszniak, line 103: `bp[0..N)` is never read — the traceback only consults `bp[t*N+j]` for `t >= 1`. Size as `(validLen - 1) * N` and shift the index, or drop a comment if the row is intentional headroom.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.h
```diff
@@ -0,0 +1,54 @@
...
+class PrivacyFilter final : public BaseModel {
+public:
+  PrivacyFilter(const std::string &modelSource,
+                const std::string &tokenizerSource,
+                std::vector<std::string> labelNames,
+                std::vector<float> viterbiBiases,
+                std::shared_ptr<react::CallInvoker> callInvoker);
+
+  [[nodiscard("Registered non-void function")]] std::vector<types::PiiEntity>
+  generate(std::string text);
```
msluszniak, line 29: Same nit as your L86 comment but on the entry point — `text` is only forwarded to `tokenizer_->encode(text)`, so `const std::string&` skips the copy.

packages/react-native-executorch/src/hooks/natural_language_processing/usePrivacyFilter.ts
```diff
@@ -0,0 +1,32 @@
...
+export const usePrivacyFilter = ({
+  model,
+  preventLoad = false,
+}: PrivacyFilterProps): PrivacyFilterType => {
+  const { error, isReady, isGenerating, downloadProgress, runForward } =
+    useModuleFactory({
+      factory: (config, onProgress) =>
+        PrivacyFilterModule.fromModelName(config, onProgress),
+      config: model,
+      deps: [model.modelName, model.modelSource, model.tokenizerSource],
```
msluszniak, line 24: `labelNames` and `viterbiBiases` are missing from the deps — both feed the constructor and rebuild the Viterbi grammar (PrivacyFilter.cpp:57), so swapping either at runtime is a no-op until the model is forcibly reloaded. The other hooks get away with the 3-key deps because they have no equivalent runtime-tunable; this one does.

packages/react-native-executorch/src/modules/natural_language_processing/PrivacyFilterModule.ts
```diff
@@ -0,0 +1,126 @@
...
+    tokenizerSource: ResourceSource,
+    labelNames: readonly string[],
+    options: {
+      viterbiBiases?: ViterbiBiases;
+      onDownloadProgress?: (progress: number) => void;
+    } = {}
+  ): Promise<PrivacyFilterModule> {
+    return PrivacyFilterModule.fromModelName(
+      {
+        modelName: 'custom',
```
msluszniak, line 102: Combined with the deps issue in `usePrivacyFilter`, two custom fine-tunes loaded back-to-back with the same `modelSource` but different `labelNames` won't trigger a reload — `modelName: 'custom'` collapses identity. Either derive a stable identity hash, or document the foot-gun.

packages/react-native-executorch/src/types/privacyFilter.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @category Types
+ */
+export interface PrivacyFilterType {
+  error: null | RnExecutorchError;
+  isReady: boolean;
+  isGenerating: boolean;
+  downloadProgress: number;
+  /**
+   * Run PII detection over the given text. Long inputs are processed in
+   * sliding 256-token windows with 128-token overlap; no truncation.
```
msluszniak: The PR description (correctly) highlights that `seq_len` is read from the model's forward input shape — but this public JSDoc hardcodes 256/128. Drop the magic numbers, or restate the policy (50% overlap, 25% edge trim) so a future 512-token export doesn't immediately drift this doc.

apps/llm/utils/piiMatching.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (exact !== -1) {
+      matches.push({
+        start: exact,
+        end: exact + e.text.length,
+        label: e.label,
+      });
+      cursor = exact + e.text.length;
+      continue;
+    }
+    const words = e.text.split(/\s+/).filter((w) => w.length > 1);
```
msluszniak, line 57: `source.indexOf("John", ...)` happily highlights the `John` inside `Johnson`. A `\b`-anchored regex per word avoids the false match.

apps/llm/app/privacy_filter/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+    setRunError(null);
+    setEntities(null);
+    setInferenceMs(null);
+    const startedAt = Date.now();
+    try {
+      const result = await filter.generate(sampleText);
+      const elapsed = Date.now() - startedAt;
+      setInferenceMs(elapsed);
+      setEntities(result);
+    } catch (e) {
```
msluszniak, line 117: The hook surfaces `RnExecutorchError` with a `.message`; this 17-line ladder is dead defense. `setRunError(e instanceof Error ? e.message : String(e))` covers everything that actually shows up here.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.h
NorbertKlockiewicz: every other model takes `std::string` by value, I am not sure but I think we discussed it once and it is enforced by JSI (not 100% sure)

packages/react-native-executorch/src/modules/natural_language_processing/PrivacyFilterModule.ts
NorbertKlockiewicz: every peer uses 'custom' string identity for fromCustomModel - I think it's kind of hardcore situation and if we want to handle it we should do it in every other model.

packages/react-native-executorch/common/rnexecutorch/models/privacy_filter/PrivacyFilter.h
msluszniak: Oh, right, nvm

packages/react-native-executorch/src/hooks/natural_language_processing/usePrivacyFilter.ts
NorbertKlockiewicz: Those values are not runtime tunable so the thing won't happen.

packages/react-native-executorch/src/modules/natural_language_processing/PrivacyFilterModule.ts
msluszniak: Yeah so the actual question is if we ultimately want to handle such case. If so, I can create an issue with this one.

NorbertKlockiewicz: I think we can create an issue as it's a legitimate bug but such a situation will probably occure once in a million.

msluszniak (approved):
LGTM :))

packages/react-native-executorch/src/modules/natural_language_processing/PrivacyFilterModule.ts
msluszniak: #1116

PR #1106, opened 2026-04-28
chore(tests): push and clean test models per-test in run_tests.sh

msluszniak:
## Description

`run_tests.sh` previously pushed every test model to the device up front (~6GB total), then ran every test. On emulators with a small `/data` partition the push step would fail with `No space left on device` before any test ran.

This PR changes the run loop to push only the models each test needs, run the test, then `rm` those models before the next test. Peak device disk usage drops to the largest single test's working set (~3GB for `LLMTests`). Tests with no model dependencies (`NumericalTests`, `RunnerTests`, `LogTests`, `FileUtilsTest`, `ImageProcessingTest`, `FrameTransformTests`) skip the push/cleanup entirely.

The test → models mapping lives in a single `models_for_test()` case statement, so adding a new test is a one-line change.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. `cd packages/react-native-executorch/common/rnexecutorch/tests`
2. `bash run_tests.sh`
3. Watch the logs — between each test you should see `Pushing models for <Test>...` and the `/data/local/tmp/rnexecutorch_tests` dir size drop back down before the next test starts.
4. Verify the suite passes end-to-end on a device that previously couldn't fit all models simultaneously.

### Screenshots

_N/A_

### Related issues

_None_

### Checklist

### Additional notes

The `models_for_test` case statement is the single source of truth for per-test asset deps; if a new test is added, its model filenames need to be added there or the test will run without them on device. The fallback `*) echo "" ;;` arm covers tests with no model deps.

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/tests/run_tests.sh
+++ b/packages/react-native-executorch/common/rnexecutorch/tests/run_tests.sh
@@ -189,6 +189,44 @@ run_test() {
   return 0
 }
 
+# Returns a space-separated list of model/asset filenames (in $MODELS_DIR) that
+# the given test executable loads at runtime. Tests not listed here have no
+# model dependencies. Adding a new test? Add its filenames below.
+models_for_test() {
+  case "$1" in
+    BaseModelTests) echo "style_transfer_candy_xnnpack_fp32.pte" ;;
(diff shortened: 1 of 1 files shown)


NorbertKlockiewicz (approved):
works as expected

PR #1114, opened 2026-04-29
build: Extend tokenizer capabilities

benITo47:
## Description

This PR introduces rebuilt binaries that contain new, updated tokenizers. 
This iteration features support for more tokenisation models (i.e. unigram, worldlevel) as well as bunch of previously unsupported pre-tokenisers, decoders, post-processors.  

### Introduces a breaking change?

- [ ] Yes
- [x] No

### Type of change

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Documentation update (improves or adds clarity to existing documentation)
- [x] Other (chores, tests, code style improvements etc.)

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Before merging, test all demo applications. See if all models that proved problematic during bumps in the past are working (i.e. kokoro, multi-method models) 
Check all LLM models, see if output is working. 

- [x] LLM app on iOS
- [x] LLM app on Android
- [x] Speech app on iOS
- [x] Speech app on Android
- [x] Text Embeddings on iOS
- [x] Text Embeddings on Android
 
### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation accordingly
- [x] My changes generate no new warnings

PR #1115, opened 2026-04-30
feat: add PARAPHRASE_MULTILINGUAL_MINILM_L12_V2 text embeddings model

### LINKED ISSUE #945: Feature request: multilingual text embeddings model

msluszniak:
## Description

Adds the `paraphrase-multilingual-MiniLM-L12-v2` sentence-transformer model — the second multilingual embeddings model after distiluse, completing #945. Ships **only the XNNPACK 8da4w variant** under `MODEL_REGISTRY.ALL_MODELS` (see "Why a single variant" below).

384-d output, max 126 tokens, 50+ languages. Tokenizer is Unigram + Precompiled normalizer + Metaspace decoder — **requires the bumped `pytorch/extension/llm/tokenizers` runtime from #1114**, so this PR blocks on that landing first and should be rebased onto main once #1114 merges.

HF repo: [software-mansion/react-native-executorch-paraphrase-multilingual-MiniLM-L12-v2](https://huggingface.co/software-mansion/react-native-executorch-paraphrase-multilingual-MiniLM-L12-v2) (`v0.9.0` tag, layout mirrors distiluse).

**Why a single variant** — 
TLDR 8da4w works faster then all and was also one of the smallest, without loss in precision.  
Longer answer:
unlike distiluse, where Core ML fp32 won iPhone thanks to ANE acceleration, benchmarks on iPhone 17 Pro + OnePlus 12 (~80-token input, 50 measured forwards after 3 warmups) showed the XNNPACK 8da4w variant Pareto-dominates the other three on both platforms: faster than XNNPACK fp32, Core ML fp32 *and* Core ML fp16 on iPhone, and ~36% smaller steady-state memory footprint than the next-best variant. Likely cause: paraphrase-multilingual-MiniLM-L12-v2 is a smaller model (~118 M params, 12 layers) where Core ML's runtime doesn't push enough work onto ANE for the precision-conversion overhead to pay off. fp16 being slower than fp32 on Core ML for this model is a tell that the runtime is falling back to slower compute units. Shipping only `_8DA4W` keeps the public surface aligned with the data; if a future Core ML or model update flips the verdict, easy to add the other variants back.

**Memory methodology note** — the new paraphrase row in `docs/docs/02-benchmarks/memory-usage.md` reports RSS / `phys_footprint` deltas from a clean app baseline (loaded − idle), captured on-device at the same conceptual point. The existing distiluse rows there (36 / 44 MB) come from an older measurement pass with a different (and not reconstructable from the diff) methodology, so the two rows are not directly comparable. A separate pass to re-measure distiluse and other rows with the same methodology would be a good follow-up.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. `cd apps/text-embeddings && npx expo run:ios` (or `run:android`).
2. Pick **"Multilingual Paraphrase (8da4w)"** in the model picker.
3. Add a sentence in one language, query with an aligned sentence in another (e.g. Polish "Słoneczko" against "It's so sunny outside!"). The cross-lingual pair should top the matches.

### Related issues

Closes the paraphrase-multilingual half of #945 (the distiluse half landed in #1098).

### Checklist

### Additional notes

Blocks on #1114.


msluszniak:
Would you rather keep 8da4w suffix or you are in favour of dropping it? cc: @chmjkb @NorbertKlockiewicz @barhanc

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1159,6 +1161,15 @@ export const DISTILUSE_BASE_MULTILINGUAL_CASED_V2_COREML = {
   tokenizerSource: DISTILUSE_BASE_MULTILINGUAL_CASED_V2_TOKENIZER,
 } as const;
 
+/**
+ * @category Models - Text Embeddings
+ */
+export const PARAPHRASE_MULTILINGUAL_MINILM_L12_V2_8DA4W = {
```
barhanc: I think for now it should be `PARAPHRASE_MULTILINGUAL_MINILM_L12_V2_QUANTIZED` so that we maintain consistent naming (and we should probably retroactively change the `DISTILUSE_BASE_MULTILINGUAL_CASED_V2_8DA4W` also). We had some discussion about model registry and naming (#612) so we can change the symbols all in one go when implementing that.

barhanc (commented):
Model works, just this small comment regarding naming.

packages/react-native-executorch/src/constants/modelUrls.ts
chmjkb: agreed, I think that answers your question @msluszniak

msluszniak: Ok, so I'll change the name for this model and for distiluse we will fix this in one holistic change.

barhanc (approved):

PR #1120, opened 2026-05-05
feat: add support for FastSAM model with point, box and text prompts

### LINKED ISSUE #555: SAM - segment anything model
Implement segment anything model. Inspired by ExecuTorch blog and Instagram implementing Squeeze sam inside their app add similar functionality in our lib.

barhanc:
## Description

Adds support for FastSAM model with required postprocessing for point, box and text (using already existing CLIP export) prompts. Also adds an example app to test these. 

Since FastSAM uses YOLO instance segmentation backbone with some clever postprocessing to imitate Facebook's SAM (see https://docs.ultralytics.com/models/fast-sam/#model-architecture), we use the existing instance segmentation C++ implementation and add TS postprocessing to minimize code duplication.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Run the Computer Vision - Segment Anything app screen and test two available models there
- [ ] You can also run Computer Vision - Instance Segmentation app screen and test the two newly added models there.
- [ ] You can also run Computer Vision - Vision Camera app screen to test real-time performance of the two new models under 'Instance Segmentation'.
- [x] Check HF pages for the exported model https://huggingface.co/software-mansion/react-native-executorch-fast-sam

### Screenshots

You can use following image for testing.

https://upload.wikimedia.org/wikipedia/commons/c/cd/Animal_diversity_October_2007.jpg

Simulator Screenshot - iPhone 17 Pro - 2026-05-06 at 23 12 00

Simulator Screenshot - iPhone 17 Pro - 2026-05-06 at 23 15 22

### Related issues

Closes #555 

### Checklist


packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1010,6 +1010,26 @@ export const SELFIE_SEGMENTATION = {
   modelSource: SELFIE_SEGMENTATION_MODEL,
 } as const;
 
+// FastSAM Instance Segmentation
+const FASTSAM_S_SEG_MODEL = `${URL_PREFIX}-fast-sam/${NEXT_VERSION_TAG}/fastsam-s/xnnpack/fastsam_s_xnnpack_fp32.pte`;
+const FASTSAM_X_SEG_MODEL = `${URL_PREFIX}-fast-sam/${NEXT_VERSION_TAG}/fastsam-x/xnnpack/fastsam_x_xnnpack_fp32.pte`;
```
msluszniak: How about quantized xnnpack / coreml for this model?

packages/react-native-executorch/src/utils/segmentAnythingPrompts.ts
```diff
@@ -0,0 +1,145 @@
...
+    throw new Error(
+      `selectByText: instances (${instances.length}) and instanceEmbeddings (${instanceEmbeddings.length}) must have the same length`
+    );
+  }
+
+  let textNormSq = 0;
+  for (let i = 0; i < textEmbedding.length; i++) {
+    const v = textEmbedding[i]!;
+    textNormSq += v * v;
+  }
```
msluszniak:
```suggestion
  for (const v of textEmbedding) {
    textNormSq += v * v;
  }
```

apps/computer-vision/app/segment_anything/index.tsx
```diff
@@ -0,0 +1,530 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
```
msluszniak, line 1: Issues regarding demo app:
- ~Clicking `Find` does not clear results of previous inference~ Done
- ~Clicking `Find` multiple times causing text box to "blink" when trying Text option.~ Done
- ~Don't we want to integrate this with vision camera, what are the current limitations?~ Done
- When I select again the same model I had at the moment, the whole inference is repeated and I need to click 'Run model' again.
- Maybe we want to add some option to segment multiple instances.

msluszniak (commented):
Do we want to add some benchmarks for this one?

packages/react-native-executorch/src/constants/modelUrls.ts
barhanc: I tried quantization and the model quality was unusable. I will try to export the model using CoreML for performance gains on iPhone.

apps/computer-vision/components/vision_camera/tasks/InstanceSegmentationTask.tsx
```diff
@@ -44,9 +49,23 @@ export default function InstanceSegmentationTask({
...
 
   const active =
-    activeModel === 'instanceSegmentationYolo26n' ? yolo26n : rfdetr;
+    activeModel === 'instanceSegmentationYolo26n'
+      ? yolo26n
+      : activeModel === 'instanceSegmentationRfdetr'
+        ? rfdetr
+        : activeModel === 'instanceSegmentationFastsamS'
+          ? fastsamS
+          : fastsamX;
```
chmjkb: Please use a lookup object instead

docs/docs/03-hooks/02-computer-vision/segment-anything.md
chmjkb, line 1: This doc diverges a bit from what we established in other docs in terms of header structure, can we align it a bit more?

apps/computer-vision/app/segment_anything/index.tsx
chmjkb, line 1: Please fix the keyboard handling on iOS in the text prompt, it feels very off

packages/react-native-executorch/src/utils/segmentAnythingPrompts.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  return best?.inst ?? null;
+}
+
+function bboxArea(bbox: Bbox): number {
+  return Math.max(bbox.x2 - bbox.x1, 0) * Math.max(bbox.y2 - bbox.y1, 0);
+}
+
```
chmjkb: this may be used in other places in code too in the future, maybe move it to commonvision

chmjkb (changes requested):
I tested the demo app on iOS and the results vere pretty mid, at least for the S version. Not sure if this is the nature of the model, but just saying

msluszniak:
> I tested the demo app on iOS and the results vere pretty mid, at least for the S version. Not sure if this is the nature of the model, but just saying

Probably the nature of the model. You can share what results you get and I can do a cross-check.

docs/docs/03-hooks/02-computer-vision/segment-anything.md
barhanc: I will polish the docs once I apply all the changes, since I'm still working on the multi-output for text prompt selector. Would you prefer that I add 'Segment Anything' section to `useInstanceSegmentation.md` or leave it in its own file but change the structure?

packages/react-native-executorch/src/utils/segmentAnythingPrompts.ts
barhanc: Do you mean `packages/react-native-executorch/src/constants/commonVision.ts`? It's under `constants` directory so I think it would be somewhat unintuitive to have helper functions there.

barhanc:
> I tested the demo app on iOS and the results vere pretty mid, at least for the S version. Not sure if this is the nature of the model, but just saying

From what I've tested, the S variant is fine for simple segmentation when objects don't overlap, but for more complex scenes it's true that artifacts show up. The X variant however worked fine on all images I tried, even ones with quite complex scenes. Did you observe bad performance on X variant also?

docs/docs/03-hooks/02-computer-vision/segment-anything.md
chmjkb: I think im fine with keeping this file, just wanted to sort out the minor inconsistency across docs

packages/react-native-executorch/src/utils/segmentAnythingPrompts.ts
chmjkb: yeah sorry i wasnt very precise, 
i think a good place for this would be something like `src/utils/vision.ts` or whatever you think fits.

chmjkb:
> > I tested the demo app on iOS and the results vere pretty mid, at least for the S version. Not sure if this is the nature of the model, but just saying
> 
> From what I've tested, the S variant is fine for simple segmentation when objects don't overlap, but for more complex scenes it's true that artifacts show up. The X variant however worked fine on all images I tried, even ones with quite complex scenes. Did you observe bad performance on X variant also?

I gave some more testing to the X version, and the point/box based detections look cool, but the prompt-based ones were pretty bad. I almost never got the result I expected, but maybe its due to poor quality of text embeddings. quick example:

Zrzut ekranu 2026-05-08 o 15 11 47

barhanc:
@msluszniak @chmjkb I've:
- updated docs by adding the section on selector use to `useInstanceSegmentation.md`, the previous file didn't really fit under `hooks/` in my opinion.
- moved the helper function `bboxArea` to `utils/commonVision.ts` as requested.
- fixed keyboard handling - hopefully it feels better now.
- fixed a problem with the way cropped images were passed to CLIP for embeddings - previously I followed exactly how it is done in ultralytics Python implementation where they simply pass the cropped image without masking; now the cropped image is masked  based on the segmentation mask, so it should work better on examples as the one posted above. There are still some problems with text prompts on some images, but these are because of CLIP model not being able to properly embed images with certain parts of the image masked.
- added optional `topk` parameter to `selectByText` that enables the user to return top-k best matching instances to the given text prompt. This is mostly for convenience and it doesn't truly solve the problem of returning multiple instances as `topk` must be passed explicitly, so e.g. it doesn't cover use cases where user would like to automatically count the number of objects matching certain prompt. The problem is that text/image embeddings are inherently contrastive and passing like a threshold doesn't really solve it. I could add a function that does something like an open-vocabulary classification where the user inputs text prompts that correspond to classes and for each segmented instance we return the best matching class based on cosine similarity, but I'm not sure if that's needed. What do you think? 

I will be also adding benchmarks shortly.

msluszniak (approved):
LGTM from my side. We can add tip to documentation, that for images with overlaps of entities, prefer fastSam-X over smaller version.

chmjkb:
> * added optional `topk` parameter to `selectByText` that enables the user to return top-k best matching instances to the given text prompt. This is mostly for convenience and it doesn't truly solve the problem of returning multiple instances as `topk` must be passed explicitly, so e.g. it doesn't cover use cases where user would like to automatically count the number of objects matching certain prompt. The problem is that text/image embeddings are inherently contrastive and passing like a threshold doesn't really solve it. I could add a function that does something like an open-vocabulary classification where the user inputs text prompts that correspond to classes and for each segmented instance we return the best matching class based on cosine similarity, but I'm not sure if that's needed. What do you think?



> @msluszniak @chmjkb I've:
> 
> * updated docs by adding the section on selector use to `useInstanceSegmentation.md`, the previous file didn't really fit under `hooks/` in my opinion.
> * moved the helper function `bboxArea` to `utils/commonVision.ts` as requested.
> * fixed keyboard handling - hopefully it feels better now.
> * fixed a problem with the way cropped images were passed to CLIP for embeddings - previously I followed exactly how it is done in ultralytics Python implementation where they simply pass the cropped image without masking; now the cropped image is masked  based on the segmentation mask, so it should work better on examples as the one posted above. There are still some problems with text prompts on some images, but these are because of CLIP model not being able to properly embed images with certain parts of the image masked.
> * added optional `topk` parameter to `selectByText` that enables the user to return top-k best matching instances to the given text prompt. This is mostly for convenience and it doesn't truly solve the problem of returning multiple instances as `topk` must be passed explicitly, so e.g. it doesn't cover use cases where user would like to automatically count the number of objects matching certain prompt. The problem is that text/image embeddings are inherently contrastive and passing like a threshold doesn't really solve it. I could add a function that does something like an open-vocabulary classification where the user inputs text prompts that correspond to classes and for each segmented instance we return the best matching class based on cosine similarity, but I'm not sure if that's needed. What do you think?
> 
> I will be also adding benchmarks shortly.
 
I think the topk should be enough, no need to over-engineer this. I'll review it now and if it looks ok then :shipit:

chmjkb (approved):

PR #1121, opened 2026-05-06
Add RNRepo to ci builds

radoslawrolka:
## Description

Adds RNRepo build-tools integration to `llm` example app so that CI builds can consume pre-built native dependencies instead of compiling them from scratch, speeding up CI.

- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Builds in CI using RNRepo should speed up the CI process. Android sees ~38% improvement across the board. iOS builds are ~15% faster.

### Checklist

### DIFF
--- a/.github/workflows/build-android-llm-example.yml
+++ b/.github/workflows/build-android-llm-example.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Build app
         working-directory: ${{ env.WORKING_DIRECTORY }}/android
         run: |
-          ./gradlew assembleDebug \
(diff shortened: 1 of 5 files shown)


radoslawrolka:
Not beneficial in the current state. The LLM app built in CI uses `rn-audio-api`, which requires opting out of Worklets (and therefore also removes Reanimated). These are some of the heaviest libraries in the build process, and they must be built sequentially, not in parallel. Because of that, I’m closing this PR for now.

Issue #1124, opened 2026-05-07
Whisper - add CoreML & Vulkan backends support


IgorSwat:
### Problem Description
Right now, all the Whisper models are utilized only within the XNNPACK backend. We could extend the support by exporting them to Vulkan & CoreML backends.

### What should be done
Export Whisper family models to CoreML & Vulkan if possible.

### Benefits to React Native ExecuTorch
Significant speed-up in speech to text module. A possibility of using bigger model (for example: `whisper-base`) with similar performance but improved accuracy compared to currently standard `whisper-tiny`.

msluszniak:
https://github.com/software-mansion/react-native-executorch/tree/%40ms/experiment-with-npu-and-gpu for a reference for coreml

IgorSwat:
Current status:
- **CoreML backend**: models exported succesfully, resulting in a huge performance boost
- **Vulkan backend**: models export properly, but fail during runtime inference with `libc++abi: terminating due to uncaught exception of type std::out_of_range: vector`. The problem occurs in preprocessing phase of the inference and a possible cause might be Vulkan's limitation on a vector size (initial audio vector contains hundreds of thousands of samples).

msluszniak:
About vulkan, what's the current status, still the same exception?

IgorSwat:
Yeah, Vulkan fails at runtime with exception inside the preprocessing model.
Apart from that it mostly works, but does not bring any significant speed-up over the XNNPACK version.

While it is possible to fix it by extracting the preprocessing implementation out of the model and to the C++ native side, but it would significantly (and unnecessarily) complicate the pipeline. Without significant performance gains I do not see a point in doing so.

msluszniak:
Understood, that's important information that vulkan does not provide any speed up here.

PR #1132, opened 2026-05-08
feat!: major performance & accuracy improvements in speech-to-text module

IgorSwat:
## Description

This PR introduces several changes to the speech-to-text module based on Whisper models:
- CoreML integration - models re-exported to CoreML backend, bringing significant performance upgrade for iOS devices.
- New streaming algorithm - eliminates duplicates in streaming output, resulting in a major quality improvement of the live streaming mode.
- Changes in demo apps: removed faulty 'voice mode' screen in LLM demo app,  refactored speech to text screen in 'speech' app by adding new CoreML models to selection bar and changing the default model for iOS devices.
- Minor code improvements in speech-to-text module 

- [x] Yes
- [ ] No

**Change**: removes predefined constants for quantized models.
**Justification**: the quantized models differ very slightly from the original ones, introducing unnecessary complexity in this case.

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Run demo app to test the live streaming mode.

### Related issues

#1124 

### Checklist


apps/speech/screens/SpeechToTextScreen.tsx
```diff
@@ -14,21 +14,25 @@ import { SafeAreaProvider, SafeAreaView } from 'react-native-safe-area-context';
...
   useSpeechToText,
   WHISPER_TINY_EN,
-  WHISPER_TINY_EN_QUANTIZED,
+  WHISPER_TINY_EN_COREML,
   WHISPER_BASE_EN,
+  WHISPER_BASE_EN_COREML,
   WHISPER_SMALL_EN,
   TranscriptionResult,
   SpeechToTextProps,
+  WHISPER_SMALL_EN_COREML,
```
msluszniak: why this is added after `TranscriptionResult` and `SpeechToTextProps`? ;p

apps/speech/package.json
```diff
@@ -20,7 +20,7 @@
     "metro-config": "^0.83.0",
     "react": "19.2.5",
     "react-native": "0.83.4",
-    "react-native-audio-api": "0.12.0",
+    "react-native-audio-api": "0.11.5",
```
msluszniak: hey, why is that? We virtually never want to downgrade packages in demo apps.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/common/types/Word.h
```diff
@@ -4,13 +4,14 @@
 
 namespace rnexecutorch::models::speech_to_text {
 
+/**
+ * Basically a different representation of token,
```
msluszniak:
```suggestion
 * Different representation of token,
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/ASR.cpp
```diff
@@ -262,11 +263,21 @@ ASR::generate(std::span<const float> waveform, const DecodingOptions &options,
...
+  // Prefill: feed each initial token individually so decode() always sees 1
+  // token
+  std::span<uint64_t> firstToken(sequenceIds.data(), 1);
+  executorch::aten::Tensor logitsTensor =
+      this->decode(firstToken, encoderFeatures, startPos);
+  ++startPos;
+  for (size_t i = 1; i < sequenceIds.size(); ++i) {
+    std::span<uint64_t> single(sequenceIds.data() + i, 1);
+    logitsTensor = this->decode(single, encoderFeatures, startPos);
+    ++startPos;
```
msluszniak:
```suggestion
  for (size_t i = 1; i < sequenceIds.size(); ++i, ++startPos) {
    std::span<uint64_t> single(sequenceIds.data() + i, 1);
    logitsTensor = this->decode(single, encoderFeatures, startPos);
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -42,122 +34,176 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+
+      // Retain only the most recent EOS entry.
+      eos_.erase(eos_.begin(), eos_.end() - 1);
+      eos_[0].tmstpend -= secondTolastEntry.tmstpend;
     }
   }
 
-  return {.committed = move_to_vector(committed),
-          .nonCommitted = move_to_vector(nonCommitted)};
+  // Return the results
```
msluszniak:
```suggestion
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -42,122 +34,176 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+      const auto &secondTolastEntry = eos_[eos_.size() - 2];
+
+      const size_t cut = static_cast<size_t>(secondTolastEntry.tmstpend *
+                                             constants::kSamplingRate);
+      const size_t lastCommittedPos = secondTolastEntry.position;
+
+      audioBuffer_.erase(audioBuffer_.begin(), audioBuffer_.begin() + cut);
+
+      // Move all words up to the last committed position (inclusive) to the
+      // committed buffer.
```
msluszniak:
```suggestion
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -42,122 +34,176 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+  // Step 2: check if the newest EOS character from transcript should be
+  // saved to eos_ vector.
+  auto lastEosIt = std::find_if(words.rbegin(), words.rend(), utils::isEos);
+  if (lastEosIt != words.rend()) {
+    size_t lastEosIndex = std::distance(words.begin(), lastEosIt.base()) - 1;
+
+    // Because of step 1, we know that if the last EOS exist in eos_,
+    // then it must be the last entry.
+    if (eos_.empty() || eos_.back().position != lastEosIndex) {
+      // Register last EOS entry
```
msluszniak:
```suggestion
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -42,122 +34,176 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+      audioCopy.begin(),
+      audioCopy.begin() + std::min(constants::kMaxSamples, audioCopy.size()));
 
-  if (transcriptions.empty()) {
-    return {.committed = {}, .nonCommitted = {}};
-  }
+  std::vector<Segment> transcriptions = asr_->transcribe(input, options);
 
   // Flatten segments into a single word sequence.
+  // This is basically our 'nonCommitted' part for now.
```
msluszniak:
```suggestion
  // This is our 'nonCommitted' part for now.
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -1,37 +1,29 @@
...
 #include "Utils.h"
 
 namespace rnexecutorch::models::speech_to_text::whisper::stream {
 
-namespace {
-std::vector<Word> move_to_vector(std::deque<Word> &container) {
-  return std::vector<Word>(std::make_move_iterator(container.begin()),
-                           std::make_move_iterator(container.end()));
+OnlineASR::OnlineASR(const ASR *asr) : asr_(asr) {
+  // Reserve an expected amount of memory for audio buffer.
```
msluszniak:
```suggestion
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -42,122 +34,176 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
-std::vector<Word> OnlineASR::finish() {
-  // We always push the last remaining hypothesis, even if it's not
-  // confirmed in second iteration, to avoid ending up with broken sentences.
-  std::deque<Word> remaining = hypothesisBuffer_.hypothesis_;
+std::vector<Word> OnlineASR::finish(const DecodingOptions &options) {
+  ProcessResult result = process(options);
+
+  // Last-tick committed delta + whatever never made it past the commit
+  // threshold.
+  std::vector<Word> residual = std::move(result.committed);
```
msluszniak:
```suggestion
  std::vector<Word> residual{std::move(result.committed)};
```

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1325,14 +1338,17 @@ export const MODEL_REGISTRY = {
     STYLE_TRANSFER_UDNIE,
```
msluszniak, line 1351: Ok, so from 0.9 we will effectively drop support from our URL to original models (neither xnnpack nor coreml), right?

msluszniak (changes requested):

apps/speech/package.json
IgorSwat: `audio-api 0.12.0` causes build fails on iOS, and I think it's the same issue @benITo47 had when testing the 1.2.0 binaries some time ago.

packages/react-native-executorch/src/constants/modelUrls.ts
IgorSwat: I don't get it - the original models are XNNPACK ones, so they will still be available.

msluszniak: `WHISPER_TINY_EN_QUANTIZED` is quantized xnnpack, WHISPER_TINY_EN is I guess full precision xnnpack, since there is no WHISPER_TINY_EN_QUANTIZED we dropped something, what exactly?

apps/speech/package.json
msluszniak: Could you please name when you have these fails? I don't have any on iOS simulator.

packages/react-native-executorch/src/constants/modelUrls.ts
IgorSwat: Well, I just think the quantized models are pointless - they weigh only a little bit less than standard float32 models, there do not bring any significant inference speed up compared to baseline, and no one really downloads them on HF. I believe their existance just introduces unnecessary noise to the module.

msluszniak: I see, I'm ok with removing some of those, now the only question is what should we remove, quantized or non-quantized. If they are just a bit smaller and just a bit faster, they are still better than original one, aren't they?

IgorSwat: Well, float32 baseline models are well tested and surely at least as accurate as quantized (and probably more accurate). If performance difference is minimal (or frankly not existing) then I don't like the idea of risking accuracy drops for some type of inputs.

msluszniak: Sure thing, that explanation is absolutely fine for me, I mostly asked because I wanted to be on the same page :))

apps/speech/package.json
IgorSwat: image

Tested on physical device.

msluszniak: @mdydek could you look at this one? Maybe you have an intuition behind this error? @IgorSwat do you have iOS 26.2 on your physical device?

IgorSwat: Yeah, I even have 26.4.

msluszniak:
Also if this PR adds breaking change, please describe it directly below `Introduces a breaking change?` section in PR body.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/ASR.cpp
```diff
@@ -262,11 +263,21 @@ ASR::generate(std::span<const float> waveform, const DecodingOptions &options,
...
-                             constants::kMaxDecodeLength)) {
-    executorch::aten::Tensor logitsTensor =
-        this->decode(sequenceIds, encoderFeatures, startPos);
 
+  // Prefill: feed each initial token individually so decode() always sees 1
+  // token
+  std::span<uint64_t> firstToken(sequenceIds.data(), 1);
+  executorch::aten::Tensor logitsTensor =
+      this->decode(firstToken, encoderFeatures, startPos);
+  ++startPos;
```
msluszniak: Please abstract it into for loop, sth like this:
```cpp
executorch::aten::Tensor logitsTensor = nullptr;
for (size_t i = 0; i < sequenceIds.size(); ++i, ++startPos) {
  ...
}
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -1,35 +1,43 @@
...
+  audioBuffer_.reserve((constants::kChunkSize + 1) * constants::kSamplingRate);
 }
-} // namespace
 
-OnlineASR::OnlineASR(const ASR *asr) : asr_(asr) {
-  // Reserve a minimal expected amount of memory for audio buffer.
-  audioBuffer_.reserve(static_cast<size_t>(2 * params::kStreamChunkThreshold *
-                                           constants::kSamplingRate));
+bool OnlineASR::isReady() const {
+  std::scoped_lock<std::mutex> lock(streamingMutex);
```
msluszniak: `std::scope_lock` generally doesn't need to be explicitly templated with mutex type, you can drop it, please apply the same to the rest of the places.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -38,126 +46,213 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
 
   // Flatten segments into a single word sequence.
+  // This is basically our 'nonCommitted' part for now.
   std::vector<Word> words;
-  words.reserve(transcriptions.front().words.size());
-
   for (auto &segment : transcriptions) {
-    words.insert(words.end(), std::make_move_iterator(segment.words.begin()),
-                 std::make_move_iterator(segment.words.end()));
+    std::move(segment.words.begin(), segment.words.end(),
```
msluszniak: `std::ranges::move` with back_inserter

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -38,126 +46,213 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+
+  // Step 1: examine all previously saved EOS points.
+  // The idea is to remove entries which have changed or no longer exist
+  // due to model correcting it's output.
+  for (size_t i = 0; i < memory_.eos.size(); i++) {
+    const auto &eos = memory_.eos[i];
+    if (eos.position >= words.size() || !utils::isEos(words[eos.position]) ||
+        (eos.position > 0 &&
+         eos.preceeding != words[eos.position - 1].content)) {
+      memory_.eos.erase(memory_.eos.begin() + i, memory_.eos.end());
```
msluszniak: Probably for loop with iterators is more accurate here,

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -38,126 +46,213 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
   }
 
-  return {.committed = move_to_vector(committed),
-          .nonCommitted = move_to_vector(nonCommitted)};
+  // Step 4: clear the buffer if it is getting too large.
+  // The idea is to use the saved EOS entries and try to cut the buffer
+  // in a 'good' spot - where it will remove a significant audio chunk, yet
+  // won't affect most recent, unfinished speech samples.
+  size_t bufferSize = audioBuffer_.size();
+  if (bufferSize > static_cast<size_t>(params::kStreamSafeBufferDuration *
```
msluszniak: Use `std::cmp_greater` instead.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -38,126 +46,213 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+
+  // Reset memory.
+  memory_.transcript.clear();
+  memory_.eos.clear();
+  memory_.toCommit.clear();
+}
+
+std::vector<Word> OnlineASR::commitAndClean(std::vector<Word> &transcript) {
+  const size_t bufferSize = audioBuffer_.size();
+  const float midBufferThreshold = params::kStreamMaxDuration / 2.0F;
```
msluszniak: Probably might be constexpr

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -38,126 +46,213 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+
+  std::vector<Word> committed;
+
+  // If we don't have any EOS entries, then we most likely have not
+  // recorded any speech. In this case we can safely cut the maximum amount of
+  // audio data.
+  if (memory_.eos.empty()) {
+    size_t cut =
```
msluszniak, line 182: same, probably constexpr

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/Utils.h
```diff
@@ -1,77 +1,22 @@
...
-                     });
-
-      if (isEqual) {
-        return calculatedSize;
-      }
-    }
-  }
-
-  return 0;
+constexpr inline bool isEos(const Word &word) {
```
msluszniak: Will this be ever used in compile time scenario so you added `constexpr` here, cause I don't think so.

msluszniak:
Side note, after merging PR with TTS and rebasing, please make sure that native tests works here after all changes.

chmjkb (approved):

chmjkb (changes requested):
approved by an accident

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/ASR.cpp
```diff
@@ -437,15 +445,22 @@ ASR::estimateWordLevelTimestampsLinear(std::span<const uint64_t> tokens,
...
     prevCharCount += wSize;
 
-    // We store punctations separately to other characters.
+    // Detect and extract trailing punctuations.
     std::string puncts = "";
     while (!w.empty() && constants::kPunctations.contains(w.back())) {
       puncts += w.back();
       w.pop_back();
     }
     std::reverse(puncts.begin(), puncts.end());
```
msluszniak: std::ranges::reverse

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -1,163 +1,258 @@
...
   }
 
-  return {.committed = move_to_vector(committed),
-          .nonCommitted = move_to_vector(nonCommitted)};
+  // Step 4: clear the buffer if it is getting too large.
+  // The idea is to use the saved EOS entries and try to cut the buffer
+  // in a 'good' spot - where it will remove a significant audio chunk, yet
+  // won't affect most recent, unfinished speech samples.
+  size_t bufferSize = audioBuffer_.size();
+  if (std::cmp_greater<size_t, size_t>(bufferSize,
```
msluszniak: since `kStreamSafeBufferMaxSamples` is now size_t `std::cmp_greater` is not needed anymore.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -773,51 +773,45 @@ export const STYLE_TRANSFER_UDNIE_QUANTIZED = {
 } as const;
 
 // S2T
-const WHISPER_TINY_EN_TOKENIZER = `${URL_PREFIX}-whisper-tiny.en/${PREVIOUS_VERSION_TAG}/tokenizer.json`;
-const WHISPER_TINY_EN_MODEL = `${URL_PREFIX}-whisper-tiny.en/${PREVIOUS_VERSION_TAG}/xnnpack/whisper_tiny_en_xnnpack.pte`;
+const WHISPER_TINY_EN_TOKENIZER = `${URL_PREFIX}-whisper-tiny.en/${VERSION_TAG}/tokenizer.json`;
+const WHISPER_TINY_EN_MODEL_XNNPACK = `${URL_PREFIX}-whisper-tiny.en/${VERSION_TAG}/xnnpack/whisper_tiny_en_xnnpack_fp32.pte`;
+const WHISPER_TINY_EN_MODEL_COREML = `${URL_PREFIX}-whisper-tiny.en/${VERSION_TAG}/coreml/whisper_tiny_en_coreml_fp32.pte`;
```
chmjkb, line 778: We used to handle the backend selection automatically, as done for the style transfer, not a big problem as this is likely going to be re-written in the mogel registry PR cc @msluszniak

chmjkb (approved):
tested the demo app and works like a charm for iOS, thank u!

packages/react-native-executorch/src/constants/modelUrls.ts
msluszniak: Yeah, I will handle it in my PR.

msluszniak:
Native tests fail to configure on this rebased branch — `tests/CMakeLists.txt:265` still references `models/speech_to_text/whisper/HypothesisBuffer.cpp`, which this PR deletes:

```
CMake Error at CMakeLists.txt:137 (add_executable):
  Cannot find source file:

    .../models/speech_to_text/whisper/HypothesisBuffer.cpp

Call Stack (most recent call first):
  CMakeLists.txt:261 (add_rn_test)
```

Please drop the `HypothesisBuffer.cpp` source line (and any `HypothesisBuffer.h` includes in the speech-to-text test) so `bash run_tests.sh` builds again.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -1,163 +1,257 @@
...
+  //
+  // This prevents the audio buffer from growing indefinitely during continuous
+  // streaming. It is particularly useful when VAD (Voice Activity Detection)
+  // is used and elements are inserted but not processed for a long time.
+  // It should not pass the condition in a normal streaming, that is when
+  // process() method is called regularly within reasonable steps of time.
+  if (audioBuffer_.size() > constants::kMaxSamples) {
+    // Note that results are not actually committed now, but saved for
+    // a later call of process().
+    memory_.toCommit = commitAndClean(memory_.transcript);
```
msluszniak: Silent drop on rapid `insertAudioChunk` calls: this assigns rather than appends, so if the buffer cap is hit twice before `process()` consumes `memory_.toCommit`, the first batch is lost. Suggest appending:

```cpp
auto pending = commitAndClean(memory_.transcript);
std::ranges::move(pending, std::back_inserter(memory_.toCommit));
```

The header comment notes this is intended for the VAD-muted path where inserts pile up without processing — which is exactly the path where two cap-hits before a process() are plausible.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -158,10 +159,10 @@ void SpeechToText::stream(std::shared_ptr<jsi::Function> callback,
     // running transcriptions too rapidly (before the audio buffer is filled
     // with significant amount of new data) can cause streamer to commit wrong
     // phrases.
-    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    std::this_thread::sleep_for(std::chrono::milliseconds(timeout));
```
msluszniak: `sleep_for(timeout)` blocks `streamStop` for up to `timeout` ms — the loop only re-checks `isStreaming_` after the sleep returns, so the final flush is delayed by the full timeout when the user stops mid-pause. A `std::condition_variable` signalled by `streamInsert` / `streamStop`, with `wait_for(timeout, pred)` checking `!isStreaming_ || readyToProcess_`, would let stop take effect immediately and also let inserts wake the loop sooner than the next tick.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.h
```diff
@@ -42,7 +42,8 @@ class SpeechToText {
 
   // Stream
   void stream(std::shared_ptr<jsi::Function> callback,
-              std::string languageOption, bool enableTimestamps);
+              std::string languageOption, bool enableTimestamps,
+              uint32_t timeout);
```
msluszniak: Parameter name mismatch with the `.cpp` definition, which uses `bool verbose` ([SpeechToText.cpp:118](https://github.com/software-mansion/react-native-executorch/blob/1aebae62ef64abe920a4324f95f39812a3d20087/packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp#L118)). Rename here to `verbose` so the declaration and definition agree (the `.cpp` name matches the JS-side `DecodingOptions.verbose`).

msluszniak (commented):
Tested green on Android (demo app + native tests). A few suggestions inline, plus a few that touch lines unchanged by this PR — listed here since the lines aren't part of the diff and can't be commented inline:

- `src/types/stt.ts:10,12,14` — `SpeechToTextModelName` union still lists `'whisper-tiny-en-quantized' | 'whisper-base-en-quantized' | 'whisper-small-en-quantized'`. The quantized constants are deleted in this PR, so these literals now type-check but cannot be constructed from any built-in. Worth dropping in the same breaking-change.
- `common/rnexecutorch/models/speech_to_text/whisper/ASR.cpp:217` (preexisting) — divisor `tokens.size() + 1` matches neither a literal mean (`scores.size()`) nor OpenAI Whisper's formula (`len(full_seq) + 1`, where `full_seq` includes SOT prefix and EOT). Worth picking one explicitly. For reference, whisper.cpp uses `sum_logprobs / result_len` (no `+1`) — [src/whisper.cpp:6602-6603](https://github.com/ggml-org/whisper.cpp/blob/master/src/whisper.cpp#L6602-L6603).
- `common/rnexecutorch/models/speech_to_text/whisper/ASR.cpp:308` (preexisting) — `std::mt19937 gen((std::random_device{}()))` lives inside the autoregressive sampling loop, so `random_device` is consulted and a fresh Mersenne state is constructed for every sampled token. Hoist to a member (or `static thread_local`) seeded once per `generate()`.
- `common/rnexecutorch/models/speech_to_text/SpeechToText.h:38` (preexisting) — `transcribeStringOnly` is declared but never defined or referenced anywhere in the package; dead API surface, safe to drop.

Non-blocking — feel free to fold what you want into this PR or a follow-up.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -773,51 +773,45 @@ export const STYLE_TRANSFER_UDNIE_QUANTIZED = {
...
   tokenizerSource: WHISPER_TINY_EN_TOKENIZER,
 } as const;
 
-/**
- * @category Models - Speech To Text
- */
-export const WHISPER_TINY_EN_QUANTIZED = {
-  modelName: 'whisper-tiny-en-quantized',
+export const WHISPER_TINY_EN_COREML = {
+  modelName: 'whisper-tiny-en',
```
msluszniak, line 811: `modelName` collides with `WHISPER_TINY_EN` above — both set `'whisper-tiny-en'`. Same pattern in `WHISPER_BASE_EN_COREML`, `WHISPER_SMALL_EN_COREML`, `WHISPER_TINY_COREML`, `WHISPER_BASE_COREML`, `WHISPER_SMALL_COREML`. The JSDoc on `SpeechToTextModelConfig.modelName` says the field is "used for telemetry and hook reload triggers" — telemetry now can't distinguish backends, and consumers can't key cache invalidation on the model identity. Either give the CoreML constants distinct names (e.g. `'whisper-tiny-en-coreml'`, with matching union members in `stt.ts`) or update the JSDoc to clarify the field is backend-agnostic.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -773,51 +773,45 @@ export const STYLE_TRANSFER_UDNIE_QUANTIZED = {
...
+  modelSource: WHISPER_TINY_EN_MODEL_XNNPACK,
   tokenizerSource: WHISPER_TINY_EN_TOKENIZER,
 } as const;
 
-/**
- * @category Models - Speech To Text
- */
-export const WHISPER_TINY_EN_QUANTIZED = {
-  modelName: 'whisper-tiny-en-quantized',
+export const WHISPER_TINY_EN_COREML = {
```
msluszniak, line 810: Missing `@category Models - Speech To Text` JSDoc block — without it the typedoc autogen will misplace this constant. Also missing on `WHISPER_BASE_EN_COREML` (line 830), `WHISPER_SMALL_EN_COREML` (line 850), and `WHISPER_TINY_COREML` (line 870); the XNNPACK siblings and `WHISPER_BASE_COREML` / `WHISPER_SMALL_COREML` have it.

packages/react-native-executorch/src/constants/modelUrls.ts
msluszniak: You can also ignore this since in the PR with model registry, I'll change the url to be platform specific.

packages/react-native-executorch/src/constants/modelUrls.ts
msluszniak: Same as above, feel free to ignore since I will handle it by myself.

msluszniak (approved):
Looks good, added some minor improvements and tested all on android. If you want, you can retest on demo app on iOS, up to you. Overall great job on this one :))

chmjkb:
I tested demos on iOS before those changes and it worked good, guess ill retest tomorrow

chmjkb (approved):
tested demo apps on iOS, works fine

PR #1134, opened 2026-05-08
feat!: multilingual text-to-speech

### LINKED ISSUE #712: Text to Speech - add new languages support
### Problem Description
As it stands for now, the RNE implementation of the Kokoro model supports 2 languages: 🇺🇸 American English and 🇬🇧 British English. Hovewer, the Kokoro model itself provides a support for many other languages, such as:
- 🇪🇸 Spanish
- 🇫🇷 French
- 🇮🇹 Italian
- 🇧🇷 Brazilian Portuguese
- 🇨🇳 Mandarin Chinese
The only significant requirement to support those languages is to provide an appropriate phonemization, similar to the one for English.

There is also a possibility of adding an external language, not supported by Kokoro originally, by finetuning the model and training a new [voice](https://huggingface.co/hexgrad/Kokoro-82M/blob/main/VOICES.md).

### What should be done
In case of a language already supported by Kokoro:
1. Implement phonemization for given language in the phonemization package

In case of an external language:
1. Implement phonemization for given language in the phonemization package
2. Select an appropriate dataset containing speech samples for given language
3. Finetune the model to produce a new voice for given language

Expands the Text to Speech function accessability for non-english users.

IgorSwat:
## Description

Introduces major changes to the text-to-speech module based on Kokoro model, including:
- **Multilingual text-to-speech** - a set of complete pipelines & voices for different languages. A complete list of (currently) supported languages can be found below.
- **Improved phonemization & speech quality** - utilizing neural phonemization model as a fallback for the old lexicon-base phonemization significantly improves speech quality, particularly for non-standard, out of dictionary words.
- **Timestamp-based audio cutting** - an improve postprocessing algorithm, eliminates artifacts introduced by .pte model, resulting in cleaner, more natural speech.
- **API changes**: prepared for voice-cloning & custom, fine-tuned versions of Kokoro model.

Supported language current status:
- 🇺🇸 American English: ✅
- 🇬🇧 British English: ✅
- 🇫🇷 French: ✅
- 🇪🇸 Spanish: ✅
- 🇵🇹/🇧🇷 Portugese: ✅
- 🇮🇹 Italian: ✅
- 🇵🇱 Polish: ✅
- 🇩🇪 German: ✅
- 🇮🇳 Hindi: ✅
- 🇯🇵 Japanese: ❌ (coming soon)
- 🇨🇳 Mandarin Chinese: ❌ (coming soon)

- [x] Yes
- [ ] No

There are 2 major breaking changes introduced by this PR:
- Changed **"synthezation from phonemes"** API. 

   Old API:
   ```
    const audioData = await tts.forwardFromPhonemes({
      phonemes:
        'ɐ mˈæn hˌu dˈʌzᵊnt tɹˈʌst hɪmsˈɛlf, kæn nˈɛvəɹ ɹˈiᵊli tɹˈʌst ˈɛniwˌʌn ˈɛls.',
    });
    ```
    
    New API:
    ```
    const audioData = await tts.forward({
      text:
        'ɐ mˈæn hˌu dˈʌzᵊnt tɹˈʌst hɪmsˈɛlf, kæn nˈɛvəɹ ɹˈiᵊli tɹˈʌst ˈɛniwˌʌn ˈɛls.',
       phonemize: false,  # Disables phonemization and treats text as phonemes
    });
    ```
- Changed predefined model - voice setups. Now both model files & voice/phonemization files are bundled together, due to languages like Polish or German having fine-tuned model weights.

  Old API:
  ```
  const model = useTextToSpeech({
    model: KOKORO_MEDIUM,
    voice: KOKORO_VOICE_AF_HEART,
  });
  ```
  
  New API:
  ```
  const model = useTextToSpeech(KOKORO_AMERICAN_ENGLISH_FEMALE_HEART);
  ```

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Play around demo speech apps.

Unit tests for RNE-specific code will be added later on.
Phonemis package has it's own, wide range of unit tests implemented (see [Phonemis repo](https://github.com/IgorSwat/Phonemis))

### Related issues

#712 

### Checklist


packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
+                             size_t end, Separator sep) -> Cost {
+                       if (end - bp > limit) {
+                         return INF;
+                       }
+
+                       Cost sepPenalty = sep == Separator::EOS     ? 1
+                                         : sep == Separator::PAUSE ? 3
+                                         : sep == Separator::WHITE ? 1000
+                                                                   : 0;
+
```
msluszniak: Kinda magic number here

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
+        [limit](Cost acc, size_t beg, int64_t prevBp, int64_t bp, size_t end,
+                Separator sep) -> Cost {
+          if (end - bp > limit) {
+            return INF;
+          }
+
+          Cost sepPenalty = sep == Separator::EOS     ? 5
+                            : sep == Separator::PAUSE ? 18
+                            : sep == Separator::WHITE ? 1000
+                                                      : 0;
```
msluszniak: Again magic numbers.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
+                     [limit](Cost acc, size_t beg, int64_t prevBp, int64_t bp,
+                             size_t end, Separator sep) -> Cost {
+                       if (end - bp > limit) {
+                         return INF;
+                       }
+
+                       Cost sepPenalty = sep == Separator::EOS     ? 1
+                                         : sep == Separator::PAUSE ? 3
+                                         : sep == Separator::WHITE ? 1000
+                                                                   : 0;
```
msluszniak: Can you please abstract lambda into separate entity and then pass it to partition function? It's hard to ready right now, believe me I took me a while to understand what is going on here. Some thing like:

```cpp
auto minBreakCostFn = [limit](Cost acc, size_t beg, int64_t prevBp, int64_t bp,
                             size_t end, Separator sep) -> Cost { ... };
return partition(input, limit, minBreakCostFn);
```
Same below.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
+    size_t start = static_cast<size_t>(currBp + 1);
+    segments.emplace_back(start, lastIdx - start);
+    lastIdx = static_cast<size_t>(currBp + 1);
+    currBp = dp[currBp].second;
   }
+  // Add the first segment
+  segments.emplace_back(0, lastIdx);
 
-  std::ranges::reverse(result);
+  std::reverse(segments.begin(), segments.end());
```
msluszniak: why not std::ranges::reverse?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
```diff
@@ -28,8 +28,9 @@ Synthesizer::Synthesizer(const std::string &modelSource,
         forwardMethods_.emplace_back(name, inputSize);
       }
     }
-    std::stable_sort(forwardMethods_.begin(), forwardMethods_.end(),
-                     [](const auto &a, const auto &b) { return a.second < b.second; });
+    std::stable_sort(
```
msluszniak: `std::ranges::stable_sort`

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -9,86 +10,95 @@ namespace rnexecutorch::models::text_to_speech::kokoro::utils {
...
-        sum / kAudioCroppingSteps >= kAudioSilenceThreshold) {
-      return i;
+    // Check if moving average exceeds threshold
+    if (processedCount >= kAudioCroppingSteps &&
+        (windowSum / kAudioCroppingSteps) >= kAudioSilenceThreshold) {
+      return currentIndex;
     }
 
-    i = reverse ? i - 1 : i + 1;
+    currentIndex = reverse ? currentIndex - 1 : currentIndex + 1;
```
msluszniak:
```suggestion
    currentIndex += reverse ? -1 : 1;
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -9,86 +10,95 @@ namespace rnexecutorch::models::text_to_speech::kokoro::utils {
...
+  }
 
-  lbound = lbound > margin ? lbound - margin : 0;
-  rbound = std::min(rbound + margin, audio.size() > 0 ? audio.size() - 1 : 0);
+  size_t lbound = findAudioBound<false>(audio);
+  size_t rbound = findAudioBound<true>(audio);
 
-  return audio.subspan(lbound, rbound >= lbound ? rbound - lbound + 1 : 0);
+  // Apply margins
+  lbound = (lbound > margin) ? lbound - margin : 0;
```
msluszniak, line 63:
```suggestion
  lbound = std::max(lbound - margin, 0);
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -9,86 +10,95 @@ namespace rnexecutorch::models::text_to_speech::kokoro::utils {
...
+  rbound = std::min(rbound + margin, audio.size() - 1);
+
+  const size_t strippedLength = (rbound >= lbound) ? (rbound - lbound + 1) : 0;
```
msluszniak, line 66:
```suggestion
  const size_t strippedLength = std::max(rbound - lbound + 1,  0);
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -9,86 +10,95 @@ namespace rnexecutorch::models::text_to_speech::kokoro::utils {
...
+  const size_t totalLength = expectedSize.value_or(phonemes.size() + 2);
+  const size_t maxPhonemes = totalLength - 2;
+  const size_t effectivePhonemeCount = std::min(maxPhonemes, phonemes.size());
+
+  // 2. Initialize with pad tokens
+  std::vector<Token> tokens(totalLength, constants::kPadToken);
+
+  // 3. Map phonemes to vocabulary tokens
+  // Starting from index 1 to leave index 0 as start-padding
+  std::transform(phonemes.begin(), phonemes.begin() + effectivePhonemeCount,
```
msluszniak, line 87: `std::ranges::transform`

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -70,8 +70,9 @@ Pod::Spec.new do |s|
   # react-native-skia. The headers are preserved by preserve_paths and
   # then made available by HEADER_SEARCH_PATHS.
   s.exclude_files = [
-    "common/rnexecutorch/tests/**/*",
-    "common/rnexecutorch/jsi/*.{h,hpp}"
+    "common/rnexecutorch/tests/**/*.{cpp}",
```
msluszniak: Incorrect change, `"common/rnexecutorch/tests/**/*"` is correct. Probably you handle this incorrectly during rebase

msluszniak (changes requested):
You should also update the code in documentation and documentation in general. Also address lint warnings, there are plenty of them that you need to add to cspell ignore.

msluszniak:
Also if this PR adds breaking change, please describe it directly below Introduces a breaking change? section in PR body.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
       pausePoints.push_back(i);
-    } else if (phoneme < 256 && std::isspace(static_cast<char>(phoneme))) {
+    } else if (c < 256 && std::isspace(static_cast<char>(c))) {
       whitePoints.push_back(i);
     }
   }
 
-  std::vector<std::u32string> result = {};
+  std::vector<std::pair<size_t, size_t>> segments;
+  int64_t currBp = dp[n - 1].second;
```
msluszniak:
```suggestion
  int64_t currBp = dp.back().second;
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
-  // Note that because of backtracking, the order is reversed.
-  int32_t end = phonemes.size() - 1;
-  while (end != -1) {
-    int32_t begin = mem[end].second + 1;
-    result.push_back(phonemes.substr(begin, end - begin + 1));
-    end = mem[end].second;
+  while (currBp != -1) {
+    size_t start = static_cast<size_t>(currBp + 1);
+    segments.emplace_back(start, lastIdx - start);
+    lastIdx = static_cast<size_t>(currBp + 1);
```
msluszniak:
```suggestion
    lastIdx = start
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
```diff
@@ -43,11 +44,10 @@ Synthesizer::Synthesizer(const std::string &modelSource,
...
 
-Result<std::vector<EValue>> Synthesizer::generate(std::span<const Token> tokens,
-                                                  std::span<bool> textMask,
-                                                  std::span<int64_t> indices,
-                                                  std::span<float> dur,
-                                                  std::span<float> ref_s) {
+Result<std::vector<EValue>> Synthesizer::generate(
+    std::span<const Token> tokens, std::span<const bool> textMask,
+    std::span<const int64_t> indices, std::span<const float> dur,
+    std::span<const float> ref_s) {
```
msluszniak: You're passing const spans to almost immediatiely const cast them and strip that const, what is the reason for that?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
```diff
@@ -62,28 +62,33 @@ Result<std::vector<EValue>> Synthesizer::generate(std::span<const Token> tokens,
...
+      make_tensor_ptr({1, constants::kVoiceRefSize},
+                      const_cast<float *>(ref_s.data()), ScalarType::Float);
 
   // Select appropriate forward method based on token count
-  auto it = std::ranges::find_if(forwardMethods_,
-      [noTokens](const auto &entry) { return static_cast<int32_t>(entry.second) >= noTokens; });
-  std::string selectedMethod = (it != forwardMethods_.end()) ? it->first : forwardMethods_.back().first;
+  auto it =
+      std::ranges::find_if(forwardMethods_, [noTokens](const auto &entry) {
+        return static_cast<int32_t>(entry.second) >= noTokens;
```
msluszniak: `std::cmp_greater_equal` instead

apps/speech/components/ModelPicker.tsx
```diff
@@ -50,23 +55,22 @@ export function ModelPicker<T>({
...
+        pageX: number,
         pageY: number
       ) => {
-        setTriggerHeight(height);
         const spaceBelow = Dimensions.get('window').height - (pageY + height);
-        setExpandUp(spaceBelow < DROPDOWN_MAX_HEIGHT);
+        const y =
+          spaceBelow >= DROPDOWN_MAX_HEIGHT
+            ? pageY + height + 2
+            : pageY - Math.min(DROPDOWN_MAX_HEIGHT, models.length * 42) - 2;
```
chmjkb, line 67: a bunch of magic numbers, please explain

apps/speech/screens/TextToSpeechScreen.tsx
```diff
@@ -10,37 +10,52 @@ import {
...
+  { label: '🇬🇧 BF Emma', value: KOKORO_BRITISH_ENGLISH_FEMALE_EMMA },
+  { label: '🇬🇧 BM Daniel', value: KOKORO_BRITISH_ENGLISH_MALE_DANIEL },
+  { label: '🇫🇷 FF Siwis', value: KOKORO_FRENCH_FEMALE_SIWIS },
+  { label: '🇪🇸 EF Dora', value: KOKORO_SPANISH_FEMALE_DORA },
+  { label: '🇪🇸 EM Alex', value: KOKORO_SPANISH_MALE_ALEX },
+  { label: '🇮🇹 IF Sara', value: KOKORO_ITALIAN_FEMALE_SARA },
+  { label: '🇮🇹 IM Nicola', value: KOKORO_ITALIAN_MALE_NICOLA },
+  { label: '🇵🇹 PF Dora', value: KOKORO_PORTUGUESE_FEMALE_DORA },
+  { label: '🇵🇹 PM Santa', value: KOKORO_PORTUGUESE_MALE_SANTA },
+  { label: '🇵🇱 PM Mateusz', value: KOKORO_POLISH_MALE_MATEUSZ },
```
chmjkb, line 55: xD

chmjkb: i wonder who the fella is

apps/speech/screens/TextToSpeechScreen.tsx
```diff
@@ -77,23 +92,19 @@ const createAudioBufferFromVector = (
 };
 
 export const TextToSpeechScreen = ({ onBack }: { onBack: () => void }) => {
-  const [selectedModel, setSelectedModel] =
-    useState<KokoroConfig>(KOKORO_MEDIUM);
-  const [selectedVoice, setSelectedVoice] = useState<VoiceConfig>(
-    KOKORO_VOICE_AF_HEART
+  const [selectedVoice, setSelectedVoice] = useState<TextToSpeechModelConfig>(
```
chmjkb: this naming seems off

packages/react-native-executorch/android/src/main/cpp/CMakeLists.txt
```diff
@@ -88,9 +88,15 @@ endif()
 
 # ------- phonemis -------
 
-set(PHONEMIS_LIBS
-  "${LIBS_DIR}/phonemis/${ANDROID_ABI}/libphonemis.a"
-)
+set(PHONEMIS_DIR "${COMMON_THIRD_PARTY_DIR}/phonemis")
+
+set(BUILD_RUNNER OFF CACHE BOOL "" FORCE)
+set(BUILD_TESTS OFF CACHE BOOL "" FORCE)
```
chmjkb: Why do we need this change?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -4,25 +4,39 @@
...
+          .phonemizer =
+              phonemis::phonemizer::Config{
+                  .lang = lang,
+                  .lexicon_filepath = lexiconSource.empty()
+                                          ? std::nullopt
+                                          : std::make_optional(lexiconSource),
+                  .nn_model_filepath =
+                      neuralModelSource.empty()
+                          ? std::nullopt
+                          : std::make_optional(neuralModelSource)}}),
```
chmjkb, line 38: this is very hard to read

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Partitioner.cpp
```diff
@@ -13,117 +13,118 @@ using namespace params::partitioning;
...
-                                    mem[breakIdx].second, breakIdx, i + 1) +
-                             penalty;
-        if (newEstimation < estimation && breakIdx > 0) {
-          estimation = newEstimation;
-          prevBreakIdx = breakIdx;
+      Separator sep = q == &eosPoints     ? Separator::EOS
+                      : q == &pausePoints ? Separator::PAUSE
+                                          : Separator::WHITE;
+      for (size_t breakIdx : (*q)) {
+        Cost cost = costFn(dp[breakIdx].first, 0, dp[breakIdx].second, breakIdx,
```
chmjkb:
```suggestion
        auto cost = costFn(dp[breakIdx].first, 0, dp[breakIdx].second, breakIdx,
```

packages/react-native-executorch/src/constants/tts/voices.ts
```diff
@@ -1,84 +1,286 @@
...
+    taggerSource: KOKORO_PHONEMIZER_EN_GB_TAGGER,
+    lexiconSource: KOKORO_PHONEMIZER_EN_GB_LEXICON,
+    neuralModelSource: KOKORO_PHONEMIZER_EN_GB_MODEL,
+  },
+} as TextToSpeechModelConfig;
+
+/**
+ * @category TTS Supported Voices
+ */
+export const KOKORO_FRENCH_FEMALE_SIWIS = {
```
chmjkb, line 163: i think you meant swiss

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
```diff
@@ -54,48 +44,52 @@ export class TextToSpeechModule {
...
-      voice.lang,
-      phonemizerPaths[0],
-      phonemizerPaths[1],
-      modelPaths[0],
-      modelPaths[1],
-      voiceDataPath
+      phonemizerConfig.lang,
+      taggerIdx >= 0 ? (paths[taggerIdx] as string) : '',
+      lexiconIdx >= 0 ? (paths[lexiconIdx] as string) : '',
+      neuralModelIdx >= 0 ? (paths[neuralModelIdx] as string) : '',
```
chmjkb: do we need those assertions?

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
IgorSwat: Both `lbound` and `rbound` are of type `size_t`, which usually is resolved to an unsigned type. Therefore, the expression `lbound - margin` could cause an underflow.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
IgorSwat: Same as above.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
msluszniak: Right right, my bad.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
msluszniak: Same here, ignore.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
IgorSwat: I think `std::ranges` version would complicate the code in this case, as we do not simply run the transform on the entire string.

msluszniak: Yeap agreed, I missed this `+ effectivePhonemeCount` part

apps/speech/components/ModelPicker.tsx
IgorSwat: Come on, this is just a frontend of a demo app. Those values are indeed random numbers which make it look good 😅

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Synthesizer.cpp
IgorSwat: Well, I wanted to make it clear that these values are immutable within the method. But as I look at it now, I also think that it looks a little bit weird.

Will be changed to non-const.

packages/react-native-executorch/android/src/main/cpp/CMakeLists.txt
IgorSwat: We don't 👽

packages/react-native-executorch/src/constants/tts/voices.ts
IgorSwat: Naah, it's Siwis - they named this voice after the freanch speech dataset 'SIWIS'.

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
IgorSwat: Yes, we do. Tagger, lexicon and neural model are all theoretically optional, so we need some conditional to decide whether we should pass en empty string or an existing value.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
IgorSwat: I agree, but there are 2 things which prevent me from making it more readable:
- Each time I try to improve it by just re-formatting indentation, the linter reverts it during committing phase
- It's hard to simplify the code logic itself, since we always need these "empty path" checks to check what comes from the JS side, and these checks generate most of the complexity of this piece of code.

msluszniak: >the linter reverts it during committing phase

This is because we have clang-format call in pre-commit hook with LLVM default format enforced.

msluszniak (approved):
Please make sure that iOS is also tested since I don't have any for testing.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
chmjkb: i mean you could just extract this to separate variables, and it would be more readable

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
chmjkb: Well, I don't think you need a type assertion (`as string`) here. You could just perform a following check:
```
const [duration, synth, voice] = paths;
  if (!duration || !synth || !voice) {
    throw new Error('Kokoro: missing required model paths');
  }
```
and you don't need to lie to typescript about it

msluszniak: @chmjkb following check which is? 😅

chmjkb: updated the comment, github ui for some reason didnt render a code block properly

chmjkb: The as string casts are lying to TS, if paths[0] is undefined at runtime, you pass undefined to a native function expecting a string. Replacing that with an upfront throw costs a couple of lines and gives you a real error message when something breaks.

barhanc (commented):
I've tested the Speech app on iOS and standard usage is fine. There are however some lurking bugs, in Text to Speech app trying to generate speech using 'AF Heart' from text written in hindi alphabet (e.g. 'शुभ प्रभात') crashes the whole app with `offset + count out of range` in `audio = audio.subspan(0, lastTokenTimestamp);` in `rnexecutorch::models::text_to_speech::kokoro::Kokoro::synthesize`

msluszniak:
> I've tested the Speech app on iOS and standard usage is fine. There are however some lurking bugs, in Text to Speech app trying to generate speech using 'AF Heart' from text written in hindi alphabet (e.g. 'शुभ प्रभात') crashes the whole app with `offset + count out of range` in `audio = audio.subspan(0, lastTokenTimestamp);` in `rnexecutorch::models::text_to_speech::kokoro::Kokoro::synthesize`

Uuu, good catch. have you tried other characters specific for a language like ü, etc.?

barhanc:
> > I've tested the Speech app on iOS and standard usage is fine. There are however some lurking bugs, in Text to Speech app trying to generate speech using 'AF Heart' from text written in hindi alphabet (e.g. 'शुभ प्रभात') crashes the whole app with `offset + count out of range` in `audio = audio.subspan(0, lastTokenTimestamp);` in `rnexecutorch::models::text_to_speech::kokoro::Kokoro::synthesize`
> 
> Uuu, good catch. have you tried other characters specific for a language like ü, etc.?

Yeah, I tried the same for german-, french-, and spanish-specific characters and there wasn't any problem.

IgorSwat:
> I've tested the Speech app on iOS and standard usage is fine. There are however some lurking bugs, in Text to Speech app trying to generate speech using 'AF Heart' from text written in hindi alphabet (e.g. 'शुभ प्रभात') crashes the whole app with `offset + count out of range` in `audio = audio.subspan(0, lastTokenTimestamp);` in `rnexecutorch::models::text_to_speech::kokoro::Kokoro::synthesize`

Fixed.

msluszniak:
@IgorSwat inspired by Bartek's finding I'm trying some other other attack to expose some problem. Will come back with my finds.

packages/react-native-executorch/src/modules/natural_language_processing/TextToSpeechModule.ts
IgorSwat: I still don't get it. These asserts are only for dumb Typescript not yelling at me that the values can be undefined - when in fact thay can't be, because there is a check for it above, but TS does not understand it 😔

chmjkb: You're right there's a check at L78. Where I'd push back is, as string rewrites the type to whatever you claim, so we lose the compile-time check too. If paths ever becomes e.g. (string | number)[], as string
just silently lies and we ship a number to native instead of TS catching it.

Also the inline // DurationPredictor source comments are there because paths[0]/[1]/[2] aren't self-documenting. We can kill both with a destructure + a quick falsy check, names replace the comments:
```
  const [duration, synth, voice] = paths;
  if (!duration || !synth || !voice) {
    throw new RnExecutorchError(
      RnExecutorchErrorCode.DownloadInterrupted,
      'Kokoro: missing required model paths.'
    );
  }

  const tagger = paths[taggerIdx] ?? '';
  const lexicon = paths[lexiconIdx] ?? '';
  const neural = paths[neuralModelIdx] ?? '';

  return await global.loadTextToSpeechKokoro(
    phonemizerConfig.lang,
    tagger, lexicon, neural,
    duration, synth, voice,
  );
```
I don't think its very productive to keep arguing, but I think it's a better aproach to not use type assertions, if possible.

msluszniak:
## TTS edge-case findings from stress testing

Ran a battery of inputs against `Kokoro::generate` (via `forward()`) and the streaming path on Android. Bartek's exact crash is no longer reproducible — `e173e9d94` fixes it. The following are still open.

### 1. `speed` parameter has no validation
| `speed` | result |
|---|---|
| `0` | throws bare `std::exception` (no message, no error code) |
| `NaN` / `Infinity` / `-1` / `1e9` | silently accepted; emits ~5500 samples regardless of text |
| `1e-6` | emits 150 945 samples (≈6.3 s) for `"Hello world"` |

`1e-6` is the most worrying — `audioLength = kTicksPerDuration * effectiveDuration` is `int32_t` (`Kokoro.cpp:349`); a small enough `speed` overflows that and the synthesizer allocates unbounded memory. Suggested guard: reject non-finite or `≤ 0` speeds at the JS boundary *and* in `Kokoro::generate`, with a real `RnExecutorchError(InvalidUserInput, …)`.

### 2. Streaming worker hangs on non-EOS content
`Kokoro.cpp:171-189`:

```cpp
size_t chunkSize = (eosIt != inputTextBuffer_.rend())
                       ? std::distance(eosIt, inputTextBuffer_.rend())
                       : 0;

if (chunkSize > 0 ||
    streamSkippedIterations >= params::kStreamMaxSkippedIterations) {
  input = inputTextBuffer_.substr(0, chunkSize);   // chunkSize still 0
  inputTextBuffer_.erase(0, chunkSize);            // erases nothing
  streamSkippedIterations = 0;                     // reset, loop forever
}
```

When `streamInsert` content has no end-of-sentence character, the buffer never drains. The skip-threshold force-flush path fires but uses `chunkSize=0` to extract the chunk, so it produces an empty `input` and resets the counter. `streamStop(false)` then waits forever for the buffer to empty. `streamStop(true)` is the only recovery.

Repros: `streamInsert('a')`, `streamInsert('hello world')`, 2000× U+200D — all permanently hang the worker.

Suggested fix:

```cpp
if (chunkSize > 0) {
  // normal flush by EOS
} else if (streamSkippedIterations >= params::kStreamMaxSkippedIterations) {
  input = inputTextBuffer_.substr(0, searchLimit);
  inputTextBuffer_.erase(0, searchLimit);
  streamSkippedIterations = 0;
} else {
  streamSkippedIterations++;
}
```

### 3. `streamStop(true)` drops in-flight audio silently
`Kokoro.cpp:137-145`:

```cpp
auto nativeCallback = [this, callback](const std::vector<float> &audioVec) {
  if (this->isStreaming_) {        // false after streamStop(true)
    this->callInvoker_->invokeAsync(...);
  }
};
```

If `streamStop(true)` lands while a chunk is mid-synthesis, the synthesizer finishes the chunk and then the callback no-ops — the audio is generated and discarded with no signal. In a captioning / live-narration context that's a silently lost sentence.

Suggested fix: deliver the chunk that completed before the stop, or surface "aborted with in-flight chunk discarded" through `onEnd`/`onCancel`.

### 4. Observational (optional): solo punctuation produces ~0.5 s of audible artifacts
Single-character inputs like `.`, `!`, `?`, `...` produce 12k–17k samples of mostly-silent audio that contains low-amplitude artifacts the model emits while filling the duration predictor's window. `stripAudio`'s silence threshold doesn't catch them, so the user hears a faint click/breath. Not a crash, not strictly wrong — flagging as observed behavior in case it's worth a guard later (e.g. early-return when all content phonemes are punctuation).

---

### Reproducing

Stress-test version of the speech app lives on branch **[`@ms/tts-stress-tests`](https://github.com/software-mansion/react-native-executorch/tree/@ms/tts-stress-tests)** — the only changed file is `apps/speech/screens/TextToSpeechScreen.tsx` (preset chip rows + force-stop button + a switch from `model.stream()` to `model.forward()` so the hook's silent period-append doesn't mask the actual model behavior).

```bash
git fetch origin
git checkout @ms/tts-stress-tests
yarn install
cd apps/speech && yarn android
adb logcat -c && adb logcat ReactNativeJS:V AndroidRuntime:E DEBUG:V libc:E '*:F'
```

Open the app → **Text To Speech** screen.

**Top row — "Test presets"** drives `forward()` (one-shot synthesis, no streaming wrapper):
- `space`, `spaces`, `newline`, `dot`, `excl`, `q`, `...` — punctuation/empty edge cases (relates to finding 4)
- `Hindi`, `Arabic`, `Chinese`, `Japanese`, `Hebrew`, `Russian`, `Korean` — script-mismatch with English voice (Bartek's family ;p)
- `emoji`, `emoji-mix`, `ZW-chars`, `NUL`, `EN+Hindi`, `diacritics` — non-vocab / mixed-script
- `speed=0`, `speed=NaN`, `speed=Inf`, `speed=-1`, `speed=1e-6`, `speed=1e9` — drives finding 1
- `noPh:EN`, `noPh:nums`, `noPh:syms` — `phonemize: false` with non-phoneme input

**Bottom row — "Streaming tests"** drives `streamInsert` + `stream()` directly (bypassing the hook's `.` append):
- `no-term:a`, `no-term:long` — finding 2; these will hang the worker (tap the red **Force stop** button to recover)
- `many-EOS` — sanity check (multiple sentences in one insert)
- `insert-flood-EOS` — concurrency / buffer growth under load (no race observed)
- `race:stop-during-synth` — finding 3
- `race:insert-during-synth` — sanity check (no data loss observed)

Tap **Force stop** any time a streaming test hangs — it calls `streamStop(true)` so you can keep testing.

**Interpreting the logs.** Each tap emits one of:

```
I ReactNativeJS: [TTS-test]   text=<json> speed=<n> phonemize=<bool>
I ReactNativeJS: [TTS-test]   forward() returned <N> samples
I ReactNativeJS: [TTS-test]   threw: <message>

I ReactNativeJS: [TTS-stream] start: <label>
I ReactNativeJS: [TTS-stream] <label> chunk #<n>: <N> samples (t=<ms>)
I ReactNativeJS: [TTS-stream] end: <label> — <chunks> chunks, <samples> samples, <ms>ms
I ReactNativeJS: [TTS-stream] threw: <label> -> <message>
```

Quick decoder:
- `returned 0 samples` → safe no-op (input had no usable phonemes).
- `returned <large N>` with weird input → check whether the model produced unintended audio (findings 1 / 4).
- `threw: std::exception` with no detail → wrap with proper `RnExecutorchError` somewhere upstream.
- `start: <label>` followed by **no chunks** and a multi-minute duration → streaming worker is hung (finding 2); you need Force stop.
- `start: <label>` followed by **no chunks** but quick end (≲1 s) → in-flight chunk dropped (finding 3) — synthesizer ran, callback no-op'd.
- `chunk #N` lines mean audio was delivered to JS. The streaming sanity tests (`many-EOS`, `insert-flood-EOS`, `race:insert-during-synth`) should each produce several chunks.

msluszniak (changes requested):
Comment above

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -133,6 +138,11 @@ std::vector<float> Kokoro::generate(std::u32string input, float speed,
 
 void Kokoro::stream(std::shared_ptr<jsi::Function> callback, float speed,
                     bool phonemize, bool stopOnEmptyBuffer) {
+  if (speed < constants::kMinValidSpeed || speed > constants::kMaxValidSpeed) {
```
msluszniak: Don't we want to split it into two separate errors, where you can tell limits explicitly, e.g. speed is to low, minimum speed is 0.1, but you passed `<xyz>`.

IgorSwat:
@msluszniak 
1. I added checks for speed parameter
2. I know this issue for a long time. The thing is, for normal (one-shot) inputs it never happens, cause we always add a dot '.' if no EOS character is present at the end. And I once changed it in exact the same way as your "reviewer" suggested, but then the LLM streaming mode often fails, because it's very hard to adjust the kStreamMaxSkippedIterations parameter to not stop the streaming prematurely, because it's dependent on LLM generation speed. So fixing it properly is a non-trivial task, and I don't think we have time for that.
3. Harmless issue, I don't see a point in complicating the API further for something like that.
4. Another harmless issue. I don't see a point in covering all the "stupid" inputs from user, unless it results in a direct crash.

msluszniak:
I agree that 4 is rather good to skip. 
Regarding 3, ok I just wanted to be sure you are aware of this buggy behaviour.

Regarding 2:
> cause we always add a dot '.'

Where do we add this dot?

IgorSwat:
> Where do we add this dot?

`useTextToSpeech.ts`, lines 107 to 112.

msluszniak:
> > Where do we add this dot?
> 
> `useTextToSpeech.ts`, lines 107 to 112.

Ok, what about textToSpeechModule? I cannot see the similar hack.

IgorSwat:
> Ok, what about textToSpeechModule? I cannot see the similar hack.

We can't do that in textToSpeechModule. And I honestly do not see a point in doing so, if it is already done in the hook.

msluszniak:
Hmmm, the very minimum we need to do is to escalate this into separate issue since it is a serious problem. After 0.9 release I will work on the solution that both solves this issue and do not break llm integration.

And by the way, hook is completely separate mechanism, if we have this hack in module that is re-used by the hook, then ok. But other way around, I disagree.

msluszniak (approved):
As said, either me or you need to work on the eos issue, but except that, I have no other things to add. Great job overall! :))

chmjkb (approved):

PR #1139, opened 2026-05-11
fix(constants): inline `LIB_VERSION` via codegen to fix compiled output

### LINKED ISSUE #1138: `LIB_VERSION` resolves to non-existent `lib/package.json` after `bob build`, breaks web/Node consumers
### Summary

`src/constants/resourceFetcher.ts` reads the library version with a runtime `require`:

```ts
export const LIB_VERSION: string = require('../../package.json').version;
```

That relative path is correct from `src/constants/` (two-up = package root), but `react-native-builder-bob`'s `module` target emits the file to `lib/module/constants/` — one level deeper — and the babel transform doesn't rewrite the relative `require`. So the compiled output ships `require('../../package.json')` unchanged, which from the new location resolves to `lib/package.json` — a file that does not exist.

### Why this only bites some consumers

- Metro for **iOS/Android** respects the package's `"react-native": "src/index"` field, so it reads the source `src/constants/resourceFetcher.ts` and the require path is fine.
- **Web** (Metro with `react-native-web`), Node, and anything else that falls back to `"main"` / `"module"` reads `./lib/module/index.js` and transitively `lib/module/constants/resourceFetcher.js` — where the require path is broken.

### Repro

In a fresh Expo app with `react-native-web`:

```bash
npx create-expo-app repro
cd repro
npx expo install react-native-executorch
# Import anything from it in app/_layout.tsx, e.g. initExecutorch
npx expo start --web
```

Bundler error:

```
Unable to resolve module ../../package.json from
node_modules/react-native-executorch/lib/module/constants/resourceFetcher.js:
None of these files exist:
  * node_modules/react-native-executorch/lib/package.json(...)
```

### Versions affected

- `0.8.4` (latest published at time of writing): confirmed
- `0.9.0` (HEAD of `main`): same source, same build layout — also affected

### Diagnosis

- Source: `packages/react-native-executorch/src/constants/resourceFetcher.ts`
- Compiled file that breaks: `packages/react-native-executorch/lib/module/constants/resourceFetcher.js`
- The sourcemap (`resourceFetcher.js.map`) even has `"sourceRoot":"../../../src"` — the build knows the source is three levels up — but the babel pass leaves the `require('../../package.json')` literal untouched.

### Suggested fix

Replace the dynamic JSON require with a build-time-synced constant: a small `scripts/sync-version.cjs` that reads `package.json#version` and writes `src/constants/libVersion.ts`, wired into `scripts.prepare` (before `bob build`) and `scripts.version` (so `npm version <bump>` keeps the committed file in sync). I have a PR ready against `main` that does this — happy to open it alongside this issue.

### Workaround until fixed

In `node_modules/react-native-executorch/lib/module/constants/resourceFetcher.js`, change `require('../../package.json')` to `require('../../../package.json')`. Persist with `patch-package` if you need it to survive `npm install`.

marnelram:
## Description

`src/constants/resourceFetcher.ts` reads the library version with a runtime `require('../../package.json').version`. That path resolves correctly from `src/constants/`, but `react-native-builder-bob` emits the file at `lib/module/constants/` — one level deeper — and babel leaves the relative `require` literal unchanged. From the compiled location it points at a non-existent `lib/package.json`, so any consumer that resolves via `main`/`module` (web, Node) fails to bundle. iOS/Android Metro is unaffected because the package's `"react-native": "src/index"` entry reads source directly.

Replace the JSON `require` with a literal `LIB_VERSION` in `constants/versions.ts`, derive the current-release tag from it, and re-export `LIB_VERSION` from `resourceFetcher.ts` for existing callers. While here, rename the version constants to match what they actually mean at the call site: `VERSION_TAG` is now the current release (derived from `LIB_VERSION`) and `PREVIOUS_VERSION_TAG` is the manual previous-release string — kept manual because the previous release isn't reliably derivable from `LIB_VERSION` across major bumps. All consumers in `src/constants` are updated; hosted URLs are unchanged.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. `yarn workspace react-native-executorch run prepare`.
2. Inspect `packages/react-native-executorch/lib/module/constants/resourceFetcher.js` — it should re-export `LIB_VERSION` from `./versions` with no `require('../../package.json')` left.
3. `yarn typecheck` passes — covers the rename across `versions.ts`, `modelUrls.ts`, `ocr/models.ts`, `tts/models.ts`, `tts/voices.ts`.
4. In a fresh Expo app with `react-native-web`, install the package from this branch and run `npx expo start --web` — the bundler no longer errors on the missing `lib/package.json`.
5. iOS/Android demo apps continue to build and report the expected version. Exercise a model pinned to `PREVIOUS_VERSION_TAG` (e.g. any Llama-3.2 / Qwen-3 / Whisper download) and one pinned to `VERSION_TAG` (e.g. Qwen-3.5, YOLO26-pose, distiluse multilingual, privacy-filter) — both should resolve to the same HuggingFace URLs as before the rename.

### Related issues

Fixes #1138.

### Checklist

### Additional notes

`LIB_VERSION` is hardcoded next to the other version constants, so it must be bumped manually alongside `package.json#version` (same release step that already touches `PREVIOUS_VERSION_TAG`). `VERSION_TAG` is derived from `LIB_VERSION`, so it updates automatically. A codegen approach was considered but not adopted to keep the fix minimal.

### DIFF
--- a/packages/react-native-executorch/src/constants/modelUrls.ts
+++ b/packages/react-native-executorch/src/constants/modelUrls.ts
@@ -1013,12 +1013,12 @@ export const SELFIE_SEGMENTATION = {
 // FastSAM Instance Segmentation
 const FASTSAM_S_SEG_MODEL =
   Platform.OS === 'ios'
-    ? `${URL_PREFIX}-fast-sam/${NEXT_VERSION_TAG}/fastsam-s/coreml/fastsam_s_coreml_fp16.pte`
-    : `${URL_PREFIX}-fast-sam/${NEXT_VERSION_TAG}/fastsam-s/xnnpack/fastsam_s_xnnpack_fp32.pte`;
+    ? `${URL_PREFIX}-fast-sam/${VERSION_TAG}/fastsam-s/coreml/fastsam_s_coreml_fp16.pte`
+    : `${URL_PREFIX}-fast-sam/${VERSION_TAG}/fastsam-s/xnnpack/fastsam_s_xnnpack_fp32.pte`;
 const FASTSAM_X_SEG_MODEL =
   Platform.OS === 'ios'
-    ? `${URL_PREFIX}-fast-sam/${NEXT_VERSION_TAG}/fastsam-x/coreml/fastsam_x_coreml_fp16.pte`
-    : `${URL_PREFIX}-fast-sam/${NEXT_VERSION_TAG}/fastsam-x/xnnpack/fastsam_x_xnnpack_fp32.pte`;
+    ? `${URL_PREFIX}-fast-sam/${VERSION_TAG}/fastsam-x/coreml/fastsam_x_coreml_fp16.pte`
+    : `${URL_PREFIX}-fast-sam/${VERSION_TAG}/fastsam-x/xnnpack/fastsam_x_xnnpack_fp32.pte`;
 
 /**
  * @category Models - Instance Segmentation
--- a/packages/react-native-executorch/src/constants/modelUrls.ts
+++ b/packages/react-native-executorch/src/constants/modelUrls.ts
@@ -3,19 +3,19 @@ import {
   PRIVACY_FILTER_NEMOTRON_LABELS,
   PRIVACY_FILTER_OPENAI_LABELS,
 } from './privacyFilterLabels';
-import { URL_PREFIX, VERSION_TAG, NEXT_VERSION_TAG } from './versions';
+import { URL_PREFIX, PREVIOUS_VERSION_TAG, VERSION_TAG } from './versions';
 
 // LLMs
 
 // LLAMA 3.2
-const LLAMA3_2_3B_MODEL = `${URL_PREFIX}-llama-3.2/${VERSION_TAG}/llama-3.2-3B/original/llama3_2_3B_bf16.pte`;
-const LLAMA3_2_3B_QLORA_MODEL = `${URL_PREFIX}-llama-3.2/${VERSION_TAG}/llama-3.2-3B/QLoRA/llama3_2-3B_qat_lora.pte`;
-const LLAMA3_2_3B_SPINQUANT_MODEL = `${URL_PREFIX}-llama-3.2/${VERSION_TAG}/llama-3.2-3B/spinquant/llama3_2_3B_spinquant.pte`;
-const LLAMA3_2_1B_MODEL = `${URL_PREFIX}-llama-3.2/${VERSION_TAG}/llama-3.2-1B/original/llama3_2_bf16.pte`;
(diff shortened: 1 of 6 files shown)


chmjkb:
cc @mkopcins, i believe we can also use that new constant for the model urls to avoid duplicating

msluszniak:
> cc @mkopcins, i believe we can also use that new constant for the model urls to avoid duplicating

Do you mean something like this one?

[packages/react-native-executorch/versions.json]
```json
{ "previousVersionTag": "v0.8.0" }
```

[packages/react-native-executorch/scripts/sync-versions.cjs]

```cjs
const fs = require('fs');
const path = require('path');

const pkg = require('../package.json');
const input = require('../versions.json');

const current = `v${pkg.version}`;
const previous = input.previousVersionTag;

const out = `// AUTO-GENERATED by scripts/sync-versions.cjs — do not edit by hand.
// Sources: package.json#version (current) + versions.json#previousVersionTag (previous).
export const URL_PREFIX =
export const VERSION_TAG = 'resolve/${previous}';
export const NEXT_VERSION_TAG = 'resolve/${current}';
export const LIB_VERSION = '${pkg.version}';
`;

const target = path.join(__dirname, '..', 'src', 'constants', 'versions.ts');
fs.writeFileSync(target, out);
console.log(`[sync-versions] LIB_VERSION=${pkg.version}, VERSION_TAG=${previous}, NEXT_VERSION_TAG=${current}`);
```

[packages/react-native-executorch/src/constants/versions.ts]
```cjs
// AUTO-GENERATED by scripts/sync-versions.cjs — do not edit by hand.
export const URL_PREFIX =
export const VERSION_TAG = 'resolve/v0.8.0';
export const NEXT_VERSION_TAG = 'resolve/v0.9.0';
export const LIB_VERSION = '0.9.0';
```

[packages/react-native-executorch/src/constants/resourceFetcher.ts]

```ts
export const DOWNLOAD_EVENT_ENDPOINT =
  'https://ai.swmansion.com/telemetry/downloads/api/downloads';
export { LIB_VERSION } from './versions';
```

[packages/react-native-executorch/package.json]
```json
"prepare": "node scripts/sync-versions.cjs && bob build",
"version": "node scripts/sync-versions.cjs && git add src/constants/versions.ts"
```

msluszniak:
Ok, we eventually decided, that we will include this lib version in the already existing file. Then, we will be able to completely remove codegen part. Regardless, way need to change the version manually during release.

msluszniak:
We decided on changing `VERSION_TAG` to `PREVIOUS_VERSION_TAG` and `NEXT_VERSION_TAG` to `VERSION_TAG`.

NorbertKlockiewicz (approved):

msluszniak:
@marnelram Could you check if this changes works for you in the current shape?

PR #1141, opened 2026-05-12
refactor: streamline error handling

### LINKED ISSUE #1119: Add a ContractNotMet error and apply it to existing codebase
see: https://github.com/software-mansion/react-native-executorch/pull/1100#discussion_r3180869184

chmjkb:
## Description

I decided that defining a specific error for not meeting the contract is probably an overkill, as the input layer can be ine, output layer can be wrong and we already have the code that can distinguish between those. I added comments and I refactored the both native and JS error handling a bit
It also fixes a few typos & bugs

- [ ] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- Trigger a download interruption (kill network mid-fetch, or pass a bad URL) for any model — confirm thrown error has code DownloadInterrupted and the expected "The download has been interrupted..." message
  - Call forward() on a module before load() resolves — confirm code ModuleNotLoaded and "...before calling forward()." message
  - Kick off a second forward() while the first is still running — confirm code ModelGenerating and "...wait until previous model run is complete." message
  - Hit one of the overridden sites to make sure custom messages still flow through: e.g. LLMController.generate() with empty messages array (should say "Messages array is empty!", code InvalidUserInput), or
  BaseOCRController.internalLoad with an unsupported language ("...not supported. Please try using other language.", code LanguageNotSupported)
  - Spot-check a couple modules across categories — LLM, OCR, SpeechToText, VAD, TextToImage, Classification — make sure load + forward still work end-to-end on device (the constructor signature change shouldn't
  affect runtime)
  - Verify error.code is still set correctly on caught errors in JS (try/catch then check e.code and e.message)

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/Error.h
+++ b/packages/react-native-executorch/common/rnexecutorch/Error.h
@@ -1,36 +1,13 @@
 #pragma once
 
 #include <executorch/runtime/core/error.h>
+#include <source_location>
 #include <stdexcept>
 #include <string>
+#include <string_view>
 #include <variant>
 
 #include <rnexecutorch/ErrorCodes.h>
-
-namespace rnexecutorch::detail {
-constexpr const char *basename(const char *path) noexcept {
-  const char *base = path;
-  while (*path) {
-    if (*path++ == '/')
-      base = path;
-  }
-  return base;
-}
-} // namespace rnexecutorch::detail
-
-// clang-format off
-#define CHECK_OK_OR_THROW_FORWARD_ERROR(result) \
-  if (!(result).ok()) \
-    throw RnExecutorchError( \
-        (result).error(), std::string("[") + ::rnexecutorch::detail::basename(__FILE__) + \
-        "] Forward pass failed (in: " + __func__ + "). Ensure the model input is correct.")
-
-#define THROW_NOT_LOADED_ERROR() \
-  throw RnExecutorchError( \
-      RnExecutorchErrorCode::ModuleNotLoaded, \
-      std::string("[") + ::rnexecutorch::detail::basename(__FILE__) + \
-      "] Model not loaded (in: " + __func__ + ")")
-// clang-format on
 
 namespace rnexecutorch {
 
@@ -59,4 +36,40 @@ public:
   }
 };
 
+namespace detail {
+
+// npos + 1 wraps to 0, so the no-slash case returns the whole path.
+constexpr std::string_view basename(std::string_view path) noexcept {
+  return path.substr(path.find_last_of('/') + 1);
+}
+
+inline std::string locationPrefix(const std::source_location &loc) {
+  return "[" + std::string(basename(loc.file_name())) + "] ";
+}
+
+[[noreturn]] inline void
+throwNotLoaded(std::source_location loc = std::source_location::current()) {
+  throw RnExecutorchError(RnExecutorchErrorCode::ModuleNotLoaded,
+                          locationPrefix(loc) + "Model not loaded (in: " +
+                              loc.function_name() + ")");
+}
+
+template <typename Result>
+inline void checkOkOrThrowForwardError(
+    const Result &result,
+    std::source_location loc = std::source_location::current()) {
+  if (!result.ok()) {
+    throw RnExecutorchError(
+        result.error(), locationPrefix(loc) +
+                            "Forward pass failed (in: " + loc.function_name() +
+                            "). Ensure the model input is correct.");
+  }
+}
+
+} // namespace detail
 } // namespace rnexecutorch
+
+#define CHECK_OK_OR_THROW_FORWARD_ERROR(result)                                \
+  ::rnexecutorch::detail::checkOkOrThrowForwardError(result)
+
+#define THROW_NOT_LOADED_ERROR() ::rnexecutorch::detail::throwNotLoaded()
--- a/packages/react-native-executorch/common/rnexecutorch/Error.h
+++ b/packages/react-native-executorch/common/rnexecutorch/Error.h
@@ -7,6 +7,31 @@
 
 #include <rnexecutorch/ErrorCodes.h>
 
+namespace rnexecutorch::detail {
+constexpr const char *basename(const char *path) noexcept {
+  const char *base = path;
+  while (*path) {
+    if (*path++ == '/')
+      base = path;
+  }
+  return base;
+}
+} // namespace rnexecutorch::detail
+
+// clang-format off
+#define CHECK_OK_OR_THROW_FORWARD_ERROR(result) \
+  if (!(result).ok()) \
+    throw RnExecutorchError( \
+        (result).error(), std::string("[") + ::rnexecutorch::detail::basename(__FILE__) + \
+        "] Forward pass failed (in: " + __func__ + "). Ensure the model input is correct.")
+
+#define THROW_NOT_LOADED_ERROR() \
+  throw RnExecutorchError( \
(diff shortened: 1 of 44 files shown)


packages/react-native-executorch/common/rnexecutorch/Error.h
```diff
@@ -7,6 +7,31 @@
...
+  if (!(result).ok()) \
+    throw RnExecutorchError( \
+        (result).error(), std::string("[") + ::rnexecutorch::detail::basename(__FILE__) + \
+        "] Forward pass failed (in: " + __func__ + "). Ensure the model input is correct.")
+
+#define THROW_NOT_LOADED_ERROR() \
+  throw RnExecutorchError( \
+      RnExecutorchErrorCode::ModuleNotLoaded, \
+      std::string("[") + ::rnexecutorch::detail::basename(__FILE__) + \
+      "] Model not loaded (in: " + __func__ + ")")
```
msluszniak: I think in c++20 there are better ways to get file name, checkout this: https://en.cppreference.com/cpp/utility/source_location/file_name, https://en.cppreference.com/cpp/utility/source_location/function_name and maybe this https://en.cppreference.com/cpp/preprocessor/line

scripts/errors.config.ts
```diff
@@ -38,8 +38,7 @@ export const errorDefinitions = {
    */
   FileReadFailed: 0x72,
   /*
-   * Thrown when the size of model output is unexpected.
-   */
+   * Thrown when the size of model output is unexpected. If you're using your custom model with any of the pre-defined modeules, please verify docs or source code for the expected model I/O contract.
   InvalidModelOutput: 0x73,
   /*
```
msluszniak: Please run this script to update the code since now `InvalidModelOutput: 0x73,` is in the comment block.

scripts/errors.config.ts
```diff
@@ -38,8 +38,7 @@ export const errorDefinitions = {
    */
   FileReadFailed: 0x72,
   /*
-   * Thrown when the size of model output is unexpected.
-   */
+   * Thrown when the size of model output is unexpected. If you're using your custom model with any of the pre-defined modeules, please verify docs or source code for the expected model I/O contract.
```
msluszniak: nit: typo `modeules` → `modules` (also on line 64).

packages/react-native-executorch/common/rnexecutorch/Error.h
```diff
@@ -7,6 +7,31 @@
...
+  }
+  return base;
+}
+} // namespace rnexecutorch::detail
+
+// clang-format off
+#define CHECK_OK_OR_THROW_FORWARD_ERROR(result) \
+  if (!(result).ok()) \
+    throw RnExecutorchError( \
+        (result).error(), std::string("[") + ::rnexecutorch::detail::basename(__FILE__) + \
```
msluszniak: Heads-up: this macro expands to a bare `if` without braces, so it's vulnerable to the dangling-else trap if anyone wraps a call in an outer `if/else` later. Either wrap the body in `do { ... } while (0)`, or — if we go the `std::source_location` route — both macros become inline templates and this concern disappears entirely.

(`THROW_NOT_LOADED_ERROR` below is a single `throw` so it's not affected.)

packages/react-native-executorch/src/errors/errorUtils.ts
```diff
@@ -14,21 +28,14 @@ export class RnExecutorchError extends Error {
    */
   public cause?: unknown;
 
-  constructor(code: number, message: string, cause?: unknown) {
-    super(message);
-    /**
-     * The error code representing the type of error.
-     */
+  constructor(code: RnExecutorchErrorCode, message?: string, cause?: unknown) {
```
msluszniak: The narrowing from `number` to `RnExecutorchErrorCode` here is silently undone by `parseUnknownError` below — `new RnExecutorchError(e.code, ...)` where `e.code: number` from the duck-typed `isRnExecutorchErrorLike`. It still compiles (enums accept numbers), so the stricter type doesn't really buy us anything. Either widen back to `number | RnExecutorchErrorCode`, or cast at the call site.

packages/react-native-executorch/src/errors/errorUtils.ts
```diff
@@ -14,21 +28,14 @@ export class RnExecutorchError extends Error {
...
+    const resolved =
+      message ??
+      DefaultErrorMessages[code] ??
+      `RnExecutorch error (code ${code})`;
```
msluszniak: If neither a message nor a default exists, the user gets `RnExecutorch error (code 117)`. Using `RnExecutorchErrorCode[code]` for the reverse-lookup would surface the enum name (e.g. `InvalidUserInput`), which is much more actionable than a raw decimal.

msluszniak:
While we're on error handling — there's a related issue I see often in the demo apps: native errors surface as `[object Object]` when stringified.

**Root cause:** in [`ModelHostObject.h:263-268`](https://github.com/software-mansion/react-native-executorch/blob/%40chmjkb/contract-not-met-error/packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h#L263-L268) (and three near-identical blocks at 333-338, 405-407, 440-443, plus one in `RnExecutorchInstaller.h:99`) we throw a **plain object**, not an `Error` instance:

```cpp
jsi::Object errorData(runtime);
errorData.setProperty(runtime, "code", e.getNumericCode());
errorData.setProperty(runtime, "message", jsi::String::createFromUtf8(runtime, e.what()));
throw jsi::JSError(runtime, jsi::Value(runtime, std::move(errorData)));
```

JSI rethrows that value verbatim, so on the JS side `e` is a bare object — `e instanceof Error` is `false`, and `String(e)` / template strings / `<Text>{e}</Text>` all coerce to `[object Object]`. The data is actually there (`e.code` and `e.message` are populated), but every JS surface that stringifies it loses it.

Most demo apps follow this pattern:

```ts
setError(e instanceof Error ? e.message : String(e));
```

which is exactly what trips. The Pose Estimation screen is the only one that handles errors properly via `e instanceof RnExecutorchError` + `switch (e.code)`.

**Suggested fix (native side):** drop a helper into `Error.h` that constructs a real `Error` and tags `code` onto it:

```cpp
[[noreturn]] inline void throwAsJsError(
    facebook::jsi::Runtime &runtime, int32_t code, const std::string &message) {
  auto errorCtor = runtime.global().getPropertyAsFunction(runtime, "Error");
  auto errorObj = errorCtor
      .callAsConstructor(runtime, jsi::String::createFromUtf8(runtime, message))
      .getObject(runtime);
  errorObj.setProperty(runtime, "code", code);
  throw jsi::JSError(runtime, jsi::Value(runtime, std::move(errorObj)));
}
```

Each catch block then collapses to:

```cpp
} catch (const RnExecutorchError &e) {
  throwAsJsError(runtime, e.getNumericCode(), e.what());
}
```

Result on JS:
- `e instanceof Error` → `true`
- `String(e)` / `${e}` → `"Error: <message>"`
- `e.code` / `e.message` still work
- `parseUnknownError` still routes through `isRnExecutorchErrorLike` (code+message check runs before `instanceof Error`), so the code is preserved end-to-end

After this, the existing `e instanceof Error ? e.message : String(e)` pattern in the demo apps will start displaying the real message without any JS-side changes. Cleaning up the demo apps to extract `.code` for richer UX could be a follow-up.

Probably worth its own PR (or folded into this one since it's already on the error-handling theme) — happy to draft it.

EDIT: Moved to a separate issue → #1150.

msluszniak (approved):

Issue #431, opened 2025-07-01
Inconsistencies in model paths on Hugging Face


msluszniak:
As mentioned in #382 I'm moving this into separate issue:
Inconsistencies in naming across hugging face repository:

* Only for `llama3_2-3B_qat_lora.pte` there is `-` between name and number of parameters, while for the rest is `_`.
* For `WHISPER_TINY_DECODER` `en` is concatenated in the first part of URL with `.` and not with `-` as in the rest of examples.
* There is word all in `efficientnet_v2_s_coreml_all` and only for that model.
* For some of the models initial name is concatenated with `-` and the model name as well `-` `phi-4-mini` and `phi-4-mini_bf16.pte` while in others there is `-` and `_` e.g. `qwen-2.5` and `qwen2_5_3b_8da4w.pte`.
* Some model names adds backend info at the beginning: `xnnpack_whisper_encoder.pte` and some on the end or almost end: `style_transfer_udnie_xnnpack.pte`.
* `ssdlite320-mobilenet-v3-large` `v3` here is concatenated with `-` but in model name not: `ssdlite320-mobilenetv3-large.pte`.
* For craft detector, number means size of the input tensor while for the other suffixed numbers means model sizes.
* Some models `B` and `M` meaning billion and million of parameters is capitalized in both first part of url and name: `smolLm-2-135M` and `smolLm2_135M_bf16.pte` and for some not: `qwen-2.5-1.5B` and `qwen2_5_0_5b_8da4w.pte`.

This should be unified in a separate task

msluszniak:
Ok, we agreed that models should be named with model name & precision and be in correct directory according to backend

NorbertKlockiewicz:
Right now there's a total mess on our hugging face repositories. There's no convention which would help us unify those, that's why I propose such rules:
- the pte file name: {modelName}\_{size}\_{backend}\_{precision}.pte (e.g yolo26_m_coreml_int8.pte)
- a directory for each backend, when entering the repo in root there should be a directory for each backend to which model was exported, a config.json, README.md and if applies tokenizers
- we should have at most two versions of model for each backend (one in original precision - based on backend support fp16 or fp32 and one quantized which should be a default
- the config.json should have an information about the methods and their input + outputs shape and types

msluszniak:
> Right now there's a total mess on our hugging face repositories. There's no convention which would help us unify those, that's why I propose such rules:
> 
[...]

PR #1148, opened 2026-05-13
feat(constants)!: switch URLs to v0.9.0 layout + add MODEL_REGISTRY

### LINKED ISSUE #612: Model constants grouped by type

msluszniak:
## Description

Refreshes every URL constant to the restructured HF layout under
`resolve/v0.9.0` and adds the typed `models` accessor.

### URL refresh

All URLs follow `<model>_<size>_<backend>_<precision>.pte`, files sit
under per-size and per-backend directories on HF.

- `modelUrls.ts` — every URL rewritten; multi-backend URLs hoisted here so the registry stays declarative. The `lfm2_5_350m_xnnpack_8w4da.pte` typo is corrected to `_8da4w.pte`.
- `ocr/models.ts`, `tts/models.ts`, `tts/voices.ts` — paths updated to the new shape.
- `versions.ts` — `VERSION_TAG → resolve/v0.9.0`; `PREVIOUS_VERSION_TAG = resolve/v0.8.0` retained for the `@deprecated` Llama QLoRA aliases.

### `models` accessor

New `constants/modelRegistry.ts` exports `models`, a typed accessor grouped one-to-one with hooks:

| Group                   | Hook                                |
| ----------------------- | ----------------------------------- |
| `llm`                   | `useLLM` (includes vision-capable LLMs like `lfm2_5_vl_*`) |
| `classification`        | `useClassification`                 |
| `privacy_filter`        | `usePrivacyFilter`                  |
| `object_detection`      | `useObjectDetection`                |
| `pose_estimation`       | `usePoseEstimation`                 |
| `semantic_segmentation` | `useSemanticSegmentation`           |
| `instance_segmentation` | `useInstanceSegmentation`           |
| `style_transfer`        | `useStyleTransfer`                  |
| `speech_to_text`        | `useSpeechToText`                   |
| `text_to_speech`        | `useTextToSpeech`                   |
| `text_embedding`        | `useTextEmbeddings`                 |
| `image_embedding`       | `useImageEmbeddings`                |
| `image_generation`      | `useTextToImage`                    |
| `vad`                   | `useVAD`                            |
| `ocr`                   | `useOCR` / `useVerticalOCR`         |

Each entry is a function — call it (optionally with `{ quant, backend }`) to get the resolved config:

```ts
models.llm.llama3_2_3b()                       // default (quantized, platform-default backend)
models.llm.llama3_2_3b({ quant: false })       // base
models.llm.lfm2_5_vl_1_6b()                    // vision-capable LLM (same hook)
models.text_embedding.distiluse_base_multilingual_cased_v2({ backend: 'coreml' })
models.ocr({ language: 'en' })                 // OCR is parameterized by language
```

- The `backend` parameter is typed to exactly the backends each model ships with — `models.llm.llama3_2_3b({ backend: 'coreml' })` is a compile-time error (xnnpack-only).
- Defaults to the quantized variant when `{ quant }` is omitted.
- `text_to_speech` exposes `kokoro_small`/`kokoro_medium` plus plain voice configs under `voices.*`.

ESLint's `camelcase` rule is relaxed to `properties: 'never'` so the snake_case property keys pass while bindings/functions stay camelCase.

### Migration

```ts
// Before                                          // After
LLAMA3_2_1B_SPINQUANT                            → models.llm.llama3_2_1b()
LFM2_5_1_2B_INSTRUCT                             → models.llm.lfm2_5_1_2b_instruct({ quant: false })
LFM2_5_VL_1_6B_QUANTIZED                         → models.llm.lfm2_5_vl_1_6b()
EFFICIENTNET_V2_S                                → models.classification.efficientnet_v2_s()
PRIVACY_FILTER_OPENAI                            → models.privacy_filter.openai()
YOLO26N_POSE                                     → models.pose_estimation.yolo26n()
WHISPER_TINY_EN                                  → models.speech_to_text.whisper_tiny_en()
{ model: KOKORO_MEDIUM, voice: KOKORO_VOICE_AF_HEART }
                                                 → { model: models.text_to_speech.kokoro_medium(),
                                                     voice: models.text_to_speech.voices.af_heart }
OCR_ENGLISH                                      → models.ocr({ language: 'en' })
MODEL_REGISTRY.LLM.LLAMA3_2_3B                   → models.llm.llama3_2_3b()
```

Individual constant imports (`LLAMA3_2_1B_SPINQUANT`, `KOKORO_MEDIUM`, etc.) still work — the new accessor is the recommended path. The flat `MODEL_REGISTRY = { ALL_MODELS: {...} }` export from `modelUrls.ts` is removed; the internal `getModelNameForUrl` lookup is preserved.

### Example apps + docs

- All example apps migrated to `models.*()`. Heavily-used groups are destructured at the top of the file (`const segmentation = models.semantic_segmentation;`).
- Picker entries compared by `modelName` to handle accessor-function values.
- `bare-rn` LLM demo switched to LFM-2.5.
- Every documentation code snippet that selected a model via a named constant is rewritten to use the typed `models.<group>.<entry>()` accessor across `03-hooks/**`, `04-typescript-api/**`, `01-fundamentals/**`, etc. The webrtc-integration page intentionally keeps the named-constant style — the snippet reads cleaner alongside imports from other libraries.
- Model Registry docs page rewritten for the new accessor; the 0.8.x version's anchor is repointed to its own version to survive the rename.

### Deprecations

- `LLAMA3_2_3B_QLORA`, `LLAMA3_2_1B_QLORA` — `@deprecated`; the .pte files stay at `v0.8.0` and the constants still resolve those URLs. Use `LLAMA3_2_*_SPINQUANT` going forward.

- [x] Yes
- [ ] No

URL paths under `${VERSION_TAG}` change — code that hardcoded `resolve/v0.8.0` URLs through the constants keeps working only if it read them at runtime. The flat `MODEL_REGISTRY` export is removed in favour of the new `models` accessor.

### Type of change

### Tested on

- [ ] iOS
- [x] Android

`yarn typecheck` and `yarn lint` clean across the monorepo. Every example app runs against the v0.9.0 HF state.

### Testing instructions

```bash
yarn typecheck
yarn lint
```

In application code:

```ts
import { models } from 'react-native-executorch';

const llm = useLLM({ model: models.llm.llama3_2_3b() });
const emb = useTextEmbeddings({
  model: models.text_embedding.distiluse_base_multilingual_cased_v2({ backend: 'coreml' }),
});
```

### Related issues

#431
#612

### Checklist


apps/llm/components/ModelPicker.tsx
```diff
@@ -24,6 +24,17 @@ type Props<T> = {
...
+// React auto-invoke them as lazy initializers — state becomes the underlying
+// config object, breaking reference equality against the accessor in MODELS.
+// Match by modelName when both sides expose one, otherwise fall back to ===.
+function sameValue<T>(a: T, b: T): boolean {
+  const am = (a as { modelName?: unknown })?.modelName;
+  const bm = (b as { modelName?: unknown })?.modelName;
+  if (typeof am === 'string' && typeof bm === 'string') return am === bm;
+  return a === b;
+}
+
```
barhanc: This code is duplicated in several files. It should probably be factored out to some common utilities file. However, in general I would be in favour of not using the model accessor as both a value and a function, so that such workarounds aren't required.

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -0,0 +1,477 @@
...
+
+/**
+ * An accessor that behaves as the platform-default config when read as a value
+ * (e.g. `MODEL_REGISTRY.LLM.LLAMA3_2_3B.modelName`) and as a function when
+ * called (e.g. `MODEL_REGISTRY.LLM.LLAMA3_2_3B({ quant: false })`).
+ */
+type Accessor<
+  C extends { modelName: string },
+  B extends Backend = Backend,
+> = C & ((opts?: ModelOpts<B>) => C);
```
barhanc: I don't think we should make the accessor behave like both a value and a function and just make the user explicitly call e.g. `MODEL_REGISTRY.LLM.LLAMA3_2_3B()` for default config (perhaps with some stylistic changes like changing the names to lowercase to indicate these are getters and not constants; the use pattern in the user code would be something like ```const LLAMA3_2_3B = models.llm.llama3_2_3b()```). I feel the current approach might generate some problems e.g. when comparing models as in example apps using `sameValue` workaround.

barhanc:
There are some problems with HF repos:
- https://huggingface.co/software-mansion/react-native-executorch-clip-vit-base-patch32/tree/v0.9.0/xnnpack doesn't have text encoder and the exported url points to the image encoder. The example app fails with `{"code": 18, "message": "The model's forward function did not succeed. Ensure the model input is correct."}`
- https://huggingface.co/software-mansion/react-native-executorch-distiluse-base-multilingual-cased-v2/tree/v0.9.0 not all models conform to the naming convention (the fp32 variants use "-" instead of "_")
- https://huggingface.co/software-mansion/react-native-executorch-paraphrase-multilingual-MiniLM-L12-v2/tree/v0.9.0 not all models conform to the naming convention (the fp32 variants use "-" instead of "_")
- https://huggingface.co/software-mansion/react-native-executorch-deeplab-v3/tree/v0.9.0 doesn't have all the variants that v0.8.0 had
- https://huggingface.co/software-mansion/react-native-executorch-fcn/tree/v0.9.0 doesn't have all the variants
- https://huggingface.co/software-mansion/react-native-executorch-fast-sam/tree/v0.9.0 silently drops the xnnpack from `fastsam-x/` and `fastsam-s/` and adds `xnnpack/` and `coreml/` without the size distinction

---

- https://huggingface.co/software-mansion/react-native-executorch-qwen-3.5/tree/v0.9.0 has both the `0_8b` and `Qwen3.5-0.8B` (and the same for 2B) directories
- https://huggingface.co/software-mansion/react-native-executorch-detector-craft/tree/v0.9.0/xnnpack has both `craft_xnnpack_int8.pte` and `xnnpack_craft.pte` files

---

That's what I managed to find, probably someone else should also have a look at the updated HF repos. I will be testing example apps now.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -891,18 +899,18 @@ export const WHISPER_SMALL = {
...
-const DEEPLAB_V3_RESNET101_QUANTIZED_MODEL = `${URL_PREFIX}-deeplab-v3/${PREVIOUS_VERSION_TAG}/deeplab-v3-resnet101/xnnpack/deeplabv3_resnet101_xnnpack_int8.pte`;
-const DEEPLAB_V3_MOBILENET_V3_LARGE_QUANTIZED_MODEL = `${URL_PREFIX}-deeplab-v3/${PREVIOUS_VERSION_TAG}/deeplab-v3-mobilenet-v3-large/xnnpack/deeplabv3_mobilenet_v3_large_xnnpack_int8.pte`;
-const LRASPP_MOBILENET_V3_LARGE_QUANTIZED_MODEL = `${URL_PREFIX}-lraspp/${PREVIOUS_VERSION_TAG}/xnnpack/lraspp_mobilenet_v3_large_xnnpack_int8.pte`;
-const FCN_RESNET50_QUANTIZED_MODEL = `${URL_PREFIX}-fcn/${PREVIOUS_VERSION_TAG}/fcn-resnet50/xnnpack/fcn_resnet50_xnnpack_int8.pte`;
-const FCN_RESNET101_QUANTIZED_MODEL = `${URL_PREFIX}-fcn/${PREVIOUS_VERSION_TAG}/fcn-resnet101/xnnpack/fcn_resnet101_xnnpack_int8.pte`;
+const DEEPLAB_V3_RESNET50_MODEL = `${URL_PREFIX}-deeplab-v3/${VERSION_TAG}/xnnpack/deeplab_v3_xnnpack_fp32.pte`;
+const DEEPLAB_V3_RESNET101_MODEL = `${URL_PREFIX}-deeplab-v3/${VERSION_TAG}/xnnpack/deeplab_v3_xnnpack_fp32.pte`;
+const DEEPLAB_V3_MOBILENET_V3_LARGE_MODEL = `${URL_PREFIX}-deeplab-v3/${VERSION_TAG}/xnnpack/deeplab_v3_xnnpack_fp32.pte`;
```
barhanc: These point to the same url, but are three different models.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -891,18 +899,18 @@ export const WHISPER_SMALL = {
...
+const LRASPP_MOBILENET_V3_LARGE_MODEL = `${URL_PREFIX}-lraspp/${VERSION_TAG}/xnnpack/lraspp_mobilenet_v3_large_xnnpack_fp32.pte`;
+const FCN_RESNET50_MODEL = `${URL_PREFIX}-fcn/${VERSION_TAG}/xnnpack/fcn_xnnpack_fp32.pte`;
+const FCN_RESNET101_MODEL = `${URL_PREFIX}-fcn/${VERSION_TAG}/xnnpack/fcn_xnnpack_fp32.pte`;
```
barhanc: Different models, same url.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -891,18 +899,18 @@ export const WHISPER_SMALL = {
...
+const DEEPLAB_V3_RESNET50_QUANTIZED_MODEL = `${URL_PREFIX}-deeplab-v3/${VERSION_TAG}/xnnpack/deeplab_v3_xnnpack_int8.pte`;
+const DEEPLAB_V3_RESNET101_QUANTIZED_MODEL = `${URL_PREFIX}-deeplab-v3/${VERSION_TAG}/xnnpack/deeplab_v3_xnnpack_int8.pte`;
+const DEEPLAB_V3_MOBILENET_V3_LARGE_QUANTIZED_MODEL = `${URL_PREFIX}-deeplab-v3/${VERSION_TAG}/xnnpack/deeplab_v3_xnnpack_int8.pte`;
```
barhanc: Different models, same url.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -891,18 +899,18 @@ export const WHISPER_SMALL = {
...
+const LRASPP_MOBILENET_V3_LARGE_QUANTIZED_MODEL = `${URL_PREFIX}-lraspp/${VERSION_TAG}/xnnpack/lraspp_mobilenet_v3_large_xnnpack_int8.pte`;
+const FCN_RESNET50_QUANTIZED_MODEL = `${URL_PREFIX}-fcn/${VERSION_TAG}/xnnpack/fcn_xnnpack_int8.pte`;
+const FCN_RESNET101_QUANTIZED_MODEL = `${URL_PREFIX}-fcn/${VERSION_TAG}/xnnpack/fcn_xnnpack_int8.pte`;
```
barhanc: DIfferent models, same url.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1097,8 +1105,8 @@ export const RF_DETR_NANO_SEG = {
 } as const;
 
 // Image Embeddings
-const CLIP_VIT_BASE_PATCH32_IMAGE_MODEL = `${URL_PREFIX}-clip-vit-base-patch32/${PREVIOUS_VERSION_TAG}/xnnpack/clip_vit_base_patch32_vision_xnnpack_fp32.pte`;
-const CLIP_VIT_BASE_PATCH32_IMAGE_QUANTIZED_MODEL = `${URL_PREFIX}-clip-vit-base-patch32/${PREVIOUS_VERSION_TAG}/xnnpack/clip_vit_base_patch32_vision_xnnpack_int8.pte`;
+const CLIP_VIT_BASE_PATCH32_IMAGE_MODEL = `${URL_PREFIX}-clip-vit-base-patch32/${VERSION_TAG}/xnnpack/clip_vit_base_patch32_xnnpack_fp32.pte`;
+const CLIP_VIT_BASE_PATCH32_IMAGE_QUANTIZED_MODEL = `${URL_PREFIX}-clip-vit-base-patch32/${VERSION_TAG}/xnnpack/clip_vit_base_patch32_xnnpack_int8.pte`;
```
barhanc: Image and text models point to the same urls.

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -0,0 +1,477 @@
...
+// the (quant × backend) matrix lives next to the registry that consumes it.
+// The Platform.OS-collapsed exports in `modelUrls.ts` remain the back-compat
+// top-level imports.
+// ─────────────────────────────────────────────────────────────────────────────
+
+const EFFICIENTNET_V2_S_VARIANTS = {
+  xnnpack: {
+    base: {
+      modelName: 'efficientnet-v2-s' as const,
+      modelSource: `${URL_PREFIX}-efficientnet-v2-s/${VERSION_TAG}/xnnpack/efficientnet_v2_s_xnnpack_fp32.pte`,
```
barhanc: Shouldn't we use here (and in the `modelSource`'s below) the urls imported from `modelUrls.ts`? Right now these are duplicated here and in the `modelUrls.ts`.

barhanc (commented):
~Regarding HF repos, the DeepLab and FCN are not categorized by size (the same case as FastSAM before).~ Other HF repos look fine :)

msluszniak:
>Regarding HF repos, the DeepLab and FCN are not categorized by size (the same case as FastSAM before). Other HF repos look fine :)

@barhanc I don't think these are sizes per se in this case. You have specified only the backbone of the final model. These models might be smaller or bigger but semantically do not indicate sizes immediately (in comparison where models from the same family have different number of parameters and the name derives from this number of is explicitly named s,m,l, xl etc.).

packages/react-native-executorch/src/constants/modelRegistry.ts
msluszniak: Yeah, currently I'm in favour of making these. The same problem with default config is going to fire on every `useMemo([accessor])`, every `useCallback(fn, [accessor])`, every `Object.is-based` selector in Redux/Zustand. And stylistically, `const LLAMA3_2_3B = models.llm.llama3_2_3b())` fits better imho.

msluszniak: cc: @chmjkb when reviewing, please mind that we must agree on something here. When decided, I can implement specific version.

apps/bare-rn/App.tsx
```diff
@@ -144,7 +144,9 @@ function App() {
   const textInputRef = useRef<TextInput>(null);
   const scrollViewRef = useRef<ScrollView>(null);
 
-  const llm = useLLM({ model: LLAMA3_2_1B_SPINQUANT });
+  const llm = useLLM({
+    model: MODEL_REGISTRY.LLM.LLAMA3_2_1B,
+  });
```
chmjkb, line 145: maybe lets use lfm instead?

apps/computer-vision/app/instance_segmentation/index.tsx
```diff
@@ -31,19 +24,24 @@ import ImageWithMasks, {
...
+  { label: 'Yolo26S', value: MODEL_REGISTRY.INSTANCE_SEGMENTATION.YOLO26S_SEG },
+  { label: 'Yolo26M', value: MODEL_REGISTRY.INSTANCE_SEGMENTATION.YOLO26M_SEG },
+  { label: 'Yolo26L', value: MODEL_REGISTRY.INSTANCE_SEGMENTATION.YOLO26L_SEG },
+  { label: 'Yolo26X', value: MODEL_REGISTRY.INSTANCE_SEGMENTATION.YOLO26X_SEG },
+  {
+    label: 'RF-DeTR Nano',
+    value: MODEL_REGISTRY.INSTANCE_SEGMENTATION.RF_DETR_NANO_SEG,
+  },
+  { label: 'FastSAM-S', value: MODEL_REGISTRY.OBJECT_DETECTION.FASTSAM_S },
+  { label: 'FastSAM-X', value: MODEL_REGISTRY.OBJECT_DETECTION.FASTSAM_X },
```
chmjkb: maybe u can do sth like const SegmentationRegistry = MODEL_REGISTRY.INSTANCE_SEGMENTATION and then the code would get a bit easier to read?

chmjkb: same goes for other apps, but its no biggie

apps/llm/components/ModelPicker.tsx
chmjkb: yeah agreed, this looks like something users can easily fall for and ideally we don't have to do this

docs/docs/05-utilities/model-registry.md
```diff
@@ -2,36 +2,123 @@
...
+```
+
+## Shape
+
+`MODEL_REGISTRY` is grouped by capability. Each leaf is callable with `{ quant, backend }` and also behaves as a plain config object when read directly.
+
+| Group                   | Examples                                                                                                      |
+| ----------------------- | ------------------------------------------------------------------------------------------------------------- |
+| `LLM`                   | `LLAMA3_2_3B`, `QWEN3_4B`, `SMOLLM2_1_1_7B`, `PHI_4_MINI_4B`, `BIELIK_V3_0_1_5B`, `LFM2_5_1_2B_INSTRUCT`, ... |
+| `VLM`                   | `LFM2_5_VL_1_6B`, `LFM2_5_VL_450M`                                                                            |
```
chmjkb: Wondering if VLM makes sense, what if we want to add audio-capable models too? maybe let's call it multimodal? wdyt?

docs/docs/05-utilities/model-registry.md
```diff
@@ -2,36 +2,123 @@
...
+| `CLASSIFICATION`        | `EFFICIENTNET_V2_S`, `PRIVACY_FILTER_OPENAI`, `PRIVACY_FILTER_NEMOTRON`                                       |
+| `OBJECT_DETECTION`      | `SSDLITE_320_MOBILENET_V3_LARGE`, `YOLO26N` … `YOLO26X`, `RF_DETR_NANO`, `FASTSAM_S`, `FASTSAM_X`             |
+| `SEMANTIC_SEGMENTATION` | `DEEPLAB_V3_RESNET50`, `LRASPP_MOBILENET_V3_LARGE`, `FCN_RESNET101`, `SELFIE_SEGMENTATION`, ...               |
+| `INSTANCE_SEGMENTATION` | `YOLO26N_SEG` … `YOLO26X_SEG`, `RF_DETR_NANO_SEG`                                                             |
+| `STYLE_TRANSFER`        | `CANDY`, `MOSAIC`, `RAIN_PRINCESS`, `UDNIE`                                                                   |
+| `SPEECH_TO_TEXT`        | `WHISPER_TINY_EN`, `WHISPER_BASE`, `WHISPER_SMALL_EN`, ...                                                    |
```
chmjkb: i think TTS is missing

docs/docs/05-utilities/model-registry.md
```diff
@@ -2,36 +2,123 @@
...
-import { MODEL_REGISTRY, LFM2_5_1_2B_INSTRUCT } from 'react-native-executorch';
+type ModelOpts<B extends Backend = Backend> = {
+  quant?: boolean; // Pick the non-quantized variant when false. Defaults to the quantized variant when one is published.
+  backend?: B; // Explicit backend; the set of allowed values is per-model.
+};
 ```
 
-### Accessing a model directly
+- `quant` defaults to the quantized variant for models that publish one (e.g. `LLAMA3_2_3B` → SpinQuant, `EFFICIENTNET_V2_S` → int8). Pass `{ quant: false }` to get the full-precision variant. For models with a single variant, `quant` is accepted but has no effect.
+- `backend` selects an explicit backend. The accessor's call signature is typed to exactly the backends the model ships with, so requesting one the model doesn't publish is a compile-time error (e.g. `MODEL_REGISTRY.LLM.LLAMA3_2_3B({ backend: 'coreml' })` does not type-check — Llama 3.2 is xnnpack-only). When `backend` is omitted, the platform-default applies (typically CoreML on iOS and XNNPACK on Android for multi-backend models).
```
chmjkb: nice 👍🏻

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -0,0 +1,475 @@
...
+ * called (e.g. `MODEL_REGISTRY.LLM.LLAMA3_2_3B({ quant: false })`).
+ */
+type Accessor<
+  C extends { modelName: string },
+  B extends Backend = Backend,
+> = C & ((opts?: ModelOpts<B>) => C);
+
+type BackendCell = {
+  base?: { modelName: string };
+  quant?: { modelName: string };
```
chmjkb, line 62: wondering if we can make it a string literal union?

chmjkb (changes requested):

packages/react-native-executorch/src/constants/modelRegistry.ts
chmjkb: but maybe its overengineering, you decide :D

packages/react-native-executorch/src/constants/modelRegistry.ts
msluszniak: Do we want to allow snake case here?
I can change it to be an option here:

```js
    // `properties: 'never'` lets the lowercase snake_case keys in `models`
    // (e.g. `models.text_to_speech.kokoro_small`, mirroring the underlying
    // `.pte` filenames) pass while still requiring camelCase for variable
    // and function declarations.
    'camelcase': ['error', { properties: 'never' }],
```

docs/docs/05-utilities/model-registry.md
msluszniak: And OCR as well, both in documentation and in the code.

docs/docs/05-utilities/model-registry.md
msluszniak: Yes, makes sense especially because models can have both vision and audio capabilities.

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -0,0 +1,514 @@
...
+    smollm2_1_1_7b: pair(M.SMOLLM2_1_1_7B, M.SMOLLM2_1_1_7B_QUANTIZED),
+    phi_4_mini_4b: pair(M.PHI_4_MINI_4B, M.PHI_4_MINI_4B_QUANTIZED),
+    lfm2_5_350m: pair(M.LFM2_5_350M, M.LFM2_5_350M_QUANTIZED),
+    lfm2_5_1_2b_instruct: pair(
+      M.LFM2_5_1_2B_INSTRUCT,
+      M.LFM2_5_1_2B_INSTRUCT_QUANTIZED
+    ),
+    bielik_v3_0_1_5b: pair(M.BIELIK_V3_0_1_5B, M.BIELIK_V3_0_1_5B_QUANTIZED),
+  },
+  multimodal: {
```
barhanc: I believe `lmm` (Large Multimodal Models) is a more fitting name.

barhanc (commented):
In docs we have many snippets that use the old API for selecting the model, these should probably be changed as well.

packages/react-native-executorch/src/constants/modelRegistry.ts
msluszniak: I didn't know this is well established acronym, will change that tomorrow.

msluszniak:
The fact that currently models from both `llm` and `lmm` are used by `useLLM` and analogical for modules seems to be a bit off, what do you think, @barhanc @chmjkb?

barhanc:
> The fact that currently models from both `llm` and `lmm` are used by `useLLM` and analogical for modules seems to be a bit off, what do you think, @barhanc @chmjkb

We can just put all these models under `llm` and leave the hooks and modules unchanged, as the multimodal models are still language models at their core just with added capabilities for other modalities (in this option as a user I would probably want some easy way to see what modalities a given model supports). The other option would be to add `useLMM` hook and the rest of API mirroring the LLM, but this would lead to a lot of duplicated code, so imo the first option is better.

msluszniak:
Agreed, I'm also in favour of moving them under llm. Also regarding this one: 
> in this option as a user I would probably want some easy way to see what modalities a given model supports

Do you have anything particular solution on your mind?

barhanc:
I guess we already have something like this in place, since the user can check it like this
```ts
const LFM2_5_VL = models.llm.lfm2_5_vl_1_6b()
console.log(LFM2_5_VL.capabilities)
```

msluszniak:
@barhanc @chmjkb I will rebase this PR once change to TTS will land on main. But you can review it and then only the rebase part.

apps/llm/app/llm/index.tsx
```diff
@@ -43,7 +43,7 @@ function LLMScreen() {
   const [isTextInputFocused, setIsTextInputFocused] = useState(false);
   const [userInput, setUserInput] = useState('');
   const [selectedModel, setSelectedModel] = useState<LLMModelSources>(
-    LLAMA3_2_1B_SPINQUANT
+    models.llm.llama3_2_1b()
```
mkopcins: Lets not use llama, use lfm instead

barhanc (approved):
LGTM 🚀

mkopcins (approved):

chmjkb (approved):

msluszniak:
TODO when rebasing after #1132 lands: unify the Whisper URLs (no separate `*_COREML` constants). That resolves two things flagged on #1132:

- `modelName` collision between `WHISPER_*` and `WHISPER_*_COREML`: https://github.com/software-mansion/react-native-executorch/pull/1132#discussion_r3275593849
- Missing `@category Models - Speech To Text` JSDoc on `WHISPER_TINY_EN_COREML`, `WHISPER_BASE_EN_COREML`, `WHISPER_SMALL_EN_COREML`, `WHISPER_TINY_COREML`: https://github.com/software-mansion/react-native-executorch/pull/1132#discussion_r3275593855

packages/react-native-executorch/src/constants/modelUrls.ts
barhanc: Shouldn't we use here (and other Whisper models) the `Platform.OS` ternary to select CoreML on iOS by default, like in other models?

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1195,22 +1184,15 @@ export const MULTI_QA_MPNET_BASE_DOT_V1 = {
 
 /**
  * @category Models - Text Embeddings
+ * @deprecated Use `MODEL_REGISTRY.TEXT_EMBEDDING.DISTILUSE_BASE_MULTILINGUAL_CASED_V2`
+ * (defaults to xnnpack on Android) or pass `{ backend: 'xnnpack' }`.
```
barhanc: This is a stale comment, the `MODEL_REGISTRY` was changed to `models`.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -1299,134 +1281,108 @@ export const FSMN_VAD = {
...
-    CLIP_VIT_BASE_PATCH32_TEXT,
-    BK_SDM_TINY_VPRED_512,
-    BK_SDM_TINY_VPRED_256,
-    FSMN_VAD,
-    PRIVACY_FILTER_OPENAI,
-    PRIVACY_FILTER_NEMOTRON,
-  },
-} as const;
+// Internal — populates the urlToModelName lookup below. Not exported: the
+// typed grouping lives in `./modelRegistry` as MODEL_REGISTRY.
```
barhanc: `MODEL_REGISTRY` -> `models`

docs/docs/05-utilities/model-registry.md
```diff
@@ -2,36 +2,142 @@
...
+```
+
+## Migration from the previous registry
+
+Earlier releases exposed `MODEL_REGISTRY` as an uppercase, dual-value/function shape. That has been replaced with a function-only, lowercase `models` export.
+
+```typescript
+// Before
+useLLM(MODEL_REGISTRY.LLM.LLAMA3_2_3B);
+useLLM(MODEL_REGISTRY.LLM.LLAMA3_2_3B({ quant: false }));
```
chmjkb: I don't think the version before accepted a quant param. This looks wrong.

chmjkb (changes requested):

docs/docs/05-utilities/model-registry.md
msluszniak: Good catch

docs/docs/01-fundamentals/01-getting-started.md
```diff
@@ -38,85 +38,45 @@ For supported React Native and Expo versions, see the [Compatibility table](../0
...
-  </TabItem>
-  <TabItem value="yarn" label="yarn">
-
-    ```bash
-    yarn add react-native-executorch-expo-resource-fetcher expo-file-system expo-asset
-    ```
-
-  </TabItem>
-</Tabs>
-
```
chmjkb, line 98: Why are we getting rid of this? I'm not very against it, but just asking

docs/docs/03-hooks/02-computer-vision/useInstanceSegmentation.md
```diff
@@ -179,12 +178,6 @@ try {
 }
 ```
 
-For detailed API Reference for `selectByPoint`, `selectByBox`, and `selectByText` see their respective documentation pages:
-
-- [`selectByPoint` API Reference](../../06-api-reference/functions/selectByPoint.md)
-- [`selectByBox` API Reference](../../06-api-reference/functions/selectByBox.md)
-- [`selectByText` API Reference](../../06-api-reference/functions/selectByText.md)
-
```
chmjkb, line 187: Why are we getting rid of this?

docs/docs/04-typescript-api/02-computer-vision/PoseEstimationModule.md
```diff
@@ -12,13 +12,13 @@ TypeScript API implementation of the [usePoseEstimation](../../03-hooks/02-compu
...
-import { PoseEstimationModule, YOLO26N_POSE } from 'react-native-executorch';
-
+import { models, PoseEstimationModule } from 'react-native-executorch';
 const imageUri = 'path/to/image.png';
 
 // Creating an instance and loading the model
-const poseEstimationModule =
-  await PoseEstimationModule.fromModelName(YOLO26N_POSE);
+const poseEstimationModule = await PoseEstimationModule.fromModelName(
+  models.pose_estimation.yolo26n()
```
chmjkb, line 20: maybe its not worth it to chagne it, but the segmentation models have a seg suffix (which i think is useless, since the .segmentation imples that its for segmentation), this model doesnt.

docs/docs/05-utilities/04-error-handling.md
```diff
@@ -68,9 +67,7 @@ llm.delete();
 
 ## Reference
 
-All errors in React Native ExecuTorch inherit from `RnExecutorchError` and include a `code` property. The code is typed as `RnExecutorchErrorCode | number`: in practice you'll almost always see a member of the enum, but the codes are generated from a shared config into both the C++ and TS sources, and if those generated files drift the raw numeric value flows through unchanged. When switching on `err.code`, always include a `default` branch.
-
-Below is a comprehensive list of all possible errors, organized by category.
+All errors in React Native ExecuTorch inherit from `RnExecutorchError` and include a `code` property from the `RnExecutorchErrorCode` enum. Below is a comprehensive list of all possible errors, organized by category.
```
chmjkb: Why are we deleting this?

docs/docs/05-utilities/04-error-handling.md
```diff
@@ -85,15 +82,15 @@ These errors occur when trying to perform operations on a model in an invalid st
...
+| `InvalidModelSource`   | Model source type is invalid         | Providing wrong type for model source (e.g., object when string expected)               | Ensure model source matches expected type                |
+| `LanguageNotSupported` | Language not supported by model      | Passing unsupported language to multilingual OCR or Speech-to-Text models               | Use a supported language or different model              |
+| `PlatformNotSupported` | Current platform is not supported    | Using features (e.g., camera frame processing) on an unsupported platform or OS version | Ensure you're running on a supported platform/OS version |
+| `WrongDimensions`      | Input tensor dimensions don't match  | Providing input with incorrect shape for the model                                      | Check model's expected input dimensions                  |
+| `UnexpectedNumInputs`  | Wrong number of inputs provided      | Passing more or fewer inputs than model expects                                         | Match the number of inputs to model metadata             |
```
chmjkb: i think this should stay

docs/docs/05-utilities/04-error-handling.md
```diff
@@ -135,12 +132,12 @@ These errors are specific to streaming transcription operations.
...
-| `InvalidModelOutput` | Model output size unexpected     | Model produces output of wrong size            | Verify model's expected I/O contract in docs or source code |
-| `ThreadPoolError`    | Threadpool operation failed      | Internal threading issue                       | Restart the model or app                                    |
-| `TokenizerError`     | Tokenizer or tokenization failed | Tokenizer initialization or processing error   | Check tokenizer files and model compatibility               |
-| `UnknownError`       | Unexpected error occurred        | 3rd-party library error or unhandled exception | Check logs for details, report if reproducible              |
+| Error Code           | Description                      | When It Occurs                                 | How to Handle                                  |
+| -------------------- | -------------------------------- | ---------------------------------------------- | ---------------------------------------------- |
+| `InvalidModelOutput` | Model output size unexpected     | Model produces output of wrong size            | Verify model is compatible with the library    |
```
chmjkb: i think this should stay

docs/docs/05-utilities/model-registry.md
```diff
@@ -2,36 +2,142 @@
...
-// Find models by name
-const whisperModels = Object.values(MODEL_REGISTRY.ALL_MODELS).filter((m) =>
-  m.modelName.includes('whisper')
-);
+`text_to_speech` is grouped by model family then by language code (`kokoro.en_us`, `kokoro.en_gb`, `kokoro.fr`, `kokoro.es`, `kokoro.it`, `kokoro.pt`, `kokoro.hi`, `kokoro.pl`, `kokoro.de`). Each leaf returns a complete Kokoro preset bundling the model, voice, and phonemizer — pass the whole result to `useTextToSpeech`. The `kokoro` level reserves room for a future TTS family without forcing a breaking rename.
+
+```typescript
+import { models, useTextToSpeech } from 'react-native-executorch';
+
+const tts = useTextToSpeech(models.text_to_speech.kokoro.en_us.heart());
```
chmjkb, line 99: Why are we using explicit property access here for languages, instead of passing it in the object? It's a bit inconsistent as OCR accepts a language param.
I guess that's because of the fact that each language has its own unique set of voices, but maybe we can unify it a bit?

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -271,10 +274,16 @@ export class LLMController {
...
+        RnExecutorchErrorCode.ModuleNotLoaded,
+        'The model is currently not loaded. Please load the model before calling forward().'
+      );
     }
     if (this._isGenerating) {
-      throw new RnExecutorchError(RnExecutorchErrorCode.ModelGenerating);
+      throw new RnExecutorchError(
+        RnExecutorchErrorCode.ModelGenerating,
+        'The model is currently generating. Please wait until previous model run is complete.'
+      );
```
chmjkb: If this is a message we commonly use, we should probably do something similar to what has recently been done in the C++ layer
but its not the scope of this pr

chmjkb (changes requested):

docs/docs/05-utilities/model-registry.md
msluszniak: >I guess that's because of the fact that each language has its own unique set of voices, but maybe we can unify it a bit?

Yes, you guessed it correctly, and no, we probably cannot unify this more. Since voice itself is bound to the language, we cannot pass language the same way we do in OCR to the particular voice. It is inconsistent with ocr because it is completely different mechanism, and that's fine imo. I think there is no point of unifying thing that simply should be different. But if you have some particular way to improve the API in mind, please feel free to send me this ;)

docs/docs/04-typescript-api/02-computer-vision/PoseEstimationModule.md
msluszniak: Yeah, I'm in favour of dropping `seg` suffix entirely.

docs/docs/03-hooks/02-computer-vision/useInstanceSegmentation.md
```diff
@@ -150,14 +148,15 @@ Instance segmentation models return a list of segmented instances. After `forwar
...
   useInstanceSegmentation,
   selectByPoint,
   selectByBox,
   selectByText,
-  FASTSAM_X,
 } from 'react-native-executorch';
-
-const model = useInstanceSegmentation({ model: FASTSAM_X });
+const model = useInstanceSegmentation({
+  model: models.object_detection.fastsam_x(),
```
barhanc: FastSAM should be in `models.instance_segmentation`

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -0,0 +1,636 @@
...
+      modelName: 'fastsam-s' as const,
+      modelSource: M.FASTSAM_S_XNNPACK_FP32_MODEL,
+    },
+  },
+  coreml: {
+    base: {
+      modelName: 'fastsam-s' as const,
+      modelSource: M.FASTSAM_S_COREML_FP16_MODEL,
+    },
+  },
```
chmjkb, line 264: in one model the base is int8, in the other fp16, is this intentional?

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -0,0 +1,636 @@
...
+      map[value.language as SupportedLanguage] = value;
+    }
+  }
+  return map;
+})();
+
+function craft({ language }: { language: SupportedLanguage }): OcrConfig {
+  const cfg = OCR_BY_LANGUAGE[language];
+  if (!cfg) {
+    throw new Error(
```
chmjkb: you should throw RnExecutorchError

packages/react-native-executorch/src/constants/modelRegistry.ts
msluszniak: How this should be handled here @barhanc both base as f32?

barhanc: On iOS fp16 is visibly faster than fp32 so it should be the default for CoreML. I think it's fine the way it is now.

chmjkb (approved):

PR #1160, opened 2026-05-20
feat: voice-activity streaming mode & inner-vad for speech-to-text module

### LINKED ISSUE #1118: Implement continuous voice activity detection
### Problem Description
Right now, the VAD module supports a static inference on given audio chunk. However, it lacks native support for continuous run i.e. streaming. As a user, having to implement it on JS/React side is not so pretty of a task.

### What should be done
Implement streaming mode within the native, C++ side, similar to the one that Whisper models have.

Continuous VAD would open the door to implementing advanced speech mode - where user does not need to perform any action to trigger speech recognition. This, combined with speech to text module, would have countless number of applications - from voice-controlled interfaces, to personal AI assistants.

IgorSwat:
## Description

This PR introduces changes focused on **voice-activity-detection** module and it's utilization within the library:
- **Native side VAD streaming** - introduces a continuous voice-activity-detection mechanism with user-friendly callback system. Example usage from demo app:
```
  await model.stream({
    onSpeechBegin: () => {...},
    onSpeechEnd: () => {...},
    options: {...},
  });
  ```
- **VAD x STT integration** - adds an option to utilize voice-activity-detection within the speech-to-text module, significantly improving the effective performance of the STT.
- **Demo apps**: introduces new screen in the **speech** demo app: `VoiceActivityDetectionScreen` and changes the behavior of `SpeechToTextScreen`, adding a toggle to switch the VAD submodule for STT on/off.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- To test the VAD streaming: run the `VoiceActivityDetectionScreen` within the `Speech` demo app.
- To test the VAD & STT integration: run the `SpeechToTextScreen` within the `Speech` demo app, with VAD toggle on.

### Related issues

#1118 

### Checklist

### CODE REVIEW

File: packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.h [hunk 1]
```diff
@@ -57,6 +60,9 @@ class OnlineASR : public schema::OnlineASR {
   // ASR module connection for transcribing the audio
   const ASR *asr_;
 
+  // VAD module connection for selecting processing (optional)
+  const VoiceActivityDetection *vad_;
```
chmjkb, line 65: maybe we can make it std optional?

packages/react-native-executorch/src/modules/natural_language_processing/VADModule.ts
```diff
@@ -70,4 +72,83 @@ export class VADModule extends BaseModule {
...
+      await new Promise<void>((r) => (waiter = r));
+    }
+  }
+
+  /**
+   * Inserts a new audio chunk into the streaming VAD session.
+   * @param waveform - The audio chunk to insert.
+   */
+  public streamInsert(waveform: Float32Array): void {
+    if (this.nativeModule == null) return;
```
chmjkb: I think we should throw here to match what is done for other APIs

packages/react-native-executorch/src/modules/natural_language_processing/VADModule.ts
```diff
@@ -70,4 +72,83 @@ export class VADModule extends BaseModule {
...
+    this.nativeModule.streamInsert(waveform);
+  }
+
+  /**
+   * Stops the current streaming VAD session.
+   */
+  public streamStop(): void {
+    if (this.nativeModule == null) return;
```
chmjkb: I think we should throw here to match what is done for other APIs

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -52,12 +55,59 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
-  // for absolute safety we can additionally clip the buffer.
-  std::span<const float> input(
-      audioCopy.begin(),
-      audioCopy.begin() + std::min(constants::kMaxSamples, audioCopy.size()));
+  std::span<const float> input;
+
+  // Allowing VAD changes logic significantly - we no longer commit and clean
+  // at max samples reached moments, but rather at the end of speech moments.
+  if (options.useVAD) {
+    std::span<float> vadInput(const_cast<float *>(audioCopy.data()),
```
msluszniak: is this const cast even needed? You have declaration of mutable `std::vector` above

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -52,12 +55,59 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+    const auto &lastSegment = speechSegments.back();
+    size_t marginSamples =
+        options.vadDetectionMargin * constants::kSamplesPerMilisecond;
+
+    if (audioCopy.size() - lastSegment.end <= marginSamples) {
+      // Speech is ongoing. Keep last 1s context and trim around current
+      // segment.
+      size_t startWithMargin =
+          std::max(0, static_cast<int>(lastSegment.start) -
+                          static_cast<int>(constants::kSamplingRate));
```
msluszniak: Why you're casting to plain ints here? if you ultimately want `size_t`

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -20,8 +21,11 @@ inline constexpr auto kPaddedWindowSize = std::bit_ceil(kWindowSize); // 512
 inline constexpr size_t kModelInputMin = 100;
 inline constexpr size_t kModelInputMax = 1000;
 inline constexpr auto kSpeechThreshold = 0.6f;
-inline constexpr size_t kMinSpeechDuration = 25;  // 250 ms
-inline constexpr size_t kMinSilenceDuration = 10; // 100 ms
-inline constexpr size_t kSpeechPad = 3;           // 30 ms
+inline constexpr size_t kMinSpeechDuration = 25;                 // 250 ms
+inline constexpr size_t kMinSilenceDuration = 10;                // 100 ms
+inline constexpr size_t kSpeechPad = 3;                          // 30 ms
```
msluszniak, line 26: Why these are in 10s of ms while other constants above are in miliseconds?

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,118 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+
+  while (isStreaming_) {
+    // Make sure that audio buffer does not exceed it's max size
+    // BEFORE infering the model, such that potentially save 1 unnecessary
+    // iteration.
+    if (audioBuffer_.size() > constants::kStreamBufferMaxSize) {
+      // Keep the most recent audio samples in.
+      size_t cut = audioBuffer_.size() - constants::kStreamBufferMinReserve;
+
+      std::scoped_lock<std::mutex> lock(audioBufferMutex_);
```
msluszniak:
```suggestion
      std::scoped_lock lock(audioBufferMutex_);
```

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,118 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+
+  // Reset the streaming state
+  isStreaming_ = false;
+  audioBuffer_.clear();
+}
+
+void VoiceActivityDetection::streamStop() { isStreaming_ = false; }
+
+void VoiceActivityDetection::streamInsert(std::span<float> audio) {
+  std::scoped_lock<std::mutex> lock(audioBufferMutex_);
```
msluszniak:
```suggestion
  std::scoped_lock lock(audioBufferMutex_);
```

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,118 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+  // Build a native callback which simply sends a boolean signal,
+  // where true corresponds to detected ongoing speech, and false corresponds to
+  // silence.
+  auto nativeCallback = [this, callback](bool speaking) {
+    callInvoker_->invokeAsync([callback, speaking](jsi::Runtime &rt) {
+      callback->call(rt, jsi::Value(speaking));
+    });
+  };
+
+  isStreaming_ = true;
```
msluszniak: `isStreaming_` should be `std::atomic<bool>` and the cleanup should be RAII:

1. **Data race** — plain `bool` read in `while (isStreaming_)` here, written from `streamStop()` on another thread. UB per the C++ memory model; the compiler may hoist the load out of the loop and it never exits.
2. **TOCTOU on entry** — the `if (isStreaming_) throw; ... isStreaming_ = true;` pattern isn't atomic. Fold guard + set into `compare_exchange_strong`.
3. **No exception safety / unsynchronized clear** — if `generate()` or `invokeAsync()` throws, the trailing `isStreaming_ = false; audioBuffer_.clear();` is skipped and the object wedges (every later `stream()` throws `StreamingInProgress`). That final `clear()` also doesn't hold `audioBufferMutex_`, racing `streamInsert()`.

No folly needed — a local RAII struct covers it:

```cpp
// header
std::atomic<bool> isStreaming_{false};

// cpp, top of stream()
bool expected = false;
if (!isStreaming_.compare_exchange_strong(expected, true)) {
  throw RnExecutorchError(RnExecutorchErrorCode::StreamingInProgress,
                          "Streaming is already in progress!");
}
struct StreamingGuard {
  VoiceActivityDetection *self;
  ~StreamingGuard() {
    self->isStreaming_ = false;
    std::scoped_lock lock(self->audioBufferMutex_);
    self->audioBuffer_.clear();
  }
} guard{this};
```

Then drop the trailing reset at the end of the function.

apps/speech/screens/SpeechToTextScreen.tsx
chmjkb, line 1: one note regarding the demo app:
I'm not sure if this was a one time thing but I find it hard to hide the keyboard after using the URL input 😅

apps/speech/screens/VoiceActivityDetectionScreen.tsx
```diff
@@ -0,0 +1,321 @@
...
+              </Text>
+            )}
+          </ScrollView>
+        </View>
+
+        <View style={styles.inputContainer}>
+          {isStreaming ? (
+            <TouchableOpacity
+              onPress={handleStopStreaming}
+              style={[styles.liveButton, styles.backgroundRed]}
```
chmjkb, line 201: I started tapping that button fast, repetitively and I got an error saying that the model is currently generating. Can we handle this gracefully, as done in the TTS screen?

packages/react-native-executorch/src/modules/natural_language_processing/SpeechToTextModule.ts
```diff
@@ -107,7 +120,8 @@ export class SpeechToTextModule {
     return await global.loadSpeechToText(
       'whisper',
```
chmjkb: I don't think we should. pass hardcoded 'whisper' here, but a modelName coming from the config.

chmjkb (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
IgorSwat: All the constants in this file are in 10s of ms.

msluszniak: Then why these are named `kWindowSizeMs` and `kHopLengthMs`, this ms suffix almost screams to me: "This value is in milliseconds". If not, then I would be very surprised ;p

IgorSwat: I mean, constants like `kWindowSizeMs` are indeed in milliseconds, but the other ones like `kModelInputMin` are in tens of milliseconds.

msluszniak: Yeah exactly, and that was my original question, `Why these are in 10s of ms while other constants above are in miliseconds?` Why is that? Can't we unify this?

IgorSwat: We can, but then we will have to divide it each time by 10.  This is how the VAD code is constructed and it was like that a long time before my changes arrived.

msluszniak: Can we create an issue for this then, since this wasn't introduced by this PR, and fix this later on for the whole code?

msluszniak:
Please also fix these warnings:
image

docs/docs/03-hooks/01-natural-language-processing/useSpeechToText.md
```diff
@@ -155,12 +156,39 @@ export default function LiveTranscriber() {
 
 ## Advanced Features
 
+### VAD Integration (Recommended for Live)
+
+Integrating **Voice Activity Detection (VAD)** as a submodule improves streaming performance by automatically removing silence. This reduces CPU usage, saves battery, and prevents hallucinations during silent periods.
```
chmjkb: im not sure this should be under advanced features, as this is pretty important, and in fact not very advanced

docs/docs/04-typescript-api/01-natural-language-processing/VADModule.md
```diff
@@ -37,11 +37,23 @@ To create a ready-to-use instance, call the static [`fromModelName`](../../06-ap
 
 The factory returns a promise that resolves to a loaded `VADModule` instance.
 
-For more information on loading resources, take a look at [loading models](../../01-fundamentals/02-loading-models.md) page.
+For more information on loading resources, take a look at the [loading models](../../01-fundamentals/02-loading-models.md) page.
 
 ## Running the model
 
-To run the model, you can use the [`forward`](../../06-api-reference/classes/VADModule.md#forward) method on the module object. Before running the model's [`forward`](../../06-api-reference/classes/VADModule.md#forward) method, make sure to extract the audio waveform you want to process. You'll need to handle this step yourself, ensuring the audio is sampled at 16 kHz. Once you have the waveform, pass it as an argument to the forward method. The method returns a promise that resolves to the array of detected speech segments.
+### Batch Processing
```
chmjkb: Maybe `transcribing files` would be more intuivite for entry-level users?

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -6,6 +6,7 @@
 namespace rnexecutorch::models::voice_activity_detection::constants {
 
 inline constexpr uint32_t kSampleRate = 16000;
+inline constexpr uint32_t kSampleRateMiliseconds = 16;
```
benITo47: Nit: Should be "kSampleRateMilliseconds"

msluszniak: And btw, I'd argue for renaming it altogether — `kSampleRateMilliseconds` reads as "sample rate, in milliseconds" but the value is actually *samples per millisecond*. `kSamplesPerMs` (or `kSamplesPerMillisecond`) reads correctly at every call site, e.g. `audioBuffer_.size() / kSamplesPerMs → ms`. Combined with the `kSampleRate / 1000` form:

```cpp
inline constexpr uint32_t kSamplesPerMs = kSampleRate / 1000;
```

benITo47 (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,125 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
```
msluszniak: This is probably a deadcode since you've already done that higher, aka `!isStreaming_.compare_exchange_strong(expected, true)`

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,125 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+
+    bool speaking = false;
+
+    // Get the last detected speech segment (if any exists) and
+    // check if it is recent enough to consider the speech as ongoing.
+    if (!detection.empty()) {
+      auto lastSegment = detection.back();
+      auto speechEnd = lastSegment.end;
+
+      uint32_t diffMs = (audioBuffer_.size() - speechEnd) /
```
msluszniak: This `audioBuffer_.size()` is read with no lock held, while `streamInsert()` can be writing concurrently. The comment a few lines up only argues that the *pointer* stays stable (true — buffer is pre-reserved), but `vector::insert` still mutates the vector's size/end internals, so reading `.size()` unsynchronized here is a data race per the C++ memory model. Either wrap in `audioBufferMutex_`, or track sample count in an `std::atomic<size_t>` that `streamInsert` updates.

packages/react-native-executorch/src/modules/natural_language_processing/VADModule.ts
```diff
@@ -70,4 +72,85 @@ export class VADModule extends BaseModule {
...
+        finished = true;
+        wake();
+      } catch (e) {
+        error = e;
+        finished = true;
+        wake();
+      }
+    })();
+
+    while (this.isStreaming && !finished) {
```
msluszniak, line 131: `stream()` resolves as soon as `this.isStreaming` flips, but the native loop only re-checks the flag at the top of the next iteration — so for up to `timeout + one inference` after `await streamStop()` returns, the native streamer is still alive, can still queue `callInvoker_->invokeAsync` callbacks, and still touches `audioBuffer_`. If the caller then runs `unload()` (or the host object is destroyed) we're in UAF / use-after-unload territory.

Two options: (a) actually join — `stream()` doesn't resolve until the native `stream()` call returns, and `streamStop()` awaits that; or (b) document explicitly that `unload()` is not safe immediately after `streamStop()` and that callbacks may fire after the promise resolves. (a) is the safer contract.

packages/react-native-executorch/src/hooks/natural_language_processing/useVAD.ts
```diff
@@ -21,5 +21,29 @@ export const useVAD = ({ model, preventLoad = false }: VADProps): VADType => {
   const forward = (waveform: Float32Array) =>
     runForward((inst) => inst.forward(waveform));
 
-  return { error, isReady, isGenerating, downloadProgress, forward };
+  const stream = (input: VADStreamingInput) =>
+    runForward((inst) => inst.stream(input));
+
+  const streamInsert = (waveform: Float32Array) =>
+    runForward((inst) => {
```
msluszniak, line 28: Both `streamInsert` and `streamStop` go through `runForward`, which gates on `isGenerating`. While `stream()` is running, `isGenerating` is true, so every `streamInsert(buffer)` call rejects with "model is currently generating" — which is what Jakub hit on the rapid-tap repro. That thread is marked resolved, but I don't see the fix in either `useVAD.ts` or `VADModule.ts`. At minimum `streamInsert` (a buffer push) must bypass `runForward` for the streaming API to function; arguably `streamStop` should bypass too — you may want to stop precisely because inference is stuck.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -6,6 +6,7 @@
...
 inline constexpr auto kMstoSecond = 0.001f;
```
msluszniak: `kMstoSecond = 0.001f` is awkward — the value is just `1/1000` and the name reads as "Mstosecond" run-together. Either rename to `kMsToSeconds` / `kMillisToSeconds`, or just inline `0.001f` at the two use sites below. Even cleaner would be to drop the floating-point round-trip entirely: `inline constexpr auto kHopLength = kHopLengthMs * kSampleRate / 1000;`

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -12,4 +13,27 @@ size_t getNonSpeechClassProbabilites(const executorch::aten::Tensor &tensor,
...
+  }
+
+  std::vector<types::Segment> mergedSegments;
+  mergedSegments.push_back(segments[0]);
+
+  for (size_t i = 1; i < segments.size(); ++i) {
+    auto &lastMerged = mergedSegments.back();
+    const auto &current = segments[i];
+
+    if (current.start - lastMerged.end <= maxMergeGap) {
```
msluszniak: `current.start` and `lastMerged.end` are both `size_t`. If a caller ever passes overlapping or out-of-order segments (`current.start < lastMerged.end`), this subtraction underflows to a huge value, the `<= maxMergeGap` check passes, and segments merge silently-wrong. The precondition (sorted, non-overlapping input) isn't documented or checked. Safest fix:

```cpp
if (current.start >= lastMerged.end &&
    current.start - lastMerged.end <= maxMergeGap)
```

Or document the contract in the header.

chmjkb (commented):
besides my previous comment, I think it looks good, great work!

packages/react-native-executorch/src/hooks/natural_language_processing/useVAD.ts
IgorSwat: If it's true, then that's a Module Factory issue - and I don't think this PR is a good place to fix it.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
msluszniak: https://github.com/software-mansion/react-native-executorch/issues/1172

packages/react-native-executorch/src/hooks/natural_language_processing/useVAD.ts
msluszniak: https://github.com/software-mansion/react-native-executorch/issues/1173

docs/docs/04-typescript-api/01-natural-language-processing/SpeechToTextModule.md
```diff
@@ -14,11 +14,16 @@ The `SpeechToTextModule` class provides a direct interface to the library's spee
...
+  FSMN_VAD,
+} from 'react-native-executorch';
 
-// Initialize the model
+// Initialize the model with VAD submodule
 const model = await SpeechToTextModule.fromModelName(
-  models.speech_to_text.whisper_tiny_en(),
+  WHISPER_TINY_EN,
+  FSMN_VAD, // Optional VAD submodule
   (progress) => {
```
msluszniak: Incorrect rebase here.

docs/docs/04-typescript-api/01-natural-language-processing/SpeechToTextModule.md
```diff
@@ -107,12 +90,14 @@ The `stream()` function accepts several optional parameters:
...
+  SpeechToTextModule,
+  WHISPER_TINY_EN,
+  FSMN_VAD,
+} from 'react-native-executorch';
 import { AudioManager, AudioRecorder } from 'react-native-audio-api';
 
-const model = await SpeechToTextModule.fromModelName(
-  models.speech_to_text.whisper_tiny_en()
-);
+const model = await SpeechToTextModule.fromModelName(WHISPER_TINY_EN, FSMN_VAD);
```
msluszniak: Same here, incorrect rebase.

docs/docs/04-typescript-api/01-natural-language-processing/SpeechToTextModule.md
```diff
@@ -155,6 +140,40 @@ model.streamStop();
...
+Integrating **Voice Activity Detection (VAD)** as a submodule improves streaming performance by automatically removing silence. This reduces CPU usage, saves battery, and prevents hallucinations during silent periods.
+
+To use it, provide the VAD model configuration when loading the module and enable `useVAD` in the stream options:
+
+```typescript
+const model = await SpeechToTextModule.fromModelName(
+  WHISPER_TINY_EN,
+  FSMN_VAD // Optional VAD submodule
+);
+
```
msluszniak, line 151: Incorrect rebase.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,124 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+      uint32_t diffMs =
+          (audioBuffer_.size() - speechEnd) / constants::kSamplesPerMs; // [ms]
+
+      speaking = diffMs <= detectionMargin;
+    }
+
+    // Send result to the JS side.
+    nativeCallback(speaking);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(timeout));
```
msluszniak: Maybe we improve this the same way as whisper with `std::conditional_variable`?

apps/speech/screens/SpeechToTextScreen.tsx
```diff
@@ -104,11 +108,16 @@ export const SpeechToTextScreen = ({ onBack }: { onBack: () => void }) => {
...
+    if (!audioURL.trim() || model.isGenerating) {
+      if (!audioURL.trim()) {
+        console.warn('Please provide a valid audio file URL');
+      }
       return;
     }
 
+    Keyboard.dismiss();
+
+    const uri = await getAudioFile(audioURL);
```
msluszniak: This should be removed, as the `getAudioFile` is already in try block which is the correct way (it may fail).

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -12,14 +12,20 @@ namespace rnexecutorch::models::speech_to_text {
 SpeechToText::SpeechToText(const std::string &modelName,
                            const std::string &modelSource,
                            const std::string &tokenizerSource,
+                           const std::string &vadSource,
```
msluszniak, line 15: This change in signature breaks build of native tests. You have to fix them.

msluszniak (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,129 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+    callInvoker_->invokeAsync([callback, speaking](jsi::Runtime &rt) {
+      callback->call(rt, jsi::Value(speaking));
+    });
+  };
+
+  while (isStreaming_) {
+    // Make sure that audio buffer does not exceed it's max size
+    // BEFORE infering the model, such that potentially save 1 unnecessary
+    // iteration.
+    if (audioBuffer_.size() > constants::kStreamBufferMaxSize) {
```
msluszniak: Data race: `audioBuffer_.size()` is read here without holding `audioBufferMutex_`, while `streamInsert` writes to the vector under the same lock. `vector::size()` racing with `vector::insert` is undefined behavior. Move the check inside the lock:

```cpp
{
  std::scoped_lock lock(audioBufferMutex_);
  if (audioBuffer_.size() > constants::kStreamBufferMaxSize) {
    size_t cut = audioBuffer_.size() - constants::kStreamBufferMinReserve;
    audioBuffer_.erase(audioBuffer_.begin(), audioBuffer_.begin() + cut);
  }
}
```

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.cpp
```diff
@@ -100,12 +71,129 @@ VoiceActivityDetection::generate(std::span<float> waveform) const {
...
+
+    std::span<float> input;
+    {
+      // Since the buffer does not reallocate it's memory, this is perfectly
+      // safe, and copying buffer's data is not necessary.
+      std::scoped_lock lock(audioBufferMutex_);
+      input = std::span(audioBuffer_);
+    }
+
+    std::vector<types::Segment> detection = generate(input);
```
msluszniak: Span lifetime hazard: the lock is released after `input = std::span(audioBuffer_)` and then `generate(input)` runs unlocked. The justifying comment relies on the 20 s reservation (`2 * kStreamBufferMaxSize`), but `streamInsert` is unbounded — if JS pushes enough audio between snapshot and `generate()` to cross capacity, the vector reallocates and `input` dangles. The pre-trim caps the buffer near 10 s, leaving „only” ~10 s of headroom in one `timeout` window — fine for mic capture, undefined behavior in adversarial conditions. Either hold the lock across `generate()`, or copy under the lock and pass the copy:

```cpp
std::vector<float> snapshot;
{
  std::scoped_lock lock(audioBufferMutex_);
  snapshot.assign(audioBuffer_.begin(), audioBuffer_.end());
}
auto detection = generate(snapshot);
```

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.cpp
```diff
@@ -12,4 +13,28 @@ size_t getNonSpeechClassProbabilites(const executorch::aten::Tensor &tensor,
...
+  std::vector<types::Segment> mergedSegments;
+  mergedSegments.push_back(segments[0]);
+
+  for (size_t i = 1; i < segments.size(); ++i) {
+    auto &lastMerged = mergedSegments.back();
+    const auto &current = segments[i];
+
+    if (current.start < lastMerged.end ||
+        current.start - lastMerged.end <= maxMergeGap) {
+      lastMerged.end = current.end;
```
msluszniak: If `current.end < lastMerged.end` (e.g. an overlapping inner segment), this overwrites with the smaller value and shrinks the merged segment. Today `postprocess()` produces monotonically increasing bounds so it doesn’t trigger, but the invariant isn’t enforced anywhere. Safer:

```cpp
lastMerged.end = std::max(lastMerged.end, current.end);
```

packages/react-native-executorch/src/modules/natural_language_processing/VADModule.ts
```diff
@@ -70,4 +72,85 @@ export class VADModule extends BaseModule {
...
+    let error: unknown;
+    let finished = false;
+
+    let waiter: (() => void) | null = null;
+    const wake = () => {
+      waiter?.();
+      waiter = null;
+    };
+
+    this.isStreaming = true;
```
msluszniak, line 105: `this.isStreaming` is set true here and only reset in `streamStop()`. If the native `stream()` returns naturally (error path, or `streamStop` originating from the C++ side via unload), the JS flag stays `true` until the next `stream()` overwrites it. Reset it in a `finally` on the inner IIFE so the module reflects the real state:

```ts
(async () => {
  try {
    await this.nativeModule.stream(...);
  } catch (e) {
    error = e;
  } finally {
    this.isStreaming = false;
    finished = true;
    wake();
  }
})();
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/SpeechToText.cpp
```diff
@@ -166,6 +185,9 @@ void SpeechToText::stream(std::shared_ptr<jsi::Function> callback,
                        [this] { return !isStreaming_.load(); });
   }
 
+  // Do not use VAD on final processing to flush and commit everything properly.
+  options.useVAD = false;
```
msluszniak: Mutating the in-flight `options` struct to disable VAD for `finish()` works but is fragile — anything later that reads `options.useVAD` will see the toggled-off value. Prefer a copy:

```cpp
StreamingOptions finishOptions = options;
finishOptions.useVAD = false;
std::vector<Word> finalWords = streamer_->finish(finishOptions);
```

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/VoiceActivityDetection.h
```diff
@@ -15,25 +16,59 @@
...
-  mutable std::mutex inference_mutex_;
+  /**
+   * When called, stops the streaming procedure.
+   */
+  void streamStop();
+
+  /**
+   * Adds a new audio chunk to the streaming buffer.
+   */
+  void streamInsert(std::span<float> audio);
```
msluszniak: `streamInsert` only reads from the span — take `std::span<const float>` to match `OnlineASR::insertAudioChunk`, signal no-write intent, and let callers pass `const` audio buffers.

packages/react-native-executorch/src/modules/natural_language_processing/SpeechToTextModule.ts
```diff
@@ -184,6 +198,8 @@ export class SpeechToTextModule {
     const verbose = !!options.verbose;
     const language = options.language || '';
     const timeout = options.timeout || 100;
+    const useVAD = options.useVAD ?? false;
+    const vadDetectionMargin = options.vadDetectionMargin || 500;
```
msluszniak: `options.vadDetectionMargin || 500` coerces `0` to `500`. Use `??` so callers can explicitly pass `0` (no margin):

```ts
const vadDetectionMargin = options.vadDetectionMargin ?? 500;
```

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/Params.h
```diff
@@ -60,4 +59,9 @@ constexpr inline float kWordsPerSecondLow = 1.5F;
  */
 constexpr inline int32_t kChunkBreakBuffer = 2; // [s]
 
+constexpr inline float kVadGapFactor = 1.2F;
```
msluszniak, line 68: What does `1.2` represent here? A short comment explaining the heuristic (it’s used as `vadDetectionMargin * kVadGapFactor` for the `mergeGap` passed to `vad_->generate`) would save future readers a git-blame trip.

msluszniak (commented):
Review from local verification (VAD native tests pass 15/15, demo app boots). A few correctness items and minor cleanups inline.

docs/docs/04-typescript-api/01-natural-language-processing/SpeechToTextModule.md
```diff
@@ -61,42 +62,20 @@ The `transcribe()` function accepts an optional configuration object:
...
 
-:::iStreaming Options
+### Streaming Options
+
 The `stream()` function accepts several optional parameters:
 
 - `language`: The language code (e.g., `'es'`, `'fr'`). Required for multilingual models.
 - `verbose`: If `true`, includes word-level timestamps and segment metadata in the result objects.
 - `timeout`: (Advanced) The interval (in milliseconds) between processing consecutive audio chunks. Lower values provide more frequent updates, while higher values reduce CPU consumption. Defaults to `100`.
+- `useVAD`: Enable the Voice Activity Detection submodule (if configured during load) to improve performance by skipping silence.
```
msluszniak, line 76: Worth a one-line note that `useVAD=true` runs VAD inference on every `timeout` tick alongside ASR while speech is ongoing. Net result is still a win (silence skips ASR entirely), but the framing here reads as if VAD is free — it isn't, just cheaper than running ASR on silence.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Constants.h
```diff
@@ -6,22 +6,26 @@
...
 inline constexpr auto kSpeechThreshold = 0.6f;
-inline constexpr size_t kMinSpeechDuration = 25;  // 250 ms
-inline constexpr size_t kMinSilenceDuration = 10; // 100 ms
-inline constexpr size_t kSpeechPad = 3;           // 30 ms
+inline constexpr size_t kMinSpeechDuration = 25;                 // 250 ms
+inline constexpr size_t kMinSilenceDuration = 10;                // 100 ms
+inline constexpr size_t kSpeechPad = 3;                          // 30 ms
+inline constexpr size_t kStreamBufferMaxSize = 10 * kSampleRate; // 10s
+inline constexpr size_t kStreamBufferMinReserve =
+    1 * kSampleRate; // 1s of audio
```
msluszniak, line 30: Missing trailing newline at EOF (`\ No newline at end of file` in the diff). Most lints/editors will flag this.

packages/react-native-executorch/common/rnexecutorch/models/voice_activity_detection/Utils.h
```diff
@@ -10,4 +11,16 @@ size_t getNonSpeechClassProbabilites(const executorch::aten::Tensor &tensor,
...
+ * equal to the specified maximum merge gap.
+ *
+ * @param segments A collection of speech segments to be merged.
+ * @param maxMergeGap The maximum allowed distance between two segments (in
+ * samples) to qualify them for a merge.
+ * @return A new collection containing the merged speech segments.
+ */
+std::vector<types::Segment>
+mergeSegments(const std::vector<types::Segment> &segments, size_t maxMergeGap);
+
```
msluszniak, line 25: Missing trailing newline at EOF.

packages/react-native-executorch/common/rnexecutorch/models/speech_to_text/whisper/OnlineASR.cpp
```diff
@@ -55,12 +58,57 @@ ProcessResult OnlineASR::process(const DecodingOptions &options) {
...
+  if (options.useVAD && vad_) {
+    auto speechSegments = vad_->generate(audioCopy, options.vadDetectionMargin *
+                                                        params::kVadGapFactor);
+
+    if (speechSegments.empty()) {
+      // Extra cleanup to speed-up future processing by removing silence.
+      if (audioCopy.size() > params::kVadDeadSamplesRemovalSamples) {
+        std::scoped_lock lock(streamingMutex);
+        size_t cut = std::min(params::kVadDeadSamplesRemovalSamples -
+                                  kSafetyMarginSamples,
```
msluszniak, line 74: `params::kVadDeadSamplesRemovalSamples - kSafetyMarginSamples` is a `size_t` subtraction that relies on the numeric relationship `9 > 3` between two `constexpr` constants defined in different headers. If someone later tweaks either constant such that the ordering inverts, this underflows silently to a huge `size_t` and the `std::min(...)` returns it (because `audioBuffer_.size()` is unlikely to exceed it), then `erase` reads garbage past the buffer — UB.

A compile-time fence next to the constants in `Params.h` would catch this at build time with zero runtime cost:

```cpp
static_assert(kVadDeadSamplesRemovalSamples >
              static_cast<size_t>(kStreamSafetyThreshold * constants::kSamplingRate),
              "kVadDeadSamplesRemovalSamples must exceed the safety margin");
```

(`static_assert` is evaluated by the compiler; it emits no runtime code, so this is appropriate for production paths.)

msluszniak (commented):
Follow-ups (cleanups + missing test coverage):

**Missing native tests** — the PR adds significant new behavior but no tests cover it:
- `utils::mergeSegments` has no unit test (new function in `Utils.cpp`).
- `VAD::stream` / `streamInsert` / `streamStop` lifecycle isn't covered — start → insert audio → boolean transitions at the `detectionMargin` boundary → stop.
- `OnlineASR::process` with `useVAD=true` is the core of the PR but has no integration test — the speech-segment-driven trim/commit path is entirely uncovered.

VADTests already runs on device (15/15 pass on a real Pixel-class arm64 in my local run), so adding tests there is the path of least resistance.

docs/docs/04-typescript-api/01-natural-language-processing/SpeechToTextModule.md
msluszniak: Nvm, forget it

msluszniak (approved):
🚀

chmjkb (approved):

benITo47 (approved):

PR #1162, opened 2026-05-21
feat: Adding support for gemma4 e2b models

### LINKED ISSUE #1062: Gemma4 support
Export gemma4 E2B and E4B models. Key to this is efficient memory management, as a huge part of model params (above 50%) is not needed at runtime, but only on first token consumption (PLE layers)

mkopcins:

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Test by running apps/llm app on llm screen (for text only model) and multimodal screen (for audio-vision-text model). Text model should work as any other llm model. Multimodal can process up-to-30sec audio chunks as well as image inputs, should be able to transcribe audio, describe pictures or similar.

### Related issues

#1062 

### Checklist

### DIFF
--- a/apps/llm/app/multimodal_llm/index.tsx
+++ b/apps/llm/app/multimodal_llm/index.tsx
@@ -34,6 +34,7 @@ const SUGGESTED_PROMPTS = [
   'Describe this scene in detail',
   'What objects can you see?',
   'What text appears in this image?',
+  'Transcribe the audio?',
 ];
 import { useLLMStats } from '../../hooks/useLLMStats';
 import { StatsBar } from '../../components/StatsBar';
@@ -63,7 +64,7 @@ function MultimodalLLMScreen() {
   const [error, setError] = useState<string | null>(null);
 
   const vlm = useLLM({
-    model: models.llm.lfm2_5_vl_1_6b(),
+    model: models.llm.gemma4_e2b(),
   });
   const tokenCount = vlm.isReady ? vlm.getGeneratedTokenCount() : 0;
   const { stats, onMessageSend } = useLLMStats(
@@ -175,7 +176,8 @@ function MultimodalLLMScreen() {
   };
 
   const sendMessage = async () => {
-    if (!userInput.trim() || vlm.isGenerating) return;
+    if (!(imageUri || audioBuffer || userInput.trim()) || vlm.isGenerating)
+      return;
     onMessageSend();
     const text = userInput.trim();
     setUserInput('');
@@ -346,14 +348,15 @@ function MultimodalLLMScreen() {
               onChangeText={setUserInput}
             />
 
-            {userInput.trim() && !vlm.isGenerating && (
-              <TouchableOpacity
-                style={styles.sendChatTouchable}
-                onPress={sendMessage}
-              >
-                <SendIcon height={24} width={24} padding={4} margin={8} />
-              </TouchableOpacity>
-            )}
+            {(imageUri || audioBuffer || userInput.trim()) &&
+              !vlm.isGenerating && (
+                <TouchableOpacity
+                  style={styles.sendChatTouchable}
+                  onPress={sendMessage}
+                >
+                  <SendIcon height={24} width={24} padding={4} margin={8} />
+                </TouchableOpacity>
+              )}
             {vlm.isGenerating && (
               <TouchableOpacity
                 style={styles.sendChatTouchable}
--- a/apps/llm/app/multimodal_llm/index.tsx
+++ b/apps/llm/app/multimodal_llm/index.tsx
@@ -12,6 +12,11 @@ import {
   View,
 } from 'react-native';
 import { launchImageLibrary } from 'react-native-image-picker';
+import {
+  AudioManager,
+  AudioRecorder,
+  AudioContext,
+} from 'react-native-audio-api';
 import { useIsFocused } from '@react-navigation/native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { useLLM, LFM2_5_VL_1_6B_QUANTIZED } from 'react-native-executorch';
@@ -46,7 +51,15 @@ function MultimodalLLMScreen() {
   const textInputRef = useRef<TextInput>(null);
   const { setGlobalGenerating } = useContext(GeneratingContext);
 
-  // Added error state
+  const [audioBuffer, setAudioBuffer] = useState<Float32Array | null>(null);
+  const [audioLabel, setAudioLabel] = useState<string | null>(null);
+  const [audioUrl, setAudioUrl] = useState('');
+  const [isFetchingAudio, setIsFetchingAudio] = useState(false);
+  const [isRecording, setIsRecording] = useState(false);
+  const [hasMicPermission, setHasMicPermission] = useState(false);
+  const recorder = useRef(new AudioRecorder());
+  const recordChunks = useRef<Float32Array[]>([]);
+
   const [error, setError] = useState<string | null>(null);
 
   const vlm = useLLM({
@@ -68,6 +81,87 @@ function MultimodalLLMScreen() {
     if (vlm.error) setError(String(vlm.error));
   }, [vlm.error]);
 
+  useEffect(() => {
+    AudioManager.setAudioSessionOptions({
+      iosCategory: 'playAndRecord',
+      iosMode: 'spokenAudio',
+      iosOptions: ['allowBluetoothHFP', 'defaultToSpeaker'],
+    });
+    (async () => {
+      const status = await AudioManager.requestRecordingPermissions();
+      setHasMicPermission(status === 'Granted');
+    })();
+  }, []);
+
+  const loadAudioFromUrl = async () => {
+    const url = audioUrl.trim();
+    if (!url) return;
+    setIsFetchingAudio(true);
+    try {
+      const ctx = new AudioContext({ sampleRate: 16000 });
+      const decoded = await ctx.decodeAudioData(url);
+      const pcm = decoded.getChannelData(0);
+      const name = url.split('/').pop() || 'audio';
+      setAudioBuffer(pcm);
+      setAudioLabel(`${name} · ${(pcm.length / 16000).toFixed(1)}s`);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e));
+    } finally {
+      setIsFetchingAudio(false);
+    }
+  };
+
+  const startRecording = async () => {
+    if (!hasMicPermission) {
+      setError('Microphone permission denied. Please enable it in Settings.');
+      return;
+    }
+    recordChunks.current = [];
+    const sampleRate = 16000;
+    recorder.current.onAudioReady(
+      { sampleRate, bufferLength: 0.1 * sampleRate, channelCount: 1 },
+      ({ buffer }) => {
+        recordChunks.current.push(new Float32Array(buffer.getChannelData(0)));
+      }
+    );
+    try {
+      const ok = await AudioManager.setAudioSessionActivity(true);
+      if (!ok) {
+        setError('Cannot start audio session');
+        return;
+      }
+      const result = recorder.current.start();
+      if (result.status === 'error') {
+        setError(`Recording problems: ${result.message}`);
+        return;
+      }
+      setIsRecording(true);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e));
+    }
+  };
+
+  const stopRecording = () => {
+    recorder.current.stop();
+    setIsRecording(false);
+    const total = recordChunks.current.reduce((n, c) => n + c.length, 0);
+    if (total === 0) return;
+    const pcm = new Float32Array(total);
+    let off = 0;
+    for (const c of recordChunks.current) {
+      pcm.set(c, off);
+      off += c.length;
+    }
+    recordChunks.current = [];
+    setAudioBuffer(pcm);
+    setAudioLabel(`Recording · ${(pcm.length / 16000).toFixed(1)}s`);
+  };
+
+  const clearAudio = () => {
+    setAudioBuffer(null);
+    setAudioLabel(null);
+  };
+
   const pickImage = async () => {
     try {
       const result = await launchImageLibrary({ mediaType: 'photo' });
@@ -88,12 +182,19 @@ function MultimodalLLMScreen() {
     textInputRef.current?.clear();
     Keyboard.dismiss();
     const currentImageUri = imageUri;
+    const currentAudio = audioBuffer;
     setImageUri(null);
+    setAudioBuffer(null);
+    setAudioLabel(null);
     try {
-      await vlm.sendMessage(
-        text,
-        currentImageUri ? { imagePath: currentImageUri } : undefined
-      );
+      const media =
+        currentImageUri || currentAudio
+          ? {
+              ...(currentImageUri ? { imagePath: currentImageUri } : {}),
+              ...(currentAudio ? { audioBuffer: currentAudio } : {}),
+            }
+          : undefined;
+      await vlm.sendMessage(text, media);
     } catch (e) {
       // Updated to set UI error instead of just console.error
       setError(e instanceof Error ? e.message : String(e));
@@ -159,6 +260,42 @@ function MultimodalLLMScreen() {
             </TouchableOpacity>
           )}
 
+          {/* Audio URL input */}
+          <View style={styles.audioUrlRow}>
+            <TextInput
+              placeholder="Audio URL (mp3/wav/…)"
+              placeholderTextColor="#C1C6E5"
+              style={styles.audioUrlInput}
+              value={audioUrl}
+              onChangeText={setAudioUrl}
+              autoCapitalize="none"
+              autoCorrect={false}
+            />
+            <TouchableOpacity
+              style={[
+                styles.audioUrlButton,
+                (!audioUrl.trim() || isFetchingAudio || vlm.isGenerating) &&
+                  styles.disabled,
+              ]}
+              onPress={loadAudioFromUrl}
+              disabled={!audioUrl.trim() || isFetchingAudio || vlm.isGenerating}
+            >
+              <Text style={styles.audioUrlButtonText}>
+                {isFetchingAudio ? '…' : 'Load'}
+              </Text>
+            </TouchableOpacity>
+          </View>
+
+          {/* Audio attachment strip */}
+          {audioLabel && (
+            <View style={styles.audioAttachmentContainer}>
+              <Text style={styles.audioAttachmentText}>🎵 {audioLabel}</Text>
+              <TouchableOpacity onPress={clearAudio}>
+                <Text style={styles.audioAttachmentClear}>✕</Text>
+              </TouchableOpacity>
+            </View>
+          )}
+
           <StatsBar stats={stats} />
           <View
             style={[
@@ -176,6 +313,17 @@ function MultimodalLLMScreen() {
               disabled={vlm.isGenerating}
             >
               <Text style={styles.imageButtonText}>📷</Text>
+            </TouchableOpacity>
+
+            {/* Mic record / stop button */}
+            <TouchableOpacity
+              style={styles.imageButton}
+              onPress={isRecording ? stopRecording : startRecording}
+              disabled={vlm.isGenerating}
+            >
+              <Text style={styles.imageButtonText}>
+                {isRecording ? '⏹️' : '🎤'}
+              </Text>
             </TouchableOpacity>
 
             <TextInput
@@ -318,6 +466,64 @@ const styles = StyleSheet.create({
     fontSize: 12,
     fontFamily: 'regular',
     color: ColorPalette.blueDark,
+  },
+  audioAttachmentContainer: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    paddingHorizontal: 16,
+    paddingVertical: 8,
+    marginHorizontal: 16,
+    marginBottom: 4,
+    borderRadius: 8,
+    borderWidth: 1,
+    borderColor: ColorPalette.blueLight,
+    backgroundColor: '#fafbff',
+  },
+  audioAttachmentText: {
+    fontSize: 13,
+    fontFamily: 'regular',
+    color: ColorPalette.blueDark,
+  },
+  audioAttachmentClear: {
+    fontSize: 16,
+    color: ColorPalette.blueDark,
+    paddingHorizontal: 8,
+  },
+  audioUrlRow: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    marginHorizontal: 16,
+    marginBottom: 4,
+  },
+  audioUrlInput: {
+    flex: 1,
+    padding: 10,
+    borderTopLeftRadius: 8,
+    borderBottomLeftRadius: 8,
+    borderWidth: 1,
+    borderColor: ColorPalette.blueLight,
+    borderRightWidth: 0,
+    fontFamily: 'regular',
+    fontSize: 13,
+    color: ColorPalette.primary,
+  },
+  audioUrlButton: {
+    paddingVertical: 10,
+    paddingHorizontal: 16,
+    backgroundColor: ColorPalette.strongPrimary,
+    borderTopRightRadius: 8,
+    borderBottomRightRadius: 8,
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  audioUrlButtonText: {
+    color: '#fff',
+    fontFamily: 'medium',
+    fontSize: 13,
+  },
+  disabled: {
+    opacity: 0.5,
   },
   bottomContainer: {
     height: 100,
--- a/apps/llm/app/multimodal_llm/index.tsx
+++ b/apps/llm/app/multimodal_llm/index.tsx
@@ -64,7 +64,7 @@ function MultimodalLLMScreen() {
   const [error, setError] = useState<string | null>(null);
 
   const vlm = useLLM({
-    model: models.llm.gemma4_e2b(),
+    model: models.llm.gemma4_e2b_multimodal(),
   });
   const tokenCount = vlm.isReady ? vlm.getGeneratedTokenCount() : 0;
   const { stats, onMessageSend } = useLLMStats(
--- a/apps/llm/app/multimodal_llm/index.tsx
+++ b/apps/llm/app/multimodal_llm/index.tsx
@@ -244,7 +244,7 @@ function MultimodalLLMScreen() {
               <Text style={styles.helloText}>Hello! 👋</Text>
               <Text style={styles.bottomHelloText}>
                 {model.capabilities.find((c) => c === 'audio')
-                  ? 'Say hi or pick an image or and ask me anything about it.'
+                  ? 'Say "Hi" or pick an image or and ask me anything about it.'
                   : 'Pick an image and ask me anything about it.'}
               </Text>
               <SuggestedPrompts
--- a/apps/llm/app/multimodal_llm/index.tsx
+++ b/apps/llm/app/multimodal_llm/index.tsx
@@ -28,6 +28,7 @@ import Spinner from '../../components/Spinner';
 import { GeneratingContext } from '../../context';
 import SuggestedPrompts from '../../components/SuggestedPrompts';
 import ErrorBanner from '../../components/ErrorBanner';
+import AudioWaveform from '../../components/AudioWaveform';
 
 const SUGGESTED_PROMPTS = [
   "What's in this image?",
@@ -291,10 +292,16 @@ function MultimodalLLMScreen() {
           {/* Audio attachment strip */}
           {audioLabel && (
             <View style={styles.audioAttachmentContainer}>
-              <Text style={styles.audioAttachmentText}>🎵 {audioLabel}</Text>
-              <TouchableOpacity onPress={clearAudio}>
-                <Text style={styles.audioAttachmentClear}>✕</Text>
-              </TouchableOpacity>
+              <View style={styles.audioAttachmentRow}>
+                <Text style={styles.audioAttachmentText}>🎵 {audioLabel}</Text>
+                <TouchableOpacity onPress={clearAudio}>
+                  <Text style={styles.audioAttachmentClear}>✕</Text>
+                </TouchableOpacity>
+              </View>
+              <AudioWaveform
+                buffer={audioBuffer}
+                style={styles.audioWaveform}
+              />
             </View>
           )}
 
@@ -471,9 +478,7 @@ const styles = StyleSheet.create({
     color: ColorPalette.blueDark,
   },
   audioAttachmentContainer: {
-    flexDirection: 'row',
-    alignItems: 'center',
-    justifyContent: 'space-between',
+    flexDirection: 'column',
     paddingHorizontal: 16,
     paddingVertical: 8,
     marginHorizontal: 16,
@@ -482,6 +487,11 @@ const styles = StyleSheet.create({
     borderWidth: 1,
     borderColor: ColorPalette.blueLight,
     backgroundColor: '#fafbff',
+  },
+  audioAttachmentRow: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
   },
   audioAttachmentText: {
     fontSize: 13,
@@ -492,6 +502,10 @@ const styles = StyleSheet.create({
     fontSize: 16,
     color: ColorPalette.blueDark,
     paddingHorizontal: 8,
+  },
+  audioWaveform: {
+    marginTop: 6,
+    minWidth: 0,
   },
   audioUrlRow: {
     flexDirection: 'row',
--- a/apps/llm/app/multimodal_llm/index.tsx
+++ b/apps/llm/app/multimodal_llm/index.tsx
@@ -12,6 +12,11 @@ import {
   View,
 } from 'react-native';
 import { launchImageLibrary } from 'react-native-image-picker';
+import {
+  AudioManager,
+  AudioRecorder,
+  AudioContext,
+} from 'react-native-audio-api';
 import { useIsFocused } from '@react-navigation/native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { models, useLLM } from 'react-native-executorch';
@@ -46,7 +51,15 @@ function MultimodalLLMScreen() {
   const textInputRef = useRef<TextInput>(null);
   const { setGlobalGenerating } = useContext(GeneratingContext);
 
-  // Added error state
+  const [audioBuffer, setAudioBuffer] = useState<Float32Array | null>(null);
+  const [audioLabel, setAudioLabel] = useState<string | null>(null);
+  const [audioUrl, setAudioUrl] = useState('');
+  const [isFetchingAudio, setIsFetchingAudio] = useState(false);
+  const [isRecording, setIsRecording] = useState(false);
+  const [hasMicPermission, setHasMicPermission] = useState(false);
+  const recorder = useRef(new AudioRecorder());
+  const recordChunks = useRef<Float32Array[]>([]);
+
   const [error, setError] = useState<string | null>(null);
 
   const vlm = useLLM({
@@ -68,6 +81,87 @@ function MultimodalLLMScreen() {
     if (vlm.error) setError(String(vlm.error));
   }, [vlm.error]);
 
+  useEffect(() => {
+    AudioManager.setAudioSessionOptions({
+      iosCategory: 'playAndRecord',
+      iosMode: 'spokenAudio',
+      iosOptions: ['allowBluetoothHFP', 'defaultToSpeaker'],
+    });
+    (async () => {
+      const status = await AudioManager.requestRecordingPermissions();
+      setHasMicPermission(status === 'Granted');
+    })();
+  }, []);
+
+  const loadAudioFromUrl = async () => {
+    const url = audioUrl.trim();
+    if (!url) return;
+    setIsFetchingAudio(true);
+    try {
+      const ctx = new AudioContext({ sampleRate: 16000 });
+      const decoded = await ctx.decodeAudioData(url);
+      const pcm = decoded.getChannelData(0);
+      const name = url.split('/').pop() || 'audio';
+      setAudioBuffer(pcm);
+      setAudioLabel(`${name} · ${(pcm.length / 16000).toFixed(1)}s`);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e));
+    } finally {
+      setIsFetchingAudio(false);
+    }
+  };
+
+  const startRecording = async () => {
+    if (!hasMicPermission) {
+      setError('Microphone permission denied. Please enable it in Settings.');
+      return;
+    }
+    recordChunks.current = [];
+    const sampleRate = 16000;
+    recorder.current.onAudioReady(
+      { sampleRate, bufferLength: 0.1 * sampleRate, channelCount: 1 },
+      ({ buffer }) => {
+        recordChunks.current.push(new Float32Array(buffer.getChannelData(0)));
+      }
+    );
+    try {
+      const ok = await AudioManager.setAudioSessionActivity(true);
+      if (!ok) {
+        setError('Cannot start audio session');
+        return;
+      }
+      const result = recorder.current.start();
+      if (result.status === 'error') {
+        setError(`Recording problems: ${result.message}`);
+        return;
+      }
+      setIsRecording(true);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e));
+    }
+  };
+
+  const stopRecording = () => {
+    recorder.current.stop();
+    setIsRecording(false);
+    const total = recordChunks.current.reduce((n, c) => n + c.length, 0);
+    if (total === 0) return;
+    const pcm = new Float32Array(total);
+    let off = 0;
+    for (const c of recordChunks.current) {
+      pcm.set(c, off);
+      off += c.length;
+    }
+    recordChunks.current = [];
+    setAudioBuffer(pcm);
+    setAudioLabel(`Recording · ${(pcm.length / 16000).toFixed(1)}s`);
+  };
+
+  const clearAudio = () => {
+    setAudioBuffer(null);
+    setAudioLabel(null);
+  };
+
   const pickImage = async () => {
     try {
       const result = await launchImageLibrary({ mediaType: 'photo' });
@@ -88,12 +182,19 @@ function MultimodalLLMScreen() {
     textInputRef.current?.clear();
     Keyboard.dismiss();
     const currentImageUri = imageUri;
+    const currentAudio = audioBuffer;
     setImageUri(null);
+    setAudioBuffer(null);
+    setAudioLabel(null);
     try {
-      await vlm.sendMessage(
-        text,
-        currentImageUri ? { imagePath: currentImageUri } : undefined
-      );
+      const media =
+        currentImageUri || currentAudio
+          ? {
+              ...(currentImageUri ? { imagePath: currentImageUri } : {}),
+              ...(currentAudio ? { audioBuffer: currentAudio } : {}),
+            }
+          : undefined;
+      await vlm.sendMessage(text, media);
     } catch (e) {
       // Updated to set UI error instead of just console.error
       setError(e instanceof Error ? e.message : String(e));
@@ -159,6 +260,42 @@ function MultimodalLLMScreen() {
             </TouchableOpacity>
           )}
 
+          {/* Audio URL input */}
+          <View style={styles.audioUrlRow}>
+            <TextInput
+              placeholder="Audio URL (mp3/wav/…)"
+              placeholderTextColor="#C1C6E5"
+              style={styles.audioUrlInput}
+              value={audioUrl}
+              onChangeText={setAudioUrl}
+              autoCapitalize="none"
+              autoCorrect={false}
+            />
+            <TouchableOpacity
+              style={[
+                styles.audioUrlButton,
+                (!audioUrl.trim() || isFetchingAudio || vlm.isGenerating) &&
+                  styles.disabled,
+              ]}
+              onPress={loadAudioFromUrl}
+              disabled={!audioUrl.trim() || isFetchingAudio || vlm.isGenerating}
+            >
+              <Text style={styles.audioUrlButtonText}>
+                {isFetchingAudio ? '…' : 'Load'}
+              </Text>
+            </TouchableOpacity>
+          </View>
+
+          {/* Audio attachment strip */}
+          {audioLabel && (
+            <View style={styles.audioAttachmentContainer}>
+              <Text style={styles.audioAttachmentText}>🎵 {audioLabel}</Text>
+              <TouchableOpacity onPress={clearAudio}>
+                <Text style={styles.audioAttachmentClear}>✕</Text>
+              </TouchableOpacity>
+            </View>
+          )}
+
           <StatsBar stats={stats} />
           <View
             style={[
@@ -176,6 +313,17 @@ function MultimodalLLMScreen() {
               disabled={vlm.isGenerating}
             >
               <Text style={styles.imageButtonText}>📷</Text>
+            </TouchableOpacity>
+
+            {/* Mic record / stop button */}
+            <TouchableOpacity
+              style={styles.imageButton}
+              onPress={isRecording ? stopRecording : startRecording}
+              disabled={vlm.isGenerating}
+            >
+              <Text style={styles.imageButtonText}>
+                {isRecording ? '⏹️' : '🎤'}
+              </Text>
             </TouchableOpacity>
 
             <TextInput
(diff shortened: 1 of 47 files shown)


packages/react-native-executorch/third-party/android/libs/executorch/arm64-v8a/libexecutorch.so
benITo47, line 1: What was changed in the binaries? Asking for the record.

msluszniak: Adding Vulkan to RNE

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.cpp
```diff
@@ -74,63 +76,68 @@ std::string LLM::generate(std::string input,
...
-  size_t searchPos = 0;
-
-  while (true) {
-    size_t found = prompt.find(imageToken, searchPos);
-    if (found == std::string::npos) {
-      if (searchPos < prompt.size()) {
-        inputs.push_back(llm::make_text_input(prompt.substr(searchPos)));
-      }
+  size_t imageIdx = 0, audioIdx = 0, pos = 0;
+  constexpr int32_t kAudioSampleRate = 16000;
```
msluszniak: What if in the future some other model will accept different sample rate or multiple sample rates? Every time we add such model, we will need to break this file that suppose to be general, for every llm. This is a code smell.

packages/react-native-executorch/common/rnexecutorch/models/llm/LLM.h
```diff
@@ -22,10 +22,16 @@ class LLM : public BaseModel {
...
-                                 std::shared_ptr<jsi::Function> callback);
+  // Audio variant: `audioWaveforms` is a parallel vector of fp32 mono 16 kHz
+  // PCM buffers (decoded upstream, same contract as SpeechToText::transcribe).
+  // The prompt is scanned for `imageToken` and/or `audioToken` placeholders;
+  // each placeholder consumes the next entry from its respective vector in
+  // order. Either set of paths/waveforms/token may be empty.
+  std::string generateMultimodal(
+      std::string prompt, std::shared_ptr<jsi::Function> callback,
+      std::vector<std::string> imagePaths = {}, std::string imageToken = "",
+      std::vector<std::vector<float>> audioWaveforms = {},
```
msluszniak: Again, what if in the future multimodal llm will accept video as an input, we need to add another positional argument and break common header. Bad code design that was there and we just extend it. But to be fair, if we plan to refactor whole codebase using this approach proposed by Bartek: see https://github.com/software-mansion/react-native-executorch/issues/1208, then I think it doesn't matter as much as if we would keep this codebase as standalone one.

packages/react-native-executorch/common/runner/encoders/audio_encoder.cpp
```diff
@@ -0,0 +1,116 @@
...
+        "Check that the .pte file matches the declared capabilities.");
+  }
+  return module_->load_method(kAudioEncoderMethod);
+}
+
+bool AudioEncoder::is_loaded() const noexcept {
+  return module_->is_method_loaded(kAudioEncoderMethod);
+}
+
+int32_t AudioEncoder::encoderTokenCount() const { return last_token_count_; }
```
msluszniak: Probably noexcept

packages/react-native-executorch/common/runner/encoders/audio_encoder.cpp
```diff
@@ -0,0 +1,116 @@
...
+  ET_CHECK_OR_RETURN_ERROR(
+      k_blocks >= kAudioBlockKMin && k_blocks <= kAudioBlockKMax,
+      InvalidArgument,
+      "AudioEncoder: waveform of %lld samples needs k_blocks=%lld; "
+      "audio_encoder accepts k in [%lld, %lld] (block=%d samples; max %.2f s "
+      "@ 16 kHz)",
+      static_cast<long long>(n_valid), static_cast<long long>(k_blocks),
+      static_cast<long long>(kAudioBlockKMin),
+      static_cast<long long>(kAudioBlockKMax),
+      static_cast<int>(kSamplesPerBlock),
```
msluszniak: Why do you mix `long long` and `int64_t`?

packages/react-native-executorch/common/runner/multimodal_decoder_runner.h
```diff
@@ -46,6 +77,24 @@ class MultimodalDecoderRunner : public TextDecoderRunner {
...
+    if (!outputs_result.ok()) {
+      return outputs_result.error();
+    }
+    auto &outputs = *outputs_result;
+    ET_CHECK_MSG(outputs.size() == 1,
+                 "Expected 1 output from text_decoder, got %zu",
+                 outputs.size());
+    ET_CHECK_MSG(outputs[0].isTensor(), "text_decoder output is not a tensor");
+    return outputs[0].toTensor();
+  }
```
msluszniak, line 96: Similar to previous comments, PLE is related to gemma 4 or similar models. If there will be gemma 5 with another quirks, then we need to break this contract again.

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -92,32 +93,52 @@ int32_t Sampler::sample_topp(T *probabilities, float coin) {
...
+    return;
+  }
+  // Partial-select the (topk_-th largest) threshold using nth_element on a
+  // copy of logits; O(n) average.
+  std::vector<T> scratch(logits, logits + vocab_size_);
+  std::nth_element(scratch.begin(), scratch.begin() + (topk_ - 1),
+                   scratch.end(), std::greater<T>());
+  const T threshold = scratch[topk_ - 1];
+  const T neg_inf = std::numeric_limits<T>::lowest();
+  for (int i = 0; i < vocab_size_; i++) {
```
msluszniak:
```suggestion
  for (size_t i = 0; i < vocab_size_; i++) {
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -126,10 +147,80 @@ Sampler::Sampler(int32_t vocab_size, float temperature, float topp,
...
+
+// Mask logits whose softmax-prob falls outside the top-p nucleus to -inf.
+// Keeps the token that crosses the threshold (HuggingFace convention).
+template <typename T> void Sampler::mask_topp(T *logits) {
+  if (topp_ <= 0.0f || topp_ >= 1.0f) {
+    return;
+  }
+  // Softmax into a scratch probs[] (do not mutate logits yet).
+  T max_val = logits[0];
+  for (int i = 1; i < vocab_size_; i++) {
```
msluszniak:
```suggestion
  for (size_t i = 1; i < vocab_size_; i++) {
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -92,32 +93,52 @@ int32_t Sampler::sample_topp(T *probabilities, float coin) {
...
+  const T threshold = scratch[topk_ - 1];
+  const T neg_inf = std::numeric_limits<T>::lowest();
+  for (int i = 0; i < vocab_size_; i++) {
+    if (logits[i] < threshold) {
+      logits[i] = neg_inf;
+    }
+  }
 }
 
 Sampler::Sampler(int32_t vocab_size, float temperature, float topp,
```
msluszniak: Why `vocab_size` is signed int?

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -126,10 +147,80 @@ Sampler::Sampler(int32_t vocab_size, float temperature, float topp,
...
+  for (int i = 1; i < vocab_size_; i++) {
+    if (logits[i] > max_val) {
+      max_val = logits[i];
+    }
+  }
+  std::unique_ptr<ProbIndex<T>[]> probindex =
+      std::make_unique<ProbIndex<T>[]>(vocab_size_);
+  T sum = 0;
+  for (int i = 0; i < vocab_size_; i++) {
+    T e = static_cast<T>(expf(static_cast<float>(logits[i] - max_val)));
```
msluszniak:
```suggestion
    T e = static_cast<T>(std::expf(static_cast<float>(logits[i] - max_val)));
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -126,10 +147,80 @@ Sampler::Sampler(int32_t vocab_size, float temperature, float topp,
...
+    probindex[i].prob = e;
+    probindex[i].index = i;
+    sum += e;
+  }
+  if (sum <= T(0)) {
+    return;
+  }
+  for (int i = 0; i < vocab_size_; i++) {
```
msluszniak: And fix other places as such
```suggestion
  for (size_t i = 0; i < vocab_size_; i++) {
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -126,10 +147,80 @@ Sampler::Sampler(int32_t vocab_size, float temperature, float topp,
...
       rng_state_(rng_seed) {}
 
+Sampler::Sampler(int vocab_size, float temperature, float topp, int32_t topk)
+    : Sampler(vocab_size, temperature, topp, topk, std::time(nullptr)) {}
+
+Sampler::Sampler(int vocab_size, float temperature, float topp,
+                 unsigned long long rng_seed)
+    : Sampler(vocab_size, temperature, topp, 0, rng_seed) {}
+
 Sampler::Sampler(int vocab_size, float temperature, float topp)
```
msluszniak: Poor design, can't we somehow avoid 5 different constructors which take or not some subset of arguments?

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -124,6 +125,37 @@ export const QWEN3_0_6B_QUANTIZED = {
   generationConfig: QWEN3_GENERATION_CONFIG,
 } as const;
 
+// GEMMA 4
+const GEMMA4_E2B_XNNPACK = `${URL_PREFIX}-gemma-4/${VERSION_TAG}/e2b/xnnpack/gemma_4_e2b_xnnpack_8da4w.pte`;
+const GEMMA4_E2B_VULKAN = `${URL_PREFIX}-gemma-4/${VERSION_TAG}/e2b/vulkan/gemma_4_e2b_vulkan_8da4w.pte`;
```
msluszniak: Out of curiosity, have you tried export for `CoreML`?

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -16,6 +16,13 @@ import { Logger } from '../common/Logger';
 import { RnExecutorchError, parseUnknownError } from '../errors/errorUtils';
 import { RnExecutorchErrorCode } from '../errors/ErrorCodes';
 
+// Audio soft-token expansion constants for Gemma4's audio_encoder.
+// Mirrors AUDIO_SAMPLES_PER_BLOCK (kSamplesPerBlock=7680) and the per-block
+// soft-token rate in audio_encoder.cpp; used to size the context budget so
+// long audio doesn't silently overflow get_max_seq_len during prefill.
+const AUDIO_SAMPLES_PER_BLOCK = 7680;
+const AUDIO_TOKENS_PER_BLOCK = 12;
```
msluszniak: Similar to cpp, `LLMController` is a general file for LLMs, but `AUDIO_SAMPLES_PER_BLOCK = 7680` is value specific for gemma 4, shouldn't be there.

msluszniak (changes requested):

packages/react-native-executorch/common/runner/multimodal_prefiller.cpp
```diff
@@ -23,91 +26,410 @@ using ::executorch::runtime::Result;
...
+      ET_CHECK_OR_RETURN_ERROR(audio_encoder_ != nullptr, InvalidState,
+                               "No audio encoder registered");
+      const long t_aud_begin = time_in_ms();
+      auto enc = audio_encoder_->encode(input);
+      ET_CHECK_OK_OR_RETURN_ERROR(enc.error(), "Audio encoding failed");
+      audio_encode_ms += time_in_ms() - t_aud_begin;
+      audio_calls += 1;
+      // Snapshot the encoder output NOW — see AudioSlot comment above for
+      // why the returned EValue's tensor metadata can't survive past the
+      // next module_->execute(). num_audio and audio_hidden are read from
```
msluszniak: As we dicussed, this must be refactored. It is hard to read, functions are way too long. Here each `if` block might be a separate function. Please create issue for the refactor if the multimodal prefiller if you don't want to handle this in the PR.

msluszniak (changes requested):
CI needs to be fixed. Also please add testing instructions.

packages/react-native-executorch/src/controllers/LLMController.ts
mkopcins: yes, absolutely agreed, those are leftovers from iterating and I missed them in self-review

packages/react-native-executorch/src/constants/modelUrls.ts
mkopcins: no, but right now mlx looks to be the better way forward

packages/react-native-executorch/common/runner/sampler.cpp
mkopcins: not a clue tbh, this is meta's code. My best guess is for no reason

msluszniak (changes requested):
You should change HF for gemma 4 models:
Both text only and multimodal does not comply with the schema:
- tokenizers and configs are repeated in backends, but in other models are in root. 
- they miss README.md.
- they miss dummy config file with `modelName`, that is used by HF for model counting.
- they are not added to collection LLM (Multimodal)

packages/react-native-executorch/src/hooks/natural_language_processing/useLLM.ts
```diff
@@ -106,7 +107,10 @@ export function useLLM({
   );
 
   const sendMessage = useCallback(
-    (message: string, media?: { imagePath?: string }) => {
+    (
+      message: string,
+      media?: { imagePath?: string; audioBuffer?: Float32Array }
```
msluszniak, line 107: This is addition to the current API, but documentation hasn't been changed. Please update docs.

msluszniak (changes requested):

packages/react-native-executorch/common/runner/multimodal_decoder_runner.h
mkopcins: yes, this is a real concern. But this requires a bigger refactor which I think is beyond this PR. We would need to refactor the entire runner/ directory to be more modular to be able to handle different contracts.

docs/docs/04-typescript-api/01-natural-language-processing/LLMModule.md
```diff
@@ -133,20 +133,24 @@ The `capabilities` field is already set on the model constant. You can also cons
...
   modelSource: require('./path/to/model.pte'),
   tokenizerSource: require('./path/to/tokenizer.json'),
   tokenizerConfigSource: require('./path/to/tokenizer_config.json'),
-  capabilities: ['vision'],
+  capabilities: ['vision', 'audio'],
 });
 ```
 
-Once loaded, pass `imagePath` to [`sendMessage`](../../06-api-reference/classes/LLMModule.md#sendmessage):
+Once loaded, pass `imagePath`Once loaded, pass `imagePath` or `audioBuffer` to [`sendMessage`](../../06-api-reference/classes/LLMModule.md#sendmessage):
```
chmjkb:
```suggestion
Once loaded, pass `imagePath` or `audioBuffer` to [`sendMessage`](../../06-api-reference/classes/LLMModule.md#sendmessage):
```

chmjkb (changes requested):

docs/plans/2026-06-01-multimodal-prefiller-refactor-design.md
chmjkb, line 1: should this be here?

docs/plans/2026-06-01-multimodal-prefiller-refactor.md
chmjkb, line 1: should this be here?

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
```diff
@@ -223,6 +226,22 @@ inline std::vector<float> getValue<std::vector<float>>(const jsi::Value &val,
...
+  std::vector<std::vector<float>> result;
+  result.reserve(length);
+  for (size_t i = 0; i < length; ++i) {
+    auto span =
+        getTypedArrayAsSpan<float>(array.getValueAtIndex(runtime, i), runtime);
+    result.emplace_back(span.begin(), span.end());
+  }
+  return result;
+}
+
```
chmjkb, line 243: Could this return vector<span<float>> and skip the copy? Or does the downstream consumer need ownership?

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -423,6 +429,8 @@ template <typename Model> class ModelHostObject : public JsiHostObject {
                         });
                     return;
                   } catch (const std::exception &e) {
+                    rnexecutorch::log(rnexecutorch::LOG_LEVEL::Info, "kappa",
+                                      e.what());
```
chmjkb: you always use this log xD

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -351,9 +353,11 @@ template <typename Model> class ModelHostObject : public JsiHostObject {
         return jsi_conversion::getJsiValue(std::move(result), runtime);
       }
     } catch (const RnExecutorchError &e) {
+      rnexecutorch::log(rnexecutorch::LOG_LEVEL::Info, "kappa", e.what());
```
msluszniak: What is kappa here?

packages/react-native-executorch/common/runner/encoders/audio_encoder.cpp
chmjkb, line 1: not going to block the PR over this, however this is getting pretty similar to vision encoder, maybe we can lift something up

packages/react-native-executorch/common/runner/base_llm_runner.cpp
chmjkb, line 1: i believe we should sanity check the inputs provided by user in setters here

packages/react-native-executorch/common/runner/constants.h
```diff
@@ -23,8 +23,22 @@ inline constexpr auto kVisionEncoderMethod = "vision_encoder";
...
 inline constexpr auto numOfAddedBoSTokens = 0;
 inline constexpr auto numOfAddedEoSTokens = 0;
 
+// Gemma4
+// PLE models only: token id that marks image placeholder slots in input_ids.
+// token_embedding run on this id produces the per-layer PLE signal for image
+// positions; the inputs_embeds output for those positions is discarded (the
+// vision encoder output replaces it).
+inline constexpr auto kImagePlaceholderId = "image_placeholder_id";
+// True iff the model exposes a per-layer-embedding (PLE) signal alongside
```
chmjkb, line 35:
```suggestion
// True if the model exposes a per-layer-embedding (PLE) signal alongside
```

packages/react-native-executorch/src/hooks/natural_language_processing/useLLM.ts
```diff
@@ -58,11 +58,12 @@ export function useLLM({
...
-          tokenizerSource: model.tokenizerSource,
-          tokenizerConfigSource: model.tokenizerConfigSource!,
-          capabilities: model.capabilities,
-          defaultGenerationConfig: model.generationConfig,
+          model: model,
+          // modelSource: model.modelSource,
+          // tokenizerSource: model.tokenizerSource,
+          // tokenizerConfigSource: model.tokenizerConfigSource!,
+          // capabilities: model.capabilities,
+          // defaultGenerationConfig: model.generationConfig,
```
msluszniak: probably should be removed

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
mkopcins: ignore, stale logs

packages/react-native-executorch/common/runner/text_prefiller.cpp
```diff
@@ -10,6 +10,7 @@
 // LLM.
 
 #include "text_prefiller.h"
+#include "rnexecutorch/Log.h"
```
chmjkb: redundant include

packages/react-native-executorch/src/hooks/natural_language_processing/useLLM.ts
chmjkb: why is this commented out?

chmjkb (changes requested):

packages/react-native-executorch/common/runner/constants.h
mkopcins: iff == if and only if

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
mkopcins: I love this one, removed now

packages/react-native-executorch/common/rnexecutorch/host_objects/JsiConversions.h
msluszniak: Nah, we cannot return vector of spans, it would create dangling memory throughout executions. What is more important here to refine is change from:
```cpp
auto span = getTypedArrayAsSpan<float>(array.getValueAtIndex(runtime, i), runtime);
result.emplace_back(span.begin(), span.end());
```

to

```cpp
jsi::Value element = array.getValueAtIndex(runtime, i);
auto span = getTypedArrayAsSpan<float>(element, runtime);
result.emplace_back(span.begin(), span.end());
```
since `array.getValueAtIndex(runtime, i)` is e temporary object and it's lifespan ends at the end of the line.

packages/react-native-executorch/common/runner/multimodal_prefiller.cpp
```diff
@@ -23,91 +26,369 @@ using ::executorch::runtime::Result;
...
+      ids.push_back(static_cast<int64_t>(t));
+    }
+  } else if (input.is_tokens()) {
+    std::vector<uint64_t> tokens = input.get_tokens();
+    for (auto t : tokens) {
+      ids.push_back(static_cast<int64_t>(t));
     }
+    ET_LOG(Error, "Unsupported MultimodalInput type");
+    return Error::NotSupported;
+  }
```
msluszniak, line 118: I guess this was the original intention
```suggestion
  } else if (input.is_tokens()) {
    std::vector<uint64_t> tokens = input.get_tokens();
    for (auto t : tokens) {
      ids.push_back(static_cast<int64_t>(t));
}
  }
    ET_LOG(Error, "Unsupported MultimodalInput type");
    return Error::NotSupported;
```

msluszniak:
Please change the PR title. Now, I'm testing demo app using testing instructions.

msluszniak (changes requested):
I got this error after sending image with audio, and then only audio, ready to apply fixes are in the comments below this comment. Here is an exact error:

Image

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -124,16 +117,20 @@ export class LLMController {
         this.nativeModule.unload();
       }
 
+      if ('audio' in (model.capabilities ?? [])) {
```
msluszniak: **Root cause of the `Cannot read property 'samplesPerBlock' of undefined` crash** (reported above on the 2nd audio message).

`capabilities` is an **array**, so the `in` operator checks for an index/key, not membership — `'audio' in ['vision', 'audio']` is always `false`, so `audioConfig` is never assigned and stays `undefined`. Use `Array.prototype.includes`:

```suggestion
      if ((model.capabilities ?? []).includes('audio')) {
```

msluszniak (commented):
Suggested fixes for the `Cannot read property 'samplesPerBlock' of undefined` error from my review above. Reproduces on the second audio message (image+audio then audio, or two consecutive audios); the first works only because `SlidingWindowContextStrategy` skips the token-counting callback while history length is 1. **Both inline suggestions below are required** to fix it; the third is optional cleanup. Verified on a physical Android device.

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -124,6 +124,41 @@ export const QWEN3_0_6B_QUANTIZED = {
...
+ * @category Models - VLM
+ */
+export const GEMMA4_E2B_MM = {
+  modelName: 'gemma4-e2b-multimodal',
+  modelSource:
+    Platform.OS === `android` ? GEMMA4_E2B_VULKAN_MM : GEMMA4_E2B_XNNPACK_MM,
+  tokenizerSource: GEMMA4_TOKENIZER,
+  tokenizerConfigSource: GEMMA4_TOKENIZER_CONFIG,
+  capabilities: ['vision', 'audio'],
+} as const;
```
msluszniak, line 160: Second half of the same bug: `GEMMA4_E2B_MM` declares `capabilities: ['vision', 'audio']` but has **no `audioConfig`**, so even after the `.includes('audio')` fix `model.audioConfig` is `undefined` and the crash persists. Add the audio config to the multimodal model:

```suggestion
  audioConfig: {
    samplesPerBlock: 7680,
    tokensPerBlock: 12,
  },
} as const;
```

packages/react-native-executorch/src/constants/modelUrls.ts
```diff
@@ -124,6 +124,41 @@ export const QWEN3_0_6B_QUANTIZED = {
...
+export const GEMMA4_E2B = {
+  modelName: 'gemma4-e2b',
+  modelSource:
+    Platform.OS === `android` ? GEMMA4_E2B_VULKAN : GEMMA4_E2B_XNNPACK,
+  tokenizerSource: GEMMA4_TOKENIZER,
+  tokenizerConfigSource: GEMMA4_TOKENIZER_CONFIG,
+  audioConfig: {
+    samplesPerBlock: 7680,
+    tokensPerBlock: 12,
+  },
```
msluszniak: Cleanup: this `audioConfig` is on the **text-only** model, which has no `audio` capability, so it's never read (and is misleading). It belongs on `GEMMA4_E2B_MM` (see the suggestion there). Remove it here:

```suggestion
```

msluszniak (commented):
We needs to add some kind of symbol to messages with audio attached, now it is not clear for the user. Also, how big is the context in this model? It seems to be really small which makes it hard to keep conversation with this model using multimodal inputs. Another thing which is concerning is that the TTFT rapidly grows when we keep sending messages.

msluszniak:
Clicking on the send button does not block UI which is not intuitive for the user that something is already happening.

mkopcins:
context is 2048 tokens

packages/react-native-executorch/common/runner/multimodal_decoder_runner.h
msluszniak: Created an issue as a follow-up, so we can resolve this comment: https://github.com/software-mansion/react-native-executorch/issues/1225

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -124,16 +117,20 @@ export class LLMController {
         this.nativeModule.unload();
       }
 
+      if ((model.capabilities ?? []).find((c) => c == 'audio')) {
```
chmjkb: fix linter warning here

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -269,25 +283,38 @@ export class LLMController {
...
     if (!this._isReady) {
       throw new RnExecutorchError(RnExecutorchErrorCode.ModuleNotLoaded);
     }
     if (this._isGenerating) {
       throw new RnExecutorchError(RnExecutorchErrorCode.ModelGenerating);
     }
+    const hasImages = !!imagePaths && imagePaths.length > 0;
+    const hasAudio = !!audioWaveforms && audioWaveforms.length > 0;
     try {
+      console.log('kappa');
```
chmjkb: kappa

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -355,29 +384,35 @@ export class LLMController {
...
+    const audioWaveforms = messages
+      .filter((m) => m.audioWaveform)
+      .map((m) => m.audioWaveform!);
     const renderedChat: string = this.applyChatTemplate(
       messages,
       this.tokenizerConfig,
       tools,
       { tools_in_user_message: false, add_generation_prompt: true }
     );
+    console.log('kappa', performance.now() - s);
```
chmjkb: kappa

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -334,6 +361,8 @@ export class LLMController {
     messages: Message[],
     tools?: LLMTool[]
   ): Promise<string> {
+    let s = performance.now();
+    console.log('kappa', performance.now() - s);
```
chmjkb: kappa

chmjkb (commented):
i'll test the app on iOS

chmjkb:
Something feels off about the demo app 😅 

IMG_0157

packages/react-native-executorch/common/runner/multimodal_prefiller.cpp
```diff
@@ -23,91 +26,369 @@ using ::executorch::runtime::Result;
...
+  std::vector<int64_t> cache_positions(static_cast<size_t>(total_len));
+  for (int64_t i = 0; i < total_len; ++i) {
+    cache_positions[static_cast<size_t>(i)] = start_pos + i;
+  }
+  const int64_t num_chunks = (total_len + chunk_cap - 1) / chunk_cap;
+  for (int64_t ci = 0; ci < num_chunks; ++ci) {
+    const int64_t chunk_start = ci * chunk_cap;
+    const int64_t chunk_end = std::min(chunk_start + chunk_cap, total_len);
+    const int64_t chunk_len = chunk_end - chunk_start;
+    prefillChunk(last_outs, embeds_buf, chunk_start, chunk_len, hidden,
```
msluszniak: **Return values of all prefill helpers are discarded.** `prefill()` ignores the `Error`/`Result` from `processMultimodalInput` (L303), `encodeImages` (L353), `encodeAudio` (L362), `initializePLE` (L367) and `prefillChunk` (here). The `ET_CHECK_OK_OR_RETURN_ERROR` inside each helper only returns from the helper, so a failed `module_->execute(kTextModelMethod, …)`, an `encode()` failure, or a dtype mismatch is silently swallowed and execution continues to `last_outs[0].toTensor()` on stale/empty outputs (wrong result or a crash) instead of surfacing a clean error to JS. Wrap each call in `ET_CHECK_OK_OR_RETURN_ERROR(...)`. **Coupled with the L112-114 issue — fix both together.**

packages/react-native-executorch/common/runner/multimodal_prefiller.cpp
```diff
@@ -23,91 +26,369 @@ using ::executorch::runtime::Result;
...
+      ids.push_back(static_cast<int64_t>(t));
     }
+  } else if (input.is_tokens()) {
+    std::vector<uint64_t> tokens = input.get_tokens();
+    for (auto t : tokens) {
+      ids.push_back(static_cast<int64_t>(t));
+    }
+  }
+  ET_LOG(Error, "Unsupported MultimodalInput type");
+  return Error::NotSupported;
```
msluszniak: **`processMultimodalInput` has no success (`Error::Ok`) return path.** After any valid branch (image/audio/text/tokens) control falls through to the `Unsupported` log + `return Error::NotSupported`. (The `is_text` failure path does return `InvalidArgument`, but no branch returns success.) Today this is masked only because the caller ignores the return (see the L382 comment); fixing that without this would make every prefill fail. Route the fall-through through an `else`:

```suggestion
  } else {
    ET_LOG(Error, "Unsupported MultimodalInput type");
    return Error::NotSupported;
  }
  return Error::Ok;
```

packages/react-native-executorch/common/runner/multimodal_prefiller.cpp
```diff
@@ -23,91 +26,369 @@ using ::executorch::runtime::Result;
...
+                                        PLEEmbeddings &ple_embeddings) {
+  auto full_ple_tok = embed_outputs[1].toTensor();
+  ple_embeddings.num_layers = static_cast<SizesType>(full_ple_tok.size(2));
+  ple_embeddings.ple_dim = static_cast<SizesType>(full_ple_tok.size(3));
+  ple_embeddings.ple_tok_dtype = full_ple_tok.scalar_type();
+  const size_t total_numel = static_cast<size_t>(full_ple_tok.numel());
+  const size_t total_bytes = full_ple_tok.nbytes();
+  ET_CHECK_OR_RETURN_ERROR(total_numel > 0, InvalidState,
+                           "ple_tok has zero elements");
+  size_t ple_elem_size = total_bytes / total_numel;
```
msluszniak: **PLE chunk offset is 0 for every chunk after the first → corrupts long (multi-chunk) prompts.** This local `ple_elem_size` shadows the struct field `PLEEmbeddings::ple_elem_size` (which stays `0`), so `prefillChunk` (L253) computes `ple_chunk_ptr` with element size `0` and reads PLE data from offset `0` for any `chunk_start > 0`. Single-chunk prompts accidentally work, which hides it. Assign the field:

```suggestion
  ple_embeddings.ple_elem_size = total_bytes / total_numel;
  const size_t ple_elem_size = ple_embeddings.ple_elem_size;
```

msluszniak (commented):
Review of the latest state (`ccb5526c1`) across the native runner, TS layer, and the example app. Inline comments group into: **confirmed correctness bugs** — the prefiller return-handling (`processMultimodalInput` has no `Error::Ok` path + `prefill()` discards every helper's return; these are coupled), the PLE element-size shadowing that zeroes the per-chunk offset on long prompts, and the `LLMModule` imperative API dropping `audioConfig` — plus robustness/API gaps and nits. A few items are phrased as questions where I couldn't confirm intent (e.g. the audio-encoder scalar). Verified against the code; happy to discuss any of them.

packages/react-native-executorch/common/runner/multimodal_prefiller.cpp
```diff
@@ -23,91 +26,369 @@ using ::executorch::runtime::Result;
...
+  // ------------------------------------------------------------
+  int64_t max_seq_len = get_max_seq_len().value_or(-1);
+  int64_t max_context_len = get_max_context_len().value_or(max_seq_len);
+  bool enable_dynamic_shape = get_enable_dynamic_shape();
+  const int64_t prefill_total_cap =
+      enable_dynamic_shape ? max_context_len : max_seq_len;
+  const int64_t decoder_chunk_size = max_seq_len;
+
+  std::vector<int64_t> ids;
+  ids.reserve(static_cast<size_t>(prefill_total_cap));
```
msluszniak, line 302: Robustness: if `get_max_seq_len()` is `nullopt`, `max_seq_len` is `-1`, `prefill_total_cap` becomes `-1`, and `ids.reserve(static_cast<size_t>(-1))` throws `length_error`/`bad_alloc` (an exception across the JSI boundary); the later `total_len <= prefill_total_cap` check is also meaningless. Guard `prefill_total_cap > 0` with `ET_CHECK_OR_RETURN_ERROR` before reserving.

packages/react-native-executorch/common/runner/multimodal_prefiller.h
```diff
@@ -18,26 +18,74 @@
...
+  int64_t num_audio;
+  int64_t audio_hidden;
+};
+
+struct PLEEmbeddings {
+  std::vector<uint8_t> ple_tok_buf;
+  aten::SizesType num_layers = 0;
+  aten::SizesType ple_dim = 0;
+  size_t ple_elem_size = 0;
+  ::executorch::aten::ScalarType ple_tok_dtype;
```
msluszniak: Header nits: `PLEEmbeddings::ple_tok_dtype` has no default initializer (the other members do) — give it one. Also `ImageSlot`/`AudioSlot`/`PLEEmbeddings` are declared in an anonymous namespace in a header (each TU gets its own copy), and the prefiller's helper methods (`processMultimodalInput`/`encodeImages`/`encodeAudio`/`prefillChunk`/`initializePLE`) are `public` though they're internal — consider a named namespace and moving them to `private`.

packages/react-native-executorch/common/runner/encoders/audio_encoder.cpp
```diff
@@ -0,0 +1,111 @@
...
+      InvalidArgument,
+      "AudioEncoder: waveform of %lld samples needs k_blocks=%lld.",
+      static_cast<long long>(n_valid), static_cast<long long>(k_blocks));
+  const int64_t n_padded = k_blocks * kSamplesPerBlock;
+
+  // Own the padded waveform for the lifetime of this call; from_blob below
+  // borrows without copying. The current export takes
+  //   forward(waveform[1, 7680*k] fp32, num_blocks: int64 scalar)
+  // — input 1 is a rank-0 Long telling the encoder how many of the K_MAX
+  // blocks contain real PCM. Passing a 2-d mask here trips "Attempted to
```
msluszniak, line 84: Question: `num_blocks_scalar_` is assigned `n_valid` (the raw valid **sample** count, up to ~480k) while the computed `k_blocks` is discarded — but the name and comment describe this scalar as a **block** count. Is the exported `audio_encoder` graph's second input the valid-sample length, or should this be `k_blocks`? If it's samples, renaming to `valid_samples_scalar_` avoids the confusion; if it's blocks, this is a bug.

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -84,69 +85,146 @@ int32_t Sampler::sample_topp(T *probabilities, float coin) {
...
-  return probindex[last_idx].index; // in case of rounding errors
+  return probindex[last_idx].index;
 }
 
-Sampler::Sampler(int32_t vocab_size, float temperature, float topp,
-                 unsigned long long rng_seed, float min_p,
-                 float repetition_penalty)
+// Mask logits outside the top-k by rank to -inf. Ties at the k-th boundary
+// are kept (matches HuggingFace TopKLogitsWarper).
+template <typename T> void Sampler::mask_topk(T *logits) {
```
msluszniak, line 126: `mask_topp` is implemented but never called — `sample()` runs `mask_topk` → `softmax` → the legacy `sample_topp` (probability space), so the documented `temperature → top-k → top-p → softmax → multinomial` pipeline doesn't match the actual order. Either wire `mask_topp` in (and drop `sample_topp`) or remove it and fix the header comment. Minor: `size_t` loop counters mixed with signed `int n0`/`last_idx`/`int32_t vocab_size_` will likely trip `-Wsign-compare`, and `last_idx == -1` (when `n0 == 0`) would make `for (size_t i = 0; i <= last_idx; …)` run to `SIZE_MAX` — latent today since `n0 >= 1`, but worth tightening.

packages/react-native-executorch/common/rnexecutorch/models/llm/Types.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <string>
+#include <vector>
```
msluszniak, line 5: `MultimodalInputs` uses `std::optional` (below) but this header only includes `<string>`/`<vector>` — it compiles via transitive includes today but is brittle. Add the include:

```suggestion
#include <optional>
#include <vector>
```

packages/react-native-executorch/src/modules/natural_language_processing/LLMModule.ts
```diff
@@ -59,10 +59,13 @@ export class LLMModule {
...
-        modelSource: namedSources.modelSource,
-        tokenizerSource: namedSources.tokenizerSource,
-        tokenizerConfigSource: namedSources.tokenizerConfigSource,
-        capabilities: namedSources.capabilities,
+        model: {
+          modelName: namedSources.modelName,
+          modelSource: namedSources.modelSource,
+          tokenizerSource: namedSources.tokenizerSource,
+          tokenizerConfigSource: namedSources.tokenizerConfigSource,
+          capabilities: namedSources.capabilities,
```
msluszniak, line 69: **Audio is silently dropped by the imperative API.** `fromModelName`'s `namedSources` type omits `audioConfig` and it isn't forwarded here, so `LLMModule.fromModelName(GEMMA4_E2B_MM)` reaches the controller with `audioConfig === undefined` and the first audio `sendMessage` throws on `this.audioConfig!.samplesPerBlock`. Forward it (and add `audioConfig?: AudioConfig` to the `namedSources` param type above — that line is outside this diff):

```suggestion
          capabilities: namedSources.capabilities,
          audioConfig: namedSources.audioConfig,
```
Also `sendMessage`/`forward` on `LLMModule` are still typed image-only (`media?: { imagePath?: string }`, `imagePaths?: string[]`), so audio can't pass through the imperative API without a cast — worth widening to match the hook.

packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -124,16 +117,20 @@ export class LLMController {
         this.nativeModule.unload();
       }
 
+      if ((model.capabilities ?? []).find((c) => c === 'audio')) {
+        this.audioConfig = model.audioConfig;
+      }
```
msluszniak: **`audioConfig` is never reset on model reload.** The controller instance is reused across `load()` calls; loading an audio model sets `this.audioConfig`, then loading a non-audio model leaves it stale (the guard skips clearing it). Make the assignment unconditional — `undefined` is correct for non-audio models and it's already gated by `capabilities` everywhere it's read:

```suggestion
      this.audioConfig = model.audioConfig;
```
Optional hardening: at load time, if `capabilities` includes `'audio'` but `model.audioConfig` is undefined, throw `InvalidConfig` — that turns the `LLMModule` crash above into a clear, early error.

apps/llm/app/multimodal_llm/index.tsx
```diff
@@ -46,11 +53,19 @@ function MultimodalLLMScreen() {
   const textInputRef = useRef<TextInput>(null);
   const { setGlobalGenerating } = useContext(GeneratingContext);
 
-  // Added error state
+  const [audioBuffer, setAudioBuffer] = useState<Float32Array | null>(null);
+  const [audioLabel, setAudioLabel] = useState<string | null>(null);
+  const [audioUrl, setAudioUrl] = useState('');
+  const [isFetchingAudio, setIsFetchingAudio] = useState(false);
+  const [isRecording, setIsRecording] = useState(false);
+  const [hasMicPermission, setHasMicPermission] = useState(false);
```
msluszniak, line 61: The screen fully unmounts on navigate-away (the `useIsFocused` wrapper renders `null`), but there's no cleanup — an active recording, the audio session, and in-flight generation all leak, and unmount-mid-generation can crash. Add a cleanup effect:

```tsx
useEffect(() => () => {
  recorder.current.stop();
  AudioManager.setAudioSessionActivity(false);
}, []);
```
and interrupt generation on unmount (read `isGenerating`/`interrupt` via a ref to avoid a stale closure).

apps/llm/app/multimodal_llm/index.tsx
```diff
@@ -68,6 +83,87 @@ function MultimodalLLMScreen() {
...
+      if (!ok) {
+        setError('Cannot start audio session');
+        return;
+      }
+      const result = recorder.current.start();
+      if (result.status === 'error') {
+        setError(`Recording problems: ${result.message}`);
+        return;
+      }
+      setIsRecording(true);
```
msluszniak, line 146: Two recording-robustness issues: (1) this `result.status === 'error'` branch is effectively dead — with no file output, `AudioRecorder.start()` always returns `{ status: 'success' }`, so a real native start failure still flips the UI to 'recording' and yields empty audio; register `recorder.current.onError(...)` instead. (2) `onAudioReady` pushes a `Float32Array` per ~0.1s with no cap, so a long recording grows memory unbounded and hands a huge buffer to `sendMessage` despite the model's ~30s window — enforce a max-duration stop (and reject oversized decoded buffers in `loadAudioFromUrl`).

apps/llm/app/multimodal_llm/index.tsx
```diff
@@ -68,6 +83,87 @@ function MultimodalLLMScreen() {
...
+      setAudioLabel(`${name} · ${(pcm.length / 16000).toFixed(1)}s`);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e));
+    } finally {
+      setIsFetchingAudio(false);
+    }
+  };
+
+  const startRecording = async () => {
+    if (!hasMicPermission) {
```
msluszniak, line 123: Mic permission is requested once on mount and this is a dead-end on denial: the button isn't disabled when `!hasMicPermission`, so tapping it only sets 'enable it in Settings' with no way to act, and there's no re-check or `Linking.openSettings()`. Re-request inside `startRecording` (await `requestRecordingPermissions()` and update state), offer `Linking.openSettings()` when `Denied`, and/or disable the button when permission is known-denied.

apps/llm/app/multimodal_llm/index.tsx
```diff
@@ -23,12 +28,14 @@ import Spinner from '../../components/Spinner';
 import { GeneratingContext } from '../../context';
 import SuggestedPrompts from '../../components/SuggestedPrompts';
 import ErrorBanner from '../../components/ErrorBanner';
+import AudioWaveform from '../../components/AudioWaveform';
 
 const SUGGESTED_PROMPTS = [
   "What's in this image?",
   'Describe this scene in detail',
   'What objects can you see?',
   'What text appears in this image?',
```
msluszniak, line 37: Nits: the empty-state copy further down still says "Pick an image and ask me anything about it" though audio is now supported; the new suggested prompt `'Transcribe the audio?'` reads oddly as a question; and the `AudioWaveform` import + `SUGGESTED_PROMPTS` const are interleaved between import statements (legal but worth tidying).

chmjkb (approved):
besides what Mateusz requested, i think everything is fine on my end

apps/llm/app/multimodal_llm/index.tsx
msluszniak: Partially addressed — the cleanup effect now stops the recorder and deactivates the audio session 👍. Still missing: **interrupt in-flight generation on unmount.** Unmounting while `vlm.isGenerating` is true can crash (per the library contract). Add to the same cleanup effect:
```tsx
if (isGeneratingRef.current) vlm.interrupt();
```
reading `isGenerating`/`interrupt` via a ref so the `[]`-effect does not capture a stale closure.

apps/llm/app/multimodal_llm/index.tsx
msluszniak: Partially addressed. Remaining:
- Typo in the new focused-state copy: `"Say hi or pick an image or and ask me anything about it."` → drop the stray words (e.g. `"Say hi, or pick an image, and ask me anything about it."`).
- The non-focused branch still says image-only — `"Pick an image and ask me anything about it."` (L248) — update it to mention audio too.
- The `'Transcribe the audio?'` suggested prompt still reads oddly as a question (e.g. `'Transcribe the audio'`).

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -84,69 +85,146 @@ int32_t Sampler::sample_topp(T *probabilities, float coin) {
...
+  if (topk_ <= 0 || topk_ >= vocab_size_) {
+    return;
+  }
+  // Partial-select the (topk_-th largest) threshold using nth_element on a
+  // copy of logits; O(n) average.
+  std::vector<T> scratch(logits, logits + vocab_size_);
+  std::nth_element(scratch.begin(), scratch.begin() + (topk_ - 1),
+                   scratch.end(), std::greater<T>());
+  const T threshold = scratch[topk_ - 1];
+  const T neg_inf = std::numeric_limits<T>::lowest();
```
msluszniak: If `std::numeric_limits<T>` is constexpr for all Ts, we can change to:
```suggestion
  constexpr T neg_inf = std::numeric_limits<T>::lowest();
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -84,69 +85,146 @@ int32_t Sampler::sample_topp(T *probabilities, float coin) {
...
+    T e = static_cast<T>(std::expf(static_cast<float>(logits[i] - max_val)));
+    probindex[i].prob = e;
+    probindex[i].index = i;
+    sum += e;
+  }
+  if (sum <= T(0)) {
+    return;
+  }
+  for (size_t i = 0; i < vocab_size_; i++) {
+    probindex[i].prob = probindex[i].prob / sum;
```
msluszniak:
```suggestion
    probindex[i].prob /= sum;
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -84,69 +85,146 @@ int32_t Sampler::sample_topp(T *probabilities, float coin) {
...
+      last_idx = i;
+      break;
+    }
+  }
+  // Mark kept indices, then -inf the rest.
+  std::vector<bool> keep(vocab_size_, false);
+  for (size_t i = 0; i <= last_idx; i++) {
+    keep[probindex[i].index] = true;
+  }
+  const T neg_inf = std::numeric_limits<T>::lowest();
```
msluszniak: Similarly, we can probably change to this one.
```suggestion
  constexpr T neg_inf = std::numeric_limits<T>::lowest();
```

msluszniak (approved):
I think all issues that I raised are now addressed. Thank you for working on this one Mateusz. 🚀

PR #1169, opened 2026-05-21
fix(llm): surface tool-call errors via sendMessage

### LINKED ISSUE #1157: Tool-calling errors are swallowed and never surface in `useLLM().error`
### Description

When `useLLM` is configured with `toolsConfig`, any error that happens on the tool-execution path is silently dropped. The error reaches neither `llm.error` nor the `await llm.sendMessage(...)` try/catch, so apps that wire an `ErrorBanner` to those signals show nothing while the JS console reports an unhandled rejection (or a `Logger.error`).

### Where it happens

Two swallowing points in the controller:

1. `parseToolCall` catches every parse failure and returns `[]`:
   https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/src/utils/llm.ts#L42-L45
   ```ts
   } catch (e) {
     Logger.error(e);
     return [];
   }
   ```

2. `LLMController.sendMessage` invokes `executeToolCallback` fire-and-forget — no `.catch`, not awaited:
   https://github.com/software-mansion/react-native-executorch/blob/main/packages/react-native-executorch/src/controllers/LLMController.ts#L427-L441
   ```ts
   if (this.toolsConfig) {
     const toolCalls = parseToolCall(response);
     for (const toolCall of toolCalls) {
       this.toolsConfig
         .executeToolCallback(toolCall)
         .then((toolResponse: string | null) => { ... });
     }
   }
   ```

### Reproduction

1. Open the `llm` demo app → **Tool calling** screen.
2. Switch the picker to a non-tool-trained model (e.g. Llama 3.2 1B QLoRA).
3. Send a tool-shaped prompt (`What events do I have today?`).
4. JS logs an error from `parseToolCall` (or an unhandled rejection if the callback throws), but the on-screen `ErrorBanner` stays empty.

### Expected

Tool-side errors should propagate so apps can display them. Concretely:
- `LLMController.sendMessage` should `await` the tool callbacks under try/catch and forward any failure into the existing error path the hook already exposes as `error`.
- `parseToolCall` parse failures are a softer case — empty result is sometimes correct (model legitimately decided not to call a tool). At minimum, distinguish "no tool call in output" from "malformed tool-call JSON" and surface the latter (e.g. via a typed warning or an `error` channel).

### Notes

- Unrelated to PR #1148 (model registry) — same bug exists on `main`.
- Found while testing the model-registry PR with Llama 3.2 1B QLoRA on the tool-calling screen.

msluszniak:
## Description

`useLLM` with `toolsConfig` silently dropped errors from the tool-execution path: `parseToolCall` swallowed every parse failure with `Logger.error`, and `LLMController.sendMessage` fired `executeToolCallback` without `await` or `.catch`. Neither `llm.error` nor the caller's `await llm.sendMessage(...)` `try/catch` ever saw them.

This PR makes those errors propagate:

- `parseToolCall` now distinguishes "no tool call in output" (returns `[]` — the model legitimately chose not to call a tool) from "malformed tool-call JSON" (throws `RnExecutorchError(InvalidModelOutput)`).
- `LLMController.sendMessage` `await`s `executeToolCallback` inside `try/catch` and rethrows via `parseUnknownError`, so failures reach the caller's `try/catch` (and the hook's `error` state if wired).

- [x] Yes
- [ ] No

Callers of `parseToolCall` that previously received `[]` for malformed tool-call JSON now get a thrown `RnExecutorchError(InvalidModelOutput)`. Same applies to `sendMessage` when `toolsConfig` is set and the tool path throws — the error now surfaces instead of being silently swallowed.

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

1. Open the `llm` demo app → **Tool calling** screen.
2. Switch the picker to a non-tool-trained model (e.g. Llama 3.2 1B QLoRA / SpinQuant).
3. Send a tool-shaped prompt (`What events do I have today?`).
4. If the tool-execution path throws (e.g. "max recursion depth reached" on tool-naive models, malformed tool-call JSON), the on-screen `ErrorBanner` now displays the error instead of staying empty.
5. Sanity check on a tool-trained model (Hammer 2.1 1.5B): tool calls still execute normally and no error banner appears.

### Related issues

Fixes #1157

### Checklist

### DIFF
--- a/packages/react-native-executorch/src/controllers/LLMController.ts
+++ b/packages/react-native-executorch/src/controllers/LLMController.ts
@@ -416,18 +416,20 @@ export class LLMController {
     }
 
     if (this.toolsConfig) {
-      const toolCalls = parseToolCall(response);
-      for (const toolCall of toolCalls) {
-        this.toolsConfig
-          .executeToolCallback(toolCall)
-          .then((toolResponse: string | null) => {
-            if (toolResponse) {
-              this.messageHistoryCallback([
-                ...this._messageHistory,
-                { content: toolResponse, role: 'assistant' },
-              ]);
-            }
-          });
+      try {
+        const toolCalls = parseToolCall(response);
+        for (const toolCall of toolCalls) {
+          const toolResponse =
+            await this.toolsConfig.executeToolCallback(toolCall);
+          if (toolResponse) {
+            this.messageHistoryCallback([
+              ...this._messageHistory,
+              { content: toolResponse, role: 'assistant' },
+            ]);
+          }
+        }
+      } catch (e) {
+        throw parseUnknownError(e);
       }
     }
 
--- a/packages/react-native-executorch/src/controllers/LLMController.ts
+++ b/packages/react-native-executorch/src/controllers/LLMController.ts
@@ -421,7 +421,7 @@ export class LLMController {
         for (const toolCall of toolCalls) {
           const toolResponse =
             await this.toolsConfig.executeToolCallback(toolCall);
-          if (toolResponse) {
+          if (toolResponse != null) {
             this.messageHistoryCallback([
               ...this._messageHistory,
               { content: toolResponse, role: 'assistant' },
--- a/packages/react-native-executorch/src/utils/llm.ts
+++ b/packages/react-native-executorch/src/utils/llm.ts
@@ -4,45 +4,56 @@ import { Schema, Validator } from 'jsonschema';
 import { jsonrepair } from 'jsonrepair';
 import { DEFAULT_STRUCTURED_OUTPUT_PROMPT } from '../constants/llmDefaults';
 import * as zCore from 'zod/v4/core';
-import { Logger } from '../common/Logger';
+import { RnExecutorchError } from '../errors/errorUtils';
+import { RnExecutorchErrorCode } from '../errors/ErrorCodes';
 
 /**
  * Parses tool calls from a given message string.
  * @category Utilities - LLM
  * @param message - The message string containing tool calls in JSON format.
- * @returns An array of `ToolCall` objects extracted from the message.
+ * @returns An array of `ToolCall` objects extracted from the message. Returns
+ *   an empty array if the model did not emit a `[...]` block (i.e. chose not
+ *   to call any tool).
+ * @throws {RnExecutorchError} `InvalidModelOutput` when a `[...]` block is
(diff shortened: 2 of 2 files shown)


packages/react-native-executorch/src/controllers/LLMController.ts
```diff
@@ -416,18 +416,20 @@ export class LLMController {
...
-                { content: toolResponse, role: 'assistant' },
-              ]);
-            }
-          });
+      try {
+        const toolCalls = parseToolCall(response);
+        for (const toolCall of toolCalls) {
+          const toolResponse =
+            await this.toolsConfig.executeToolCallback(toolCall);
+          if (toolResponse) {
```
barhanc: I know this wasn't touched in this PR, but shouldn't this be explicitly `if (toolResponse != null) {`, so that the empty string is still treated as a proper tool response?

packages/react-native-executorch/src/utils/llm.ts
```diff
@@ -4,45 +4,56 @@ import { Schema, Validator } from 'jsonschema';
...
  * Parses tool calls from a given message string.
  * @category Utilities - LLM
  * @param message - The message string containing tool calls in JSON format.
- * @returns An array of `ToolCall` objects extracted from the message.
+ * @returns An array of `ToolCall` objects extracted from the message. Returns
+ *   an empty array if the model did not emit a `[...]` block (i.e. chose not
+ *   to call any tool).
+ * @throws {RnExecutorchError} `InvalidModelOutput` when a `[...]` block is
+ *   present but cannot be parsed as JSON — distinct from the model
+ *   legitimately deciding not to invoke a tool.
```
barhanc, line 19: Does this not make it a breaking change?

barhanc (approved):
Tested and it works fine, just two tiny comments that don't really influence the approve :)

packages/react-native-executorch/src/controllers/LLMController.ts
msluszniak: Good catch — fixed in ddf96fd. The callback can legitimately return an empty string, so we now only skip the history append when it is null.

packages/react-native-executorch/src/utils/llm.ts
msluszniak: You're right — it is a breaking change for callers that relied on the previous swallow-and-log behavior. Updated the PR description to flag it.

PR #1170, opened 2026-05-21
fix(computer-vision): align vision_camera segmentation overlay with preview

### LINKED ISSUE #1158: vision_camera demo: semantic segmentation overlay misaligned in portrait
### Description

In the `computer-vision` demo app's **vision_camera** screen, the semantic-segmentation overlay (FCN ResNet50 in the repro below, but the code path is shared across all variants) lands rotated and displaced relative to the actual subject in the camera preview. The mask shape covers roughly the right area but is offset and warped — e.g. a TV in the center of the frame produces a purple blob shifted up-and-left and stretched.

### Where it happens

`apps/computer-vision/components/vision_camera/tasks/SegmentationTask.tsx:156-165`:

```ts
// Sensor frames are landscape-native, so width/height are swapped
// relative to portrait screen orientation.
const screenW = frame.height;
const screenH = frame.width;
const maskW =
  argmax.length === screenW * screenH
    ? screenW
    : Math.round(Math.sqrt(argmax.length));
const maskH =
  argmax.length === screenW * screenH
    ? screenH
    : Math.round(Math.sqrt(argmax.length));
```

Two issues:

1. **Square-mask fallback.** When `argmax.length !== screenW * screenH`, the code assumes the mask is square (`sqrt(argmax.length)` used for both width and height). For models that output at a non-square spatial resolution — or when the swap on lines 156–157 disagrees with the actual frame orientation — `maskW` and `maskH` get a single sqrt value, the mask is built with the wrong aspect, and Skia's `fit="cover"` then center-crops/stretches it onto the portrait canvas. Net result: rotated/displaced overlay.
2. **Hard-coded landscape→portrait swap.** Lines 156–157 assume `frame.width`/`frame.height` always report landscape-native sensor dimensions. That isn't stable across vision-camera versions and rotation states. When the swap is wrong, branch 1 misses and we fall into the square fallback.

### Reproduction

1. Open `computer-vision` demo → vision_camera screen.
2. Select **Segment → FCN ResNet50** (or any DeepLab variant).
3. Aim the camera at a clear, asymmetric subject (e.g. a TV).
4. The colored overlay appears in the right *area* but rotated/displaced relative to the actual subject.

### Suggested fix

- Get the true mask dimensions from the model output rather than guessing. The hook returns spatial dims for each task (or we can expose them) — use those directly.
- Stop swapping `frame.width`/`frame.height` blindly. Use `frame.orientation` or a vision-camera helper to determine the actual on-screen aspect, then map model coords → display coords explicitly (a transform on the `<SkiaImage>`, not just `fit="cover"`).
- Once dims are correct, the `fit` mode should match what's used for the camera preview itself, so the mask aligns pixel-for-pixel.

### Notes

- Same code on `main` — unrelated to PR #1148 (model registry).
- Found while testing the model-registry PR on Android (Samsung, 1080×2340 portrait).

msluszniak:
## Description

In the `computer-vision` demo's vision_camera screen the semantic-segmentation overlay landed rotated and offset relative to the camera preview (FCN ResNet50 / any DeepLab variant). Two layers fixed:

- `SegmentationTask.tsx`: the `argmax.length === screenW * screenH` branch was unreachable (`runOnFrame` runs with `resizeToInput=false`, so the mask is always at *model output* resolution) and the `<SkiaImage fit="cover">` then stretched it across the whole portrait canvas. Drop the dead branch, infer the (square) mask side from `sqrt(length)`, and draw the SkiaImage into the camera preview's cover-fit rect with `fit="fill"` — same pattern as `OCRTask`/`ObjectDetectionTask`. Mask + preview now line up pixel-for-pixel in portrait.
- `vision_camera/index.tsx`: pin `orientationSource` to `'interface'` while the segmentation task is active. The activity is portrait-locked and the native side rotates the mask into that coord system via `inverseRotateMat`, but `'device'` made the preview rotate with the phone, leaving the mask drifted by 90° in landscape. Other tasks keep `'device'` since their bbox/point coords tolerate the device-rotated preview fine.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. Open the `computer-vision` demo → vision_camera screen.
2. Select **Segment → FCN ResNet50** (or any DeepLab variant).
3. Aim the camera at a clear asymmetric subject (e.g. a TV).
4. The colored overlay should line up with the actual subject in the preview, not rotated/displaced.
5. Tilt the phone to landscape — the mask should keep tracking the same object as in portrait.
6. Switch to another task (Classify / Detect / OCR / …) and back to Segment — preview should keep rotating with the device for those, and re-pin to the activity orientation for Segment.

### Related issues

Fixes #1158.

### Checklist

### Additional notes

iOS not retested with this patch — the C++ rotation already handles iOS correctly and the JS change is platform-agnostic, but a quick sanity check on a portrait iPhone before merge wouldn't hurt.

### DIFF
--- a/apps/computer-vision/app/vision_camera/index.tsx
+++ b/apps/computer-vision/app/vision_camera/index.tsx
@@ -234,7 +234,14 @@ export default function VisionCameraScreen() {
         device={device}
         outputs={frameOutput ? [frameOutput] : []}
         isActive={isFocused}
-        orientationSource="device"
+        // Segmentation draws a 2D mask that the native side rotates into the
+        // activity's coord system (portrait, since the activity is locked).
+        // Pin the preview to that same coord system so mask + preview can't
+        // drift apart when the phone is tilted. Other tasks render coords
+        // (bboxes/points) and tolerate the device-rotated preview fine.
+        orientationSource={
+          activeTask === 'segmentation' ? 'interface' : 'device'
+        }
         onError={(e) => {
           console.warn('[Camera] onError', e);
           setError(e.message);
--- a/apps/computer-vision/components/vision_camera/tasks/SegmentationTask.tsx
+++ b/apps/computer-vision/components/vision_camera/tasks/SegmentationTask.tsx
@@ -153,14 +153,16 @@ export default function SegmentationTask({
           const result = segRof(frame, isFrontCamera, [], false);
           if (result?.ARGMAX) {
             const argmax: Int32Array = result.ARGMAX;
-            // Native rotates the mask into screen-space (see
-            // `inverseRotateMat`). Derive screen-space dims from
-            // `frame.orientation`: portrait orientations ("left"/"right")
-            // swap sensor-native width/height, landscape ones keep them.
-            const orient = frame.orientation;
-            const isScreenPortrait = orient === 'left' || orient === 'right';
-            const screenW = isScreenPortrait ? frame.height : frame.width;
-            const screenH = isScreenPortrait ? frame.width : frame.height;
+            // Both the preview and the mask live in a portrait coord system:
+            // the activity is portrait-locked (so CameraX's PreviewView always
+            // renders the preview in portrait orientation regardless of how
+            // the device is physically tilted), and the native side runs
+            // `inverseRotateMat` which always converts the mask into the same
+            // portrait coord system. Treat sensor-native dims as portrait by
+            // swapping height/width — same convention as the sibling OCR and
+            // ObjectDetection tasks.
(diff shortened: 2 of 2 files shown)


msluszniak:
The original problem is solved, but I spotted, that flip by 90 or 270 deg makes mask off again. TODO: fix the horizontal orientation as well

barhanc (approved):
Tested on both iOS and Android and it does fix the described problem.

PR #1187, opened 2026-05-26
feat: add blazeface support

chmjkb:

- [ ] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/rnexecutorch/models/instance_segmentation/BaseInstanceSegmentation.cpp
+++ b/packages/react-native-executorch/common/rnexecutorch/models/instance_segmentation/BaseInstanceSegmentation.cpp
@@ -57,7 +57,8 @@ TensorPtr BaseInstanceSegmentation::buildInputTensor(const cv::Mat &image) {
 std::vector<types::Instance> BaseInstanceSegmentation::runInference(
     const cv::Mat &image, double confidenceThreshold, double iouThreshold,
     int32_t maxInstances, const std::vector<int32_t> &classIndices,
-    bool returnMaskAtOriginalResolution, const std::string &methodName) {
+    bool returnMaskAtOriginalResolution, const std::string &methodName,
+    bool useWeightedNms) {
 
   std::scoped_lock lock(inference_mutex_);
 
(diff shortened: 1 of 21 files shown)


packages/react-native-executorch/common/rnexecutorch/models/object_detection/ObjectDetection.h
```diff
@@ -14,6 +14,16 @@ namespace models::object_detection {
 using executorch::extension::TensorPtr;
 using executorch::runtime::EValue;
 
+/// Affine transform from model-input pixel coords back to source-image coords:
+/// `x_src = x_model * scaleX + offsetX`. Covers both plain stretch (offsets
+/// zero) and letterbox (offsets carry the centre-pad).
+struct BoxTransform {
```
benITo47, line 20: Maybe move it to utils? This might be used in other model pipes as well.

benITo47 (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/instance_segmentation/BaseInstanceSegmentation.cpp
```diff
@@ -131,11 +135,13 @@ std::vector<types::Instance> BaseInstanceSegmentation::generateFromFrame(
 std::vector<types::Instance> BaseInstanceSegmentation::generateFromPixels(
     JSTensorViewIn tensorView, double confidenceThreshold, double iouThreshold,
     int32_t maxInstances, std::vector<int32_t> classIndices,
-    bool returnMaskAtOriginalResolution, std::string methodName) {
+    bool returnMaskAtOriginalResolution, std::string methodName,
+    bool useWeightedNms) {
 
   cv::Mat image = extractFromPixels(tensorView);
   return runInference(image, confidenceThreshold, iouThreshold, maxInstances,
-                      classIndices, returnMaskAtOriginalResolution, methodName);
```
msluszniak, line 138: We are modifying the code that should be general (BASE instance segmentation) and we bloat it with some positional parameters that are useless for other models. This is super bad architectural design what we have right now. We need to discuss this internally how we should tackle that. Because this is not the problem of this PR only, but almost all models. I created RFC in discussion section how to deal with this one. See https://github.com/software-mansion/react-native-executorch/discussions/1189

msluszniak (changes requested):

packages/react-native-executorch/common/rnexecutorch/models/instance_segmentation/BaseInstanceSegmentation.cpp
chmjkb: I feel like we already agreed internally on the underlying issue, I drafted an internal design doc on this ~2 months ago proposing pretty much the same fix: ordered preprocessor pipeline from JS, (postprocessing can be a bit trickier) new models added via config rather than new C++. We just didn't really spend time to sit down and implement it. 

For this PR: extending BaseInstanceSegmentation was symmetry with ObjectDetection, not because I think the pattern is right. Once we commit to the pipeline design (especially on the native side, as the TS partially does it) useWeightedNms collapses into a postprocessor field anyway.

Is this blocking for this PR? I'm happy to take on the refactor, since we're past the release, but I don't feel like it's the scope of this PR.

msluszniak: It is not a scope of this PR for sure, but yeah, I would consider it a blocking. The refactor should be top priority now. We should discuss this internally how we should proceed with the refactor: either direct PRs to main or PRs to dedicated branch and then one to main. Also imo, we need to have the file with architecture outline in the repository, not in internal resources. We can start by pushing markdown with architectural outline abstracted from design docs.

It is better to start with refactor of instance segmentation and object detection, and test on this model if addition of brand new model goes smoothly rather than merging this and refactor immediately after.

barhanc:
I think we can close this PR in favour of #1280 which added the same functionality to the rewrite branch.

msluszniak:
Closing as suggested in the comment above. If we, for some reason, want to add this PR before v0.10.0, we can re-open this one.

PR #1196, opened 2026-05-28
fix(hooks): allow VAD streamInsert/streamStop during streaming

### LINKED ISSUE #1173: VAD streaming: streamInsert/streamStop blocked by isGenerating gate — useModuleFactory lacks a side-channel primitive
## Summary

`useVAD` exposes `stream`, `streamInsert`, and `streamStop`, but all three go through `runForward`, which rejects calls while `isGenerating === true`. While `stream()` is in flight, `isGenerating` is `true`, so every subsequent `streamInsert(buffer)` rejects with *"model is currently generating"* — i.e. the streaming API cannot actually be used as a stream.

A rapid-tap repro of this behaviour was already discussed during PR #1160 review; the thread was resolved without a code change.

## Repro

```ts
const vad = useVAD({ model });
vad.stream({ /* ... */ });          // starts inference, isGenerating = true
vad.streamInsert(chunk1);            // rejects: "model is currently generating"
vad.streamInsert(chunk2);            // rejects
```

## Root cause

In [`useVAD.ts`](packages/react-native-executorch/src/hooks/natural_language_processing/useVAD.ts):

```ts
const stream       = (input)    => runForward((inst) => inst.stream(input));
const streamInsert = (waveform) => runForward((inst) => { inst.streamInsert(waveform); return Promise.resolve(); });
const streamStop   = ()         => runForward((inst) => { inst.streamStop();           return Promise.resolve(); });
```

`runForward` is the same gate used by `forward` and intentionally serialises full inferences. `streamInsert` is a buffer push, not an inference — it must not share the gate.

## Why this is more general than `useVAD`

`useModuleFactory` exposes a single primitive, `runForward`, which serialises calls via `isGenerating`. That's correct for one-shot inference, but a streaming-capable module needs *two* call paths:

- **Gated** (only one running at a time): `forward`, `stream`.
- **Side-channel** (allowed *during* a running stream): `streamInsert` (buffer push), `streamStop` (interrupt signal).

The other streaming hooks — [`useTextToSpeech`](packages/react-native-executorch/src/hooks/natural_language_processing/useTextToSpeech.ts) and [`useSpeechToText`](packages/react-native-executorch/src/hooks/natural_language_processing/useSpeechToText.ts) — quietly worked around this by *not* using `useModuleFactory` at all and calling the module instance directly for the side-channel methods. `useVAD` is the first hook to migrate streaming methods onto `useModuleFactory` and so is the first to hit the trap. Any future streaming hook will hit it too.

## Fix options

1. **Local fix in `useVAD`** — bypass `runForward` for `streamInsert`/`streamStop` and call `instance.streamInsert/streamStop` directly (matching the existing TTS/STT shape). Smallest diff; doesn't help future hooks.
2. **Extend `useModuleFactory`** — add a non-gating primitive (e.g. `runSideChannel(fn)` that only checks `isReady`, never `isGenerating`). Then streaming hooks express the two regimes explicitly, and TTS/STT can migrate onto `useModuleFactory` cleanly later.

Additionally, `streamStop` arguably should never be gated even in the local-fix path — a caller may want to stop *because* inference is stuck.

## Context

Follow-up to PR #1160 review: https://github.com/software-mansion/react-native-executorch/pull/1160#discussion_r3282441407

msluszniak:
## Description

`useVAD` exposed `stream`, `streamInsert`, and `streamStop` all through `useModuleFactory.runForward`, which rejects calls while `isGenerating === true`. Once `stream()` was in flight, every `streamInsert(chunk)` rejected with *"model is currently generating"* — making the streaming API unusable in practice.

Add a `runSideChannel` primitive to `useModuleFactory` that only checks `isReady` (never `isGenerating` and never flips it). Streaming hooks now have two regimes:

- Gated (`runForward`): `forward`, `stream` — one at a time.
- Side-channel (`runSideChannel`): `streamInsert`, `streamStop` — allowed *during* a running stream.

`useVAD.streamInsert` / `streamStop` now use `runSideChannel`. `useTextToSpeech` and `useSpeechToText` aren't affected (they avoid `useModuleFactory` entirely for their side-channel methods); they're candidates for a follow-up migration onto the new primitive.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

End-to-end on the `apps/speech` Voice Activity Detection screen:

1. Build and install the speech app on an Android device (mic required; the screen disables Start on simulators).
2. Tap **Start VAD Stream** multiple times very quick, speak and pause. With this PR: `SPEAKING` / `SILENT` events appear in the log as expected. Without it (`git checkout HEAD~1 -- packages/react-native-executorch/src/hooks/{useModuleFactory.ts,natural_language_processing/useVAD.ts}` and reload Metro): `streamInsert` rejects on every 100 ms audio chunk and no SPEAKING transition fires and error occurs.

### Related issues

Fixes #1173

### DIFF
--- a/packages/react-native-executorch/src/hooks/natural_language_processing/useVAD.ts
+++ b/packages/react-native-executorch/src/hooks/natural_language_processing/useVAD.ts
@@ -9,14 +9,20 @@ import { useModuleFactory } from '../useModuleFactory';
  * @returns Ready to use VAD model.
  */
 export const useVAD = ({ model, preventLoad = false }: VADProps): VADType => {
-  const { error, isReady, isGenerating, downloadProgress, runForward } =
-    useModuleFactory({
-      factory: (config, onProgress) =>
-        VADModule.fromModelName(config, onProgress),
-      config: model,
-      deps: [model.modelName, model.modelSource],
-      preventLoad,
-    });
+  const {
+    error,
+    isReady,
+    isGenerating,
+    downloadProgress,
+    runForward,
+    runSideChannel,
+  } = useModuleFactory({
+    factory: (config, onProgress) =>
+      VADModule.fromModelName(config, onProgress),
+    config: model,
+    deps: [model.modelName, model.modelSource],
+    preventLoad,
+  });
 
   const forward = (waveform: Float32Array) =>
     runForward((inst) => inst.forward(waveform));
@@ -25,16 +31,9 @@ export const useVAD = ({ model, preventLoad = false }: VADProps): VADType => {
     runForward((inst) => inst.stream(input));
 
   const streamInsert = (waveform: Float32Array) =>
-    runForward((inst) => {
-      inst.streamInsert(waveform);
-      return Promise.resolve();
-    });
+    runSideChannel((inst) => inst.streamInsert(waveform));
 
-  const streamStop = () =>
-    runForward((inst) => {
-      inst.streamStop();
-      return Promise.resolve();
-    });
+  const streamStop = () => runSideChannel((inst) => inst.streamStop());
 
   return {
     error,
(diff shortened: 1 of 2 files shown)


benITo47 (approved):
Tested on iOS - fix works  ✅

PR #1197, opened 2026-05-28
fix(tts): add streamFlush() API and drain Kokoro buffer on streamStop

### LINKED ISSUE #1153: Kokoro::stream hangs on non-EOS-terminated buffer; streamStop(false) never returns
## Summary

`Kokoro::stream` hangs the streaming worker when the input buffer holds content that doesn't end in an end-of-sentence character. The buffer never drains, `streamStop(false)` waits forever, and only `streamStop(true)` recovers.

## Repro

Build the speech app from branch [`@ms/tts-stress-tests`](https://github.com/software-mansion/react-native-executorch/tree/@ms/tts-stress-tests) (preset chips reproduce this in one tap; see `no-term:a` / `no-term:long`). Or directly:

```ts
model.streamInsert('a');
const p = model.stream({ text: '', ... });
await new Promise((r) => setTimeout(r, 6000));
model.streamStop(false);  // never returns
await p;
```

Same hang with `'hello world'`, 2000× U+200D, or any content that ends without `.?!;`.

## Root cause

[`Kokoro.cpp:171-189`](https://github.com/software-mansion/react-native-executorch/blob/%40is/multilingual-tts/packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp#L171-L189):

```cpp
size_t chunkSize = (eosIt != inputTextBuffer_.rend())
                       ? std::distance(eosIt, inputTextBuffer_.rend())
                       : 0;

if (chunkSize > 0 ||
    streamSkippedIterations >= params::kStreamMaxSkippedIterations) {
  input = inputTextBuffer_.substr(0, chunkSize);   // chunkSize still 0
  inputTextBuffer_.erase(0, chunkSize);            // erases nothing
  streamSkippedIterations = 0;                     // reset, loop forever
}
```

When no EOS exists in the buffer, `chunkSize = 0`. The force-flush threshold (`streamSkippedIterations >= kStreamMaxSkippedIterations`) fires correctly, but the extraction still uses `chunkSize` and pulls zero characters. Counter resets, loop continues, buffer never drains.

## Why the obvious fix is unsafe

Switching the force-flush branch to use the searchable window length:

```cpp
} else if (streamSkippedIterations >= params::kStreamMaxSkippedIterations) {
  input = inputTextBuffer_.substr(0, searchLimit);
  inputTextBuffer_.erase(0, searchLimit);
  streamSkippedIterations = 0;
}
```

…breaks the **LLM streaming** mode. `kStreamMaxSkippedIterations` is wall-clock-divided-by-`kStreamPause`, so on slow LLM token rates it trips before the LLM has even produced a sentence, forcing mid-word flushes and degrading speech quality. Tuning the threshold to satisfy both regimes is brittle — context noted in [#1134 comment](https://github.com/software-mansion/react-native-executorch/pull/1134#issuecomment-4491389865).

## Why this isn't just a theoretical concern

The current JS hook (`useTextToSpeech.ts:108-111`) hides this by auto-appending `.` when calling `stream({ text })`. Two places that bypass the rescue:

- Callers using `streamInsert` directly (incremental LLM tokens, dictation, partial captions).
- A streaming caller whose final chunk doesn't end with EOS (LLM truncated, network error, user pressed stop mid-sentence). The trailing un-terminated suffix sits in the buffer permanently.

In both cases `streamStop(false)` blocks forever with no diagnostic.

## Options that don't break LLM streaming

Each is a smaller change than re-tuning the skip counter:

1. **Caller-declared flush intent.** `streamInsert(text, { canFlush?: boolean })` — LLM mode passes `false`, normal apps pass `true`. Caller knows its own pacing.
2. **Explicit `streamFlush()` API.** Caller signals "I'm done feeding for now, partition what's left." LLM mode never calls it; normal apps call it before `streamStop(false)`. No threshold tuning.
3. **Wall-clock idle timeout.** Track time since last `streamInsert`. If the buffer has content and N seconds passed without new inserts, flush. LLM streams keep inserting fast enough to never trip it.

## Mitigations if a proper fix is out of scope

- **Document the hazard** in `streamInsert` / `streamStop` JSDoc — currently nothing in the API surface hints that `streamStop(false)` can hang indefinitely.
- **JS-side safety net** in the hook's `stream()` wrapper: time `streamStop(false)`, fall back to `streamStop(true)` with `console.warn` after some threshold. Zero cost to LLM streaming; prevents downstream apps from soft-locking.

## References

- PR #1134 review comment with full stress-test findings: https://github.com/software-mansion/react-native-executorch/pull/1134#issuecomment-4491389865
- Stress-test app branch: https://github.com/software-mansion/react-native-executorch/tree/@ms/tts-stress-tests

msluszniak:
## Description

`Kokoro::stream` hung when the buffer held content without an end-of-sentence character: the timer-based force-flush (after `kStreamMaxSkippedIterations` idle iterations) tried to extract 0 characters, so the loop reset its skip counter without draining anything and `streamStop(false)` never returned. The same threshold also fired mid-token for LLM-style streaming, partitioning sentences before they finished.

Replace the threshold with caller-driven flushing:

- New `streamFlush()` API. Caller signals "drain what's currently buffered, EOS or not" — typically right before `streamStop(false)` for normal apps that ended on un-terminated content.
- `streamStop(false)` now drains automatically (equivalent to `streamFlush()` plus auto-stop on empty buffer), so it's guaranteed to return even if the trailing tail has no EOS.
- LLM-style callers feeding partial tokens never call `streamFlush()` — model punctuation drives natural EOS partitioning, and the residual tail is drained by `streamStop(false)` when generation completes.

The skipped-iteration force-flush and its tuning constant `kStreamMaxSkippedIterations` are removed.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. Build & run `apps/speech` on Android.
2. Open `Text to Speech - LLM Streaming`.
3. Tap Generate and let the story play through.
4. Tap Stop mid-stream — verify it returns promptly without hanging.

### Related issues

Fixes #1153

### DIFF
--- a/apps/speech/screens/TextToSpeechLLMScreen.tsx
+++ b/apps/speech/screens/TextToSpeechLLMScreen.tsx
@@ -140,6 +140,9 @@ export const TextToSpeechLLMScreen = ({ onBack }: TextToSpeechLLMProps) => {
     } catch (e) {
       console.error('Generation failed:', e);
     } finally {
+      // LLM finished — partition any trailing un-terminated tail so it gets
+      // synthesized before the stream closes.
+      tts.streamFlush();
       tts.streamStop(false);
       await ttsPromise;
 
--- a/docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
+++ b/docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
@@ -85,7 +85,7 @@ The module provides two ways to generate speech using either raw text or pre-gen
 
 1.  [**`forward({ text, speed, phonemize })`**](../../06-api-reference/interfaces/TextToSpeechType.md#forward): Generates the complete audio waveform at once. Returns a promise resolving to a `Float32Array`.
 2.  [**`stream({ speed, phonemize, stopAutomatically, onNext, ... })`**](../../06-api-reference/interfaces/TextToSpeechType.md#stream): An async generator-like functionality (managed via callbacks like `onNext`) that yields chunks of audio as they are computed.
-    This is ideal for reducing the "time to first audio" for long sentences. You can also dynamically insert text during the generation process using `streamInsert(text)` and stop it with `streamStop(instant)`.
+    This is ideal for reducing the "time to first audio" for long sentences. You can also dynamically insert text during the generation process using `streamInsert(text)`, force-partition trailing content without an end-of-sentence character via `streamFlush()`, and stop the stream with `streamStop(instant)`.
 
 :::tip Recommendation
 In most cases, the **`stream()`** method is recommended over `forward()`. It significantly reduces latency by allowing audio playback to begin as soon as the first chunk is synthesized, rather than waiting for the entire text to be processed.
--- a/docs/docs/04-typescript-api/01-natural-language-processing/TextToSpeechModule.md
+++ b/docs/docs/04-typescript-api/01-natural-language-processing/TextToSpeechModule.md
@@ -49,7 +49,7 @@ The module provides a way to generate speech using either raw text or pre-genera
 
 1.  [**`forward(text, speed, phonemize)`**](../../06-api-reference/classes/TextToSpeechModule.md#forward): Generates the complete audio waveform at once. Returns a promise resolving to a `Float32Array`.
     - `phonemize` defaults to `true`. When set to `false`, the input is expected to be a string of IPA phonemes.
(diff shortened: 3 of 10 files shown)


msluszniak:
@IgorSwat 
Now I'm in favour of changing direction to:
>Explicit streamFlush() API. Caller signals "I'm done feeding for now, partition what's left." LLM mode never calls it; normal apps call it before streamStop(false). No threshold tuning.

because this will not introduce breaking change as the current state.

IgorSwat (approved):

msluszniak (commented):
I need to add documentation for this function for flushing.

PR #1203, opened 2026-05-29
test: add TS API contract test suite scaffold

msluszniak:
## Description

Adds a Jest-based scaffold and twelve API contract test suites under [packages/react-native-executorch/__tests__/api/](packages/react-native-executorch/__tests__/api/), with `typecheck:tests` + `test` scripts wired into a new CI job. Catches drift across the public TypeScript surface (modules, hooks, types, registry, error codes, voices, URLs) without exercising native code.

### Introduces a breaking change?

- [ ] Yes
- [x] No

### Type of change

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Documentation update (improves or adds clarity to existing documentation)
- [x] Other (chores, tests, code style improvements etc.)

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

```sh
yarn workspace react-native-executorch typecheck:tests
yarn workspace react-native-executorch test
```

Expected: 12 test suites, 971 passed, 9 skipped (documented in #1202), 1 snapshot. The new `test` job in `.github/workflows/ci.yml` runs the same on PRs.

### Screenshots

N/A — no UI changes.

### Related issues

Closes #1018. Findings catalogued in #1202.

### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the documentation accordingly
- [x] My changes generate no new warnings

### Additional notes

Tests do not touch native code: \`global.loadXxx\` and \`ResourceFetcher\` are stubbed at the JS layer. Module classes are discovered via \`Object.entries(RNE)\` and the hook return contract is enforced via \`satisfies\` (tsc-checked).

Inconsistencies the tests surface are encoded as explicit exception sets (\`SKIPS_BASE_MODULE\`, \`SKIPS_STATIC_FACTORY\`, \`SKIPS_DELETE\`, \`HOOK_NAME_EXCEPTIONS\`, \`PARAMETERIZED_ACCESSORS\`) so the suite stays green on today's codebase. Each exception entry maps to a bullet in #1202 and can be removed as the underlying API is normalized.

PR #1204, opened 2026-05-29
test+refactor!: API contract test scaffold + fixes for #1202

### LINKED ISSUE #1018: Add API test in TypeScript
Currently, there are only sanity tests in cpp code. We also need some TS testing.

### LINKED ISSUE #1202: API consistency findings surfaced by TS contract tests
The new API contract tests added in #1018 (branch [`@ms/api-tests-scaffold`](https://github.com/software-mansion/react-native-executorch/tree/@ms/api-tests-scaffold)) surfaced several inconsistencies across the public TypeScript surface. They are currently encoded as explicit exception sets in the test files so the suite passes today; this issue tracks the cleanup work.

## Modules that don't extend `BaseModule`

`BaseModule` provides the shared `delete()` / `forwardET()` / `getInputShape()` plumbing and the `generateFromFrame` JSI binding hook. The following modules skip it and re-implement (or omit) parts of that contract:

- `OCRModule`
- `VerticalOCRModule`
- `LLMModule`
- `SpeechToTextModule`
- `TextToSpeechModule`
- `TokenizerModule`

Listed in `SKIPS_BASE_MODULE` in `__tests__/api/moduleContracts.test.ts`.

## Modules without a static `from*` factory

Every other module uses `Module.fromModelName(...)` / `Module.fromCustomModel(...)` as the construction entry point:

- `ExecutorchModule` — constructs blank and exposes an instance `load()` method instead.
- `TokenizerModule` — no factory at all.

Listed in `SKIPS_STATIC_FACTORY`.

## Hook naming inconsistency

Every hook drops the `Module` suffix from its module class name (`ClassificationModule` → `useClassification`, `LLMModule` → `useLLM`) **except**:

- `ExecutorchModule` → `useExecutorchModule` (keeps the suffix).

Listed in `HOOK_NAME_EXCEPTIONS`.

## Informational: parameterized registry accessor

`models.ocr.craft({ language })` requires a `language` argument, while every other accessor in the registry is either fully parameterless or takes only optional `ModelOpts`. This is intentional (OCR pipelines are picked per language), but worth flagging because the test walker has to special-case it. Listed in `PARAMETERIZED_ACCESSORS` in `__tests__/api/modelRegistry.test.ts`.

## How to act on this

Each bullet above corresponds to an entry in one of the exception sets. As an inconsistency is fixed at the source, the matching entry can be deleted and the contract test will start enforcing the rule for that module.

Refs #1018.

msluszniak:
## Description

Bundles two related changes so the API surface and its contract tests land together:

1. **API contract test scaffold** (under `packages/react-native-executorch/__tests__/api/`) — 12 Jest suites + a CI `test` job. Catches drift across the public TypeScript surface (modules, hooks, types, registry, error codes, voices, URLs) without exercising native code.
2. **Source-level fixes for every actionable finding in #1202** so the contract suite enforces every rule unconditionally — no exception sets, 0 skipped tests.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

```sh
yarn workspace react-native-executorch typecheck:tests
yarn workspace react-native-executorch test
yarn typecheck
yarn lint
```

Expected: 12 suites green, 980 tests passed, 0 skipped, 1 snapshot. The two demo apps that exercise the renamed APIs (`apps/speech`, `apps/computer-vision/text_to_image`) build and run with unchanged behaviour.

### Screenshots

N/A — no UI changes.

### Related issues

Closes #1018 (TS API tests). Closes #1202 (findings the tests surfaced). Supersedes #1203.

### Checklist

### Additional notes

**Breaking API changes:**

- `ExecutorchModule` and `TokenizerModule` now construct via static factories (`fromModelSource`, `fromModelName`); the instance `load()` is removed. The `useModule` hook helper is deleted alongside them.

**Renames kept as deprecated aliases** (callers don't have to migrate yet, but `@deprecated` JSDoc flags new usage):

- `useExecutorch` is the new canonical name; `useExecutorchModule` is re-exported as a deprecated alias.
- `useTextToSpeech({ model, preventLoad })` is the canonical single-arg form; the previous `useTextToSpeech(model, { preventLoad })` two-arg signature is preserved as a deprecated overload.
- `useTextToImage().forward` is the canonical hook return field (matches `TextToImageModule.forward`); `.generate` is preserved as a deprecated alias.

**Non-breaking fixes:**

- `OCRModule`, `VerticalOCRModule`, `LLMModule`, `SpeechToTextModule`, `TextToSpeechModule`, `TokenizerModule` now extend `BaseModule`. `TokenizerModule` consequently gains a working `delete()` (was leaking the native tokenizer handle).

**Docs + skill files** updated for the renames (hand-written only; `docs/docs/06-api-reference/**` regenerates from JSDoc on the next typedoc build).

### Test suite layout

| Test file | Layer |
| --- | --- |
| moduleContracts | Module → BaseModule, `from*` factory, `useXxx` hook export |
| modelRegistry | Accessor return shape, unique modelName per category |
| hookContracts | `useXxx` return type ⊇ `{error, isReady, isGenerating, downloadProgress}` (compile-time) |
| modelUrls | Every URL is `https://` on the software-mansion HF org |
| errorCodes | Unique codes, working reverse lookup, non-empty default messages |
| ttsVoices | Voice variable-name region ↔ `phonemizerConfig.lang` ↔ phonemizer URL paths |
| apiSurface | Snapshot of public exports |
| hookPropsContract | Every `*Props` has `preventLoad?: boolean`; every hook takes a single object arg (compile-time) |
| registryHookCompatibility | Registry sample per category assignable to the hook's `model` prop (compile-time) |
| modulePrototype | Every module exposes a public method; `delete()` reachable; `BaseModule` surface snapshot |
| moduleConstruction | Mocks `ResourceFetcher`, constructs each module via its `from*` factory, asserts instance + `delete()` (runtime) |
| moduleHookSignatureAlignment | Module prototype method signatures align with their hook return fields (compile-time) |


packages/react-native-executorch/__tests__/api/hookContracts.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import type {
+  useExecutorch,
+  useLLM,
+  usePrivacyFilter,
+  useSpeechToText,
+  useTextEmbeddings,
+  useTextToSpeech,
+  useTokenizer,
+  useVAD,
+} from '../../src';
```
msluszniak, line 22: Why these are splitted?

packages/react-native-executorch/__tests__/api/moduleHookSignatureAlignment.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    returns: true as EqualReturn<LLMModule['generate'], LLMType['generate']>,
+  },
+  styleTransfer_forward: {
+    inputs: true as EqualParam<
+      StyleTransferModule['forward'],
+      StyleTransferType['forward']
+    >,
+    returns: true as EqualReturn<
+      StyleTransferModule['forward'],
+      StyleTransferType['forward']
```
msluszniak, line 121: llm has generate but the rest has forward, why? We need to think if we want to keep it this way.

EDIT:
Ok, the problem with LLM is that LLMModule exposes both generate and forward and generate from useLLM utilizes LLMModule's generate. I'm ok with both current and changed version, just want to make sure everybody is ok with the chosen version.

msluszniak (commented):
Also why CI workflow wasn't triggered by this PR?

msluszniak:
Ok, tested and all functionalities that are changed worked: demo apps + custom app that covered `useExectorch` and `useTokenizer`.

packages/react-native-executorch/src/hooks/general/useExecutorch.ts
```diff
@@ -0,0 +1,32 @@
...
+import { TensorPtr } from '../../types/common';
+import { useModuleFactory } from '../useModuleFactory';
+
+/**
+ * React hook for managing an arbitrary Executorch module instance.
+ * @category Hooks
+ * @param props - Configuration object containing `modelSource` and optional `preventLoad` flag.
+ * @returns Ready to use Executorch module.
+ */
+export const useExecutorch = ({
```
benITo47, line 15: This rename is a breaking change. 
Maybe we can export an alias for the patch releases. 
"export const useExecutorchModule = useExecutorch"
We could get rid of it when 1.0 rolls out

msluszniak: We can do that, but please mind the fact that this is not the only breaking change in this PR.

benITo47: I know, I'm still reading the changes though.
I feel like the hook and generate->forward renames have the most impact for the end user. We should handle it gracefully if possible

msluszniak: I'm in favour of making aliases for `generate` and `useExecutorchModule` and `useTextToSpeech` with different signature.

msluszniak: >llm has generate but the rest has forward, why? We need to think if we want to keep it this way.

This way we can also rename this function in llm from generate to forward if we want, similarly to textToImage.

benITo47: Yeah, two birds with one stone kind of situation. Let's go with the aliases til 1.0

msluszniak: Ok, the problem with LLM is that LLMModule exposes both `generate` and `forward`, and `generate` from useLLM utilizes LLMModule's `generate`. I'm okay with both current and changed version, just want to make sure everybody is ok with the chosen version.

packages/react-native-executorch/__tests__/api/moduleConstruction.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  ObjectDetectionModule,
+  PoseEstimationModule,
+  PrivacyFilterModule,
+  ResourceFetcher,
+  SemanticSegmentationModule,
+  StyleTransferModule,
+  TextEmbeddingsModule,
+  TextToImageModule,
+  VADModule,
+} from '../../src';
```
benITo47, line 22: Why are OCRModule, VerticalOCRModule, SpeechToTextModule, TextToSpeechModule, and TokenizerModule constructions omitted here - is this deliberate or simple overlook?

msluszniak: My fault, pushed fix.

benITo47 (approved):
I have no remarks other than the already fixed ones. 
I'm approving but someone should still take a second look

msluszniak:
We'll add completely different setup for API testing on `rne-rewrite`

PR #1215, opened 2026-06-01
fix(ci): gate node_modules diagnostic on existing step id

### LINKED ISSUE #1210: [CI] npm-publish.yml: "Show packed node_modules" gated on a non-existent step id
## Problem description

In `.github/workflows/npm-publish.yml`, the guard that checks whether `node_modules` leaked into the published tarball uses:

```yaml
- name: Check if any node_modules were packed
  id: check_node_modules
  run: ! grep --silent -E "node_modules/.+" build.log
```

…but the diagnostic step is gated on a step id that does not exist:

```yaml
- name: Show packed node_modules
  if: failure() && steps.node_modules.outcome == 'failure'   # <-- should be steps.check_node_modules
  run: ! grep -E "node_modules/.+" build.log
```

There is no step with `id: node_modules`, so `steps.node_modules.outcome` is always undefined and the condition can never be true. The diagnostic never runs. If a regression ever bundles `node_modules` into the npm package, the publish job fails (correctly, via `check_node_modules`) but the log that's supposed to show *which* paths leaked is silently skipped.

(The diagnostic's own `run` is otherwise fine: `! grep -E ...` prints the matching lines, then inverts the exit code — so it would surface the offending paths if it were reached.)

## Proposed solution

Fix the step id reference:

```yaml
- name: Show packed node_modules
  if: failure() && steps.check_node_modules.outcome == 'failure'
  run: grep -E "node_modules/.+" build.log
```

## Alternative solutions

Drop the diagnostic step entirely and rely on `Show build log`, which already prints the full `build.log` on build failure — though it's gated on `steps.build.outcome`, not the node_modules check, so it wouldn't fire for this specific failure mode either. Fixing the id is the smaller, correct change.

## Benefits to React Native ExecuTorch

Restores a working safety diagnostic in the daily publish pipeline, so an accidental `node_modules` inclusion is debuggable straight from CI logs instead of requiring a local repro.

## Additional context

`.github/workflows/npm-publish.yml` (the three consecutive `Check…` / `Show build log` / `Show packed node_modules` steps in the `build` job).

msluszniak:
## Description

The "Show packed node_modules" diagnostic in `.github/workflows/npm-publish.yml` was gated on `steps.node_modules.outcome == 'failure'`, but no step has `id: node_modules` — the check step's id is `check_node_modules`. The condition was therefore always undefined and the diagnostic never ran. On a `node_modules` leak the publish job failed (correctly, via `check_node_modules`) but the log showing *which* paths leaked was silently skipped.

This fixes the id reference and drops the exit-code inversion (`!`) so the step prints the matching paths and reports success.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

CI-only change. Verify the workflow still parses and that the `Show packed node_modules` step references `steps.check_node_modules.outcome`.

### Related issues

Closes #1210

### Checklist

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -90,9 +90,8 @@ jobs:
 
       - name: Show packed node_modules
         working-directory: ${{ env.EXECUTORCH_DIR }}
-        if: failure() && steps.node_modules.outcome == 'failure'
-        run: >-
-          ! grep -E "node_modules/.+" build.log
+        if: failure() && steps.check_node_modules.outcome == 'failure'
+        run: grep -E "node_modules/.+" build.log
 
       - name: Add package name to env
         working-directory: ${{ env.EXECUTORCH_DIR }}


msluszniak:
I'm closing this one since we add different CI setup on `rne-rewrite` branch

Issue #1213, opened 2026-06-01
Tracking: remove @deprecated symbols at 1.0.0


msluszniak:
## Problem description

This issue is the tracking point for `@deprecated` public API symbols that are still shipped and must be removed when breaking changes are batched at the 1.0.0 release. Without a single tracker, deprecated surface tends to accumulate and undermines the "1.0.0 = API stability" guarantee if it's still carrying aliases.

Symbols to remove at 1.0.0 (each with its replacement):

- [ ] `src/types/llm.ts:353` — `topp` alias → use `topP`.
- [ ] `src/constants/modelUrls.ts` (~lines 30, 64) — non-SpinQuant Llama constants (`LLAMA3_2_3B`, `LLAMA3_2_1B`) → use the SpinQuant variants.
- [ ] `src/constants/modelUrls.ts` (~lines 561, 568) — non-quantized LFM2.5-VL constants → use the `*_QUANTIZED` variants.
- [ ] `src/constants/modelUrls.ts:1204` — `distiluse_base_multilingual_cased_v2` → use the `models.text_embedding.*()` accessor.

(List captured from `grep -rn "@deprecated" src` against `main` @ v0.8.4. Re-run before the 1.0.0 cut to catch any added since, and check the box as each symbol is removed.)

Additionally:
- [ ] Remove all unused repositories from Hugging Face.

## Proposed solution

Attach a `1.0.0` milestone to this issue (and/or a `breaking-change` label to each PR that removes a symbol) so the entries above are removed together as part of the 1.0.0 API-stability commitment rather than lingering indefinitely. Optionally add a lightweight check that flags newly added `@deprecated` markers without a target-removal note, so the list stays current.

## Alternative solutions

Leave deprecations in place and rely on ad-hoc cleanup. Rejected: deprecated surface tends to accumulate and undermines the "1.0.0 = API stability" guarantee if it's still carrying aliases.

## Benefits to React Native ExecuTorch

Keeps the 1.0.0 public surface clean and the stability commitment credible; gives consumers a single reference for what's going away and what to migrate to.

PR #1222, opened 2026-06-08
chore(deps)!: use non-legacy Expo FS in Resource Fetcher, bump SDK version to 56 in demo apps

### LINKED ISSUE #885: Rewrite `expo-resource-fetcher` to the new API
We need to rewrite our expo resource fetcher from legacy file system implementation to the newer one.

### LINKED ISSUE #1017: Resource fetcher - aggregated issue
This issue aggregates all potential improvements for resource fetcher.

- [ ] #885
- [x] #815
- [x] #807

chmjkb:
## Description

This PR bumps Expo SDK to 56 across all demo apps using Expo. It also refactors the Expo Resource Fetcher to use the aforementioned version. For users using <56, a /legacy import for ExpoResourceFetcher is still available.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

 - In an app on Expo SDK <56 (e.g. SDK 54/55), install this package and import { ExpoResourceFetcher } from 'react-native-executorch-expo-resource-fetcher'. Confirm it throws RnExecutorchError with code ResourceFetcherFileSystemApiUnavailable (0xbc) and a message pointing at the /legacy entry point — not the cryptic File.createDownloadTask is not a function.
  - In the same old-SDK app, switch the import to react-native-executorch-expo-resource-fetcher/legacy and confirm fetch, pause/resume/cancel, deleteResources, and listDownloadedFiles all still work.
  - In an app on Expo SDK 56+, import from the default entry point and confirm the guard does not fire and downloads work end-to-end.
  - Run yarn install at the repo root and confirm no new peer-dep warnings on the SDK-56 demo apps under apps/.

### Checklist


msluszniak:
I'll check why CI has diffs on comments in *.h files. This is definitely related to my addition to CI which suppose to check if error codes are regenerated.

chmjkb:
> I'll check why CI has diffs on comments in *.h files. This is definitely related to my addition to CI which suppose to check if error codes are regenerated.

im pretty sure pre-commit does a lint, and yarn codegen:errors returns something that is not yet linted properly:)
so changed on this branch are linted, while the ones generated by the codegen are not, hence the diff. not sure tho

msluszniak:
~Yeah, makes sense.~ That's not the reason. This branch has stale main tip and it caused the CI failure.

msluszniak (changes requested):
Actually the reason why CI fails is because you are not rebased on main, please rebase it first ;)

msluszniak:
**Can we drop the `react-native-executorch` devDependency bump?**

In `packages/react-native-executorch/package.json` this PR bumps the core package's dev baseline to RN 0.85:

```diff
- "@types/react": "~19.1.10",
- "react": "19.1.0",
- "react-native": "0.81.5",
+ "@types/react": "~19.2.0",
+ "react": "19.2.3",
+ "react-native": "0.85.3",
```

I don't think the core package should move off 0.81, for a few reasons:

- **It's not needed for this PR.** The only RNE-core change here is the new `ResourceFetcherFileSystemApiUnavailable` enum (8 lines across `ErrorCodes.ts` + `ErrorCodes.h`) — no RN 0.85 APIs. All the actual `expo-file-system` work lives in `packages/expo-resource-fetcher`, which already carries its own `react-native@0.85.3` / `expo@^56` devDeps. That's the right place for it, since `File.createDownloadTask` is SDK-56-only.

- **`peerDependencies` are `react-native: "*"`,** so there's no install-time floor on consumers. The dev RN version is effectively the *only* thing pinning what we type-check and test core against — bumping it to 0.85 silently drops our 0.81 coverage with nothing to catch a regression.

- **It works against this PR's own goal.** `expo-resource-fetcher/legacy` exists for Expo SDK <56 users, who are on RN 0.81/0.83 — never 0.85. If core RNE is only developed against 0.85, the exact users `/legacy` is for are running a core that's no longer tested on their RN. Same for `bare-resource-fetcher` (untouched here, still `peer react-native: "*"` / `dev 0.81.5`) and `apps/bare-rn` (pinned `0.81.5`).

Could we revert just those three devDep lines and keep the SDK-56 / RN-0.85 bumps scoped to the demo apps and `expo-resource-fetcher`? The core package staying on 0.81 is what makes the legacy + bare lanes meaningful.

chmjkb:
> **Can we drop the `react-native-executorch` devDependency bump?**
> 
> In `packages/react-native-executorch/package.json` this PR bumps the core package's dev baseline to RN 0.85:
> 
> ```diff
> - "@types/react": "~19.1.10",
> - "react": "19.1.0",
> - "react-native": "0.81.5",
> + "@types/react": "~19.2.0",
> + "react": "19.2.3",
> + "react-native": "0.85.3",
> ```
> 
> I don't think the core package should move off 0.81, for a few reasons:
> 
> * **It's not needed for this PR.** The only RNE-core change here is the new `ResourceFetcherFileSystemApiUnavailable` enum (8 lines across `ErrorCodes.ts` + `ErrorCodes.h`) — no RN 0.85 APIs. All the actual `expo-file-system` work lives in `packages/expo-resource-fetcher`, which already carries its own `react-native@0.85.3` / `expo@^56` devDeps. That's the right place for it, since `File.createDownloadTask` is SDK-56-only.
> * **`peerDependencies` are `react-native: "*"`,** so there's no install-time floor on consumers. The dev RN version is effectively the _only_ thing pinning what we type-check and test core against — bumping it to 0.85 silently drops our 0.81 coverage with nothing to catch a regression.
> * **It works against this PR's own goal.** `expo-resource-fetcher/legacy` exists for Expo SDK <56 users, who are on RN 0.81/0.83 — never 0.85. If core RNE is only developed against 0.85, the exact users `/legacy` is for are running a core that's no longer tested on their RN. Same for `bare-resource-fetcher` (untouched here, still `peer react-native: "*"` / `dev 0.81.5`) and `apps/bare-rn` (pinned `0.81.5`).
> 
> Could we revert just those three devDep lines and keep the SDK-56 / RN-0.85 bumps scoped to the demo apps and `expo-resource-fetcher`? The core package staying on 0.81 is what makes the legacy + bare lanes meaningful.

While I agree that this PR maybe is not the best place to bump it, since it involves no internal changes I think opening up another PR is frankly a waste of time. 

Also, I've seen Reanimated and RNGH bump to latest RN versions, so I think it might make sense to bump it. I guess it would make sense to use the latest version to keep up with eventual breaking changes as they come.

msluszniak:
Ok, as these are devDeps, we can indeed follow the latest version. Let's keep it.

msluszniak (commented):
A few code-level notes from a pass over the resource-fetcher refactor (separate from the devDep comment above). All non-blocking except possibly the first, which is a genuine question about the new download API's error behavior.

packages/expo-resource-fetcher/src/handlers.ts
```diff
@@ -153,32 +135,22 @@ export async function handleRemote(
...
-        downloads.delete(source);
-        reject(
-          new RnExecutorchError(
-            RnExecutorchErrorCode.ResourceFetcherDownloadFailed,
-            `Failed to fetch resource from '${uri}', status: ${result?.status}`
-          )
+    .then(async (downloadedFile) => {
+      // null means the task was paused before completion — resume() will continue it.
+      // missing handle means the task was canceled — cancel() already rejected.
+      if (!downloadedFile) {
```
msluszniak: Question: does the new `downloadAsync()` reject on non-2xx responses (404/500), or can it resolve with a `File` containing the error body? The old implementation explicitly guarded `result.status` against `OK`/`PARTIAL_CONTENT`; here the only non-success case handled is `null` (pause). If a 404/500 body can resolve successfully, it gets `move`d to `fileUri` and cached as a "valid" model — and because `handleRemote` short-circuits on `new File(fileUri).exists`, every later fetch then returns the corrupt file (`wasDownloaded: false`) until a manual delete. If the new API doesn't reject on non-2xx out of the box, it'd be worth restoring an explicit status check (same applies to `resume()` in ResourceFetcher.ts).

packages/expo-resource-fetcher/src/handlers.ts
```diff
@@ -153,32 +135,22 @@ export async function handleRemote(
...
+        throw new RnExecutorchError(
```
msluszniak: Minor readability: routing the paused case (`downloadAsync()` resolving `null`) through `throw DownloadInterrupted` just to swallow it in `.catch` when `status === PAUSED` is a little hard to follow. An explicit early `return` here when paused (mirroring how `resume()` handles it in ResourceFetcher.ts) would avoid constructing a throwaway error.

packages/expo-resource-fetcher/src/handlers.ts
```diff
@@ -153,32 +135,22 @@ export async function handleRemote(
...
       }
+      const downloadHandle = downloads.get(source);
+      if (!downloadHandle) return;
 
       try {
-        await moveAsync({ from: cacheFileUri, to: fileUri });
+        await downloadedFile.move(new File(fileUri));
       } catch (error) {
         downloads.delete(source);
         reject(error);
```
msluszniak: Consistency nit: the download `.catch` below wraps failures in `RnExecutorchError(ResourceFetcherDownloadFailed, …, cause)`, but this `move`-failure path rejects with the raw `error`. Worth wrapping it too so callers get a consistent error code/shape.

packages/expo-resource-fetcher/src/ResourceFetcherUtils.ts
```diff
@@ -92,21 +92,20 @@ export namespace ResourceFetcherUtils {
...
+        `Failed to create directory at ${RNEDirectory}`,
+        error
+      );
     }
   }
 
   export async function checkFileExists(fileUri: string) {
-    const fileInfo = await getInfoAsync(fileUri);
-    return fileInfo.exists;
+    return new File(fileUri).exists;
```
msluszniak: `checkFileExists` looks unused now — all call sites were inlined to `new File(...).exists`, so this could be removed. If it's kept, worth noting it's file-only: `new File(dir).exists` isn't equivalent to the old `getInfoAsync` for directory paths (which is exactly why `createDirectoryIfNoExists` correctly switched to `Directory`).

packages/expo-resource-fetcher/src/legacy/ResourceFetcher.ts
```diff
@@ -0,0 +1,195 @@
+/**
```
msluszniak, line 1: This `legacy/` tree duplicates ~500 lines from `src/` (ResourceFetcher / handlers / Utils), differing only in the `expo-file-system` import and a few call shapes. The pause/resume/cancel orchestration is subtle and now lives in two copies that will drift over time. Could the orchestration be shared (some already lives in `BaseResourceFetcherClass`) with just the FS primitives injected via a thin adapter? At minimum a cross-link comment + a "keep these in sync" note would help.

packages/expo-resource-fetcher/src/ResourceFetcher.ts
```diff
@@ -23,24 +23,29 @@
...
-  DownloadStatus,
-} from './ResourceFetcherUtils';
+
+if (
+  typeof Directory !== 'function' ||
+  typeof File !== 'function' ||
+  typeof (File as unknown as { createDownloadTask?: unknown })
+    .createDownloadTask !== 'function'
+) {
+  throw new RnExecutorchError(
```
msluszniak, line 50: This top-level `throw` runs at module-eval (bundle load / app start), not at fetcher use, and it sits between two import groups (imports are hoisted, so it runs after all of them regardless of placement). Consider moving the check into the class constructor / a lazy init so the module stays import-safe and the error surfaces at instantiation with a cleaner stack. Purely stylistic.

packages/expo-resource-fetcher/src/handlers.ts
chmjkb: the new API rejects on non-2xx natively, so the explicit status guard is no longer needed.

packages/expo-resource-fetcher/src/legacy/ResourceFetcher.ts
chmjkb: In general I agree with the concern. I've deduped what's shareable without new abstraction, the API-independent logic (getType, getFilesSizes, core util re-exports) now lives once in sharedUtils.ts, and the subtle orchestration (fetch loop, pause/resume/cancel coordination) was already shared via BaseResourceFetcherClass; what's forked is the FS-binding layer. 

Before going further though, I think we should first agree on how long the legacy path lives, because the answer decides the refactor. Expo has already deprecated the legacy FS API and will likely drop it entirely; as people move to newer SDKs, the audience for /legacy only shrinks. If we agree it's temporary on our end (which I think makes more sense), then investing in an FS-primitives adapter isn't worth spending time. At deletion time we'd be inlining it back out.  I don't think we should be maintaining a path whose underlying API is frozen and abandoned upstream is cost without a growing user base.
 
So my proposal: mark /legacy as @deprecated now, agree on a concrete removal point (e.g.  v1.0.0), and until then treat it as frozen, cross-link comments added to both handlers.ts files.

msluszniak: Ok, we can go with the deprecated path. Please add markers to this issue: https://github.com/software-mansion/react-native-executorch/issues/1213.

msluszniak (commented):
I'll test all demo apps, testing instruction worked. Just one small thing, please resolve this warning: 

Image

EDIT: I tested and all apps worked. Please fix this warning and it's good to go for me.

chmjkb:
> I'll test all demo apps, testing instruction worked. Just one small thing, please resolve this warning:
> 
> Image
> EDIT: I tested and all apps worked. Please fix this warning and it's good to go for me.

should be resolved now!

msluszniak (approved):

PR #1232, opened 2026-06-15
perf: speed up top-p sampling for large vocabularies

NorbertKlockiewicz:
## Description

Optimizes token sampling for large-vocabulary models (e.g. Gemma 4 E2B, 262k vocab), where the previous full-vocabulary sort in top-p dominated per-token latency.

Two changes in `sampler.cpp`:

- **`mask_topp`**: replaces the `O(n log n)` sort over all logits with a logit-space histogram (`kBins=2048`) that locates the nucleus threshold in two `O(n)` passes — no sort, no per-token vocab-sized allocation. Binning in logit space (rather than probability space) keeps uniform resolution for both peaked and flat distributions.
- **`softmax`**: skips `exp()` on logits already masked to `lowest()` by top-k/top-p. The result underflows to zero anyway, and the call is slow on device.

On an iPhone 17 Pro with Gemma 4 E2B (int4), per-token sampling drops from ~45 ms to ~10 ms. The histogram approximates the exact sort-based nucleus; the resulting sampled distribution is statistically equivalent (verified the kept-mass fraction stays within <1% of the exact nucleus across peaked, flat, and sharp distributions).

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

1. Run an LLM with a large vocabulary and a non-zero temperature with `topP` set (e.g. Gemma 4 E2B with `temperature: 0.8`, `topP: 0.9`).
2. Generate a long response and observe tokens/sec.
3. Confirm output remains coherent and sampling is unchanged in character (still stochastic, not greedy).

Greedy decoding (`temperature: 0`) is unaffected — it bypasses this path entirely.

### Checklist

### Additional notes

The histogram is an approximation bounded by bin granularity (`kBins=2048` over a `kRange=40` logit span). This is intentional: exact top-p over a 262k vocab where the nucleus can exceed 100k tokens is inherently expensive, and the sampling outcome is statistically indistinguishable from the exact version.

### DIFF
--- a/packages/react-native-executorch/common/runner/sampler.cpp
+++ b/packages/react-native-executorch/common/runner/sampler.cpp
@@ -36,6 +36,8 @@
 #include <algorithm>
 #include <ctime>
 #include <limits>
+#include <ranges>
+#include <span>
 #include <vector>
 
 namespace executorch {
@@ -149,30 +151,24 @@ template <typename T> void Sampler::mask_topp(T *logits) {
   if (topp_ <= 0.0f || topp_ >= 1.0f) {
     return;
   }
-  constexpr int kBins = 2048;
+  constexpr int32_t kBins = 2048;
   constexpr float kRange = 40.0f;
 
-  float max_val = static_cast<float>(logits[0]);
-  for (size_t i = 1; i < vocab_size_; i++) {
-    float v = static_cast<float>(logits[i]);
-    if (v > max_val) {
-      max_val = v;
-    }
-  }
+  std::span<const T> logit_span{logits, static_cast<size_t>(vocab_size_)};
+  const float max_val =
+      static_cast<float>(*std::ranges::max_element(logit_span));
 
+  // Logit/exp math accumulates over the full 262k vocab; T (bf16) lacks the
+  // precision to sum without large error, so float/double are used here.
   std::vector<double> bin_mass(kBins, 0.0);
   double total = 0.0;
   for (size_t i = 0; i < vocab_size_; i++) {
     float d = static_cast<float>(logits[i]) - max_val;
     float e = std::expf(d);
     total += e;
-    int b = static_cast<int>((d + kRange) / kRange * kBins);
-    if (b < 0) {
-      b = 0;
-    } else if (b >= kBins) {
-      b = kBins - 1;
-    }
-    bin_mass[b] += e;
+    int32_t bin = static_cast<int32_t>((d + kRange) / kRange * kBins);
+    bin = std::clamp(bin, 0, kBins - 1);
+    bin_mass[bin] += e;
   }
   if (total <= 0.0) {
     return;
@@ -182,11 +178,11 @@ template <typename T> void Sampler::mask_topp(T *logits) {
   // kept (HuggingFace "keep the token that crosses" convention).
   const double target = static_cast<double>(topp_) * total;
   double acc = 0.0;
-  int keep_bin = 0;
-  for (int b = kBins - 1; b >= 0; --b) {
-    acc += bin_mass[b];
+  int32_t keep_bin = 0;
+  for (int32_t bin = kBins - 1; bin >= 0; --bin) {
+    acc += bin_mass[bin];
     if (acc >= target) {
-      keep_bin = b;
+      keep_bin = bin;
       break;
     }
   }
--- a/packages/react-native-executorch/common/runner/sampler.cpp
+++ b/packages/react-native-executorch/common/runner/sampler.cpp
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
   }
 }
 
-// Mask logits whose softmax-prob falls outside the top-p nucleus to -inf.
-// Keeps the token that crosses the threshold (HuggingFace convention).
+// Mask logits outside the top-p nucleus to -inf. Approximates the exact
+// sort-based nucleus with a histogram over (logit - max): two O(n) passes, no
+// sort. Binning in logit (not probability) space keeps uniform resolution for
+// peaked and flat distributions alike. kRange=40 spans exp() down to ~4e-18.
 template <typename T> void Sampler::mask_topp(T *logits) {
   if (topp_ <= 0.0f || topp_ >= 1.0f) {
     return;
   }
-  // Softmax into a scratch probs[] (do not mutate logits yet).
-  T max_val = logits[0];
+  constexpr int kBins = 2048;
+  constexpr float kRange = 40.0f;
+
+  float max_val = static_cast<float>(logits[0]);
   for (size_t i = 1; i < vocab_size_; i++) {
-    if (logits[i] > max_val) {
-      max_val = logits[i];
+    float v = static_cast<float>(logits[i]);
+    if (v > max_val) {
+      max_val = v;
     }
   }
-  std::unique_ptr<ProbIndex<T>[]> probindex =
-      std::make_unique<ProbIndex<T>[]>(vocab_size_);
-  T sum = 0;
+
+  std::vector<double> bin_mass(kBins, 0.0);
+  double total = 0.0;
   for (size_t i = 0; i < vocab_size_; i++) {
-    T e = static_cast<T>(std::expf(static_cast<float>(logits[i] - max_val)));
-    probindex[i].prob = e;
-    probindex[i].index = i;
-    sum += e;
+    float d = static_cast<float>(logits[i]) - max_val;
+    float e = std::expf(d);
+    total += e;
+    int b = static_cast<int>((d + kRange) / kRange * kBins);
+    if (b < 0) {
+      b = 0;
+    } else if (b >= kBins) {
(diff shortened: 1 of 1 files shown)


packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
+  constexpr int kBins = 2048;
+  constexpr float kRange = 40.0f;
+
+  float max_val = static_cast<float>(logits[0]);
   for (size_t i = 1; i < vocab_size_; i++) {
-    if (logits[i] > max_val) {
-      max_val = logits[i];
+    float v = static_cast<float>(logits[i]);
+    if (v > max_val) {
+      max_val = v;
```
mkopcins: use span instead (cant do proper suggestion:/)
```suggestion
    std::span<const T> s{logits, vocab_size_};
    *std::ranges::max_element(s); 
```

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
+
+  std::vector<double> bin_mass(kBins, 0.0);
+  double total = 0.0;
   for (size_t i = 0; i < vocab_size_; i++) {
-    T e = static_cast<T>(std::expf(static_cast<float>(logits[i] - max_val)));
-    probindex[i].prob = e;
-    probindex[i].index = i;
-    sum += e;
+    float d = static_cast<float>(logits[i]) - max_val;
+    float e = std::expf(d);
```
mkopcins: I know this is the smae naming scheme as it was, but change it to something that is longer than single letter

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
+      b = 0;
+    } else if (b >= kBins) {
+      b = kBins - 1;
+    }
+    bin_mass[b] += e;
   }
-  if (sum <= T(0)) {
+  if (total <= 0.0) {
     return;
   }
```
mkopcins: this seems always false

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
-    sum += e;
+    float d = static_cast<float>(logits[i]) - max_val;
+    float e = std::expf(d);
+    total += e;
+    int b = static_cast<int>((d + kRange) / kRange * kBins);
+    if (b < 0) {
+      b = 0;
+    } else if (b >= kBins) {
+      b = kBins - 1;
+    }
```
mkopcins:
```suggestion
    b = max(min(kBins-1,b),0);
```
or use std::clamp

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
 template <typename T> void Sampler::mask_topp(T *logits) {
   if (topp_ <= 0.0f || topp_ >= 1.0f) {
     return;
   }
-  // Softmax into a scratch probs[] (do not mutate logits yet).
-  T max_val = logits[0];
+  constexpr int kBins = 2048;
+  constexpr float kRange = 40.0f;
+
+  float max_val = static_cast<float>(logits[0]);
```
msluszniak: Why we use floats and doubles instead of `T` across this whole function?

mkopcins (changes requested):

packages/react-native-executorch/common/runner/sampler.cpp
msluszniak: use `std::clamp` here

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
+// sort-based nucleus with a histogram over (logit - max): two O(n) passes, no
+// sort. Binning in logit (not probability) space keeps uniform resolution for
+// peaked and flat distributions alike. kRange=40 spans exp() down to ~4e-18.
 template <typename T> void Sampler::mask_topp(T *logits) {
   if (topp_ <= 0.0f || topp_ >= 1.0f) {
     return;
   }
-  // Softmax into a scratch probs[] (do not mutate logits yet).
-  T max_val = logits[0];
+  constexpr int kBins = 2048;
```
msluszniak: Why it is signed int?

packages/react-native-executorch/common/runner/sampler.cpp
```diff
@@ -141,57 +141,61 @@ template <typename T> void Sampler::mask_topk(T *logits) {
...
+  double total = 0.0;
   for (size_t i = 0; i < vocab_size_; i++) {
-    T e = static_cast<T>(std::expf(static_cast<float>(logits[i] - max_val)));
-    probindex[i].prob = e;
-    probindex[i].index = i;
-    sum += e;
+    float d = static_cast<float>(logits[i]) - max_val;
+    float e = std::expf(d);
+    total += e;
+    int b = static_cast<int>((d + kRange) / kRange * kBins);
```
msluszniak: when you want to use signed int, prefer explicit naming like `int32_t`

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Done — switched to `std::ranges::max_element` over a `std::span`.

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Done — renamed to `bin`.

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Done — using `std::clamp`.

NorbertKlockiewicz: Done — using `std::clamp`.

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Done — `bin`/`keep_bin`/`kBins` are now `int32_t`.

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Made it explicit `int32_t` (it's a loop/bin index and `std::clamp(bin, 0, kBins - 1)` wants matching signed args). Switched to unsigned would force casts against the signed bounds.

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: It's actually reachable: `bin = (d + kRange) / kRange * kBins` and the max token has `d == 0`, giving `bin == kBins` (== 2048), so the upper bound is hit. It's moot now though — replaced with `std::clamp`.

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Intentional: the running `total`/`bin_mass` sums accumulate `exp()` over the full 262k vocab, and `T` here is bf16 — summing that many values in bf16 loses too much precision (catastrophic cancellation). float/double accumulation keeps the threshold accurate. Added a comment noting this.

msluszniak: Ok, when T is bf16, but when T is double then there are casts in tge code that are loss-propagating. This should be done the way that guaratee that we only upcast.

mkopcins (approved):

packages/react-native-executorch/common/runner/sampler.cpp
NorbertKlockiewicz: Good catch — corrected. Reworked it to compute in `acc_t = std::conditional_t<std::is_same_v<T, double>, double, float>`: every logit is widened to `acc_t` (or left as-is), never narrowed, so there's no loss-propagating cast for any logit dtype. `acc_t` is `float` for all current instantiations (Float/Half/BFloat16/UInt16) and only becomes `double` if a `double` logit path is ever added. The reason we don't just accumulate in `T` is bf16 specifically: summing `exp()` over the 262k vocab in bf16 saturates (measured ~40–99% error), whereas float is accurate to well under the bin granularity.

msluszniak:
Code is ok, now testing

msluszniak (approved):
I ran it on my android with different topp and other params described in testing steps and it worked.

PR #1236, opened 2026-06-16
chore(ios): simulator fallback for mlx backend

NorbertKlockiewicz:
## Description

The MLX backend cannot link or run against the iOS **Simulator** Metal SDK (`_MTLTensorDomain` / `_MTLIOErrorDomain` are absent from the simulator Metal SDK). This PR removes the dead simulator MLX binary and makes model selection automatically fall back to XNNPACK on the iOS simulator.

Changes:

- **`modelRegistry.ts`** — `resolveBackend` gains a simulator policy (`applySimulatorPolicy`): when a model resolves to the `mlx` backend on the iOS simulator (`react-native-device-info` `isEmulatorSync()`), it falls back to `xnnpack` if the model ships it. An **explicitly requested** `mlx` backend on the simulator is not silently swapped — it throws a clear, actionable error instead of failing cryptically at native model load. Models with no simulator-compatible backend (multimodal Gemma) also throw at selection time, before any download.
- Multimodal Gemma is converted to a proper variant map (`mlx` / `vulkan`) so it flows through the resolver; the dead `Platform.OS`-ternary consts it used are removed.
- **`react-native-device-info`** added as a dependency (used by the resolver).
- **`libbackend_mlx_simulator.a` deleted** and the xcframework rebuilt — the `ios-arm64-simulator` slice is MLX-free; the `ios-arm64` (device) slice keeps MLX.

The fallback/throw happens when the model accessor is called (eagerly), i.e. before any model download starts.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

1. Run the LLM example app on the **iOS Simulator**, select Gemma 4 E2B (text), and confirm it loads the XNNPACK model instead of failing on MLX.
2. On the simulator, calling `models.llm.gemma4_e2b({ backend: 'mlx' })` throws a clear error (MLX requires a physical device); `models.llm.gemma4_e2b()` falls back to XNNPACK.
3. On a **physical device**, Gemma 4 E2B still resolves to MLX and runs as before.
4. Multimodal Gemma on the simulator throws a descriptive error at selection time (no XNNPACK multimodal build exists).

### Checklist

### Additional notes

`react-native-device-info` is a native module, so consuming apps must have it linked (pod installed) — it ships with the example apps.

### DIFF
--- a/packages/react-native-executorch/src/constants/modelRegistry.ts
+++ b/packages/react-native-executorch/src/constants/modelRegistry.ts
@@ -79,17 +79,10 @@ type ConfigOf<V> = Extract<
 >;
 type BackendsOf<V> = Extract<keyof V, Backend>;
 
-const BACKEND_ORDER: Backend[] = ['xnnpack', 'coreml', 'mlx', 'vulkan', 'qnn'];
-
-function firstBackend(variants: AnyVariantMap): Backend {
-  for (const b of BACKEND_ORDER) {
-    if (variants[b]) return b;
-  }
-  throw new RnExecutorchError(
-    RnExecutorchErrorCode.Internal,
-    'Model variant map is empty.'
-  );
-}
+const PLATFORM_PREFERENCE: Partial<Record<typeof Platform.OS, Backend[]>> = {
+  ios: ['coreml', 'mlx', 'xnnpack'],
+  android: ['vulkan', 'qnn', 'xnnpack'],
+};
 
 function applySimulatorPolicy(
   backend: Backend,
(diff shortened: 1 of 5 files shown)


packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -90,26 +91,47 @@ function firstBackend(variants: AnyVariantMap): Backend {
...
   requested: Backend | undefined
 ): Backend {
-  if (requested) return requested;
+  if (requested) return applySimulatorPolicy(requested, variants, true);
   if (platformDefaults) {
     if (Platform.OS === 'ios' && platformDefaults.ios)
-      return platformDefaults.ios;
+      return applySimulatorPolicy(platformDefaults.ios, variants, false);
     if (Platform.OS === 'android' && platformDefaults.android) {
       return platformDefaults.android;
```
msluszniak: The simulator policy is applied by decorating individual `return` sites, and the asymmetry shows: the `requested`, `ios`, `default`, and `firstBackend` paths get wrapped, but the `android` branch doesn't.

It's functionally harmless (`applySimulatorPolicy` early-returns unless `Platform.OS === 'ios'`, so wrapping android would be a no-op), but it reads as if android is *intentionally* exempt from a policy when it's really just unreachable. A future maintainer can't tell the difference, and if the policy ever grows a non-iOS case the android branch silently won't get it.

Would prefer separating *selection* from *validation* and applying the policy once at a single chokepoint, so every path (android included) flows through the same gate uniformly:

```ts
function resolveBackend(
  variants: AnyVariantMap,
  platformDefaults: PlatformDefaults<Backend> | undefined,
  requested: Backend | undefined
): Backend {
  const explicit = requested !== undefined;
  const backend = selectBackend(variants, platformDefaults, requested);
  return applySimulatorPolicy(backend, variants, explicit);
}

// pure selection — no simulator awareness
function selectBackend(variants, platformDefaults, requested): Backend {
  if (requested) return requested;
  if (platformDefaults) {
    if (Platform.OS === 'ios' && platformDefaults.ios) return platformDefaults.ios;
    if (Platform.OS === 'android' && platformDefaults.android) return platformDefaults.android;
    if (platformDefaults.default) return platformDefaults.default;
  }
  if (Platform.OS === 'ios' && variants.coreml) return 'coreml';
  if (Platform.OS === 'android' && variants.xnnpack) return 'xnnpack';
  return firstBackend(variants);
}
```

This also computes `explicit` once instead of hand-setting the boolean at each call site, and keeps the selection logic readable.

msluszniak (approved):
Fallback works, and error with explicit MLX on sim raises correct error:

Image

Issue #1208, opened 2026-06-01
[RNE Rewrite] - Aggregated issue


msluszniak:
### Discussed in https://github.com/software-mansion/react-native-executorch/discussions/1189

<div type='discussions-op-text'>

<sup>Originally posted by **msluszniak** May 26, 2026</sup>

# RNE refactor

As we discussed internally, we need to refactor lib to be open for non-dectructing additions to the lib, be more modular, improve speed of the core functionalities of the lib. Here is a PoC that we will use to progress with the issue: 

- https://github.com/barhanc/rnet-poc/tree/main

For an overview of the PoC you can read [design doc](https://github.com/barhanc/rnet-poc/blob/main/README.md). For getting familiar with the new implementation style and guidelines you can use your favourite coding agent ran inside the linked repo and the [skills](https://github.com/barhanc/rnet-poc/blob/main/README.md) provided there. Of course the guidelines are only tips and recommendations based on the lessons learnt from the PoC implementations, not hard constraints, so if you feel that some of them make the implementation clunky, ugly, not performant, etc. feel free to ignore them (and open issue, so we can all discuss it).

Below there is a comprehensive list that will help us keep up to date parity process with what has been already done.

## ML Tasks

For each task one should:

- Implement the required native C++ operations (e.g. for manipulating image tensor) / bindings (e.g. for LLM runner). Typically you shouldn't need to touch the `core/` C++ implementations of model and tensor objects unless there is really no way around it.
- Implement the TS-side helpers (which may or may not wrap the C++ operations or just be implemented in pure TS) for the specific task.
- Implement the actual task pipeline in TS in the appropriate `tasks/` directory using the implemented and added primitives.
- Implement the hook wrapper over the task in `hooks/` directory.
- Update `models.ts` and `constants.ts` with appropriate constants for pre-exported models for your pipeline.
- Implement / update the example app to showcase the task pipeline and facilitate user testing.
- [PLACEHOLDER] Tests, tests, tests, ...
- Add documentation for your implementation.

### Computer Vision

- [x] #1237 
- [x] #1238
- [x] #1239
- [x] #1240
- [x] #1241
- [x] #1242
- [x] #1243
- [x] #1244

### NLP

- [x] #1245
- [x] #1246
- [x] #1248

### Speech

- [x] #1249
- [x] #1250
- [x] #1251

### Embeddings

- [x] #1247

### Other

[...]

PR #1230, opened 2026-06-12
style: add clang-format config and reformat C++ sources

msluszniak:
## Description

Replaces the implicit default LLVM style (which aggressively wrapped function signatures, chained ternaries and string concatenations) with a custom `.clang-format`: LLVM base, `ColumnLimit: 100`, `Standard: c++20` and `InsertNewlineAtEOF`. A `.clang-format-ignore` excludes the vendored libraries (`ada`, `pfft`, `runner`) and the generated `ErrorCodes.h`, and is picked up automatically by the existing clang-format pre-commit hook (no `lefthook.yml` change needed). All first-party C/C++/Objective-C sources are reformatted accordingly — the diff is whitespace-only.

### Introduces a breaking change?

- [ ] Yes
- [x] No

### Type of change

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Documentation update (improves or adds clarity to existing documentation)
- [x] Other (chores, tests, code style improvements etc.)

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

Whitespace-only change (clang-format never alters tokens). Re-running `clang-format` over the tree produces no further diff (idempotent), and vendored/generated files are left untouched.

### Related issues

Closes #1217

### Checklist

- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the documentation accordingly
- [x] My changes generate no new warnings

### Additional notes

This is a large but mechanical diff; in-flight C++ branches should rebase onto it to avoid conflicts.

PR #1256, opened 2026-06-17
chore: scaffold rne-rewrite target branch (#1255)

### LINKED ISSUE #1255: [RNE Rewrite] Setup target branch for all rewrite PRs
To start the rewrite we need to setup the target branch for all rewrite PRs. More specifically we need to branch from main and then:
- nuke `apps/`, `packages/react-native-executorch`
- setup empty `packages/react-native-executorch`
- update CI
- update GH meta: linters, formatters, precommit, etc.

msluszniak:
## Description

Sets up the long-lived `rne-rewrite` branch that all RNE rewrite PRs will target (see #1208). Keeps the repo's yarn monorepo layout, establishes the **core first**, and leaves per-task functionality to follow as separate block PRs.

- Removes `apps/` and the legacy `packages/react-native-executorch`.
- Stands up the new core package by porting the `rnet-poc` lower-level API: `cpp/core` (tensor/model/dtype/install/utils), `src/core`, and the native bridge. `MyLib` → `rnexecutorch` rename throughout. OpenCV and task extensions are intentionally deferred to the per-task PRs.
- CI: deletes the `apps/llm` native build workflows and the stale error-code verification step; runs CI on `rne-rewrite`.
- Meta: scopes lint/typecheck/build to the core package; the `*-resource-fetcher` and `-webrtc` packages stay on disk but are excluded from CI until their surface is reworked.

- [ ] Yes
- [x] No

(Targets the `rne-rewrite` dev branch; the published package on `main` is untouched.)

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

CI is TypeScript-only here (no native compile). From the repo root:

```
yarn install
yarn workspaces foreach --all --exclude react-native-executorch-bare-resource-fetcher --exclude react-native-executorch-expo-resource-fetcher --exclude react-native-executorch-webrtc --topological-dev run prepare
yarn typecheck
yarn lint
```

All four pass green. On-device native build is a follow-up.

### Related issues

Closes #1255. Part of #1208.

### Checklist

### Additional notes

- Native ExecuTorch `third-party` binaries are **not** committed (see `packages/react-native-executorch/third-party/README.md`). CI doesn't need them; provisioning them (ideally via the repo's on-demand artifact mechanism) is a follow-up. OpenCV is dropped since CV is a later block.
- The `common/phonemis` submodule is retained (unchanged from `main`) for later use by the TTS block.
- The error-code codegen system is parked (the POC has none); `scripts/generate-errors.ts` stays dormant for later reintroduction.
- Root submodules (`tokenizers-cpp`, `googletest`) are left in place for likely reuse by the LLM block; they aren't checked out in CI.
- `react-native-worklets` emits peer-dep warnings (`@babel/core`, `@react-native/metro-config`) — harmless app-integration peers, not needed to build the lib.


barhanc:
- The `docs/docs` directory should also probably be nuked, because it will have to be rewritten. 
- Also, don't we want to set the clang formatter config in this PR, because right now I see the `.cpp` files are still formatted in the MSVC style as the PoC, but #1230  is the style we want.
- Is it possible to set-up branch protection rules similar to `main` on the rewrite development branch?

msluszniak:
Thanks — addressed in the latest push:

**1. clang-format ✅** Brought in `.clang-format` (and `.clang-format-ignore`) from #1230 and reformatted the ported C++ to the LLVM-based style — the PoC/MSVC formatting is gone. Note the `.clang-format-ignore` patterns (`common/ada`, `common/pfft`, `common/runner`, `ErrorCodes.h`) don't exist on this branch yet; I kept them verbatim for parity so it lines up when those land / when #1230 reaches `main`.

**2. docs ✅ (partial — want your call)** Nuked `docs/docs`. I left `docs/versioned_docs/` (~7.7 MB), `docs/versioned_sidebars/` and `docs/versions.json` in place since they're old released-version snapshots and you only called out `docs/docs` — but by the same "it'll be rewritten" logic they're obsolete here too. Want me to nuke those as well (and trim `sidebars.js`/`docusaurus.config.js` accordingly)? The docs build only runs on `main`, so it won't gate this branch either way.

**3. Branch protection ⚠️ needs an admin** I can't do this from here — the branch-protection API returns 404 for my token (no admin scope). A maintainer would need to mirror `main`'s ruleset onto `rne-rewrite` via **Settings → Branches** (or the rulesets API). Happy to provide the exact ruleset JSON if useful.

.clang-format
barhanc, line 1: Personally I would prefer something more similar to: `{ BasedOnStyle: LLVM, UseTab: Never, IndentWidth: 4, TabWidth: 4, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, FixNamespaceComments: false }`. It is basically the default VS Code one, but doesn't put the `{` on the new line. Additionally regarding the column limit option, I find that clang formatter often makes very awkward line breaks to satisfy the column limit that often are not very readable, and it would be better to set this to no column limit and let the developers handle line breaks for very long lines. What do you think?

barhanc:
- We should also replace `skills/` with https://github.com/barhanc/rnet-poc/tree/main/.agents/skills (and let's put them under `.agents/` because otherwise e.g. antigravity doesn't see them). The current skills tell only how to use the library, the PoC skills tell how to build/extend it.
- I'm not sure what to do with the `scripts/error.config.ts` and `scripts/generate-errors.ts`. Do we need them and if so how to integrate them into the rewrite?
- `.clang-format-ignore` can be deleted for now as it only specifies stale paths that are nonexistent in the rewrite.
- We have to edit the categories in `.eslintrc.js`.
- In `.gitmodules` we have the tokenizers-cpp entry, but don't we bundle our custom tokenizers inside ExecuTorch now as per #781 ?
- ~The README, RELEASE.md and generic docs stuff like sidebar, etc. will have to be changed in the future. This, we should probably put in some TODO, either in the repo itself or in the tracking issue, so that we don't forget about it.~ (Added to tracking issue list)

.clang-format
msluszniak: Ok, so I think we can get a union of these two setups and disable fixed column limit. How about that?

msluszniak:
>We should also replace skills/ with https://github.com/barhanc/rnet-poc/tree/main/.agents/skills (and let's put them under .agents/ because otherwise e.g. antigravity doesn't see them). The current skills tell only how to use the library, the PoC skills tell how to build/extend it.

Agreed. I will change it. 

>I'm not sure what to do with the scripts/error.config.ts and scripts/generate-errors.ts. Do we need them and if so how to integrate them into the rewrite?

I think that ultimately we want to have some custom error types etc. So we probably want something like that, but we may think if we can add some improvements idk. I would add a separate point in aggregated issue for now, and we can eventually copy this solution in the end.

>.clang-format-ignore can be deleted for now as it only specifies stale paths that are nonexistent in the rewrite.

Yeap, will delete.

>In .gitmodules we have the tokenizers-cpp entry, but don't we bundle our custom tokenizers inside ExecuTorch now as per https://github.com/software-mansion/react-native-executorch/pull/781 ?

Yes, but there are kept in current main because it was treated as source of truth in native tests for our implementation of tokenizers. So I guess we can keep it for now and write similar tests based on them. But if you want I can delete them as well.

.clang-format
barhanc: Yeah, let's do that.

msluszniak: Applied in `d3960e2a5`. Final `.clang-format` is the union of both setups with the fixed column limit disabled:

```yaml
BasedOnStyle: LLVM
Standard: c++20
UseTab: Never
IndentWidth: 4
TabWidth: 4
ColumnLimit: 0
AccessModifierOffset: -4
AllowShortIfStatementsOnASingleLine: false
IndentCaseLabels: false
FixNamespaceComments: false
InsertNewlineAtEOF: true
```

Re-ran clang-format over the ported C++ — now 4-space indent, attached braces, no column-limit wrapping. All sources pass `clang-format --dry-run -Werror`.

barhanc (commented):
- The clang formatter should be run on the source code copied from PoC, since now we first formatted it using the previous config and then the current one.
- In third-party/ README it would be helpful to add a link from where we can get the binaries https://github.com/barhanc/rnet-poc/tree/main/third-party.

Other than that looks perfect!

msluszniak:
Both done in `dfda78ce2`: re-derived the C++ from the pristine PoC and ran clang-format once (the column-100 wraps are gone), and added the binaries source link (PoC `third-party/`) to the README.

barhanc (approved):

PR #1264, opened 2026-06-18
[RNE Rewrite] feat!: add classification task and common utilities for manipulating tensors in CV tasks

### LINKED ISSUE #1207: Add utilities for inspecting .pte files directly
While trying to debug problems with Disitiluse model I felt the process would be much more developer friendly, if we had some utilities to inspect model files directly, like input/output tensor schemas, supported backends, etc.  The scope of this issue is to add such functionality to react-native-executorch.

### LINKED ISSUE #1237: [RNE Rewrite] CV - add classification pipeline implementation
Implement classification pipeline using the new approach to achieve parity with current implementation. For PoC implementation see: https://github.com/barhanc/rnet-poc/blob/main/src/extensions/cv/tasks/classification.ts.

barhanc:
## Description

This PR adds: 
- common utilities such as `useModel` and `useResourceDownload` hook which are needed for all models
- common utilities and scaffolding for computer vision tasks (including native operations for image tensors manipulation)
- scaffolding for computer vision example app
- classification task with testing screen in the example app
- inspection utilities and screen in the example app to facilitate easier exports in the future for other tasks

It also introduces some fixes:
- flexible pages support on Android
- iOS bridge initialization for the RN new architecture
- configures prettierrc to 100 column width
- adds documentation to core functionalities (both on TS and native side)

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Build the example computer-vision app and confirm everything builds correctly on both iOS and Android.
- [ ] Run the classification task and validate the results.
- [ ] Check the inspection screen (try to inspect some models from our HF).

### Screenshots

N/A

### Related issues

Closes #1237 
Closes #1207 

### Checklist


apps/computer-vision/app/classification/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+    value: models.classification.EFFICIENTNET_V2_S.XNNPACK_INT8,
+  },
+  {
+    label: 'EfficientNetV2-S (XNNPACK FP32)',
+    value: models.classification.EFFICIENTNET_V2_S.XNNPACK_FP32,
+  },
+  {
+    label: 'EfficientNetV2-S (CoreML FP16)',
+    value: models.classification.EFFICIENTNET_V2_S.COREML_FP16,
+  },
```
msluszniak, line 32: If we add different backend variants, we need to protect platform specific one with graceful errors, so if anyone on android chooses coreml, we should explicitly inform about limitations.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+    if (interp == "nearest")
+        return ::cv::INTER_NEAREST;
+    if (interp == "area")
+        return ::cv::INTER_AREA;
+    if (interp == "linear")
+        return ::cv::INTER_LINEAR;
+    if (interp == "cubic")
+        return ::cv::INTER_CUBIC;
+    if (interp == "lanczos")
+        return ::cv::INTER_LANCZOS4;
```
msluszniak: I prefer to never use control flow instructions without `{}`. I will put clang-tidy enforcement: https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html for this one in the future. For now, I suggest just adding them ;)

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+                                    static_cast<double>(dst_h) / src_h);
+
+            int new_w = static_cast<int>(std::round(src_w * scale));
+            int new_h = static_cast<int>(std::round(src_h * scale));
+            int off_x = (new_w - dst_w) / 2;
+            int off_y = (new_h - dst_h) / 2;
+
+            ::cv::Mat scaled;
+            ::cv::resize(src_mat, scaled, ::cv::Size(new_w, new_h), 0, 0, interp_flag);
+            scaled(::cv::Rect(off_x, off_y, dst_w, dst_h)).copyTo(dst_mat);
```
msluszniak: These are perfect parts to be encapsulated into separate functions. When scanning through code I don't need to know the details of operations that gives me correct shapes for resize.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+        if (!opts.hasProperty(rt, "mode") || !opts.getProperty(rt, "mode").isString()) {
+            throw jsi::JSError(rt, "resize: options.mode is required and must be a string");
+        }
+
+        if (!opts.hasProperty(rt, "interpolation") || !opts.getProperty(rt, "interpolation").isString()) {
+            throw jsi::JSError(rt, "resize: options.interpolation is required and must be a string");
+        }
+
+        if (!opts.hasProperty(rt, "padValue") || !opts.getProperty(rt, "padValue").isNumber()) {
```
msluszniak, line 101: Why `padValue` is always required if we use it only when `mode == "letterbox"`?

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+        if (!dst->data_) {
+            throw jsi::JSError(rt, "resize: dst tensor has been disposed");
+        }
+
+        int src_h = src->shape_[0];
+        int src_w = src->shape_[1];
+        int channels = src->shape_[2];
+        int dst_h = dst->shape_[0];
+        int dst_w = dst->shape_[1];
```
msluszniak: I think we always want to put explicit types like `int32_t` etc. and we need to think if given value should be signed or unsigned value.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+        ::cv::Mat src_mat(src_h, src_w, cv_type, src->data_.get());
+        std::vector<::cv::Mat> channels;
+        ::cv::split(src_mat, channels);
+
+        int hw = src_h * src_w;
+        size_t elem_size = rnexecutorch::core::types::elementSize(src->dtype_);
+        uint8_t *dst_ptr = dst->data_.get();
+
+        for (int i = 0; i < src_c; ++i) {
```
msluszniak: should be unisigned type, `size_t`

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+        int h = src->shape_[1];
+        int w = src->shape_[2];
+
+        bool dstMatch = false;
+        if (dst->shape_.size() == 3 &&
+            dst->shape_[0] == c &&
+            dst->shape_[1] == h &&
+            dst->shape_[2] == w) {
+            dstMatch = true;
+        }
```
msluszniak: You should assign value of the condition to the value immediately, not check and assign.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+        if (!opts.hasProperty(rt, "beta")) {
+            throw jsi::JSError(rt, "normalize: options.beta is required");
+        }
+
+        std::vector<double> alpha(c);
+        std::vector<double> beta(c);
+
+        auto alphaVal = opts.getProperty(rt, "alpha");
+        if (alphaVal.isNumber()) {
+            std::fill(alpha.begin(), alpha.end(), alphaVal.asNumber());
```
msluszniak:
```suggestion
            std::ranges::fill(alpha, alphaVal.asNumber());
```

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+            for (int i = 0; i < c; ++i) {
+                auto val = arr.getValueAtIndex(rt, i);
+                if (!val.isNumber()) {
+                    throw jsi::JSError(rt, "normalize: options.beta array must contain only numbers");
+                }
+                beta[i] = val.asNumber();
+            }
+        } else {
+            throw jsi::JSError(rt, "normalize: options.beta must be a number or an array of numbers");
+        }
```
msluszniak: Duplication

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+    switch (dtype) {
+    case rnexecutorch::core::types::DType::uint8:
+        return CV_8U;
+    case rnexecutorch::core::types::DType::int32:
+        return CV_32S;
+    case rnexecutorch::core::types::DType::float32:
+        return CV_32F;
+    }
+    throw std::invalid_argument("unsupported dtype");
+}
```
msluszniak: These should be placed in anonymous namespace, not static functions. Same for other static from here.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+        if (!args[2].isNumber()) {
+            throw jsi::JSError(rt, "softmax: axis must be a number");
+        }
+        int axis = static_cast<int>(args[2].asNumber());
+
+        const int rank = static_cast<int>(src->shape_.size());
+        if (axis < 0) {
+            axis += rank;
+        }
```
msluszniak: Could you explain how it works?

packages/react-native-executorch/cpp/extensions/math/operations.cpp
```diff
@@ -0,0 +1,288 @@
...
+                for (size_t d = 0; d < axisDim; ++d) {
+                    float val = srcData[o * axisDim * inner + d * inner + i];
+                    if (val > max_val) {
+                        max_val = val;
+                        max_idx = static_cast<int32_t>(d);
+                    }
+                }
+                dstData[o * inner + i] = max_idx;
+            }
+        }
```
msluszniak, line 301: I would be in favour of little rewrite here to make loop over i be the most inner one. This will help make it vectorized:
```cpp
int32_t *dstData = reinterpret_cast<int32_t *>(dst->data_.get());
std::vector<float> maxVals(inner);

for (size_t o = 0; o < outer; ++o) {
    const float *srcSlab = srcData + o * axisDim * inner;
    int32_t *dstRow = dstData + o * inner;

    for (size_t i = 0; i < inner; ++i) {
        maxVals[i] = -std::numeric_limits<float>::infinity();
        dstRow[i] = 0;
    }

    for (size_t d = 0; d < axisDim; ++d) {
        const float *srcRow = srcSlab + d * inner;
        for (size_t i = 0; i < inner; ++i) {
            const float val = srcRow[i];
            if (val > maxVals[i]) {
                maxVals[i] = val;
                dstRow[i] = static_cast<int32_t>(d);
            }
        }
    }
}
```

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,613 @@
...
+            throw jsi::JSError(rt, "resize: options.interpolation is required and must be a string");
+        }
+
+        if (!opts.hasProperty(rt, "padValue") || !opts.getProperty(rt, "padValue").isNumber()) {
+            throw jsi::JSError(rt, "resize: options.padValue is required and must be a number");
+        }
+
+        auto mode = opts.getProperty(rt, "mode").asString(rt).utf8(rt);
+        auto interp = opts.getProperty(rt, "interpolation").asString(rt).utf8(rt);
+        double padValue = opts.getProperty(rt, "padValue").asNumber();
```
msluszniak, line 107: What is the convention in naming variables, some are named using camel case, other using snake case like: `src_h` etc.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+        for (size_t outerIndex = 0; outerIndex < outer; ++outerIndex) {
+            for (size_t innerIndex = 0; innerIndex < inner; ++innerIndex) {
+                const size_t base = outerIndex * axisDim * inner + innerIndex;
+
+                float maxValue = -std::numeric_limits<float>::infinity();
+                for (size_t axisIndex = 0; axisIndex < axisDim; ++axisIndex) {
+                    maxValue = std::max(maxValue, srcData[base + axisIndex * inner]);
+                }
```
msluszniak, line 177: maybe use `std::max_element`?

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
```diff
@@ -0,0 +1,108 @@
...
+  const classifyWorklet = (
+    input: ImageBuffer,
+    options?: { topk?: number }
+  ): Classification<L>[] => {
+    'worklet';
+    const tInput = preprocessor.process(input);
+    model.execute('forward', [tInput], [tLogits]);
+
+    const probas = tLogits
+      .through(softmax, tProbas) //
```
msluszniak:
```suggestion
      .through(softmax, tProbas)
```

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
```diff
@@ -0,0 +1,108 @@
...
+    model,
+    'forward',
+    [SymbolicTensor('float32', [1, 3, 'H', 'W'], [3, 'H', 'W'])],
+    [SymbolicTensor('float32', [1, 'N'], ['N'])]
+  );
+  const inpShape = meta.inputTensorMeta[0]!.shape;
+  const outShape = meta.outputTensorMeta[0]!.shape;
+
+  const tensors = [
+    tensor('float32', outShape), //
```
msluszniak:
```suggestion
    tensor('float32', outShape),
```

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
```diff
@@ -0,0 +1,108 @@
+import type { WorkletRuntime } from 'react-native-worklets';
+
+import { tensor } from '../../../core/tensor';
+import { loadModel } from '../../../core/model';
+import { validateModelSchema, SymbolicTensor } from '../../../core/modelSchema';
+import { wrapAsync } from '../../../core/runtime';
+
+import { softmax } from '../../math';
+import { type ImageBuffer } from '../image';
```
msluszniak:
```suggestion
import type { tImageBuffer } from '../image';
```

packages/react-native-executorch/src/extensions/cv/ops/image.ts
```diff
@@ -0,0 +1,170 @@
+import { rnexecutorchJsi } from '../../../native/bridge';
+import { type Tensor } from '../../../core/tensor';
+import { type ImageFormat } from '../image';
```
msluszniak:
```suggestion
import type { Tensor } from '../../../core/tensor';
import type { ImageFormat } from '../image';
```

packages/react-native-executorch/src/extensions/cv/tasks/preprocessing.ts
```diff
@@ -0,0 +1,94 @@
+import { tensor, type Tensor } from '../../../core/tensor';
+import { matchShape } from '../../../core/modelSchema';
+
+import { type ImageBuffer } from '../image';
```
msluszniak:
```suggestion
import type { ImageBuffer } from '../image';
```

packages/react-native-executorch/src/extensions/cv/tasks/preprocessing.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+export function createImagePreprocessor(
+  opts: ImagePreprocessorOptions,
+  outputShape: number[]
+): {
+  process: (input: ImageBuffer) => Tensor;
+  dispose: () => void;
+} {
+  if (!matchShape(outputShape, [1, 3, 'H', 'W'], [3, 'H', 'W']))
+    throw new Error(`preprocessor: got shape [${outputShape}], required [1,3,H,W] or [3,H,W]`);
```
msluszniak: Let's keep the same convention with brackets in TS.

packages/react-native-executorch/src/extensions/cv/tasks/preprocessing.ts
```diff
@@ -0,0 +1,94 @@
...
+ * method.
+ */
+export function createImagePreprocessor(
+  opts: ImagePreprocessorOptions,
+  outputShape: number[]
+): {
+  process: (input: ImageBuffer) => Tensor;
+  dispose: () => void;
+} {
+  if (!matchShape(outputShape, [1, 3, 'H', 'W'], [3, 'H', 'W']))
```
msluszniak: Can we move these arrays into named vars, and then reuse in error message using some `toString` or similar method?

packages/react-native-executorch/src/extensions/cv/tasks/preprocessing.ts
```diff
@@ -0,0 +1,94 @@
...
+    throw new Error(`preprocessor: got shape [${outputShape}], required [1,3,H,W] or [3,H,W]`);
+
+  const targetH = outputShape.at(-2)!;
+  const targetW = outputShape.at(-1)!;
+  const tensors = [
+    tensor('uint8', [targetH, targetW, 3]),
+    tensor('uint8', [3, targetH, targetW]),
```
msluszniak: abstract 3 to num channels or sth

packages/react-native-executorch/src/extensions/math.ts
```diff
@@ -0,0 +1,45 @@
+import { rnexecutorchJsi } from '../native/bridge';
+import { type Tensor } from '../core/tensor';
```
msluszniak:
```suggestion
import type { Tensor } from '../core/tensor';
```

msluszniak (commented):
Maybe we change example app icon to something meaningful for CV, I thought about it for our current apps, but haven't executed eventually.

packages/react-native-executorch/src/utils.ts
```diff
@@ -1,5 +1,62 @@
...
+    const model = loadModel(localPath);
+    const methodNames = model.getMethodNames();
+
+    const methods: { name: string; meta: ModelMethodMeta }[] = [];
+    for (const method of methodNames) {
+      const meta = model.getMethodMeta(method);
+      methods.push({ name: method, meta });
+    }
+
+    model.dispose();
```
msluszniak: Shouldn't this `model.dispose();` be in finally block?

packages/react-native-executorch/cpp/extensions/math/operations.cpp
```diff
@@ -0,0 +1,288 @@
...
+        if (!args[0].isObject() || !args[0].asObject(rt).isHostObject<TensorHostObject>(rt)) {
+            throw jsi::JSError(rt, "sigmoid: src must be a Tensor");
+        }
+
+        if (!args[1].isObject() || !args[1].asObject(rt).isHostObject<TensorHostObject>(rt)) {
+            throw jsi::JSError(rt, "sigmoid: dst must be a Tensor");
+        }
+
+        auto src = args[0].asObject(rt).getHostObject<TensorHostObject>(rt);
+        auto dst = args[1].asObject(rt).getHostObject<TensorHostObject>(rt);
```
msluszniak, line 30: Maybe we add the macro to avoid repetitions in this PR?

packages/react-native-executorch/src/hooks/useResourceDownload.ts
```diff
@@ -0,0 +1,104 @@
...
+ * @experimental Subject to change once the temporary react-native-fs dependency is replaced.
+ * See {@link https://github.com/software-mansion/react-native-executorch/issues/1253 | Issue #1253}.
+ * @param source The remote URL or local path to the resource. If it's already a
+ * local path, it is returned immediately as is.
+ * @param preventLoad If true, prevents checks and downloads, resetting the hook
+ * state.
+ * @returns An object containing the local file path, the download progress
+ * percentage, and any download error.
+ */
+export function useResourceDownload(source?: string, preventLoad?: boolean) {
```
msluszniak, line 32: Two components mounting with the same source both see `exists(dest) === false` -> both `downloadFile` into the same `${dest}.partial` -> both `moveFile(tmp, dest)` — the second `moveFile throws` (tmp already consumed) and you can get a corrupted cache file. I know that we want to switch to our resource fetcher. Just flagging FYI.

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
```diff
@@ -0,0 +1,108 @@
...
+ * clear all native memory.
+ * @category Typescript API
+ * @typeParam L The type representing the classification labels.
+ * @param config Classifier task configuration containing path and options.
+ * @param runtime Optional worklet runtime thread on which to run the model
+ * execution.
+ * @returns A promise resolving to an object containing classification and
+ * disposal controls.
+ */
+export async function createClassifier<L>(
```
msluszniak, line 54: `classifierOpts.labels[index]!` silently yields `undefined` labels if the label vocab is shorter than the output `N`. Since we read outShape from meta, assert `labels.length === outShape[last]` at construction. (`slice(0, options?.topk)` with a negative `topk` silently drops the tail.)

apps/computer-vision/app/classification/index.tsx
barhanc: This is already somewhat handled on the native side. Right now trying to execute the CoreML on Android will throw this error
```cpp
throw jsi::JSError(rt, "execute: Method '" + methodName + "' execution failed: " + errorMsg +
                           ". This may be due to missing required backends - use getMethodMeta()" +
                           " to check required backends and getExecuTorchRegisteredBackends()" +
                           " to check which backends are registered in the runtime.");
```
which is informative for us as developers, but I agree might not be great for app users. Do you want me to change the model list to have `Platform.select`?

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Basically we assume in the native code that all arguments are required and always explicitly passed and deal with optional values only on the TS side. This assures that it is clear for API consumers what the default values of optional params are without looking into cpp code.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
barhanc: Basically like in numpy, if you pass `-1` it is the last axis (i.e. `len(shape) - 1`), `-2` is the second to last and so on. There is a check right below that throws if the resulting axis is out of range. I can add a comment there.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Good question. From what I've seen the JSI codebases usually use camelCase, but standard in cpp is snake case for stuff like std functions. I think we should probably use camelCase only except for the `install_jsMethod` top level functions, because I think there this mixed style is more readable than `installJSMethod`. If you agree, I will rename variables to be in camelCase where it applies.

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
barhanc: Without the empty comment the formatter collapses it into 
```ts
const probas = tLogits.through(softmax, tProbas).getData(new Float32Array(tProbas.numel));
```
which is less readable imo.

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
barhanc: Without the empty comment the formatter collapses it into 
```ts
const tensors = [tensor('float32', outShape), tensor('float32', outShape)] as const;
```
which is less readable imo.

apps/computer-vision/app/classification/index.tsx
msluszniak: Yes, I would change it since demo app should have as low bar for the new users as possible. Knowing what is backend, which one is platform-specific etc. is too high abstraction imo.

packages/react-native-executorch/src/extensions/cv/tasks/preprocessing.ts
barhanc: Do you mean to change the behaviour inside the `matchShape` function or in this code snippet like
```ts
  const shapes = [
    [1, 3, 'H', 'W'],
    [3, 'H', 'W'],
  ] as const;
  if (!matchShape(outputShape, ...shapes))
     throw new Error(...)
```

packages/react-native-executorch/cpp/extensions/math/operations.cpp
msluszniak: Ok, now it makes sense. I think a comment here, even very short like: "Support for negative indices like in NumPy" or similar would be fine. Btw, do we have this behaviour documented? And generally, do we want to push documentation on every feature or create one, aggregated PR for docs?

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
msluszniak: Maybe there is an eslint exclusion for linting the next line, will be clearer than empty comment imo. This is unexpected for reader of the code and the first though, even if it's shorter and sleeker, is that this is some artefact after old comment.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
barhanc: On the TS side it is only documented as such
```ts
 // @param axis The dimension along which softmax is computed. Defaults to -1
 // (last dimension).
 ```
 I believe the JSDocs we should add on every feature but the actual docs under `docs` can be moved to an aggregated PR.

File: packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp (hunk 13 above)
Comment (msluszniak) [reply packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp:107]: Hmm, abbreviated java script in camel case is Js, not JS (or at least claude told me so ;p) so installJsMethod would be perfectly readable imo and will keep convention without any unexpected exceptions. So I'm in favour of changing to camelCase only.

File: packages/react-native-executorch/src/extensions/cv/tasks/classification.ts (hunk 15 above)
Comment (barhanc) [reply packages/react-native-executorch/src/extensions/cv/tasks/classification.ts]: I could do
```ts
    // prettier-ignore
    const probas = tLogits
      .through(softmax, tProbas)
      .getData(new Float32Array(tProbas.numel));
```
but I'm not sure if that's cleaner.

packages/react-native-executorch/src/extensions/cv/tasks/preprocessing.ts
msluszniak: Yeah, like in the snippet.

packages/react-native-executorch/src/extensions/cv/tasks/classification.ts
msluszniak: In my opinion it is, we can ask other team mates for opinion about this one ;). Here you explicitly say, you don't want this part to be formatted. In the first case, you implicitly doing the same by adding empty comment to effectively omit formatting.

barhanc: Yeah you're right, I will change it to be explicit about the intention.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Ok, agreed. What about `case` statements though?

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Done

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
msluszniak: Switch statement is ok without braces, because you can do fall-through from one step to another and this does not make scoping of variables so they are visible in other cases. So this is the biggest change between block of if vs switch statement.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: I renamed only the local variables for now. I actually think the mixed case for `install_` functions is better, e.g. for case sensitive searches over the codebase to find cpp implementation of TS-exposed function because it directly has that TS function's name. We can change it in the future in a small-scoped PR, so that we don't bloat this one.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Done. I changed the `int` type in every `->shape` to explicitly `int32_t`. It must be signed `int32_t` because ExecuTorch uses that underneath and we want the ranges to match.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: I think these are so short, that it's not really useful to put them in separate functions.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
barhanc: Let's maybe do that in a separate smaller-scoped PR so that we don't bloat this one.  I will open an issue for that.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
barhanc: I don't think it would work because we don't iterate over contiguous elements but jump every `inner` number of elements.

packages/react-native-executorch/cpp/extensions/math/operations.cpp
barhanc: Done

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
msluszniak: How about this: extract just the geometry into a small helper and keep the mode-specific cv calls inline — the branches then read at a glance and the rounding/centering becomes unit-testable:

```cpp
namespace {
struct FitBox {
    int32_t w, h, offX, offY;
};

// inner=true (letterbox): fit inside dst, center the padding
// inner=false (crop):     cover dst, center the crop
FitBox computeFit(int32_t srcW, int32_t srcH, int32_t dstW, int32_t dstH, bool inner) {
    double scaleW = static_cast<double>(dstW) / srcW;
    double scaleH = static_cast<double>(dstH) / srcH;
    double scale = inner ? std::min(scaleW, scaleH) : std::max(scaleW, scaleH);

    int32_t w = static_cast<int32_t>(std::round(srcW * scale));
    int32_t h = static_cast<int32_t>(std::round(srcH * scale));
    int32_t sign = inner ? 1 : -1; // letterbox centers padding, crop centers the crop
    return {w, h, sign * (dstW - w) / 2, sign * (dstH - h) / 2};
}
} // namespace
```

Then e.g. letterbox becomes:

```cpp
auto [new_w, new_h, off_x, off_y] = computeFit(src_w, src_h, dst_w, dst_h, /*inner=*/true);
dst_mat.setTo(::cv::Scalar::all(padValue));
::cv::Mat roi = dst_mat(::cv::Rect(off_x, off_y, new_w, new_h));
::cv::resize(src_mat, roi, roi.size(), 0, 0, interp_flag);
```

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
msluszniak: Ok, sure thing.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
msluszniak: Don't we want to expand this to the other places?

packages/react-native-executorch/cpp/extensions/math/operations.cpp
msluszniak: Oh, I missed that, agreed.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Do you mean like in other files, because in `image_ops.cpp` I changed it everywhere not only in the marked snippet?

packages/react-native-executorch/src/utils.ts
barhanc: Yes, good catch.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
msluszniak: Yeap, I mean other files.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
barhanc: Added an ad hoc fix that appends timestamp + random number to tmp file, so that the corrupted file issue is fixed.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Refactored

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: Fixed also in `operations.cpp` where applicable. Didn't find the problem anywhere else.

apps/computer-vision/app/classification/index.tsx
barhanc: Fixed

barhanc:
Regarding example apps visual aspects, we can open an issue regarding the general overhaul of example apps, but I don't think it is a priority right now.

msluszniak:
> Regarding example apps visual aspects, we can open an issue regarding the general overhaul of example apps, but I don't think it is a priority right now.

Agreed. Let me do another quick pass over the code ;)

packages/react-native-executorch/cpp/extensions/math/operations.cpp
```diff
@@ -0,0 +1,306 @@
...
+            throw jsi::JSError(rt, "softmax: axis dimension must be greater than zero");
+        }
+
+        size_t outer = 1;
+        for (size_t i = 0; i < axis; ++i) {
+            outer *= static_cast<size_t>(src->shape_[i]);
+        }
+
+        size_t inner = 1;
+        for (size_t i = axis + 1; i < rank; ++i) {
```
msluszniak: The `size_t` refactor left the bounds signed, so `i < axis` / `i < rank` are now signed/unsigned comparisons (`-Wsign-compare`). Since we're on C++20, `std::cmp_less` from `<utility>` is the clean fix:

```cpp
for (size_t i = 0; std::cmp_less(i, axis); ++i) { ... }
for (size_t i = axis + 1; std::cmp_less(i, rank); ++i) { ... }
```

Alternatively make `rank`/normalized `axis` `size_t` after the range check. Same applies to the argmax loops (L271 / L274).

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -0,0 +1,616 @@
...
+
+        ::cv::Mat srcMat(srcH, srcW, cvType, src->data_.get());
+        std::vector<::cv::Mat> channels;
+        ::cv::split(srcMat, channels);
+
+        int32_t hw = srcH * srcW;
+        size_t elemSize = rnexecutorch::core::types::elementSize(src->dtype_);
+        uint8_t *dstPtr = dst->data_.get();
+
+        for (size_t i = 0; i < srcC; ++i) {
```
msluszniak: Same mixed-sign comparison here — `size_t i < int32_t srcC`. `std::cmp_less(i, srcC)` (with `#include <utility>`) keeps the `size_t` counter without the `-Wsign-compare` warning:

```cpp
for (size_t i = 0; std::cmp_less(i, srcC); ++i) { ... }
```

Same in `toChannelsLast` (L482).

packages/react-native-executorch/src/utils.ts
```diff
@@ -1,5 +1,66 @@
...
+    const methods: { name: string; meta: ModelMethodMeta }[] = [];
+    for (const method of methodNames) {
+      const meta = model.getMethodMeta(method);
+      methods.push({ name: method, meta });
+    }
+
+    return { source, methods };
+  } catch (e) {
+    throw e;
+  } finally {
```
msluszniak: Now that disposal is in `finally`, this `catch (e) { throw e; }` is a redundant rethrow — `try { … } finally { … }` alone behaves identically and clears the `no-useless-catch` lint.

```suggestion
  } finally {
```

apps/computer-vision/components/ModelPicker.tsx
```diff
@@ -0,0 +1,98 @@
...
+          const isDisabled = option.disabled;
+          return (
+            <TouchableOpacity
+              key={index}
+              style={[
+                styles.chip,
+                isSelected && styles.chipActive,
+                isDisabled && styles.chipDisabled,
+              ]}
+              onPress={() => !isDisabled && onValueChange(option.value)}
```
msluszniak: `disabled={isDisabled}` already blocks the press, so the `!isDisabled &&` guard here is redundant:

```suggestion
              onPress={() => onValueChange(option.value)}
```

msluszniak (approved):

PR #1266, opened 2026-06-18
feat: add MLX builds for LFM2.5 and privacy-filter models

NorbertKlockiewicz:
## Description

Adds MLX (Apple GPU) builds for the LFM2.5 family and the privacy-filter
models, wiring the HF-hosted MLX variants into the model registry so they
default to MLX on iOS (with the existing XNNPACK builds as the
simulator / Android fallback):

- LFM2.5 text — 350M, 1.2B
- LFM2.5-VL — 450M, 1.6B
- Privacy filter — openai, nemotron

Runner changes needed to drive the new builds:

- `vision_encoder` now reads its declared input dtype from method metadata
  and converts the preprocessed fp32 pixels to match (fp32 passthrough,
  bf16, or fp16) instead of hardcoding `Float`. Required for bf16 MLX
  vision encoders; the fp32 path stays zero-copy.
- The multimodal prefiller image-embed splice now handles `fp32 <-> bf16`
  vision/text-embed dtype pairs (a hybrid where the vision encoder is fp32
  and the decoder embeds are bf16), in addition to the existing
  fp32<->fp16 cases.
- New `convert_from_float` helper in `runner/util.h`.

Review order: `modelUrls.ts` + `modelRegistry.ts` (the wiring) first, then
the three `common/runner` files (the dtype handling).

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS

### Testing instructions

1. On a physical iOS device, load each model via its accessor with the
   default backend (resolves to MLX on iOS), e.g.
   `models.llm.lfm2_5_vl_450m()`, `models.privacy_filter.openai()`.
2. Verify the privacy filters classify PII and the LFM2.5 / LFM2.5-VL
   models generate correctly (the VL models with an image).
3. MLX requires a physical device; on the simulator the accessors fall
   back to XNNPACK (or pass `{ backend: 'xnnpack' }` explicitly).

### Checklist

### DIFF
--- a/packages/react-native-executorch/common/runner/util.h
+++ b/packages/react-native-executorch/common/runner/util.h
@@ -171,6 +171,30 @@ convert_to_bfloat16(const ::executorch::extension::TensorPtr &src_tensor) {
 }
 
 /**
+ * Helper function to convert a float tensor to float16 (Half).
+ * Creates a new tensor with Half dtype and copies/converts the data.
+ */
+inline ::executorch::runtime::Result<::executorch::extension::TensorPtr>
+convert_to_float16(const ::executorch::extension::TensorPtr &src_tensor) {
+  ET_CHECK_OR_RETURN_ERROR(
+      src_tensor->scalar_type() == ::executorch::aten::ScalarType::Float,
+      InvalidArgument,
+      "Float16 conversion only supported from Float source data");
+
+  const auto num_elements = static_cast<size_t>(src_tensor->numel());
+  const float *float_data = src_tensor->const_data_ptr<float>();
+
+  auto half_tensor = ::executorch::extension::empty_like(
+      src_tensor, ::executorch::aten::ScalarType::Half);
+  auto *half_data = half_tensor->mutable_data_ptr<::executorch::aten::Half>();
+  for (size_t i = 0; i < num_elements; ++i) {
+    half_data[i] = ::executorch::aten::Half(float_data[i]);
+  }
+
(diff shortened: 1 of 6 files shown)


packages/react-native-executorch/common/runner/util.h
```diff
@@ -170,6 +170,38 @@ convert_to_bfloat16(const ::executorch::extension::TensorPtr &src_tensor) {
...
+  ET_CHECK_OR_RETURN_ERROR(dtype == ScalarType::Half, InvalidArgument,
+                           "Unsupported target dtype %hhd for pixel conversion",
+                           static_cast<int8_t>(dtype));
+  const auto num_elements = static_cast<size_t>(src_tensor->numel());
+  const float *float_data = src_tensor->const_data_ptr<float>();
+  auto half_tensor =
+      ::executorch::extension::empty_like(src_tensor, ScalarType::Half);
+  auto *half_data = half_tensor->mutable_data_ptr<::executorch::aten::Half>();
+  for (size_t i = 0; i < num_elements; ++i) {
+    half_data[i] = ::executorch::aten::Half(float_data[i]);
```
msluszniak: If we handle here only float16 (float, bf16 are handled earlier, doubles aren't handled here I guess), we can abstract this into separate function `convert_to_float16` and reject other dtypes.

msluszniak:
When running privacy filter on iOS, I got this error:
image

mkopcins (approved):

PR #1269, opened 2026-06-19
feat: LFM2.5 text-embedding & ColBERT (MLX/XNNPACK) with prompts and multi-vector output

NorbertKlockiewicz:
## Description

Adds two LFM2.5 retrieval models from Liquid AI and the API needed to use them, through the **existing** `useTextEmbeddings` hook — one native runner, one hook, no new public surface beyond optional model-config fields:

- **LFM2.5-Embedding-350M** — dense bi-encoder (CLS pooling, dim 1024). Trained with asymmetric `query: `/`document: ` prompts.
- **LFM2.5-ColBERT-350M** — late-interaction retriever (`Linear(1024→128)` per token). Trained with `[Q] `/`[D] ` prompts.

Both run on MLX on iOS (physical device) and XNNPACK on Android, quantized (MLX int4, XNNPACK 8da4w).

To support them without breaking the existing API, the model config grew three optional fields and `forward` became config-driven:

- `prompts` — when present, `forward` requires a `role` (`'query' | 'document'`) and auto-prepends the matching prompt.
- `multiVector` — when `true`, `forward` returns a per-token `EmbeddingResult` (`vectors`, `numTokens`, `embeddingDim`, `tokenIds`); otherwise it returns a single pooled `Float32Array` as before.
- `skipListIds` — punctuation token ids the consumer excludes from MaxSim scoring.

The library auto-applies the role prompts (the matching `query: `/`[Q] ` prefix is prepended in `forward`), but late-interaction scoring (MaxSim) stays the consumer's concern — it runs wherever the vectors are stored. The example app demonstrates one way to score (its own local `maxSim`), and the ColBERT demo is folded into the unified text-embeddings screen, picking the scorer from the model's config.

Native side: `TextEmbeddings::generate` returns the raw `[numTokens, embeddingDim]` matrix as an `EmbeddingResult`; the TS layer reduces it. The empty `BaseEmbeddings` base class was removed (`TextEmbeddings` now extends `BaseModel` directly), and output-shape validation was extracted into `TextEmbeddings::buildResult`.

**Review order:** start with the TS types (`types/textEmbeddings.ts` — `ForwardFn`/`ForwardReturn` discriminated on the model config), then the module/hook (`TextEmbeddingsModule.ts`, `useTextEmbeddings.ts`), then the native `TextEmbeddings.cpp`/`Types.h`, then the registry/URLs and the example screen.

- [ ] Yes
- [x] No

`forward` stays non-breaking: pooled models still return `Float32Array`. The new return type and `role` requirement only apply to models that opt in via config.

### Type of change

- [x] Bug fix (latent: existing models now add CLS/SEP special tokens — see Additional notes)

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. Open the `text-embeddings` example app.
2. Pick **LFM2.5 Embedding** (MLX on a physical iOS device, XNNPACK on Android/simulator) and run the example queries — weather → "sunny", match → home-team sentences should rank top.
3. Pick **LFM2.5 ColBERT (late-interaction)** — same corpus, scored with MaxSim; ordering should match.
4. Existing pooled models (MiniLM, MPNet, …) keep working unchanged.

C++ unit tests: `TextEmbeddingsTests` (incl. new `EmbeddingResult` metadata / `tokenIds` assertions) compiles and links under the Android NDK toolchain. The suite is cross-compiled, so it is not executed on the host in this setup.

### Checklist

### Additional notes

MLX requires a physical iOS device — the MLX delegate does not run on the simulator (use XNNPACK there). The two models are hosted on the Software Mansion Hugging Face org; docs are updated for both `next` and the `0.9.x` versioned set.


apps/text-embeddings/app/text-embeddings/index.tsx
```diff
@@ -323,11 +345,68 @@ const styles = StyleSheet.create({
...
+  resultText: {
+    flex: 1,
+    fontSize: 14,
     color: '#334155',
   },
+  resultScore: {
+    fontSize: 14,
+    fontWeight: '600',
+    color: '#0F172A',
+    fontVariant: ['tabular-nums'],
```
msluszniak, line 391: add `nums` to cspell or exclude demo apps from cspell.

docs/versioned_docs/version-0.9.x/03-hooks/01-natural-language-processing/useTextEmbeddings.md
```diff
@@ -120,3 +142,53 @@ function App() {
...
+      const dOff = di * dim;
+      let dot = 0;
+      for (let k = 0; k < dim; k++) {
+        dot += query.vectors[qOff + k] * doc.vectors[dOff + k];
+      }
+      if (dot > best) best = dot;
+    }
+    if (best !== -Infinity) score += best;
+  }
+  return score;
```
msluszniak: I saw you used exactly the same function in demo app. Don't we want to expose it as a helper?

docs/versioned_docs/version-0.9.x/04-typescript-api/01-natural-language-processing/TextEmbeddingsModule.md
msluszniak, line 1: I know that we need to be reactive and upload the newest models to our lib to be available immediately, but I don't like the mechanism that it goes via patches. We should think of something better for the future to not edit released documentation etc.

docs/versioned_docs/version-0.9.x/06-api-reference/classes/TextEmbeddingsModule.md
msluszniak, line 1: Why did you publish api reference?

NorbertKlockiewicz: Is the api reference generated for a versions different than "next" too? I thought it's not thus why I published it.

docs/versioned_docs/version-0.9.x/04-typescript-api/01-natural-language-processing/TextEmbeddingsModule.md
NorbertKlockiewicz: I agree, it's especially hard when there's the need to update docs. Users on versions before the patch will look at v0.9.0 docs and it will differ from the things that library ships for them. I don't really have an idea for that so here we are to make a decision if we ship it in the patch or defer to 0.10.0.

docs/versioned_docs/version-0.9.x/03-hooks/01-natural-language-processing/useTextEmbeddings.md
NorbertKlockiewicz: It probably won't hurt to expose it as a util, right now I did it the same way a dotProduct function is done to be consistent, but we can expose those two as a helper for text embeddings.

packages/react-native-executorch/src/utils/textEmbeddings.ts
```diff
@@ -17,3 +18,28 @@ export const dotProduct = (a: Float32Array, b: Float32Array) => {
...
+      let dot = 0;
+      for (let k = 0; k < dim; k++) {
+        dot += (query.vectors[qOff + k] ?? 0) * (doc.vectors[dOff + k] ?? 0);
+      }
+      if (dot > best) best = dot;
+    }
+    if (best !== -Infinity) score += best;
+  }
+  return score;
+};
```
mkopcins, line 43: wouldnt it make sense to have it in the lib?

docs/docs/03-hooks/01-natural-language-processing/useTextEmbeddings.md
```diff
@@ -60,7 +66,21 @@ You need more details? Check the following resources:
 
 ## Running the model
 
-To run the model, you can use the [`forward`](../../06-api-reference/interfaces/TextEmbeddingsType.md#forward) method. It accepts one argument, which is a string representing the text you want to embed. The function returns a promise, which can resolve either to an error or an array of numbers representing the embedding.
+To run the model, you can use the [`forward`](../../06-api-reference/interfaces/TextEmbeddingsType.md#forward) method. It accepts the text to embed and, for models trained with asymmetric prompts, an optional `role`. The return type depends on the model:
+
+- **Pooled models** (the default, e.g. MiniLM, MPNet, LFM2.5-Embedding) resolve to a single `Float32Array` — one normalized vector for the whole input.
+- **Multi-vector models** (`multiVector: true`, e.g. LFM2.5-ColBERT) resolve to an [`EmbeddingResult`](../../06-api-reference/interfaces/EmbeddingResult.md) with the per-token vectors (`vectors`, `numTokens`, `embeddingDim`, `tokenIds`).
```
mkopcins, line 72: maybe we could add a link to something explaining it what is the difference? maybe to liquid blog

docs/docs/03-hooks/01-natural-language-processing/useTextEmbeddings.md
```diff
@@ -120,3 +142,53 @@ function App() {
...
+      let dot = 0;
+      for (let k = 0; k < dim; k++) {
+        dot += query.vectors[qOff + k] * doc.vectors[dOff + k];
+      }
+      if (dot > best) best = dot;
+    }
+    if (best !== -Infinity) score += best;
+  }
+  return score;
+};
```
mkopcins: If we decide to move it inside our lib this is to be removed

docs/versioned_docs/version-0.9.x/04-typescript-api/01-natural-language-processing/TextEmbeddingsModule.md
mkopcins: are docs the problem or the entire flow? I dont think there is an option to avoid patches if we need to make code changes alongside. Waiting and batching models in a regular update makes it unnecessarily slow

packages/react-native-executorch/src/constants/modelRegistry.ts
```diff
@@ -260,6 +260,61 @@ const GEMMA4_E2B_MM_VARIANTS = {
...
+  },
+};
+
+const LFM_COLBERT_SKIP_LIST = [
+  510, 511, 512, 513, 514, 515, 516, 517, 518, 519, 520, 521, 522, 523, 524,
+  535, 536, 537, 538, 539, 540, 541, 568, 569, 570, 571, 572, 573, 600, 601,
+  602, 603,
+];
+
+const LFM_COLBERT_PROMPTS = { query: '[Q] ', document: '[D] ' };
```
mkopcins: I dont think models specific things should be here, move them outside like it is done for tts or ocr

mkopcins (changes requested):

docs/versioned_docs/version-0.9.x/06-api-reference/classes/TextEmbeddingsModule.md
msluszniak: Oh, right.

docs/versioned_docs/version-0.9.x/04-typescript-api/01-natural-language-processing/TextEmbeddingsModule.md
msluszniak: I would make them pre-releases of the `next` version, not the patches of the already released version. Till release of v.0.10, we can keep this flow, but after that, we should release pre-0.11 or so, do not update versioned docs and make it smoother.

mkopcins (approved):

PR #1273, opened 2026-06-22
[RNE Rewrite] chore: clean up and unify example app layout, styling, and fix dependencies

barhanc:
## Description

  - Refactors the computer vision example app and standardizes layout, utilties for loading images, etc. 
  - Adds reusable UI components extracted to ensure visual consistency across all demo screens. 
  - Configures package JSON/lockfile alignment and integrates the  `react-native-worklets`  patch.

NOTE: This is not a visual app overhaul PR, rather it fixes some stuff we forgot about (i.e. the worklets patch) and establishes common primitives for building example screens in fast and efficient manner.

  - [ ] Yes
  - [x] No

### Type of change

### Tested on

  - [x] iOS
  - [x] Android

### Testing instructions

  - [ ] Build and run the  computer-vision  example app on an iOS or Android device/simulator.
  - [ ] Open the Classification and Model Inspector screens to verify the UI layout remains aligned and responsive.

### Screenshots
  N/A

### Related issues
  N/A

### Checklist

### Additional notes

  N/A


apps/computer-vision/app/classification/index.tsx
```diff
@@ -66,226 +46,120 @@ export default function ClassificationScreen() {
   const handlePickImage = async (useCamera: boolean) => {
     const asset = await getImage(useCamera);
     if (asset?.uri) {
-      setImageUri(asset.uri);
+      const img = await loadSkImage(asset.uri);
```
msluszniak: `loadSkImage` can reject (manipulate/decode failure, or the explicit `throw`), but `handlePickImage` has no try/catch — the error escapes uncaught instead of reaching the `setError` UI. Previously decoding happened inside `runClassification`'s try/catch. Wrap this in try/catch and `setError`.

apps/computer-vision/app/classification/index.tsx
```diff
@@ -66,226 +46,120 @@ export default function ClassificationScreen() {
...
+      setSkiaImage(img);
```
msluszniak: `skiaImage` is a native `SkImage`, but the previous one is never `dispose()`d when replaced here (nor on unmount), leaking decoded bitmaps on repeated picks.

apps/computer-vision/app/classification/index.tsx
```diff
@@ -66,226 +46,120 @@ export default function ClassificationScreen() {
...
   const runClassification = async (sync: boolean) => {
-    if (!imageUri || !classify || !classifyWorklet) return;
-    if (!sync) setLoading(true);
+    if (!skiaImage || !classify || !classifyWorklet) return;
+    if (!sync) setIsProcessing(true);
     setError(null);
     try {
-      const inputBuffer = await loadImageBuffer(imageUri);
+      const buffer = {
+        data: skiaImage.readPixels() as Uint8Array,
```
msluszniak: `readPixels()` is typed `Uint8Array | null`; the cast hides a possible `null` reaching the sync `classifyWorklet` path, which bypasses the try/catch. Pre-existing, but worth a guard.

apps/computer-vision/utils.ts
```diff
@@ -1,39 +1,54 @@
...
 
-    return output.assets[0];
-  } catch (err) {
-    console.error(err);
+  return pickerResult.assets[0];
+};
+
+export const loadSkImage = async (uri: string, targetWidth = 800): Promise<SkImage> => {
+  const result = await ImageManipulator.manipulate(uri)
+    .resize({ width: targetWidth })
```
msluszniak: `.resize({ width: 800 })` upscales images narrower than 800px, feeding the classifier interpolated pixels. Consider clamping to the source width.

patches/react-native-worklets+0.9.1.patch
```diff
@@ -0,0 +1,748 @@
...
++  if (typeName_.empty()) {
++    return arrayBuffer;
++  }
++
++  auto constructor = rt.global().getPropertyAsFunction(rt, typeName_.c_str());
++  return constructor.callAsConstructor(rt, arrayBuffer, static_cast<double>(byteOffset_), static_cast<double>(length_));
+ }
+ 
+ SerializableObject::SerializableObject(jsi::Runtime &rt, const jsi::Object &object)
+diff --git a/node_modules/react-native-worklets/Common/cpp/worklets/SharedItems/Serializable.cpp.orig b/node_modules/react-native-worklets/Common/cpp/worklets/SharedItems/Serializable.cpp.orig
```
msluszniak: This `Serializable.cpp.orig` block (~433 lines) is an accidental backup captured during patch generation — harmless to the build (`.orig` isn't globbed) but it pollutes node_modules and is ~58% of the patch. Regenerate the patch without it.

msluszniak (commented):
Ran the computer-vision example app on a physical Android device (Samsung S26 Ultra) — it builds and runs correctly. A few findings inline.

patches/react-native-worklets+0.9.1.patch
barhanc: This is the patch copied 1:1 from PoC, are we sure these lines are not needed?

msluszniak: I asked exactly the same question an agent instance and it implied:

```
Decisive proof is in the git blob hashes. Here's the full chain of evidence:

1. The .orig is a brand-new file the patch creates (patches/...+0.9.1.patch:42-47):

new file mode 100644
index 0000000..3114a09
--- /dev/null
+++ b/node_modules/.../Serializable.cpp.orig
@@ -0,0 +1,433 @@      ← all 433 lines are pure additions
2. The blob hashes prove it's a verbatim copy of the original Serializable.cpp:

The real Serializable.cpp edit header reads index 3114a09..26cc787 → original blob 3114a09, patched blob 26cc787.
The .orig file is created as index 0000000..3114a09 → its content hashes to 3114a09.
Same hash 3114a09 = the .orig is byte-for-byte the pre-patch Serializable.cpp.

3. The naming + co-location confirm the mechanism: .orig is the conventional backup suffix that patch/git/merge tools drop next to a file before editing it. So what happened: whoever generated this patch ran the editor/patch tool (creating Serializable.cpp.orig as a backup), then ran npx patch-package react-native-worklets, which diffed the whole node_modules subtree and swept the leftover .orig into the patch alongside the real one-function change to Serializable.cpp (lines 160-172, adding the typed-array reconstruction).

So it's not a file the patch needs — it's a duplicate of the original that got captured accidentally. Harmless to the build (.orig isn't globbed by the CMake/podspec *.cpp patterns), but it's 433 of the patch's 748 lines and gets written into every node_modules copy on install. The fix is to delete the stray .orig and re-run patch-package to regenerate cleanly.
```

So it seems that when I initially added it in PoC, it has unused bloat in it.

msluszniak: Also, I check how many lines were in my patch, and there were only 309 of them: https://github.com/barhanc/rnet-poc/commit/16b6a63d96c21995a930c8f43111529165d75f77#diff-2c2857a5e4ae60c74ad79dd1de19a037a608c8f7a284cf7ac9ffa7f002499438 so it seems there are more than 400 lines added to this patch.

barhanc: Fixed. Good catch.

msluszniak (approved):

Issue #1253, opened 2026-06-16
[RNE Rewrite] Integrate ResourceFetcher with refactor


barhanc:
The PoC, we are basing this rewrite on, uses a crude, ad hoc solution for downloading and loading model files, see: https://github.com/barhanc/rnet-poc/blob/main/src/hooks/useModelDownload.ts. We should instead use the ResourceFetcher package. The PR referencing this issue should also add something like a `download` function so that it is possible to use `create<Task>` api with pre-exported models defined in `models.ts` as `create<Task>(download(models.<task>.MODEL_CONSTANT))`. 

Here is an example of minimal storage implementation using 'react-native-blob-util' https://github.com/callstackincubator/ai/blob/main/packages/llama/src/storage.ts.

msluszniak:
**Design decision: which fetcher architecture for the rewrite**

Considered four options:

**1. Port the old two-package flow (`expo-resource-fetcher` + `bare-resource-fetcher` + adapter DI)** ❌
- ~1750 LOC across 3 packages, adapter ceremony via `initExecutorch`, two divergent implementations to maintain.
- Its main justification — "Expo users shouldn't need extra native deps" — is moot: RNE ships native code, so Expo Go was never supported; every Expo user is on a dev client where any autolinked dep works.
- `pauseFetching`/`resumeFetching` throws `PlatformNotSupported` on bare Android and nobody complained — evidence the API surface it preserves isn't actually needed.

**2. Keep the current `react-native-fs` hook** ❌
- `@dr.pogodin/react-native-fs` is a maintained fork but with ~13× lower adoption than blob-util (~62k vs ~817k weekly downloads), and it's a filesystem lib where download is a side feature.

**3. Own native downloader (URLSession/OkHttp behind our JSI layer)** ⏸ deferred
- Zero peer deps and best control, but the rewrite's Android module is JSI-install-only, so it's the most plumbing-heavy option (~3–5 days both platforms). Not worth blocking the rewrite on; can be swapped in later behind the same interface.

**4. Single fetcher on `react-native-blob-util`** ✅ **my candidate**
- Purpose-built network→file API (`config({path}).fetch()` streams straight to disk), actively maintained, de facto ecosystem standard — same lib as the callstack example linked above.

**Scope for v1:**
- `download(source, onProgress?): Promise<string>` — promise-first so `create<Task>(download(models.X))` works; hooks wrap it.
[...]

PR #1274, opened 2026-06-22
[RNE Rewrite] feat: add tokenizer pipeline (#1248)

### LINKED ISSUE #1248: [RNE Rewrite] Add tokenizer pipeline implementation
Implement tokenization pipeline (basically expose tokenizers to TS) using the new approach to achieve parity with current implementation

msluszniak:
## Description

Adds the tokenizer pipeline (issue #1248) using the new worklet-based architecture, with functional parity to the current `TokenizerModule`.

A new `nlp` extension exposes a `loadTokenizer` JSI primitive (top-level on `__rnexecutorch_jsi__`, like `loadModel`) returning a `Tokenizer` host object backed by `tokenizers::HFTokenizer`. On top of it sits a `createTokenizer(config, runtime?)` async factory (async + `*Worklet` variants + `dispose`) and a `useTokenizer` hook. Methods: `encode`, `decode`, `getVocabSize`, `idToToken`, `tokenToId` — same semantics as today (special tokens follow the `tokenizer.json` post_processor). The `*Worklet` variants let an upcoming text-embeddings task tokenize → build tensors → run forward within a single worklet.

- C++: `cpp/extensions/nlp/{tokenizer,install}.{h,cpp}`, wired into `RnExecutorch.cpp`.
- TS: `src/extensions/nlp/{ops,tasks}/tokenizer.ts`, `src/hooks/useTokenizer.ts`, exports in `index.ts`, example `models.tokenizer.ALL_MINILM_L6_V2`.
- Build: tokenizer header-search paths added to `android/CMakeLists.txt` and the podspec — `pytorch/tokenizers/include` plus the bundled libs its public headers pull in (`nlohmann/json`, `re2`, and re2's `abseil` dep). Symbols link from the prebuilt `libexecutorch`. Documented in `third-party/README.md`.
- Demo: a dedicated `apps/nlp` example app with a Tokenizer screen that drives the full pipeline on device.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

CI is TypeScript-only here (native isn't compiled in CI); `yarn typecheck`, root `yarn lint`, and `yarn prepare` (bob build) all pass.

To exercise the native tokenizer end-to-end via the demo app:

1. Provision the ExecuTorch `third-party` artifacts into `packages/react-native-executorch/third-party/` (see `third-party/README.md`). The existing PoC bundle already ships the llm/tokenizers extension (headers + symbols), so no rebuild is needed.
2. `yarn && cd apps/nlp && yarn ios` — the iOS simulator works since the tokenizer is pure CPU (no GPU/Metal).
3. Open the **Tokenizer** screen. It loads `all-MiniLM-L6-v2` and auto-runs `encode` / `decode` / `getVocabSize` / `idToToken` / `tokenToId`, asserting: `encode("Hello world")` = `[7592, 2088]`, `decode` round-trips to `"hello world"`, `getVocabSize()` = `30522`, and `tokenToId(idToToken(id))` is the identity. All three assertions should read **PASS** (also logged as `[TokenizerTest]`).

The screen genuinely drives the new code — `useTokenizer` → `createTokenizer` → the `loadTokenizer` JSI primitive → the native `TokenizerHostObject` / `HFTokenizer` — so a green run validates the whole pipeline, not just the types. Verified locally on iOS (iPhone 16 Pro Max, Xcode 26.5): all assertions pass.

The `apps/nlp` app is intentionally minimal and exists only to prove this pipeline; it can be dropped after approval if you'd prefer not to keep a demo app in-tree.

### Related issues

#1248, part of #1208

### Checklist

### Additional notes

The C++ mirrors the current `TokenizerModule`, which is backed by `pytorch/tokenizers` (`tokenizers::HFTokenizer`) from the ExecuTorch llm/tokenizers extension. This PR consumes the same headers (the third-party bundle under `extension/llm/tokenizers`) and prebuilt symbols (from `libexecutorch`); it does not use the `tokenizers-cpp` submodule. Tokenizer download currently uses the temporary `react-native-fs`-based `useResourceDownload` introduced in #1264 (to be replaced by the ResourceFetcher in #1253).


packages/react-native-executorch/android/CMakeLists.txt
```diff
@@ -35,6 +37,13 @@ target_compile_definitions(${CMAKE_PROJECT_NAME} PRIVATE
 target_include_directories(${CMAKE_PROJECT_NAME} PRIVATE
     ../cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/../third-party/include
+    # pytorch/tokenizers headers (and the third-party libs they pull in:
+    # nlohmann/json, re2 and its abseil dependency) ship inside the ExecuTorch
+    # llm extension bundle
```
barhanc: I don't think this comment is necessary.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
```diff
@@ -0,0 +1,265 @@
...
+                if (!val.isNumber()) {
+                    throw jsi::JSError(rt, "decode: Expected tokens[" + std::to_string(i) + "] to be a number");
+                }
+                tokens.push_back(static_cast<uint64_t>(val.asNumber()));
+            }
+
+            auto result = self->tokenizer_->decode(tokens, skipSpecialTokens);
+            if (!result.ok()) {
+                throw jsi::JSError(rt, "decode: Failed to decode tokens: error " +
+                                           std::to_string(static_cast<int32_t>(result.error())));
```
barhanc: Same as in `encode`.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+            if (!self->tokenizer_) {
+                throw jsi::JSError(rt, "idToToken: Tokenizer has been disposed");
+            }
+
+            auto tokenId = static_cast<uint64_t>(args[0].asNumber());
+            auto result = self->tokenizer_->id_to_piece(tokenId);
+            if (!result.ok()) {
+                throw jsi::JSError(rt, "idToToken: Failed to convert id to token: error " +
+                                           std::to_string(static_cast<int32_t>(result.error())));
```
barhanc: Same as in `encode` / `decode`.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+            if (!self->tokenizer_) {
+                throw jsi::JSError(rt, "tokenToId: Tokenizer has been disposed");
+            }
+
+            auto token = args[0].asString(rt).utf8(rt);
+            auto result = self->tokenizer_->piece_to_id(token);
+            if (!result.ok()) {
+                throw jsi::JSError(rt, "tokenToId: Failed to convert token to id: error " +
+                                           std::to_string(static_cast<int32_t>(result.error())));
```
barhanc: Same as in `encode` / `decode`.

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
```diff
@@ -0,0 +1,52 @@
+import { rnexecutorchJsi } from '../../../native/bridge';
+
+declare const tokenizerBrand: unique symbol;
+
+/**
+ * A native HuggingFace-compatible tokenizer instance, backed by a JSI host
+ * object living on the worklet runtime it was loaded on.
+ *
+ * All methods are synchronous and worklet-callable, mirroring the {@link Model}
+ * and {@link Tensor} primitives. For app-level usage prefer the asynchronous
```
barhanc: The comment about mirroring Model/Tensor seems out of place for user facing API.

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
```diff
@@ -0,0 +1,52 @@
...
+ * these calls onto the worklet runtime for you.
+ * @category Types
+ */
+export type Tokenizer = {
+  readonly path: string;
+
+  /** Encodes a string into an array of token ids (special tokens follow the tokenizer.json post_processor). */
+  encode(text: string): number[];
+  /** Decodes an array of token ids back into a string. */
+  decode(tokens: number[], skipSpecialTokens: boolean): string;
```
barhanc: Perhaps `skipSpecialTokens` should have a default value?

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
```diff
@@ -0,0 +1,52 @@
...
+   * @internal
+   */
+  readonly [tokenizerBrand]: never;
+};
+
+/**
+ * Loads a HuggingFace tokenizer from a local `tokenizer.json` file.
+ *
+ * This is a worklet-compatible primitive — it must be invoked on the worklet
+ * runtime (e.g. via {@link wrapAsync}) and returns a {@link Tokenizer} host
```
barhanc: This comment is misleading, the `loadTokenizer` can be just as well called synchronously on the JS thread.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
```diff
@@ -0,0 +1,74 @@
...
+import { wrapAsync } from '../../../core/runtime';
+import { loadTokenizer } from '../ops/tokenizer';
+
+/**
+ * Model configuration required to instantiate a tokenizer task runner.
+ * @category Types
+ */
+export type TokenizerConfig = {
+  readonly tokenizerPath: string;
+};
```
barhanc: Since the config contains only one field, perhaps it can be just inlined into the constructor, unless we plan to add more config options in the future?

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
```diff
@@ -0,0 +1,52 @@
...
+declare const tokenizerBrand: unique symbol;
+
+/**
+ * A native HuggingFace-compatible tokenizer instance, backed by a JSI host
+ * object living on the worklet runtime it was loaded on.
+ *
+ * All methods are synchronous and worklet-callable, mirroring the {@link Model}
+ * and {@link Tensor} primitives. For app-level usage prefer the asynchronous
+ * {@link createTokenizer} factory or the `useTokenizer` hook, which marshal
+ * these calls onto the worklet runtime for you.
```
barhanc: I'm not sure if we want to link/mention higher-level API in lower-level docs, seems slightly out of place and adds coupling (even if only in docs) that we need to change in case something in the implementation changes.

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @category Types
+ */
+export type Tokenizer = {
+  readonly path: string;
+
+  /** Encodes a string into an array of token ids (special tokens follow the tokenizer.json post_processor). */
```
barhanc: These docs should follow standard template with `@param`, `@returns`, etc.

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -32,6 +32,13 @@ Pod::Spec.new do |s|
     "HEADER_SEARCH_PATHS" => [
       "\"$(PODS_TARGET_SRCROOT)/cpp\"",
       "\"$(PODS_TARGET_SRCROOT)/third-party/include\"",
+      # pytorch/tokenizers headers (and the third-party libs they pull in:
+      # nlohmann/json, re2 and its abseil dependency) ship inside the ExecuTorch
+      # llm extension bundle
```
barhanc: Same as in CMakeList.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
```diff
@@ -0,0 +1,74 @@
...
+  const encodeWorklet = (text: string): number[] => {
+    'worklet';
+    return tokenizer.encode(text);
+  };
+
+  const decodeWorklet = (tokens: number[], skipSpecialTokens: boolean = true): string => {
+    'worklet';
+    if (tokens.length === 0) {
+      return '';
+    }
```
barhanc: Can't we put this logic inside C++ code? Then we don't need this boilerplate and can return async function as `wrapAsync(tokenizer.decode,  runtime)`

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const dispose = () => tokenizer.dispose();
+
+  return {
+    encode: wrapAsync(encodeWorklet, runtime),
+    decode: wrapAsync(decodeWorklet, runtime),
+    getVocabSize: wrapAsync(getVocabSizeWorklet, runtime),
+    idToToken: wrapAsync(idToTokenWorklet, runtime),
+    tokenToId: wrapAsync(tokenToIdWorklet, runtime),
```
barhanc: These are extremely lightweight functions and I would consider it an antipattern to call them asynchronously on a background thread, since the cost of passing the arguments through the JS runtime boundary is much greater than executing these functions. They should just be returned as
```ts
getVocabSize: tokenizer.getVocabSize,
idToToken: tokenizer.idToToken,
tokenToId: tokenizer.tokenToId,
```

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const idToTokenWorklet = (id: number): string => {
+    'worklet';
+    return tokenizer.idToToken(id);
+  };
+
+  const tokenToIdWorklet = (token: string): number => {
+    'worklet';
+    return tokenizer.tokenToId(token);
+  };
```
barhanc: This boilerplate with wrapping `Tokenizer` methods in function with `'worklet'` directive is not necessary. You can just return e.g. `tokenToId: tokenizer.tokenToId` and it will work in any workletized function.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const dispose = () => tokenizer.dispose();
```
barhanc: This should be the first function defined in constructor, so it is easy to see that there are no native memory leaks.

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
barhanc, line 1: `ops/` isn't a very good name for this directory. It's imo too generic. For CV it works because it contains operations for common CV transforms. Here in nlp we will have directories like `llm/` for LLM related stuff, `privacy-filter` and so on, so this should imo either be in root `nlp/` as a standalone `tokenizer.ts` or nested in `nlp/tokenizer/tokenizer.ts`.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
```diff
@@ -0,0 +1,74 @@
...
+    encode: wrapAsync(encodeWorklet, runtime),
+    decode: wrapAsync(decodeWorklet, runtime),
+    getVocabSize: wrapAsync(getVocabSizeWorklet, runtime),
+    idToToken: wrapAsync(idToTokenWorklet, runtime),
+    tokenToId: wrapAsync(tokenToIdWorklet, runtime),
+    encodeWorklet,
+    decodeWorklet,
+    getVocabSizeWorklet,
+    idToTokenWorklet,
+    tokenToIdWorklet,
```
barhanc: I don't really see the point of exposing both the async and worklet functions. See my file comment, but if we want to leave this as a task, then imo we should just return `encode`, `decode` as async functions and the rest as sync.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+            if (!self->tokenizer_) {
+                throw jsi::JSError(rt, "encode: Tokenizer has been disposed");
+            }
+
+            auto text = args[0].asString(rt).utf8(rt);
+            auto result = self->tokenizer_->encode(text, kNumAddedBosTokens, kNumAddedEosTokens);
+            if (!result.ok()) {
+                throw jsi::JSError(rt, "encode: Failed to encode input: error " +
+                                           std::to_string(static_cast<int32_t>(result.error())));
```
barhanc: Please use `std::string errorMsg = executorch::runtime::to_string(result.error());` in JS exception string instead of error code.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
barhanc, line 1: The problem I see with this file is that there is really no orchestration logic here, no task per se---it's just a boilerplate wrapper over `ops/tokenizer.ts`. Now, if we want to be 100% functionally backward-compatible with our old API and must expose a *hook* `useTokenizer` then imo this file should just be basically something like this (because this is all we need for hook)
```ts
export async function createTokenizer(config: TokenizerConfig, runtime?: WorkletRuntime) {
  const { tokenizerPath } = config;
  const tokenizer = await wrapAsync(loadTokenizer, runtime)(tokenizerPath);
  const dispose = () => tokenizer.dispose();
  return {
    encode: wrapAsync(tokenizer.encode, runtime),
    decode: wrapAsync(tokenizer.decode, runtime),
    getVocabSize: tokenizer.getVocabSize,
    idToToken: tokenizer.idToToken,
    tokenToId: tokenizer.tokenToId,
    dispose,
  };
}
```
But I'm questioning if this is even needed at all. The `ops/tokenizer.ts` is the abstraction we need for implementing stuff like embeddings, privacy-filter, etc. (the pattern is similar to the `LLMRunner` use in the LLMChat task in PoC). Additionally, the tasks should be like whole pipelines that typical users can just take and do something end-to-end with. It's hard to imagine why would a typical user who only uses the hooks API want a bare tokenizer without anything else, and for power users who create their own pipelines there is still `ops/tokenizer.ts`, it is not internal-only. 

I'm fine with both approaches but if we decide to keep the tokenization as a task, then let's only do it as this very minimal wrapper above, and in end-to-end tasks like embeddings / privacy-filter let's use the `ops/tokenizer.ts` abstraction. Also small nit, the task should be called 'tokenization'.

barhanc (changes requested):
I only went over the library implementation. Tomorrow I will take a look at the example app and test it.

packages/react-native-executorch/cpp/extensions/nlp/install.cpp
```diff
@@ -0,0 +1,8 @@
+#include "install.h"
+#include "tokenizer.h"
+
+namespace rnexecutorch::extensions::nlp {
+void install(facebook::jsi::Runtime &rt, facebook::jsi::Object &module) {
+    tokenizer::install_loadTokenizer(rt, module);
```
barhanc: These should be installed under `nlp` submodule (same pattern as CV)
```cpp
void install(facebook::jsi::Runtime &rt, facebook::jsi::Object &module) {
    jsi::Object nlpModule = jsi::Object(rt);
    // ...
    module.setProperty(rt, "nlp", nlpModule);
}
```

packages/react-native-executorch/src/extensions/nlp/ops/tokenizer.ts
```diff
@@ -0,0 +1,52 @@
...
+ * This is a worklet-compatible primitive — it must be invoked on the worklet
+ * runtime (e.g. via {@link wrapAsync}) and returns a {@link Tokenizer} host
+ * object bound to that runtime.
+ * @category Typescript API
+ * @param tokenizerPath Absolute local path to a `tokenizer.json` file.
+ * @returns The loaded tokenizer.
+ */
+export function loadTokenizer(tokenizerPath: string): Tokenizer {
+  'worklet';
+  return rnexecutorchJsi.loadTokenizer(tokenizerPath) as Tokenizer;
```
barhanc: The `loadTokenizer` method should be under `rnexecutorchJsi.nlp`.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
msluszniak: I couldn't do it directly since tokenizers returns different error enum. I added small mapping to handle this one.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
msluszniak: For now I changed into a field. I we decide it is not sufficient, we can reintroduce the config.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenizer.ts
msluszniak: For now decided to keep it for now, mostly for `useTokenizer`. But I'm totally ok to change it and delete it completely if we eventually decide so.

packages/react-native-executorch/src/extensions/nlp/tasks/tokenization.ts
```diff
@@ -0,0 +1,28 @@
+import type { WorkletRuntime } from 'react-native-worklets';
+
+import { wrapAsync } from '../../../core/runtime';
+import { loadTokenizer } from '../tokenizer';
+
+/**
+ * Loads a tokenizer and exposes its operations with lifetime management, for
+ * use by the `useTokenizer` hook and as a building block for higher-level NLP
+ * tasks (embeddings, privacy filter, ...).
```
barhanc: This comment seems unnecessary and for building end-to-end tasks we should use the `tokenizer.ts` directly not this thin wrapper.

apps/nlp/app/tokenizer/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+    setRunning(true);
+    setRunError(null);
+    setIds(null);
+    setRoundTrip(null);
+    setVocabSize(null);
+    setChecks([]);
+    try {
+      const tokenIds = await encode(text);
+      const decoded = await decode(tokenIds, true);
+      const vocab = await getVocabSize();
```
barhanc: Stale`await` and similarly for `idToToken` and `tokenToId`.

barhanc (approved):
Tested and works good! One more thing I noticed (that isn't blocking on this PR because it will probably become clearer when we add some downstream tasks that use tokenizer) is the serialization cost of `number[]` returned by the async `encode`. I worry it might be quite large for long texts. I guess it's not a problem for tasks like embeddings or privacy-filter since there we would use the `tokenizer.ts` directly and tokens array would never pass the thread boundary since the tokens would then get consumed by some other model and the final returned object might be much lighter. Worth keeping in mind though. For now since the hook is mostly to test the tokenizer implementation it's fine.

msluszniak:
> Tested and works good! One more thing I noticed (that isn't blocking on this PR because it will probably become clearer when we add some downstream tasks that use tokenizer) is the serialization cost of `number[]` returned by the async `encode`. I worry it might be quite large for long texts. I guess it's not a problem for tasks like embeddings or privacy-filter since there we would use the `tokenizer.ts` directly and tokens array would never pass the thread boundary since the tokens would then get consumed by some other model and the final returned object might be much lighter. Worth keeping in mind though. For now since the hook is mostly to test the tokenizer implementation it's fine.

I'll add this point to the `TODO.md`. Maybe you have an idea how this might be fixed so we can quickly prototype this soon?

barhanc:
I guess returning TypedArray instead of `number[]` is an easy fix, since with patch the transfer of ArrayBuffer between threads is fast.

msluszniak:
@barhanc last question, do you want to keep this demo app? I build it mainly for testing purpose. It is not interactive at all, so maybe we shouldn't keep it at all?

barhanc:
For now it's fine. When we have some other screens like embeddings we can remove it.

PR #1275, opened 2026-06-22
[RNE Rewrite] feat: add semantic segmentation task

### LINKED ISSUE #1242: [RNE Rewrite] CV - add semantic segmentation pipeline implementation
Implement semantic segmentation pipeline using the new approach to achieve parity with current implementation. For PoC see: https://github.com/barhanc/rnet-poc/blob/main/src/extensions/cv/tasks/semanticSegmentation.ts.

barhanc:
## Description

- Adds semantic segmentation task and required native operations for it.
- Adds computer-vision example app screen for semantic segmentation.
- Fixes the performance regression in the argmax introduced in #1264. The initial implementation from PoC was more efficient for the default `axis=-1` case as the internal loop was over contiguous elements. The linked PR changed it so that it was more efficient for `axis=0` case, however since the default is `axis=-1` this caused a performance regression in the semantic segmentation task.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Build the computer-vision example app.
- [ ] Test the semantic segmentation screen.

### Related issues

Closes #1242 

### Checklist

### DIFF
--- a/apps/computer-vision/app/segmentation/index.tsx
+++ b/apps/computer-vision/app/segmentation/index.tsx
@@ -120,6 +120,9 @@ function SegmentationContent() {
         alphaType: AlphaType.Premul,
       };
       const nextImage = Skia.Image.MakeImage(info, outData, buffer.width * 4);
+      if (!nextImage) {
+        throw new Error('Failed to create overlay image from output data');
+      }
       setSegmentationImage(nextImage);
     } catch (e: any) {
       setError(e.message || String(e));
--- a/apps/computer-vision/app/segmentation/index.tsx
+++ b/apps/computer-vision/app/segmentation/index.tsx
@@ -1,0 +1,184 @@
+import React, { useState, useRef } from 'react';
+import { View, Text, ScrollView } from 'react-native';
+import { commonStyles } from '../../theme';
+import {
+  Skia,
+  ColorType,
+  AlphaType,
+  useImage,
+  type SkImage as SkiaImageType,
+} from '@shopify/react-native-skia';
+import { useSemanticSegmenter, models } from 'react-native-executorch';
+import ScreenWrapper from '../../components/ScreenWrapper';
+import { ModelPicker, type ModelOption } from '../../components/ModelPicker';
+import { getImage } from '../../utils';
+import { ImageViewport } from '../../components/ImageViewport';
+import { ModelStatus } from '../../components/ModelStatus';
+import { LatencyIndicator } from '../../components/LatencyIndicator';
+import { Button } from '../../components/Button';
+
+const SEGMENTATION_OPTIONS: ModelOption[] = [
+  {
+    label: 'Selfie Seg (FP32)',
+    value: models.semanticSegmentation.SELFIE_SEGMENTATION.XNNPACK_FP32,
+  },
+  {
+    label: 'LRASPP MobileNet V3 (FP32)',
+    value: models.semanticSegmentation.LRASPP_MOBILENET_V3_LARGE.XNNPACK_FP32,
+  },
+  {
+    label: 'LRASPP MobileNet V3 (INT8)',
+    value: models.semanticSegmentation.LRASPP_MOBILENET_V3_LARGE.XNNPACK_INT8,
+  },
+];
+
+function SegmentationContent() {
+  const [selectedModel, setSelectedModel] = useState<any>(SEGMENTATION_OPTIONS[0].value);
+  const [imageUri, setImageUri] = useState<string | null>(null);
+  const [isProcessing, setIsProcessing] = useState(false);
+  const [latency, setLatency] = useState<number | null>(null);
+  const [segmentationImage, setSegmentationImage] = useState<SkiaImageType | null>(null);
+  const [error, setError] = useState<string | null>(null);
+
+  const isProcessingRef = useRef(false);
+
+  const skiaImage = useImage(imageUri, (err) => setError(err.message || String(err)));
+
+  const {
+    isReady,
+    downloadProgress,
+    error: loadError,
+    segment,
+    segmentWorklet,
+  } = useSemanticSegmenter(selectedModel);
+
+  const handlePickImage = async (useCamera: boolean) => {
+    setError(null);
+    try {
+      const uri = await getImage(useCamera);
+      if (uri) {
+        setImageUri(uri);
+        setLatency(null);
+        setSegmentationImage(null);
+      }
+    } catch (e: any) {
+      setError(e.message || String(e));
+    }
+  };
+
+  const runSegmentation = async (sync: boolean) => {
+    if (isProcessingRef.current) return;
+    if (!skiaImage || !segment || !segmentWorklet) return;
+
+    isProcessingRef.current = true;
+    setIsProcessing(true);
+    setLatency(null);
+    setError(null);
+    setSegmentationImage(null);
+
+    try {
+      const pixels = skiaImage.readPixels();
+      if (!pixels) {
+        throw new Error('Failed to read pixels from image');
+      }
+      if (!(pixels instanceof Uint8Array)) {
+        throw new Error('Expected Uint8Array from readPixels');
+      }
+      const buffer = {
+        data: pixels,
+        width: skiaImage.width(),
+        height: skiaImage.height(),
+        format: 'rgba' as const,
+        layout: 'hwc' as const,
+      };
+
+      const start = Date.now();
+      const { buffer: outBuffer } = sync ? segmentWorklet(buffer) : await segment(buffer);
+      setLatency(Date.now() - start);
+
+      const outData = Skia.Data.fromBytes(outBuffer.data);
+      const info = {
+        width: buffer.width,
+        height: buffer.height,
+        colorType: ColorType.RGBA_8888,
+        alphaType: AlphaType.Premul,
+      };
+      const nextImage = Skia.Image.MakeImage(info, outData, buffer.width * 4);
+      setSegmentationImage(nextImage);
+    } catch (e: any) {
+      setError(e.message || String(e));
+    } finally {
+      isProcessingRef.current = false;
+      setIsProcessing(false);
+    }
+  };
+
+  const activeError = loadError ? String(loadError) : error;
+
+  return (
+    <ScrollView
+      style={commonStyles.container}
+      contentContainerStyle={commonStyles.contentContainer}
+    >
+      <Text style={commonStyles.description}>
+        Upload or capture an image to partition it into multiple segments using semantic
+        segmentation.
+      </Text>
+
+      <ModelPicker
+        label="Model"
+        options={SEGMENTATION_OPTIONS}
+        selectedValue={selectedModel}
+        onValueChange={(model) => {
+          setSelectedModel(model);
+          setLatency(null);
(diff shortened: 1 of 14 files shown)


barhanc:
Two models are missing from `models.ts`: DeepLabV3 and FCN, since the forward returns auxiliary tensor. I will reexport them but don't know under what tag should I put them,  is`v0.10.0` ok?

msluszniak:
> Two models are missing from `models.ts`: DeepLabV3 and FCN, since the forward returns auxiliary tensor. I will reexport them but don't know under what tag should I put them, is`v0.10.0` ok?

Yeah, v0.10.0 sounds ok.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -614,4 +614,90 @@ void install_normalize(jsi::Runtime &rt, jsi::Object &module) {
...
+        auto src = args[0].asObject(rt).getHostObject<TensorHostObject>(rt);
+        auto dst = args[1].asObject(rt).getHostObject<TensorHostObject>(rt);
+
+        if (src->dtype_ != rnexecutorch::core::types::DType::int32) {
+            throw jsi::JSError(rt, "applyColormap: src must be int32");
+        }
+        if (dst->dtype_ != rnexecutorch::core::types::DType::uint8) {
+            throw jsi::JSError(rt, "applyColormap: dst must be uint8");
+        }
+        if (dst->numel_ != src->numel_ * 4) {
```
msluszniak: I guess 4 might be replaced here and in other places in this file as `constexpr size_t numChannels = 4;` or so.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -614,4 +614,90 @@ void install_normalize(jsi::Runtime &rt, jsi::Object &module) {
...
+            if (idx < 0 || static_cast<size_t>(idx) >= numColors) {
+                throw jsi::JSError(rt, "applyColormap: tensor contains class index (" +
+                                           std::to_string(idx) + ") that exceeds provided colormap size (" +
+                                           std::to_string(numColors) + ")");
+            }
+
+            dstData[i * 4 + 0] = lut[idx][0];
+            dstData[i * 4 + 1] = lut[idx][1];
+            dstData[i * 4 + 2] = lut[idx][2];
+            dstData[i * 4 + 3] = lut[idx][3];
```
msluszniak: maybe this one when you introduce `numChannels`? Loop will be automatically unrolled on constexpr numChannels. Just add correct formatting.
```suggestion
for (size_t c = 0; c < numChannels; ++c) {
  dstData[i * numChannels + c] = lut[idx][c];
}
```

packages/react-native-executorch/cpp/extensions/math/operations.cpp
```diff
@@ -277,26 +277,19 @@ void install_argmax(jsi::Runtime &rt, jsi::Object &module) {
...
+                float maxVal = -std::numeric_limits<float>::infinity();
+                int32_t maxIdx = 0;
+                for (size_t d = 0; d < axisDim; ++d) {
+                    const float val = srcData[o * axisDim * inner + d * inner + i];
+                    if (val > maxVal) {
+                        maxVal = val;
+                        maxIdx = static_cast<int32_t>(d);
                     }
                 }
+                dstData[o * inner + i] = maxIdx;
```
msluszniak, line 294: Why do we eventually change this one. Didn't it give performance gain?

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -0,0 +1,195 @@
+import type { WorkletRuntime } from 'react-native-worklets';
+
+import { tensor } from '../../../core/tensor';
+import { loadModel } from '../../../core/model';
+import { validateModelSchema, SymbolicTensor } from '../../../core/modelSchema';
+import { wrapAsync } from '../../../core/runtime';
+
+import { type ImageBuffer } from '../image';
```
msluszniak:
```suggestion
import type { ImageBuffer } from '../image';
```

packages/react-native-executorch/cpp/extensions/math/operations.cpp
barhanc: The change from #1264 caused a performance regression for default `axis=-1` case as in the PR description.

msluszniak: Oh, ok.

msluszniak: Maybe we can add some comment in here to not "optimize" it wrongly in the future then?

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -0,0 +1,195 @@
...
+  ): SemanticSegmentationResult<L> => {
+    'worklet';
+    const tInput = preprocessor.process(input);
+    model.execute('forward', [tInput], [tOutput]);
+
+    let returnColormap: ColorMap<L> | undefined;
+    if (nClasses > 1) {
+      if (colormap) {
+        returnColormap = Object.fromEntries(
+          opts.labels.map((l) => [l, colormap[l] ?? [0, 0, 0, 0]])
```
msluszniak, line 180: When a partial `colormap` is passed, unspecified labels fall back to transparent `[0,0,0,0]` instead of the default palette — so overriding one class blanks out all the others. Should this fall back to `defaultColormap[i]`?

apps/computer-vision/app/segmentation/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+      setLatency(Date.now() - start);
+
+      const outData = Skia.Data.fromBytes(outBuffer.data);
+      const info = {
+        width: buffer.width,
+        height: buffer.height,
+        colorType: ColorType.RGBA_8888,
+        alphaType: AlphaType.Premul,
+      };
+      const nextImage = Skia.Image.MakeImage(info, outData, buffer.width * 4);
```
msluszniak, line 122: `Skia.Image.MakeImage` can return `null`; it's passed straight to `setSegmentationImage`, so a failure leaves a blank overlay while latency still reads as success. Worth a null check → `setError`, like the `readPixels` guards above.

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -0,0 +1,195 @@
...
+  const targetW = outShape.at(-1)!;
+
+  // Generate highly distinct, high-contrast colors, see:
+  // https://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
+  const defaultColormap = opts.labels.map((_, i) => {
+    if (i === 0) return [0, 0, 0, 0] as const;
+    return [...hslToRgb((i * 137.5) % 360, 95, 50), 255] as const;
+  });
+
+  if (nClasses > 1 && opts.labels.length !== nClasses) {
```
msluszniak, line 144: This check only runs for `nClasses > 1`, so a binary model (e.g. SELFIE: 2 labels, 1 output channel (?)) never validates `labels` and never uses them for coloring. Intentional, or worth a guard? If we believe that every model will return at least 2 values (e.g. for selfie foreground, background) it might be as it is. However, idk how universal it is.

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -0,0 +1,195 @@
...
+    throw new Error(
+      `Model outputs ${nClasses} classes, but ${opts.labels.length} labels were provided in the configuration.`
+    );
+  }
+
+  const tensors = [
+    tensor('float32', outShape),
+    tensor('float32', [nClasses, targetH, targetW]),
+    tensor('float32', [nClasses, targetH, targetW]),
```
msluszniak, line 153: `tSigmoid` is allocated for the model's lifetime but only used on the binary path (the multiclass path never touches it). Could be allocated per-path to avoid holding an unused `[nClasses,H,W]` buffer.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -614,4 +614,89 @@ void install_normalize(jsi::Runtime &rt, jsi::Object &module) {
...
+            auto color = colorVal.asObject(rt).asArray(rt);
+            if (color.size(rt) != numRgbaChannels) {
+                throw jsi::JSError(rt, "applyColormap: colormap entry must be an RGBA color array of size 4");
+            }
+            for (size_t c = 0; c < numRgbaChannels; ++c) {
+                auto channelVal = color.getValueAtIndex(rt, c);
+                if (!channelVal.isNumber()) {
+                    throw jsi::JSError(rt, "applyColormap: colormap channel value must be a number");
+                }
+                lut[i][c] = static_cast<uint8_t>(channelVal.asNumber());
```
msluszniak: This only guards `isNumber()`; out-of-range or NaN channel values wrap/UB silently through `static_cast<uint8_t>`. Minor since palettes are internal, but a 0–255 range check would be safer.

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
barhanc: This is intentional, if the user passes colormap that only specifies color for e.g. 'cat', then the model only segments cats from the image. Probably worth adding to docs, which reminds me, we didn't add jsdoc for the closure bundle returned by `create<Task>` constructor, I will later open PR that adds them also to other models.

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
barhanc: This is intentional, since for a model like selfie segmentation the raw logit values returned by the model can be used e.g. for soft background removal, that's why we return different things for these cases. Definitely requires jsdoc comment.

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
barhanc: Not worth the additional code logic imo. Right now the tensor allocation code is clean and the possible performance/memory benefits from introducing the logic are negligible.

msluszniak: Ok, understood.

msluszniak:
1. Clicking button `Run Async` causes weird jump when screen is scrolled to the bottom on my android. 
2. Total Pipeline Latency overlaps with android system bar at the very bottom of the app.

barhanc:
> 1. Clicking button `Run Async` causes weird jump when screen is scrolled to the bottom on my android.
> 2. Total Pipeline Latency overlaps with android system bar at the very bottom of the app.

Should be fixed now.

msluszniak (approved):

PR #1283, opened 2026-06-25
[RNE Rewrite] feat: on-demand native lib download and backend splitting

msluszniak:
## Description

Ports PR #1039's on-demand native-lib mechanism into the rewrite flow. Apps declare what they need via a `react-native-executorch` block (`backends` / `libs` / `features`) in `package.json`; the postinstall `scripts/download-libs.js` expands `features` via the forward-looking `FEATURE_MAP`, writes `rne-build-config.json`, and downloads only the requested per-backend artifacts from GitHub Releases into `third-party/`. The podspec and `android/build.gradle.kts` read that config to gate `RNE_ENABLE_*` and link backends as separate libs (`Xnnpack`/`CoreML`/`MLX` xcframeworks, `lib*_executorch_backend.so`); MLX is the iOS device slice only.

Adapted to the rewrite: `cpp/` layout (opencv kept as an extensible source list, guarded in `RnExecutorch.cpp`), Kotlin-DSL gradle config read, single `android/CMakeLists.txt`. The scripts, `FEATURE_MAP`, and artifact contract are kept identical to #1039 so future rewrite work stays code-only. CI (TS-only) skips the download via `RNET_SKIP_DOWNLOAD` in the setup action.

Backed by `software-mansion-labs/executorchrne-split-build` (already ET 1.3.1). Release artifacts (device-only MLX) are built/uploaded separately; committed headers / `executorch.jar` / `ExecutorchLib` wrapper are provisioned alongside.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

1. Config flow: set an app's `react-native-executorch.features` (e.g. `["llm"]`), run `yarn install`, and check `rne-build-config.json` reflects it (`enableOpencv:false, enableCoreml:false, enableXnnpack:true`). Setting legacy `extras` errors.
2. Download flow: `RNET_BASE_URL=https://github.com/software-mansion/react-native-executorch/releases/download/v0.0.0-rewrite-libs-test RNET_TARGET=android-arm64-v8a INIT_CWD=<app> node scripts/download-libs.js` — tarballs extract (checksum-verified) into `third-party/android/libs/...` exactly where `CMakeLists.txt`/podspec expect.
3. CI gates: `yarn lint`, `yarn workspace react-native-executorch prepare`, `yarn typecheck` (all green; install with `RNET_SKIP_DOWNLOAD=1`).

Test artifacts uploaded to the `v0.0.0-rewrite-libs-test` pre-release: Android freshly built from `rne-split-build` (ET 1.3.1, split backends, arm64-v8a + x86_64); iOS reused from the same rne-split-build 1.3.1 lineage (#1039) with MLX stripped to the device slice only. Full download/extract/checksum flow verified across iOS + both Android ABIs. On-device app build still to be run.

### Related issues

#1208

### Checklist

### Additional notes

Draft: all of third-party/ (headers + binaries) is downloaded on demand, not committed — headers ship as a platform-independent headers.tar.gz alongside the binary artifacts (built with device-only MLX), uploaded separately.

Header provenance: `headers.tar.gz` is assembled from the `rne-split-build` ExecuTorch CMake **install include tree** (`cmake-out*/include/`: `executorch`, `pytorch/{c10,torch,headeronly}`, the tokenizer third-party headers `absl`/`re2`/`nlohmann`, and `cpuinfo`/`pthreadpool` — all platform-independent, identical across ABIs/SDKs) plus the `opencv2` headers from the OpenCV prebuilt (same source as the `opencv-rne` pod / Android static `libopencv_*.a`, not built from executorch).

Docs getting-started is deferred (no rewrite getting-started page exists yet).


packages/react-native-executorch/scripts/download-libs.js
barhanc, line 1: Please fix the CI/lint warning `'url' is already declared in the upper scope on line 335 column 19`.

packages/react-native-executorch/cpp/RnExecutorch.cpp
```diff
@@ -1,18 +1,26 @@
 #include "RnExecutorch.h"
 
 #include "core/install.h"
-#include "extensions/cv/install.h"
 #include "extensions/math/install.h"
 #include "extensions/nlp/install.h"
 
+// The cv extension links against OpenCV. When opencv is opted out
+// (RNE_ENABLE_OPENCV unset by the build config), its sources are excluded from
+// compilation, so guard the include + install call too.
```
barhanc: Comment not needed imo, the flag is clear enough.

packages/react-native-executorch/third-party/.gitignore
barhanc, line 1: The `executorch.jar` is not ignored.

packages/react-native-executorch/scripts/package-release-artifacts.sh
barhanc, line 1: Why are these scripts nested inside `packages/react-native-executorch` and not in the monorepo root (like the error gen scripts)?

packages/react-native-executorch/react-native-executorch.podspec
```diff
@@ -10,37 +29,128 @@ Pod::Spec.new do |s|
...
+  physical_ldflags = [
+    '$(inherited)',
+    "\"#{executorch_binaries_path}/libthreadpool_ios.a\"",
+  ]
+  simulator_ldflags = [
+    '$(inherited)',
+    "\"#{executorch_binaries_path}/libthreadpool_simulator.a\"",
+  ]
+
+  xnnpack_xcframework_path = File.expand_path('$(PODS_TARGET_SRCROOT)/third-party/ios/XnnpackBackend.xcframework', __dir__)
```
barhanc: `File.expand_path` on paths containing Xcode environment variables like `$(PODS_TARGET_SRCROOT)` produces malformed path configurations.

packages/react-native-executorch/scripts/download-libs.js
```diff
@@ -0,0 +1,439 @@
...
+
+function download(url, dest) {
+  return new Promise((resolve, reject) => {
+    const file = fs.createWriteStream(dest);
+    const get = (currentUrl) => {
+      const client = currentUrl.startsWith('http://') ? http : https;
+      const headers = {};
+      if (process.env.GITHUB_TOKEN && !currentUrl.startsWith('http://')) {
+        headers.Authorization = `Bearer ${process.env.GITHUB_TOKEN}`;
+      }
```
barhanc:
```suggestion
      if (process.env.GITHUB_TOKEN && currentUrl.includes('github.com')) {
        headers.Authorization = `Bearer ${process.env.GITHUB_TOKEN}`;
      }
```

barhanc:
- Also, did you check the example app size with and without the split and how much it decreases it?
- Would you be able to check that MLX backend gets correctly bundled for the computer-vision app (I don't have an iOS  physical device available)?

---

- Do we have some custom script for vendoring the executorch headers, because the headers downloaded using the split script do not match the ones copied directly from executorch repo used in rnet-poc, e.g. all llm runner helpers are missing?

packages/react-native-executorch/scripts/package-release-artifacts.sh
msluszniak: `download-libs.js` must live in the package: it runs at the package's postinstall, ships in the npm tarball, and resolves `third-party/` relative to itself (unlike the dev-only error-gen scripts at root). `package-release-artifacts.sh` is its build-side pair (produces exactly the layout `download-libs.js` consumes), so it's colocated. Happy to move the release script to the monorepo root if you'd prefer.

msluszniak:
> * Also, did you check the example app size with and without the split and how much it decreases it?
> * Would you be able to check that MLX backend gets correctly bundled for the computer-vision app (I don't have an iOS  physical device available)?
> 
> * Do we have some custom script for vendoring the executorch headers, because the headers downloaded using the split script do not match the ones copied directly from executorch repo used in rnet-poc, e.g. all llm runner helpers are missing?

1. Haven't check yet, will do this in a minute.
2. Unfortunately, neither do I, so we need to defer this check till Monday.
3. That's my fault. There is no such script and that's the reason why I missed headers. I will add such script and update published assets.

msluszniak:
@barhanc I added mentioned script, now working on memory comparison for cv demo app.

msluszniak:
## App-size impact of the backend split

Measured on the `computer-vision` demo (release APK, `arm64-v8a`). iOS figures are artifact-derived (no IPA built yet). Native `.so` are stored uncompressed, so they land in the APK at full size.

### CV app — split vs all-backends

| | vs complete bundle | vs RNE-only files |
|---|---|---|
| **Android** (measured) | 78.3 → 68.3 MB (−10.0 MB, **−13%**) | 25.7 → 15.7 MB (−10.0 MB, **−39%**) |
| **iOS** (estimated) | ~0 | ~0 |

iOS is ~0 here because the CV app uses *every* iOS backend (xnnpack + coreml + **mlx** via `keypointDetection`). Android RNE-only footprint = `libexecutorch` core 11.3 + `libxnnpack…` 2.6 + `libRnExecutorch` (+ statically-linked opencv) 1.9, plus Vulkan 10.0 in the no-split build.

### What the split drops, per backend (the cross-app value)

| Component | Platform | Dropped | Notes |
|---|---|---|---|
| **Vulkan** | Android | **10.0 MB** | dynamic `.so`, full size |
| **MLX** | iOS | **~13 MB `.a` + 2.7 MB `mlx.metallib`** | `-force_load` → not stripped; metallib is a non-strippable resource |
| **CoreML** | iOS | ~1.8 MB `.a` | `-force_load` |
| **XNNPACK** | both | 2.6 / 7 MB | almost always needed |
| **opencv** | both | ~1.5 MB effective | static + dead-stripped, so ≪ its 13 MB of `.a` |
| **phonemis** | both | ~1.5–2.5 MB (once Kokoro TTS lands) | source-compiled into the lib; the 22 MB espeak `data/` is stripped from the npm package |

packages/react-native-executorch/react-native-executorch.podspec
barhanc, line 1: Trying to build for iOS simulator I get
```shell
error: Build input files cannot be found: '/third-party/ios/CoreMLBackend.xcframework/ios-arm64-simulator/libCoreMLBackend.a', '/third-party/ios/XnnpackBackend.xcframework/ios-arm64-simulator/libXnnpackBackend.a'. Did you forget to declare these files as outputs of any script phases or custom
  build rules which produce them? (in target 'computervision' from project 'computervision')
```
but I have these files downloaded using script. It seems that there is some problem with path resolution. Defining paths like this fixed it for me
```podspec
  executorch_binaries_path = "#{__dir__}/third-party/ios/libs/executorch"
  xnnpack_xcframework_path = "#{__dir__}/third-party/ios/XnnpackBackend.xcframework"
  coreml_xcframework_path  = "#{__dir__}/third-party/ios/CoreMLBackend.xcframework"
  mlx_xcframework_path     = "#{__dir__}/third-party/ios/MLXBackend.xcframework"
```

barhanc (changes requested):

packages/react-native-executorch/react-native-executorch.podspec
msluszniak: Good catch, applied. Root cause: these feed `-force_load` in OTHER_LDFLAGS, linked at the app target where `$(PODS_TARGET_SRCROOT)` (pod-scoped) is undefined → expands to empty → `/third-party/…`. Switched the four linker paths to `#{__dir__}/…` (HEADER_SEARCH_PATHS keep the variable — they're the pod's own compile settings). Fixed in 24e497697.

barhanc (approved):

barhanc:
Tested iOS on device and the backends work.

Issue #1271, opened 2026-06-19
[RNE Rewrite] Setup clangd and clang-tidy setup for non-3rd party c++ code


msluszniak:
We need to setup clangd with strong compile flags to make it easier to improve code during the rewrite process. The first step is to research which compile flags we should use. Then setup clangd. As a follow up, we need to setup clang-tidy static checks as a first step to potential clang-tidy CI.

PR #1285, opened 2026-06-25
[RNE Rewrite] chore: set up clangd for C++ sources

msluszniak:
## Description

Sets up a committed [clangd](https://clangd.llvm.org/) config for the core package's C/C++ sources so the rewrite gets code intelligence and a shared, strict warning set, and enforces that set at commit time. First step of #1271; clang-tidy CI is a follow-up (#1286).

- `compile_flags.txt` — include roots / `-std=c++20` / defines, with third-party headers (ExecuTorch/torch/JSI) as `-isystem` so their warnings don't pollute our diagnostics. Relative paths anchored to the package dir, so it is portable with no per-machine generation.
- `.clangd` — strict warning set layered on top (incl. `-Wconversion`/`-Wsign-conversion`), with clangd's include cleaner disabled (ExecuTorch umbrella headers misreport includes).
- `scripts/check-cpp-warnings.sh` + `lefthook.yml` — a `pre-commit` command that compiles staged `cpp/` sources with the same warning set and aborts the commit on any warning, keeping the editor and the commit gate in sync. It skips gracefully (never blocks) when no compiler or the provisioned headers are available; bypass with `git commit --no-verify`.
- `.gitignore` — track the shared config while keeping generated `compile_commands.json` local.
- `CONTRIBUTING.md` — prerequisites (provisioned `third-party/include` + `yarn install`), editor setup, and the pre-commit behavior.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

1. Provision `packages/react-native-executorch/third-party/include` and run `yarn install`.
2. Open a file under `packages/react-native-executorch/cpp` in an editor with the clangd extension (MS C/C++ IntelliSense disabled); confirm the `<executorch/...>`/`<jsi/jsi.h>` includes resolve and the current sources are clean.
3. Confirm the strict flags fire by introducing a deliberate violation, e.g. in `cpp/core/dtype.cpp`:
   ```cpp
   int rne_probe(float f) {
       int casted = (int)f;   // expect -Wold-style-cast + -Wconversion
       int unused = 7;        // expect -Wunused-variable
       return casted;
   }
   ```
   clangd should flag these on our code (third-party headers stay quiet thanks to `-isystem`).
4. `git add` that change and `git commit` — the pre-commit hook aborts with the warning printed. Verified end-to-end via lefthook: warning → `cpp-warnings ❌` (exit 1); clean → `✔️` (exit 0). Revert afterwards.

### Related issues

#1271

### Checklist

### DIFF
--- a/packages/react-native-executorch/scripts/check-cpp-warnings.sh
+++ b/packages/react-native-executorch/scripts/check-cpp-warnings.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+#
+# Pre-commit guard: fail the commit when staged C/C++ sources in the core package
+# produce compiler warnings under the project's clangd warning set — i.e. the
+# include flags in compile_flags.txt plus the -W flags in .clangd. This mirrors
+# what clangd shows in the editor, so a commit can't introduce a warning that the
+# editor would have flagged.
+#
+# It skips gracefully (without blocking the commit) when a C++ compiler or the
+# provisioned ExecuTorch/JSI headers are not available, so contributors who don't
+# build the native code are never blocked. Bypass explicitly with `git commit
+# --no-verify`. Override the compiler with CXX=/path/to/clang++.
+#
+# Usage: check-cpp-warnings.sh <staged file> [<staged file> ...]
+set -uo pipefail
+
+PKG_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+PKG_PREFIX="packages/react-native-executorch/"
+
+# Keep only staged files inside this package's cpp/ tree, relative to PKG_DIR.
+rel_files=()
+for f in "$@"; do
+  case "$f" in
+    "${PKG_PREFIX}"cpp/*) rel_files+=("${f#"$PKG_PREFIX"}") ;;
+    cpp/*)                rel_files+=("$f") ;;
+  esac
+done
+[ "${#rel_files[@]}" -eq 0 ] && exit 0
+
+CXX="${CXX:-clang++}"
+if ! command -v "$CXX" >/dev/null 2>&1; then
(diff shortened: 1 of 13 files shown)


packages/react-native-executorch/scripts/check-cpp-warnings.sh
```diff
@@ -0,0 +1,68 @@
...
+
+if [ ! -d third-party/include/executorch ] ||
+   [ ! -f ../../node_modules/react-native/ReactCommon/jsi/jsi/jsi.h ]; then
+  echo "ℹ C++ warning check skipped: provision third-party/include and run 'yarn install' to enable it."
+  exit 0
+fi
+
+# Compilation database (includes / std / defines) + the .clangd warning set.
+db_flags=()
+while IFS= read -r line; do
```
barhanc: Doesn't this assume that `compile_flags.txt` must have a new empty line at the end?

packages/react-native-executorch/scripts/check-cpp-warnings.sh
barhanc, line 1: We should have additional check that the expected `.clangd` and `compile_flags.txt` files exist.

packages/react-native-executorch/scripts/check-cpp-warnings.sh
msluszniak: Good catch — the read loop now uses `|| [ -n "$line" ]` so the last flag is read even without a trailing newline (commit 927f6151f).

packages/react-native-executorch/scripts/check-cpp-warnings.sh
msluszniak: Added — the script now skips with a notice if `compile_flags.txt` or `.clangd` is missing (commit 927f6151f).

barhanc (approved):

PR #1286, opened 2026-06-25
[RNE Rewrite] chore: add clang-tidy static checks

### LINKED ISSUE #1271: [RNE Rewrite] Setup clangd and clang-tidy setup for non-3rd party c++ code

msluszniak:
## Description

Adds clang-tidy static analysis for the core package's C/C++ sources. Second step of #1271 (clangd setup landed in #1285). Rebased onto `rne-rewrite`, so this is clang-tidy-only.

- `.clang-tidy` — a high-signal check set enabled group by group (one commit each): `bugprone-*`, `performance-*`, `clang-analyzer-*`, `cppcoreguidelines-*`, `google-*`, `llvm-*`, `misc-*`, `modernize-*`, `readability-*`. Each group's narrow exclusion list (raw-buffer pointer math, `reinterpret_cast`, `#pragma once`, JSI multiple-inheritance, opinionated/noisy checks) is documented inline with rationale. Analyzes only this package's own `cpp/` headers; third-party is `-isystem`.
- `scripts/clang-tidy.sh` + the `lint:cpp` package script — run clang-tidy over `cpp/` using `compile_flags.txt` (the same database clangd uses), failing on any finding.
- `.github/workflows/clang-tidy.yml` — CI gate, **dormant** behind the `ENABLE_CLANG_TIDY` repo variable. It provisions headers through the on-demand download flow (`download-libs.js`, #1283) rather than a stub, since clang-tidy needs the ExecuTorch headers to parse the sources.

The fixes each group surfaced are split across the per-group commits (scoped `DType` enum, member-init lists, `explicit` ctor, `const`-correctness cleanup, `use-auto`/`emplace`/`cmp_*` modernizations, `NOLINT`s for the intentional cases).

Depends on #1283 for `download-libs.js`; enable the workflow once that has merged and a release carrying `headers.tar.gz` exists. The provisioning step sets `RNET_HEADERS_ONLY=1` (clang-tidy needs only headers) — see PR thread re: honoring that flag in `download-libs.js`.

Verified locally (Homebrew LLVM clang-tidy): `lint:cpp` is green on all sources and exits non-zero on an injected finding.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

1. Provision `packages/react-native-executorch/third-party/include` and run `yarn install`.
2. From the repo root: `CLANG_TIDY=$(brew --prefix llvm)/bin/clang-tidy yarn workspace react-native-executorch lint:cpp` — expect 0 findings.
3. Confirm the gate has teeth: introduce a bug in a `cpp/` file, e.g. `double r = a / b;` (ints) for `bugprone-integer-division`, and re-run — expect a non-zero exit. Revert afterwards.

### Related issues

#1271

### Checklist

### DIFF
--- a/packages/react-native-executorch/cpp/core/dtype.cpp
+++ b/packages/react-native-executorch/cpp/core/dtype.cpp
@@ -54,8 +54,7 @@ size_t elementSize(DType dtype) {
     switch (dtype) {
     case DType::uint8:
         return 1;
-    // int32 and float32 are both 4 bytes; the identical branches are intentional.
-    // NOLINTNEXTLINE(bugprone-branch-clone)
+    // NOLINTNEXTLINE(bugprone-branch-clone): int32 and float32 are both 4 bytes; the identical branches are intentional.
     case DType::int32:
         return 4;
     case DType::float32:
--- a/packages/react-native-executorch/cpp/core/dtype.cpp
+++ b/packages/react-native-executorch/cpp/core/dtype.cpp
@@ -54,6 +54,8 @@ size_t elementSize(DType dtype) {
     switch (dtype) {
     case DType::uint8:
         return 1;
+    // int32 and float32 are both 4 bytes; the identical branches are intentional.
+    // NOLINTNEXTLINE(bugprone-branch-clone)
     case DType::int32:
         return 4;
     case DType::float32:
--- a/packages/react-native-executorch/cpp/core/tensor.cpp
+++ b/packages/react-native-executorch/cpp/core/tensor.cpp
@@ -8,11 +8,11 @@ namespace jsi = facebook::jsi;
 TensorHostObject::TensorHostObject(const std::vector<std::int32_t> &shape, rnexecutorch::core::types::DType dtype)
     : dtype_(dtype),
       shape_(shape),
-      numel_(std::accumulate(shape.begin(), shape.end(), static_cast<size_t>(1), std::multiplies<size_t>())) {
+      numel_(std::accumulate(shape.begin(), shape.end(), static_cast<size_t>(1), std::multiplies<>())) {
     const auto elemSize = rnexecutorch::core::types::elementSize(dtype);
 
     size_ = numel_ * elemSize;
-    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays) — owning runtime-sized byte buffer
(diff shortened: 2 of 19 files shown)


packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -5,14 +5,14 @@
...
+      shape_(shape),
+      numel_(std::accumulate(shape.begin(), shape.end(), static_cast<size_t>(1), std::multiplies<>())) {
     const auto elemSize = rnexecutorch::core::types::elementSize(dtype);
 
-    dtype_ = dtype;
-    shape_ = shape;
-    numel_ = std::accumulate(shape.begin(), shape.end(), size_t(1), std::multiplies<size_t>());
-
     size_ = numel_ * elemSize;
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays) — owning runtime-sized byte buffer
```
barhanc:
```suggestion
    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays): owning runtime-sized byte buffer
```

packages/react-native-executorch/cpp/core/dtype.cpp
```diff
@@ -54,6 +54,8 @@ size_t elementSize(DType dtype) {
     switch (dtype) {
     case DType::uint8:
         return 1;
+    // int32 and float32 are both 4 bytes; the identical branches are intentional.
+    // NOLINTNEXTLINE(bugprone-branch-clone)
```
barhanc:
```suggestion
    // NOLINTNEXTLINE(bugprone-branch-clone): int32 and float32 are both 4 bytes; the identical branches are intentional.
```
Let's use the convention
```cpp
// NOLINT...(...): explanation
```

barhanc (commented):
One tiny thing I noticed regarding `.clangd` is that it flags `#pragma once` in header files. Maybe we could also fix that in this PR.

msluszniak:
> One tiny thing I noticed regarding `.clangd` is that it flags `#pragma once` in header files. Maybe we could also fix that in this PR.

Ok, here I explicitly turned of suggestion about LLVM style header guards so it's just for clangd.

barhanc (approved):

PR #1287, opened 2026-06-26
[RNE Rewrite] feat: support offset and length options in Tensor.copyTo

barhanc:
## Description

Extends  Tensor.copyTo()  to support optional  offset  and  length  configuration parameters. This allows copying a slice/sub region of a source tensor into a destination tensor, which is useful for tasks such as extracting rows from a batched tensor.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] If you want rebuild example apps and check that nothing is broken.

### Checklist

### DIFF
--- a/packages/react-native-executorch/cpp/core/tensor.cpp
+++ b/packages/react-native-executorch/cpp/core/tensor.cpp
@@ -39,8 +39,8 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) 
     if (nameStr == "copyTo") {
         auto self = shared_from_this();
         auto fnBody = [self](jsi::Runtime &rt, const jsi::Value & /*thisVal*/, const jsi::Value *args, size_t count) -> jsi::Value {
-            if (count != 1) {
-                throw jsi::JSError(rt, "copyTo: Usage: copyTo(dst)");
+            if (count != 1 && count != 2) {
+                throw jsi::JSError(rt, "copyTo: Usage: copyTo(dst, options?)");
             }
 
             if (!args[0].isObject() || !args[0].asObject(rt).isHostObject<TensorHostObject>(rt)) {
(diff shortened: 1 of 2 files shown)


packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -71,11 +71,34 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name)
...
+                }
+            }
+
+            size_t copyLen = self->numel_ - srcOffset;
+
+            if (count == 2 && args[1].isObject()) {
+                auto optsObj = args[1].asObject(rt);
+                if (optsObj.hasProperty(rt, "length")) {
+                    copyLen = static_cast<size_t>(optsObj.getProperty(rt, "length").asNumber());
+                }
```
msluszniak, line 86: Why these are done in two separate ifs?

barhanc: They have to be separated like that so we can have the default behaviour like the one described in docs. If you want I can refactor the condition in `if` into separate boolean like `hasOptions` so it's not duplicated.

msluszniak: Wait, I don't understand why you cannot write it like that:

```cpp
            size_t srcOffset = 0;
            size_t copyLen = self->numel_;
            if (count == 2 && args[1].isObject()) {
                auto optsObj = args[1].asObject(rt);
                if (optsObj.hasProperty(rt, "offset")) {
                    srcOffset = static_cast<size_t>(optsObj.getProperty(rt, "offset").asNumber());
                }
                
                copyLen -= srcOffset;
                
                if (optsObj.hasProperty(rt, "length")) {
                    copyLen = static_cast<size_t>(optsObj.getProperty(rt, "length").asNumber());
                }
            }
```

barhanc: I guess that works as well :) Will fix.

msluszniak (changes requested):
Demo apps with semantic segmentation and style transfer does not work with this change, I got error: `Value is undefined, expected an Object`. Other models, namely object detection, kepoint detection and classification works correctly.

barhanc:
> Demo apps with semantic segmentation and style transfer does not work with this change, I got error: `Value is undefined, expected an Object`. Other models, namely object detection, kepoint detection and classification works correctly.

It works fine for me, did you do a clean rebuild?

msluszniak:
Yeah, stale caches.

msluszniak (approved):

PR #1292, opened 2026-06-30
[RNE Rewrite] Add text and image embeddings pipelines

### LINKED ISSUE #1247: [RNE Rewrite] Add image and text embeddings pipelines
Implement text and image embeddings pipelines using the new approach to achieve parity with current implementation.

msluszniak:
## Description

Adds text and image embeddings pipelines to the new architecture, achieving parity with the old flow. Embeddings are pure-TypeScript tasks (pooling + L2-norm stay baked into the `.pte`): text tokenizes and runs `forward`; image reuses the existing image preprocessor. To run the existing int64-input embedding models unchanged, this adds an `int64`/`Long` tensor dtype to the core (the tensor data path is byte-oriented, so it is a small `dtype.{h,cpp}` + `tensor.ts` change).

Text inputs are fed at their exact token length (no padding). `model.execute` validates dynamically-shaped inputs against the `[min, max, step]` bounds exposed by an optional per-method `get_dynamic_dims_<methodName>` companion; methods without it keep exact per-dimension validation. This fixes scale-sensitive pooling heads (e.g. DistilUSE's tanh projection), which padding otherwise corrupts.

Includes `createTextEmbedder` / `createImageEmbedder` tasks, `useTextEmbedder` / `useImageEmbedder` hooks, `models.textEmbeddings` / `models.imageEmbeddings` registry entries, an interactive text-embeddings demo in `apps/nlp`, and a CLIP zero-shot image-embeddings demo in `apps/computer-vision`.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

- `nlp` app → **Text Embeddings**: seeds a sentence library; type a query and **Find similar** to rank by cosine similarity, switch models via the chips. Verified on a physical Android device (arm64): all-MiniLM-L6-v2 returns 384-dim L2-normalized embeddings (~25 ms/forward on XNNPACK); DistilUSE ranks correctly with a wide similarity spread (previously compressed by padding).
- `computer-vision` app → **Image Embeddings**: pick an image and rank editable text labels via CLIP zero-shot (image vs. text embeddings). Verified on device.

### Related issues

#1247

### Checklist

### Additional notes

All text- and image-embedding models are re-exported with the per-method `get_dynamic_dims_<methodName>` companion and pinned to `v0.10.0`: all-MiniLM-L6-v2, all-mpnet-base-v2, multi-qa MiniLM/MPNet, paraphrase-ML, DistilUSE, and CLIP text (plus the DistilUSE/paraphrase CoreML variants). CLIP image has a static input, so it needs no companion and is unchanged.

### DIFF
--- a/apps/computer-vision/app/classification/index.tsx
+++ b/apps/computer-vision/app/classification/index.tsx
@@ -4,7 +4,7 @@ import { commonStyles, ColorPalette } from '../../theme';
 import { useImage } from '@shopify/react-native-skia';
 import { useClassifier, models } from 'react-native-executorch';
 import ScreenWrapper from '../../components/ScreenWrapper';
-import { getImage } from '../../utils';
+import { getImage, skImageToBuffer } from '../../utils';
 import { ModelPicker, type ModelOption } from '../../components/ModelPicker';
 import { ImageViewport } from '../../components/ImageViewport';
 import { ModelStatus } from '../../components/ModelStatus';
@@ -64,20 +64,7 @@ function ClassificationContent() {
     if (!sync) setIsProcessing(true);
     setError(null);
     try {
-      const pixels = skiaImage.readPixels();
-      if (!pixels) {
-        throw new Error('Failed to read pixels from image');
-      }
-      if (!(pixels instanceof Uint8Array)) {
-        throw new Error('Expected Uint8Array from readPixels');
-      }
-      const buffer = {
-        data: pixels,
-        width: skiaImage.width(),
-        height: skiaImage.height(),
-        format: 'rgba' as const,
-        layout: 'hwc' as const,
-      };
+      const buffer = skImageToBuffer(skiaImage);
       const start = Date.now();
       const output = sync
         ? classifyWorklet(buffer, { topk: 5 })
--- a/apps/computer-vision/app/classification/index.tsx
+++ b/apps/computer-vision/app/classification/index.tsx
@@ -1,10 +1,11 @@
 import React, { useState } from 'react';
 import { View, Text, StyleSheet, ScrollView, Platform } from 'react-native';
-import { commonStyles, ColorPalette } from '../../theme';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+import { commonStyles, ColorPalette, theme } from '../../theme';
 import { useImage } from '@shopify/react-native-skia';
 import { useClassifier, models } from 'react-native-executorch';
 import ScreenWrapper from '../../components/ScreenWrapper';
-import { getImage, skImageToBuffer } from '../../utils';
+import { getImage } from '../../utils';
 import { ModelPicker, type ModelOption } from '../../components/ModelPicker';
 import { ImageViewport } from '../../components/ImageViewport';
 import { ModelStatus } from '../../components/ModelStatus';
@@ -28,6 +29,7 @@ const MODEL_OPTIONS: ModelOption[] = [
 ];
 
 function ClassificationContent() {
+  const insets = useSafeAreaInsets();
   const [selectedModel, setSelectedModel] = useState<any>(MODEL_OPTIONS[0].value);
   const [imageUri, setImageUri] = useState<string | null>(null);
   const [isProcessing, setIsProcessing] = useState(false);
@@ -64,7 +66,20 @@ function ClassificationContent() {
     if (!sync) setIsProcessing(true);
     setError(null);
     try {
-      const buffer = skImageToBuffer(skiaImage);
+      const pixels = skiaImage.readPixels();
+      if (!pixels) {
+        throw new Error('Failed to read pixels from image');
+      }
+      if (!(pixels instanceof Uint8Array)) {
+        throw new Error('Expected Uint8Array from readPixels');
+      }
+      const buffer = {
+        data: pixels,
+        width: skiaImage.width(),
+        height: skiaImage.height(),
+        format: 'rgba' as const,
+        layout: 'hwc' as const,
+      };
       const start = Date.now();
       const output = sync
         ? classifyWorklet(buffer, { topk: 5 })
@@ -84,7 +99,10 @@ function ClassificationContent() {
   return (
     <ScrollView
       style={commonStyles.container}
-      contentContainerStyle={commonStyles.contentContainer}
+      contentContainerStyle={[
+        commonStyles.contentContainer,
+        { paddingBottom: insets.bottom + theme.spacing.large },
+      ]}
     >
       <Text style={commonStyles.description}>
         Upload or capture an image to identify objects using a classifier.
--- a/apps/nlp/app/text-embeddings/index.tsx
+++ b/apps/nlp/app/text-embeddings/index.tsx
@@ -33,27 +33,19 @@ const STARTER_SENTENCES = [
   'The cat sleeps on the warm windowsill.',
 ];
 
-// True cosine similarity. Most of these models output L2-normalized embeddings
-// (so a plain dot product would do), but some — e.g. DistilUSE — do not, so we
-// divide by the norms to stay correct regardless of whether the model normalizes.
+// These models output L2-normalized embeddings, so cosine similarity is the dot
+// product.
 const cosine = (a: Float32Array, b: Float32Array) => {
   let dot = 0;
-  let na = 0;
-  let nb = 0;
   for (let i = 0; i < a.length; i++) {
     dot += a[i]! * b[i]!;
-    na += a[i]! * a[i]!;
-    nb += b[i]! * b[i]!;
   }
-  const denom = Math.sqrt(na) * Math.sqrt(nb);
-  return denom > 0 ? dot / denom : 0;
+  return dot;
 };
 
 type Entry = { sentence: string; embedding: Float32Array };
 type Match = { sentence: string; similarity: number };
 
-// A model switch disposes the previous task while a forward() may still be
-// running; the resulting "disposed" rejection is expected and not worth showing.
 const isDisposedError = (msg: string) => /disposed/i.test(msg);
 
 function TextEmbeddingsContent() {
@@ -70,11 +62,7 @@ function TextEmbeddingsContent() {
 
   const ready = isReady && !!forward;
 
-  // Seed the library with starter sentences whenever a model finishes loading.
-  // `forward`'s identity changes only when the underlying model instance does,
-  // so this re-seeds on every model switch. The `cancelled` flag stops a pending
-  // forward() from a superseded model — whose tokenizer/model the hook has since
-  // disposed — from writing stale state or surfacing a "disposed" error.
+  // Re-seed the library with starter sentences whenever the model changes.
   useEffect(() => {
     if (!ready || !forward) return;
     let cancelled = false;
--- a/apps/nlp/app/text-embeddings/index.tsx
+++ b/apps/nlp/app/text-embeddings/index.tsx
@@ -1,0 +1,410 @@
+import React, { useEffect, useState } from 'react';
+import {
+  View,
+  Text,
+  TextInput,
+  ScrollView,
+  StyleSheet,
+  TouchableOpacity,
+  KeyboardAvoidingView,
+  Platform,
+} from 'react-native';
+import { useTextEmbeddings, models, type TextEmbeddingsModel } from 'react-native-executorch';
+import ScreenWrapper from '../../components/ScreenWrapper';
+import { ModelStatus } from '../../components/ModelStatus';
+import { Button } from '../../components/Button';
+import { theme } from '../../theme';
+
+const MODELS: { label: string; value: TextEmbeddingsModel }[] = [
+  { label: 'MiniLM L6', value: models.textEmbeddings.ALL_MINILM_L6_V2 },
+  { label: 'MPNet Base', value: models.textEmbeddings.ALL_MPNET_BASE_V2 },
+  { label: 'MultiQA MiniLM', value: models.textEmbeddings.MULTI_QA_MINILM_L6_COS_V1 },
+  { label: 'MultiQA MPNet', value: models.textEmbeddings.MULTI_QA_MPNET_BASE_DOT_V1 },
+  { label: 'Paraphrase ML', value: models.textEmbeddings.PARAPHRASE_MULTILINGUAL_MINILM_L12_V2 },
+  { label: 'DistilUSE ML', value: models.textEmbeddings.DISTILUSE_BASE_MULTILINGUAL_CASED_V2 },
+  { label: 'CLIP Text', value: models.textEmbeddings.CLIP_VIT_BASE_PATCH32_TEXT },
+];
+
+const STARTER_SENTENCES = [
+  'The weather is lovely today.',
+  "It's so sunny outside!",
+  'He drove to the stadium.',
+  'A man is eating a piece of bread.',
+  'The cat sleeps on the warm windowsill.',
+];
+
+// True cosine similarity. Most of these models output L2-normalized embeddings
+// (so a plain dot product would do), but some — e.g. DistilUSE — do not, so we
+// divide by the norms to stay correct regardless of whether the model normalizes.
+const cosine = (a: Float32Array, b: Float32Array) => {
+  let dot = 0;
+  let na = 0;
+  let nb = 0;
+  for (let i = 0; i < a.length; i++) {
+    dot += a[i]! * b[i]!;
+    na += a[i]! * a[i]!;
+    nb += b[i]! * b[i]!;
+  }
+  const denom = Math.sqrt(na) * Math.sqrt(nb);
+  return denom > 0 ? dot / denom : 0;
+};
+
+type Entry = { sentence: string; embedding: Float32Array };
+type Match = { sentence: string; similarity: number };
+
+// A model switch disposes the previous task while a forward() may still be
+// running; the resulting "disposed" rejection is expected and not worth showing.
+const isDisposedError = (msg: string) => /disposed/i.test(msg);
+
+function TextEmbeddingsContent() {
+  const [selected, setSelected] = useState(0);
+  const { isReady, downloadProgress, error, forward } = useTextEmbeddings(MODELS[selected]!.value);
+
+  const [library, setLibrary] = useState<Entry[]>([]);
+  const [input, setInput] = useState('');
+  const [matches, setMatches] = useState<Match[] | null>(null);
+  const [queryText, setQueryText] = useState('');
+  const [busy, setBusy] = useState(false);
+  const [runError, setRunError] = useState<string | null>(null);
+  const [embedMs, setEmbedMs] = useState<number | null>(null);
+
+  const ready = isReady && !!forward;
+
+  // Seed the library with starter sentences whenever a model finishes loading.
+  // `forward`'s identity changes only when the underlying model instance does,
+  // so this re-seeds on every model switch. The `cancelled` flag stops a pending
+  // forward() from a superseded model — whose tokenizer/model the hook has since
+  // disposed — from writing stale state or surfacing a "disposed" error.
+  useEffect(() => {
+    if (!ready || !forward) return;
+    let cancelled = false;
+    (async () => {
+      setBusy(true);
+      setRunError(null);
+      try {
+        const entries: Entry[] = [];
+        for (const sentence of STARTER_SENTENCES) {
+          const embedding = await forward(sentence);
+          if (cancelled) return;
+          entries.push({ sentence, embedding });
+        }
+        setLibrary(entries);
+      } catch (e: any) {
+        if (!cancelled) setRunError(e?.message ?? String(e));
+      } finally {
+        if (!cancelled) setBusy(false);
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, [forward, ready]);
+
+  const selectModel = (i: number) => {
+    if (i === selected) return;
+    setSelected(i);
+    setLibrary([]);
+    setMatches(null);
+    setQueryText('');
+    setEmbedMs(null);
+  };
+
+  const findSimilar = async () => {
+    if (!forward || !input.trim() || library.length === 0) return;
+    setBusy(true);
+    setRunError(null);
+    try {
+      const start = Date.now();
+      const q = await forward(input.trim());
+      setEmbedMs(Date.now() - start);
+      const ranked = library
+        .map(({ sentence, embedding }) => ({ sentence, similarity: cosine(q, embedding) }))
+        .sort((a, b) => b.similarity - a.similarity);
+      setQueryText(input.trim());
+      setMatches(ranked);
+    } catch (e: any) {
+      const msg = e?.message ?? String(e);
+      if (!isDisposedError(msg)) setRunError(msg);
+    } finally {
+      setBusy(false);
+    }
+  };
+
+  const addToLibrary = async () => {
+    if (!forward || !input.trim()) return;
+    setBusy(true);
+    setRunError(null);
+    try {
+      const start = Date.now();
+      const embedding = await forward(input.trim());
+      setEmbedMs(Date.now() - start);
+      setLibrary((prev) => [...prev, { sentence: input.trim(), embedding }]);
+      setInput('');
+      setMatches(null);
+    } catch (e: any) {
+      const msg = e?.message ?? String(e);
+      if (!isDisposedError(msg)) setRunError(msg);
+    } finally {
+      setBusy(false);
+    }
+  };
+
+  const removeAt = (i: number) => {
+    setLibrary((prev) => prev.filter((_, idx) => idx !== i));
+    setMatches(null);
+  };
+
+  const clearLibrary = () => {
+    setLibrary([]);
+    setMatches(null);
+    setQueryText('');
+  };
+
+  return (
+    <KeyboardAvoidingView
+      style={styles.flex}
+      behavior={Platform.OS === 'ios' ? 'padding' : undefined}
+    >
+      <ScrollView style={styles.container} contentContainerStyle={styles.content}>
+        <View style={styles.card}>
+          <Text style={styles.cardTitle}>Text Embeddings</Text>
+          <Text style={styles.cardDescription}>
+            Semantic search playground. Build a library of sentences, then find the ones closest in
+            meaning to your query using cosine similarity over the embeddings.
+          </Text>
+
+          <Text style={styles.fieldLabel}>Model</Text>
+          <ScrollView
+            horizontal
+            showsHorizontalScrollIndicator={false}
+            contentContainerStyle={styles.chipRow}
+          >
+            {MODELS.map((m, i) => (
+              <TouchableOpacity
+                key={m.label}
+                style={[styles.chip, i === selected && styles.chipActive]}
+                onPress={() => selectModel(i)}
+              >
+                <Text style={[styles.chipText, i === selected && styles.chipTextActive]}>
+                  {m.label}
+                </Text>
+              </TouchableOpacity>
+            ))}
+          </ScrollView>
+
+          <ModelStatus
+            isReady={isReady}
+            downloadProgress={downloadProgress}
+            error={error ? error.message : null}
+            modelTypeLabel="model"
+          />
+        </View>
+
+        {runError && (
+          <View style={styles.errorContainer}>
+            <Text style={styles.errorText}>{runError}</Text>
+          </View>
+        )}
+
+        <View style={styles.card}>
+          <View style={styles.libraryHeader}>
+            <Text style={styles.sectionTitle}>Sentence library ({library.length})</Text>
+            {library.length > 0 && (
+              <TouchableOpacity onPress={clearLibrary}>
+                <Text style={styles.clearLink}>Clear</Text>
+              </TouchableOpacity>
+            )}
+          </View>
+          {library.length === 0 ? (
+            <Text style={styles.emptyText}>
+              {ready ? 'Library is empty — add a sentence below.' : 'Waiting for the model…'}
+            </Text>
+          ) : (
+            library.map((item, i) => (
+              <View key={`${item.sentence}-${i}`} style={styles.libraryRow}>
+                <Text style={styles.librarySentence}>{item.sentence}</Text>
+                <TouchableOpacity onPress={() => removeAt(i)} hitSlop={8}>
+                  <Text style={styles.removeBtn}>✕</Text>
+                </TouchableOpacity>
+              </View>
+            ))
+          )}
+        </View>
+
+        <View style={styles.card}>
+          <Text style={styles.sectionTitle}>Try your sentence</Text>
+          <TextInput
+            style={styles.input}
+            value={input}
+            onChangeText={setInput}
+            autoCapitalize="none"
+            placeholder="Type a sentence…"
+            placeholderTextColor="#999"
+            multiline
+          />
+          <View style={styles.buttonRow}>
+            <Button
+              title="Find similar"
+              onPress={findSimilar}
+              disabled={!ready || !input.trim() || library.length === 0}
+              loading={busy}
+            />
+            <Button
+              title="Add to library"
+              variant="secondary"
+              onPress={addToLibrary}
+              disabled={!ready || !input.trim()}
+            />
+          </View>
+          {embedMs !== null && <Text style={styles.statsText}>Embedding time: {embedMs} ms</Text>}
+        </View>
+
+        {matches && (
+          <View style={styles.card}>
+            <Text style={styles.sectionTitle}>Ranked by similarity</Text>
+            <Text style={styles.queryText}>to “{queryText}”</Text>
+            {matches.map((m, i) => {
+              const pct = Math.max(0, Math.min(1, m.similarity)) * 100;
+              return (
+                <View key={`${m.sentence}-${i}`} style={styles.matchRow}>
+                  <View style={styles.matchHeader}>
+                    <Text style={[styles.matchSentence, i === 0 && styles.matchTop]}>
+                      {m.sentence}
+                    </Text>
+                    <Text style={styles.matchScore}>{m.similarity.toFixed(3)}</Text>
+                  </View>
+                  <View style={styles.barTrack}>
+                    <View
+                      style={[styles.barFill, { width: `${pct}%` }, i === 0 && styles.barFillTop]}
+                    />
+                  </View>
+                </View>
+              );
+            })}
+          </View>
+        )}
+      </ScrollView>
+    </KeyboardAvoidingView>
+  );
+}
+
+export default function TextEmbeddingsScreen() {
+  return (
+    <ScreenWrapper>
+      <TextEmbeddingsContent />
+    </ScreenWrapper>
+  );
+}
+
+const styles = StyleSheet.create({
+  flex: { flex: 1 },
+  container: { flex: 1, backgroundColor: theme.colors.background },
+  content: { padding: theme.spacing.large, paddingBottom: 40 },
+  card: {
+    backgroundColor: theme.colors.cardBackground,
+    borderRadius: theme.radius.large,
+    padding: 20,
+    marginBottom: 20,
+    borderWidth: 1,
+    borderColor: theme.colors.lightBorder,
+  },
+  cardTitle: {
+    fontSize: theme.typography.title.fontSize,
+    fontWeight: theme.typography.title.fontWeight,
+    color: theme.colors.strongPrimary,
+    marginBottom: 8,
+  },
+  cardDescription: {
+    fontSize: 14,
+    color: theme.colors.textMuted,
+    lineHeight: 20,
+    marginBottom: 16,
+  },
+  fieldLabel: {
+    fontSize: 12,
+    fontWeight: '600',
+    color: theme.colors.textPlaceholder,
+    textTransform: 'uppercase',
+    letterSpacing: 0.5,
+    marginBottom: 8,
+  },
+  chipRow: { gap: 8, paddingBottom: 4, marginBottom: 12 },
+  chip: {
+    paddingHorizontal: 14,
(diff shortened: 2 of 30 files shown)


apps/nlp/app/text-embeddings/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+  let na = 0;
+  let nb = 0;
+  for (let i = 0; i < a.length; i++) {
+    dot += a[i]! * b[i]!;
+    na += a[i]! * a[i]!;
+    nb += b[i]! * b[i]!;
+  }
+  const denom = Math.sqrt(na) * Math.sqrt(nb);
+  return denom > 0 ? dot / denom : 0;
+};
```
msluszniak: Is this still needed after fixing distiluse?

apps/nlp/app/text-embeddings/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+
+type Entry = { sentence: string; embedding: Float32Array };
+type Match = { sentence: string; similarity: number };
+
+// A model switch disposes the previous task while a forward() may still be
+// running; the resulting "disposed" rejection is expected and not worth showing.
```
msluszniak: I guess this comment is not needed.

apps/nlp/app/text-embeddings/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+  const [runError, setRunError] = useState<string | null>(null);
+  const [embedMs, setEmbedMs] = useState<number | null>(null);
+
+  const ready = isReady && !!forward;
+
+  // Seed the library with starter sentences whenever a model finishes loading.
+  // `forward`'s identity changes only when the underlying model instance does,
+  // so this re-seeds on every model switch. The `cancelled` flag stops a pending
+  // forward() from a superseded model — whose tokenizer/model the hook has since
+  // disposed — from writing stale state or surfacing a "disposed" error.
```
msluszniak: Same here, this comment is redundant, I can read the code which tells me more.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -218,10 +221,21 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
+            // Validates a tensor against the method signature. When `dimBounds`
+            // is null, every dimension must match the signature exactly; this is
+            // used for output tensors (which we pre-allocate and memcpy into, so
+            // their shape must be exact) and for inputs of statically-shaped
+            // models. When `dimBounds` is provided, each dimension is instead
+            // checked against the model-declared [min, max, step] range, allowing
+            // dynamically-shaped inputs (e.g. a variable sequence length) while
+            // still rejecting out-of-range shapes before they reach the runtime.
+            // A static dimension is encoded as {n, n, 1}, so it still requires an
+            // exact match.
```
msluszniak: Again, comment as such not needed. We need to instead document the behaviour of how `get_dynamic_dims` is used, how tensors are validated etc.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -243,6 +272,33 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
                 }
             };
 
+            // Statically-shaped models validate input dimensions exactly. A model
+            // exported with dynamic `forward` inputs declares the allowed ranges
+            // via an optional `get_dynamic_dims` method that returns an int64
+            // [D, 3] tensor of [min, max, step] rows, one per dimension of
+            // forward's tensor inputs, flattened in input order. Absent (or a
+            // Bool false) means all input dimensions are static.
```
msluszniak: Again, this shouldn't be a comment in the code, I guess.

apps/nlp/app/text-embeddings/index.tsx
msluszniak, line 1: We have textEmbeddings example, but we don't have image embeddings one, we need to add this.

apps/nlp/app/text-embeddings/index.tsx
msluszniak: No — all of these models L2-normalize their output (DistilUSE included), so I dropped the norm division and made it a plain dot product.

apps/nlp/app/text-embeddings/index.tsx
msluszniak: Removed.

apps/nlp/app/text-embeddings/index.tsx
msluszniak: Removed.

packages/react-native-executorch/cpp/core/model.cpp
msluszniak: Moved the get_dynamic_dims contract and how inputs/outputs are validated into the ModelHostObject class docs; the inline comment is now a one-liner pointing there.

packages/react-native-executorch/cpp/core/model.cpp
msluszniak: Trimmed to a short one-liner; the behaviour is documented in the ModelHostObject class docs.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -214,6 +259,18 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
 
+            // Per-dimension [min, max, step] bounds for forward's tensor inputs,
+            // parsed once from the optional get_dynamic_dims method and cached.
+            // Empty for statically shaped methods, which then validate exactly.
+            auto boundsIt = self->dynamicInputBounds_.find(methodName);
+            if (boundsIt == self->dynamicInputBounds_.end()) {
+                auto parsed = methodName == "forward" ? parseDynamicDims(rt, *self->etModule_)
+                                                      : std::vector<std::array<int64_t, 3>>{};
+                boundsIt = self->dynamicInputBounds_.emplace(methodName, std::move(parsed)).first;
+            }
```
msluszniak: shouldn't initialization of the dynamic dims be done in constructor?

msluszniak: Done — it now parses `get_dynamic_dims` once in the constructor (after load) into `dynamicInputBounds_`; `execute` just looks it up, so it stays immutable and needs no locking.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -223,7 +278,36 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
+                            rangeDim.max = static_cast<int32_t>(row[1]);
+                            if (row[2] > 1) {
+                                rangeDim.step = static_cast<int32_t>(row[2]);
+                            }
+                            expectedShape.emplace_back(rangeDim);
+                        }
+                        boundsOffset += rank;
+                        tensorHostObject = tensor::fromJs(rt, ctx, val, expectedDtype,
+                                                          std::optional<tensor::SymbolicShape>(std::move(expectedShape)));
+                    }
```
barhanc, line 372:
```suggestion

                    std::shared_ptr<TensorHostObject> tensorHostObject;
                    if (self->dynamicInputShapes_.contains(methodName)) {
                        auto expectedShape = self->dynamicInputShapes_[methodName][i];
                        tensorHostObject = tensor::fromJs(rt, ctx, val, expectedDtype, expectedShape);
                    } else {
                        tensorHostObject = tensor::fromJs(rt, ctx, val, expectedDtype, tensorMeta.sizes());
                    }
```

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -251,6 +335,12 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
                 }
             }
 
+            if (!dynamicInputBounds.empty() && boundsOffset != dynamicInputBounds.size()) {
+                throw jsi::JSError(rt, std::format("execute: get_dynamic_dims declares more dimensions ({}) "
+                                                   "than forward's tensor inputs use ({})",
+                                                   dynamicInputBounds.size(), boundsOffset));
+            }
+
```
barhanc:
```suggestion
```

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -214,6 +260,15 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
 
+            // Per-dimension [min, max, step] bounds parsed from get_dynamic_dims
+            // at construction. Absent for statically shaped methods, which then
+            // validate exactly.
+            const std::vector<std::array<int64_t, 3>> noBounds;
+            auto boundsIt = self->dynamicInputBounds_.find(methodName);
+            const auto &dynamicInputBounds =
+                boundsIt != self->dynamicInputBounds_.end() ? boundsIt->second : noBounds;
+            size_t boundsOffset = 0;
+
```
barhanc:
```suggestion
```

packages/react-native-executorch/cpp/core/model.cpp
barhanc: We want to minimize changes to the `execute` method required for dynamic inputs validation so that the code is future-proof, e.g. when ExecuTorch adds native support or we will want to change how the dynamic shapes are parsed.

packages/react-native-executorch/src/core/tensor.ts
```diff
@@ -50,10 +50,10 @@ export type Tensor = {
   /**
    * Writes data from a typed array into this tensor's native buffer.
    * @param src The source typed array. Its size in bytes must match the
-   * tensor's size.
+   * tensor's size. Use a `BigInt64Array` for `int64` tensors.
```
barhanc, line 53:
```suggestion
   * tensor's size.
```

packages/react-native-executorch/cpp/core/model.h
```diff
@@ -31,6 +34,8 @@ class ModelHostObject : public jsi::HostObject,
     std::string modelPath_;
     std::unique_ptr<executorch::extension::Module> etModule_;
     std::mutex mutex_;
+
+    std::unordered_map<std::string, std::vector<std::array<int64_t, 3>>> dynamicInputBounds_;
```
barhanc:
```suggestion
    std::unordered_map<std::string, std::vector<tensor::SymbolicShape>> dynamicInputShapes_;
```
Let's use the `tensor::SymbolicShape` directly so that we don't have to build it on every execute call and can simplify the `execute` code.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -73,6 +115,10 @@ ModelHostObject::ModelHostObject(const std::string &modelPath)
         const std::string errorMsg = executorch::runtime::to_string(error);
         throw std::runtime_error(std::format("Failed to load model: {}", errorMsg));
     }
+
+    if (auto bounds = parseDynamicDims(*etModule_); !bounds.empty()) {
+        dynamicInputBounds_.emplace("forward", std::move(bounds));
+    }
```
barhanc, line 214:
```suggestion

    auto methodNames = unwrap("Failed to get method names", etModule_->method_names());
    for (const auto &methodName : methodNames) {
        auto dynamicShapes = parseDynamicInputShapes(*etModule_, methodName);
        if (dynamicShapes) {
            dynamicInputShapes_.emplace(methodName, std::move(*dynamicShapes));
        }
    }
```
We don't want to hardcode the `"forward"` method. If we assume that in the export stage each dynamic method has to have a companion method `get_dynamic_dims_<methodName>` then we can create a generic mapping for all methods.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -57,6 +61,44 @@ jsi::Object tensorMetaToJs(jsi::Runtime &rt, const executorch::runtime::TensorIn
...
+        const int64_t step = data[r * 3 + 2];
+        if (maxDim < minDim || step < 1) {
+            throw std::runtime_error(std::format("get_dynamic_dims row {} is invalid: expected "
+                                                 "min <= max and step >= 1 but got [{}, {}, {}]",
+                                                 r, minDim, maxDim, step));
+        }
+        bounds.push_back({minDim, maxDim, step});
+    }
+    return bounds;
+}
```
barhanc, line 190:
```suggestion
template <typename T>
T unwrap(const std::string &ctx, executorch::runtime::Result<T> result) {
    if (!result.ok()) {
        throw std::runtime_error(std::format("{}: {}", ctx, executorch::runtime::to_string(result.error())));
    }
    return std::move(result.get());
}

/**
 * Parses compile-time dynamic dimension constraints for the inputs of a module
 * method.
 *
 * @note This is a temporary workaround. ExecuTorch (.pte) model metadata
 * natively serializes only the static upper-bound limits of dynamic/symbolic
 * dimensions, and does not expose the active dynamic range (min, max, step) at
 * runtime.
 *
 * To use this feature, the .pte model must be exported with a companion method
 * named "get_dynamic_dims_<methodName>" (e.g., "get_dynamic_dims_forward").
 *
 * Python export requirements:
 * 1. The companion method must take no arguments.
 * 2. It must return a list of outputs matching the number of Tag::Tensor inputs
 *    of the target method (scalar inputs are excluded).
 * 3. Each output must be a 2D int32 tensor of shape [rank, 3], where each row
 *    represents [min, max, step] constraints for the corresponding dimension of
 *    that input tensor.
 *
 * @param module The ExecuTorch extension module to query.
 * @param methodName The name of the target module method (e.g. "forward").
 * @return A vector of SymbolicShape objects, or std::nullopt if the companion
 *         method is not defined. If returned, the vector's size is guaranteed
 *         to equal methodMeta.num_inputs(), containing parsed SymbolicShapes
 *         for tensor inputs and empty SymbolicShapes for non-tensor inputs.
 */
std::optional<std::vector<rnexecutorch::core::tensor::SymbolicShape>>
parseDynamicInputShapes(executorch::extension::Module &module, const std::string &methodName) {
    using executorch::aten::ScalarType;

    const auto getDynamicShapesMethodName = std::format("get_dynamic_dims_{}", methodName);
    const auto ctx = getDynamicShapesMethodName + ": ";

    auto methodNames = unwrap(ctx + "failed to get method names", module.method_names());
    if (methodName == getDynamicShapesMethodName ||
        !methodNames.contains(getDynamicShapesMethodName)) {
        return std::nullopt;
    }

    auto methodMeta = unwrap(std::format("{}failed to get meta for method '{}'", ctx, methodName),
                             module.method_meta(methodName));

    size_t expectedTensorInputs = 0;
    for (size_t i = 0; i < methodMeta.num_inputs(); ++i) {
        auto tag = unwrap(std::format("{}failed to get tag for input [{}]", ctx, i), methodMeta.input_tag(i));
        if (tag == executorch::runtime::Tag::Tensor) {
            expectedTensorInputs++;
        }
    }

    auto result = unwrap(ctx + "failed to execute", module.execute(getDynamicShapesMethodName));
    if (result.size() != expectedTensorInputs) {
        throw std::runtime_error(std::format("{}number of outputs returned ({}) does not match the number of "
                                             "tensor inputs declared by method '{}' ({})",
                                             ctx, result.size(), methodName, expectedTensorInputs));
    }

    std::vector<tensor::SymbolicShape> dynamicShapes;
    dynamicShapes.reserve(methodMeta.num_inputs());
    size_t tensorIndex = 0;

    for (size_t i = 0; i < methodMeta.num_inputs(); ++i) {
        auto tag = unwrap(std::format("{}failed to get tag for input [{}]", ctx, i),
                          methodMeta.input_tag(i));

        if (tag != executorch::runtime::Tag::Tensor) {
            // Emplace an empty shape for non-tensor inputs to maintain a 1-to-1
            // index alignment between dynamicShapes and the model's inputs vector.
            dynamicShapes.emplace_back();
            continue;
        }

        const auto &out = result.at(tensorIndex);
        if (!out.isTensor()) {
            throw std::runtime_error(std::format("{}output[{}] is not a tensor", ctx, tensorIndex));
        }

        auto inputMeta = unwrap(std::format("{}failed to get tensor meta for input [{}]", ctx, i),
                                methodMeta.input_tensor_meta(i));
        const auto rank = inputMeta.sizes().size();
        const auto shapeTensor = out.toTensor();

        if (
            shapeTensor.dim() != 2 ||
            shapeTensor.size(1) != 3 ||
            shapeTensor.size(0) != static_cast<ssize_t>(rank) ||
            shapeTensor.scalar_type() != ScalarType::Int) {
            throw std::runtime_error(std::format("{}output[{}] expected to be a 2D int32_t tensor of shape [{}, 3]",
                                                 ctx, tensorIndex, rank));
        }

        const auto *shape = shapeTensor.const_data_ptr<int32_t>();
        tensor::SymbolicShape symbolicShape;
        symbolicShape.reserve(rank);

        for (size_t axis = 0; axis < rank; ++axis) {
            const auto minDim = shape[axis * 3 + 0];
            const auto maxDim = shape[axis * 3 + 1];
            const auto step = shape[axis * 3 + 2];
            if (minDim < 0 || maxDim < minDim || step < 1) {
                throw std::runtime_error(std::format("{}output[{}], axis {} is invalid: "
                                                     "expected 0 <= min <= max and step >= 1 but got [{}, {}, {}]",
                                                     ctx, tensorIndex, axis, minDim, maxDim, step));
            }
            if (maxDim > inputMeta.sizes()[axis]) {
                throw std::runtime_error(std::format("{}output[{}], axis {} max dimension ({}) "
                                                     "exceeds model metadata upper limit ({})",
                                                     ctx, tensorIndex, axis, maxDim, inputMeta.sizes()[axis]));
            }

            symbolicShape.emplace_back(tensor::RangeDim{.min = minDim, .max = maxDim, .step = step});
        }

        dynamicShapes.push_back(std::move(symbolicShape));
        ++tensorIndex;
    }

    return dynamicShapes;
}
```

  - **Encapsulation** — all parsing, validation, and the ExecuTorch-metadata quirks now live inside `parseDynamicInputShapes`, returning a clean `SymbolicShape` per input. The execute path no longer reconstructs shapes from a flat bounds list; it just indexes by input position.
  - **Generality** — iterates over every method in `method_names()` instead of hardcoding `forward`, so multi-method modules (encoder/decoder, embeddings, etc.) get dynamic-shape support for free.
  - **Safety** — structural 1:1 alignment with inputs (empty placeholder for non-tensor inputs) removes the fragile offset-walking logic and its guard branches. All shape/rank/dtype checks plus the model-metadata cross-check now fail fast at load time rather than on first inference.
  - **Future-proof** — the per-input `[rank, 3]` contract and per-method naming (`get_dynamic_dims_<method>`) make the Python export self-documenting and easy to extend without touching the runtime mapping.

packages/react-native-executorch/src/extensions/cv/tasks/imageEmbeddings.ts
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Releases all allocated native resources.
+   */
+  dispose: () => void;
+  /**
+   * Asynchronously computes the embedding vector for the given input image.
+   * @param input The input image buffer.
+   * @returns A promise resolving to the embedding vector.
+   */
+  forward: (input: ImageBuffer) => Promise<Float32Array>;
```
barhanc: A more descriptive name like `embed` would be better imo than generic `forward`.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbeddings.ts
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Releases all allocated native resources.
+   */
+  dispose: () => void;
+  /**
+   * Asynchronously computes the embedding vector for the given input text.
+   * @param input The input text to embed.
+   * @returns A promise resolving to the embedding vector.
+   */
+  forward: (input: string) => Promise<Float32Array>;
```
barhanc: Same as in image embeddings, more descriptive name would be better imo.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbeddings.ts
```diff
@@ -0,0 +1,119 @@
...
+    // mask is all ones since every position is a real token.
+    const idsData = new BigInt64Array(len);
+    const maskData = new BigInt64Array(len);
+    for (let i = 0; i < len; i++) {
+      idsData[i] = BigInt(ids[i]!);
+      maskData[i] = 1n;
+    }
+
+    const tokenIds = tensor('int64', [1, len], idsData);
+    const attentionMask = tensor('int64', [1, len], maskData);
```
barhanc: Please use the `t<Name>` naming convention for tensor variables.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbeddings.ts
```diff
@@ -0,0 +1,119 @@
...
+ * pass and returns the raw embedding vector.
+ * @category Typescript API
+ * @param config Text embeddings task configuration containing the model and
+ * tokenizer paths.
+ * @param runtime Optional worklet runtime thread on which to run the model
+ * execution.
+ * @returns A promise resolving to an object containing the embedding and
+ * disposal controls.
+ */
+export async function createTextEmbeddings(
```
barhanc: Probably should be named `...Embedder` to match other tasks. Same with file name, perhaps `...Embedding.ts` (no 's') would be more consistent.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -0,0 +1,119 @@
...
+    model.dispose();
+  };
+
+  const forwardWorklet = (input: string): Float32Array => {
+    'worklet';
+    const ids = tokenizer.encode(input);
+    if (ids.length === 0) {
+      throw new Error('createTextEmbeddings: input tokenized to zero tokens');
+    }
+    const len = Math.min(ids.length, maxSeqLen);
```
barhanc, line 97: Worth documenting the truncating behaviour on long inputs.

packages/react-native-executorch/src/extensions/nlp/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './tokenizer';
 export * from './tasks/tokenization';
+export * from './tasks/textEmbeddings';
```
barhanc: Tasks shouldn't be explicitly exported from `/extensions/<domain>/index.ts`.

barhanc (commented):
Checked only lib implementation. I will take a look at app code on Monday.

---

Regarding `core/` dynamic input support changes, I've added some suggestions that imo make the code more future-proof and easier to change.

---

Regarding TS side everything is fine, just some small nits. I was thinking though if we don't want to take the opportunity of this refactor and beef-up the TS embeddings side a bit more, something like unifying both image and text embeddings into one pipeline that on top of exposing simple `embed` method would also expose methods implementing a small vector search data structure like `insert`, `clear`, `find`, etc. I don't know how much effort would that be so it's your call. The current implementation is correct.

msluszniak:
@barhanc regarding:

1. Merging text and image pipelines currently I'm slightly towards not doing this. I see the following pros and cons of doing so:


    PROS:
    - Single `embed` entry and shared result type.
    - Removes a bit of duplication (but only the trivial one).

    CONS:
    - Breaks the split between image embeddings (`extensions/cv`) and text embeddings (`extensions/nlp`). Maybe it's incorrect one and both should be under nlp, idk.
    - API gets a bit overloaded, input needs to be `string | ImageBuffer` union and config will be union of `tokenizerPath | preprocessorOpts`.
    - text = tokenizer + int64 ids/mask + truncation; image = image preprocessor + float NCHW.
    - Even CLIP has text and image part and each is a separate `*.pte` file.

2. Regarding small vector DB functionality, I would do it in a separate PR, but not as a part of the embedding flow, but as a small, separate functionality. That is because otherwise it would introduce stateful functionality inside the generally stateless pipelines and couples them. It is also a separate scope from embeddings themselves.

If you think that there are other things to consider regarding these two, please let me know and I will re-think it. In a meantime, I will address inline comments :).

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -18,11 +21,12 @@
...
 
 template <typename T>
-T unwrap(jsi::Runtime &rt, const std::string &ctx, executorch::runtime::Result<T> result) {
+T unwrap(const std::string &ctx, executorch::runtime::Result<T> result) {
     if (!result.ok()) {
-        throw jsi::JSError(rt, std::format("{}: {}", ctx, executorch::runtime::to_string(result.error())));
+        throw std::runtime_error(std::format("{}: {}", ctx, executorch::runtime::to_string(result.error())));
     }
     return std::move(result.get());
 }
```
barhanc, line 32: This should be an overload of the `unwrap`. Right now we throw `std::runtime_error` where `jsi::JSError` should be thrown.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -98,7 +229,7 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
                 throw jsi::JSError(rt, "getMethodNames: Model has been disposed");
             }
 
-            auto methodNames = unwrap(rt, "getMethodNames", self->etModule_->method_names());
+            auto methodNames = unwrap("getMethodNames", self->etModule_->method_names());
```
barhanc: Here we should use the `unwrap` with `(rt, ...)` signature that throws `jsi::JSError`. And same below.

packages/react-native-executorch/cpp/core/model.cpp
barhanc, line 1: We can also quickly add support for validation of dynamic outputs by changing `tensorMeta.sizes()` to `output.toTensor().sizes()` in the outputs' loop (it works as I verified in the STT implementation).

packages/react-native-executorch/src/hooks/useTextEmbedder.ts
```diff
@@ -0,0 +1,45 @@
...
+    localModelPath && localTokenizerPath
+      ? { modelPath: localModelPath, tokenizerPath: localTokenizerPath }
+      : null,
+    [localModelPath, localTokenizerPath]
+  );
+
+  return {
+    isReady: !!model,
+    error: modelResource.downloadError || tokenizerResource.downloadError || error,
+    downloadProgress: (modelResource.downloadProgress + tokenizerResource.downloadProgress) / 2,
```
barhanc:
```suggestion
    downloadProgress: modelResource.downloadProgress,
```
Typically the model size >> tokenizer size, so the arithmetic mean is not a good heuristic, rather just report only the model size, at least until the resource fetcher addition that might actually expose a method for getting the size of the resource beforehand.

apps/computer-vision/app/classification/index.tsx
barhanc, line 1: Why was this change only added to classification when other CV screens use the same pattern?

barhanc:
Tested apps and they work fine. Also I see that on huggingface we also have LFM text embeddings, why are they not added here?

barhanc:
One more thing, we should probably add some information about the dynamic methods to skills and jsdoc so that agents are aware of the companion method requirement, this is adhoc solution until we have a proper documentation.

packages/react-native-executorch/cpp/core/model.cpp
msluszniak: Done — added a `jsi::JSError` overload of `unwrap`; the `std::runtime_error` one stays only for the load-time constructor path, where there's no `rt`.

packages/react-native-executorch/cpp/core/model.cpp
msluszniak: Done — `getMethodNames`/`getMethodMeta`/`execute` now all use the `(rt, …)` overload.

packages/react-native-executorch/cpp/core/model.cpp
msluszniak: Good idea — done. Outputs now validate against `output.toTensor().sizes()` instead of the static metadata.

packages/react-native-executorch/src/hooks/useTextEmbedder.ts
msluszniak: Applied.

apps/computer-vision/app/classification/index.tsx
msluszniak: It was an unrelated cleanup that I'd only applied here — reverted it to keep this PR scoped to embeddings.

msluszniak:
@barhanc I addressed the comments. Regarding LFM2.5 there are two things:
- I added mlx delegate, but haven't tested it yet since I don't have iOS device with me. Xnnpack version is tested and works correctly.
- LFM2.5 supports two modes: query and document. So I had two options, either do nothing with API and document that user must prepend with either `query:` or `document:` to each input using this model or add an option for passing mode using the API. I decided to choose the second option.

packages/react-native-executorch/src/core/modelSchema.ts
```diff
@@ -158,6 +158,17 @@ function validateTags(
  * On success it returns the method's {@link ModelMethodMeta}, which can be used
  * to read concrete input/output tensor shapes for pre-allocating scratch
  * tensors.
+ *
+ * @remarks
```
barhanc:
```suggestion
 * @remarks
```
Fixes lint warning 'Expected only 0 lines after block description'.

packages/react-native-executorch/src/core/modelSchema.ts
```diff
@@ -158,6 +158,17 @@ function validateTags(
...
+ * A symbolic (string) input dimension here only relaxes *validation*. For a
+ * dimension that must genuinely vary at runtime (e.g. sequence length), the
+ * `.pte` must additionally be exported with a companion method
+ * `get_dynamic_dims_<methodName>` (e.g. `get_dynamic_dims_forward`) that
+ * returns, per tensor input, an `int32` `[rank, 3]` tensor of `[min, max, step]`
+ * bounds — ExecuTorch metadata only serializes the static upper bound, so the
+ * runtime reads the active range from this companion. Without it, the method
+ * only accepts the exact shape it was exported with. The reported upper bound is
+ * still read from `meta.inputTensorMeta` (e.g. to truncate over-long inputs).
```
barhanc:
```suggestion
 * still read from `meta.inputTensorMeta`.
```

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -285,7 +435,10 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
 
                     auto tensorMeta = unwrap(rt, ctx + ": tensor meta", methodMeta.output_tensor_meta(index));
                     auto expectedDtype = fromScalarType(rt, ctx, tensorMeta.scalar_type());
-                    auto tensorHostObject = tensor::fromJs(rt, ctx, val, expectedDtype, tensorMeta.sizes());
+                    // Validate the JS placeholder against the tensor's actual
+                    // produced shape rather than the static metadata, so methods
+                    // with dynamically-shaped outputs are handled correctly.
```
barhanc:
```suggestion
```
Comment not needed imo.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -18,7 +21,19 @@
...
+T unwrap(const std::string &ctx, executorch::runtime::Result<T> result) {
+    if (!result.ok()) {
+        throw std::runtime_error(std::format("{}: {}", ctx, executorch::runtime::to_string(result.error())));
+    }
+    return std::move(result.get());
+}
+
+// Overload for JSI host-function contexts: surfaces the failure as a JS error
+// instead of a std::runtime_error. The plain overload above is for load-time
+// paths (e.g. the constructor) where no jsi::Runtime is available.
```
barhanc:
```suggestion
```
Comment not needed imo.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -0,0 +1,133 @@
...
+export type TextEmbedderModel = {
+  readonly modelPath: string;
+  readonly tokenizerPath: string;
+  /**
+   * Optional prompt prefix prepended to every input before tokenization. Some
+   * models are trained with a task instruction (e.g. LFM2.5-Embedding uses
+   * `'query: '` for queries and `'document: '` for passages). Can be overridden
+   * per call via the `prompt` argument of `embed` / `embedWorklet`.
+   */
+  readonly prompt?: string;
```
barhanc: I would change it to `defaultPrompt` and the comment is not needed imo.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Releases all allocated native resources.
+   */
+  dispose: () => void;
+  /**
+   * Asynchronously computes the embedding vector for the given input text.
+   * Inputs longer than the model's maximum sequence length are truncated.
+   * @param input The input text to embed.
+   * @param prompt Optional prompt prefix overriding the model's configured
+   * `prompt` for this call (e.g. `'query: '` vs `'document: '`).
```
barhanc:
```suggestion
   * `prompt` for this call.
```
Without the LFM context the example doesn't tell anything imo.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  const dispose = () => {
+    tensors.forEach((t) => t.dispose());
+    tokenizer.dispose();
+    model.dispose();
+  };
+
+  const embedWorklet = (input: string, prompt?: string): Float32Array => {
+    'worklet';
+    // Prepend the per-call prompt override, else the model's configured prompt.
```
barhanc:
```suggestion
```

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -0,0 +1,133 @@
...
+    const text = (prompt ?? defaultPrompt ?? '') + input;
+    const ids = tokenizer.encode(text);
+    if (ids.length === 0) {
+      throw new Error('createTextEmbedder: input tokenized to zero tokens');
+    }
+    // Truncate inputs longer than the model's maximum sequence length; the
+    // model has no way to attend beyond it.
```
barhanc:
```suggestion
```

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -0,0 +1,133 @@
...
+    const len = Math.min(ids.length, maxSeqLen);
+
+    // Feed the exact token length with no padding. The model resizes its dynamic
+    // sequence input to match. Padding would change the result for pooling heads
+    // that are sensitive to it (e.g. DistilUSE's tanh projection). The attention
+    // mask is all ones since every position is a real token.
```
barhanc:
```suggestion
```

packages/react-native-executorch/src/hooks/useTextEmbedder.ts
```diff
@@ -0,0 +1,47 @@
...
+      ? { modelPath: localModelPath, tokenizerPath: localTokenizerPath }
+      : null,
+    [localModelPath, localTokenizerPath]
+  );
+
+  return {
+    isReady: !!model,
+    error: modelResource.downloadError || tokenizerResource.downloadError || error,
+    // The tokenizer is tiny relative to the model, so an average would misreport
+    // progress; surface the model download progress directly.
```
barhanc:
```suggestion
```

packages/react-native-executorch/src/models.ts
```diff
@@ -528,6 +530,68 @@ const YOLO26_XLARGE_SEG_640_XNNPACK_FP32: InstanceSegmenterModel<'xyxy', CocoCla
...
+const DISTILUSE_BASE_MULTILINGUAL_CASED_V2_EMBEDDINGS: TextEmbedderModel = {
+  modelPath: `${BASE_URL}-distiluse-base-multilingual-cased-v2/${NEXT_VERSION_TAG}/xnnpack/distiluse_base_multilingual_cased_v2_xnnpack_8da4w.pte`,
+  tokenizerPath: `${BASE_URL}-distiluse-base-multilingual-cased-v2/${NEXT_VERSION_TAG}/tokenizer.json`,
+};
+const CLIP_VIT_BASE_PATCH32_TEXT_EMBEDDINGS: TextEmbedderModel = {
+  modelPath: `${BASE_URL}-clip-vit-base-patch32/${NEXT_VERSION_TAG}/xnnpack/clip_vit_base_patch32_text_xnnpack_fp32.pte`,
+  tokenizerPath: `${BASE_URL}-clip-vit-base-patch32/${NEXT_VERSION_TAG}/tokenizer.json`,
+};
+// Asymmetric retrieval model: prefix queries with `query: ` (the default) and
+// passages with `document: ` via `embed(text, 'document: ')`.
```
barhanc: This should be a `/** */` doc in the `models` object so that it's visible in public api.
```ts
    /**
     * Asymmetric retrieval model: prompt queries with `query: ` (the default)
     * and passages with `document: ` via `embed(text, 'document: ')`.
     */
    LFM2_5_EMBEDDING_350M: {
      ...LFM2_5_EMBEDDING_350M_EMBEDDINGS,
      XNNPACK_8DA4W: LFM2_5_EMBEDDING_350M_EMBEDDINGS,
      MLX_INT4: LFM2_5_EMBEDDING_350M_MLX_INT4,
    },
```

barhanc (commented):
I also don't have an iOS device available to check the MLX, so it has to wait til Monday. Other than that just some small nits, in general looks great.

barhanc (approved):
🚀

PR #1295, opened 2026-06-30
[RNE Rewrite] OCR bucketed

### LINKED ISSUE #1240: [RNE Rewrite] CV - add OCR pipeline implementation
Implement OCR pipeline using the new approach to achieve parity with current implementation.

benITo47:
## Description

 Adds a unified,  OCR + document understanding pipeline to react-native-executorch. Two detector/recognizer families ship out of the box - EasyOCR  and PaddleOCR PP-OCRv6 - behind one createOCR task and a useOCR hook, plus a higher-level document pipeline (createDocumentOCR / useDocumentOCR) that orchestrates orientation correction, UVDoc dewarp, PP-DocLayoutV3 region layout, SLANet table-structure recognition, and reading-order assembly into HTML.

Models use a bucketed PTE contract: each model ships per-size detect_<S> / recognize_<W> methods and the pipeline snaps each image to the closest bucket.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. OCR screen - pick PaddleOCR (XNNPACK/Vulkan) or EasyOCR; select an image from the gallery → Run OCR. Verify detected regions (green overlay) and the recognized text list with per-region confidence/latency.
2. Toggle Vertical text and run a stacked/column image (e.g. signage, container codes) - confirm horizontal reads are unaffected and columns are read top-to-bottom.
3. Document Pipeline screen - run a document photo; verify orientation correction, optional dewarp, layout regions, table HTML, and reading-order output. 

### Checklist


msluszniak:
Firstly, fix the PR description since it is struckthrough and fix clang tidy.

packages/react-native-executorch/cpp/core/model.cpp
barhanc, line 1: Changes to the core should first be introduced in a separate PR. Also let's wait with this one because we will be adding support for dynamic methods and as I understand this is required because of the bucketed approach.

packages/react-native-executorch/cpp/extensions/cv/box_ops.cpp
barhanc, line 1: What is the reason this file changed? It looks to me these are purely stylistic changes.

packages/react-native-executorch/cpp/extensions/cv/image_ops.h
```diff
@@ -9,4 +9,9 @@ void install_toChannelsFirst(facebook::jsi::Runtime &rt, facebook::jsi::Object &
 void install_toChannelsLast(facebook::jsi::Runtime &rt, facebook::jsi::Object &module);
 void install_normalize(facebook::jsi::Runtime &rt, facebook::jsi::Object &module);
 void install_applyColormap(facebook::jsi::Runtime &rt, facebook::jsi::Object &module);
+void install_gridSample(facebook::jsi::Runtime &rt, facebook::jsi::Object &module);
+// Perspective-crop an oriented quad of `src` into the `dst` canvas (crop +
+// resize-to-height + pad/align), HWC uint8. Generic image op; the OCR recognizer
+// uses it to lift each text quad into a fixed-height recognizer canvas.
```
barhanc: Please remove those comments, the JSDoc and the comments inside the particular C++ function implementation should be enough. This creates unnecessary coupling if anything in the implementation changes and we forget to change the comment it might create confusion.

packages/react-native-executorch/cpp/extensions/cv/utils.h
```diff
@@ -18,4 +20,43 @@ inline int dtypeToCvDepth(rnexecutorch::core::types::DType dtype) {
...
+
+// Optional boolean (defaults when absent) — used for flags a caller may omit.
+inline bool getBoolPropOr(facebook::jsi::Runtime &rt, const facebook::jsi::Object &opts,
+                          const char *name, bool fallback) {
+    if (!opts.hasProperty(rt, name) || !opts.getProperty(rt, name).isBool()) {
+        return fallback;
+    }
+    return opts.getProperty(rt, name).asBool();
+}
+
```
barhanc: These are not CV specific. If you want to add JSI conversion functions, these should be added to core.

packages/react-native-executorch/src/core/model.ts
```diff
@@ -119,6 +119,21 @@ export interface Model {
...
 
+  /**
+   * Unloads a single previously-executed method, freeing its memory-planned
+   * activation arena (and, on graph-compiling backends like CoreML, its
+   * compiled graph). The method transparently reloads on its next `execute`.
+   *
+   * Use this to bound native memory when many distinct methods are executed
+   * over a session — e.g. bucketed OCR, where each `detect_<S>`/`recognize_<W>`
+   * size that is ever run would otherwise stay resident for the model's
+   * lifetime.
```
barhanc: Please remove this comment.

packages/react-native-executorch/src/extensions/cv/ops/boxes.ts
```diff
@@ -125,6 +110,31 @@ export function scaleBox<F extends BoxFormat>(
...
 }
 
+/**
+ * Computes the axis-aligned bounding box (`xyxy`) enclosing a set of points,
+ * e.g. the corners of an oriented OCR {@link Quad}.
+ * @category Utils
+ * @param points The points to bound (need not be ordered).
+ * @returns The enclosing bounding box in `xyxy` format.
+ */
+export function boundingBoxOf(points: readonly Point[]): BoundingBox<'xyxy'> {
```
barhanc: Any reason this couldn't take `boxFormat: F` parameter instead of returning xyxy box always.

packages/react-native-executorch/src/extensions/cv/ops/image.ts
```diff
@@ -209,3 +210,194 @@ export function applyColormap(
...
   return rnexecutorchJsi.cv.applyColormap(src, dst, colormap);
 }
+
+/**
+ * Detector heatmap postprocessing mode. `'craft'` consumes a half-resolution
+ * interleaved `[..,Hd,Wd,2]` (text, affinity) map; `'dbnet'` consumes a
+ * full-resolution `[..,H,W]` probability map.
+ * @category Types
+ */
+export type TextBoxMode = 'craft' | 'dbnet';
```
barhanc: There should be no model specific references *anywhere* outside `models.ts`. This is regarding this type and the options objects below. Also these additions shouldnt be in `image.ts`, maybe in `textBoxes.ts` or something.

packages/react-native-executorch/src/extensions/cv/ops/image.ts
```diff
@@ -209,3 +210,194 @@ export function applyColormap(
...
+ * @param src The source image tensor in HWC uint8 layout. Shape [H,W,C].
+ * @param dst The pre-allocated recognizer canvas in HWC uint8 layout, sized
+ * `[recH, bucketW, C]`.
+ * @param quad Eight numbers `[x0,y0,..,x3,y3]` (TL,TR,BR,BL) in `src` pixels.
+ * @param opts Content width, alignment, and padding configuration.
+ * @returns The destination tensor `dst`.
+ */
+export function warpQuad(src: Tensor, dst: Tensor, quad: number[], opts: WarpQuadOptions): Tensor {
+  'worklet';
+  // Defaults live here (the native op requires every field) — see add-native-extension.
```
barhanc: Please clean up those kind of comments.

packages/react-native-executorch/src/extensions/cv/ops/image.ts
```diff
@@ -209,3 +210,194 @@ export function applyColormap(
...
+ * recognizer canvas `dst`, folding crop + resize-to-height + bucket-pad.
+ * @category Typescript API
+ * @param src The source image tensor in HWC uint8 layout. Shape [H,W,C].
+ * @param dst The pre-allocated recognizer canvas in HWC uint8 layout, sized
+ * `[recH, bucketW, C]`.
+ * @param quad Eight numbers `[x0,y0,..,x3,y3]` (TL,TR,BR,BL) in `src` pixels.
+ * @param opts Content width, alignment, and padding configuration.
+ * @returns The destination tensor `dst`.
+ */
+export function warpQuad(src: Tensor, dst: Tensor, quad: number[], opts: WarpQuadOptions): Tensor {
```
barhanc, line 257: Why isn't quad passed here as `Quad` type and unpacked inside?

packages/react-native-executorch/src/extensions/cv/ops/image.ts
```diff
@@ -209,3 +210,194 @@ export function applyColormap(
...
+  /**
+   * Apply a per-timestep softmax so `values` are probabilities (CRNN logits).
+   * Leave `false` for heads that are already softmaxed (SVTR).
+   */
+  readonly softmax?: boolean;
+};
+
+/**
+ * Per-timestep argmax + max value over recognizer logits, computed natively on
+ * the tensor buffer (avoids copying the whole `[T, V]` tensor into JS).
```
barhanc: Maybe it's my lack of understanding of OCR pipeline but this doc doesn't really say much to me what this function does.

packages/react-native-executorch/src/extensions/cv/ops/image.ts
```diff
@@ -209,3 +210,194 @@ export function applyColormap(
...
+    indices.push(flat[i]!);
+    values.push(flat[i + 1]!);
+  }
+  return { indices, values };
+}
+
+/**
+ * Resamples `src` through a backward sampling field (the `torch.grid_sample`
+ * step of a geometric dewarp) into the pre-allocated `dst`, natively via
+ * `cv::remap`.
```
barhanc: Again the docs don't say much. It seems that to understand what this does one must first read `torch.grid_sample` docs in PyTorch and then `cv::remap` docs from OpenCV. Can't we explain it better without referencing two other huge libraries?

packages/react-native-executorch/src/extensions/cv/ops/points.ts
```diff
@@ -9,6 +9,48 @@ export type Point = {
...
+ * @category Utils
+ * @param v The value to clamp.
+ * @param lo The lower bound.
+ * @param hi The upper bound.
+ * @returns `v` constrained to `[lo, hi]`.
+ */
+export function clamp(v: number, lo: number, hi: number): number {
+  'worklet';
+  return v < lo ? lo : v > hi ? hi : v;
+}
```
barhanc: This shouldn't be a public function. If you need this just write a small helper and don't expose it. Also I think writing `Math.max(lo, Math.min(val, hi))` isn't that much hassle.

packages/react-native-executorch/src/extensions/cv/ops/points.ts
```diff
@@ -9,6 +9,48 @@ export type Point = {
...
+
+/**
+ * Per-axis scale and offset that map a `to`-space coordinate back into
+ * `from`-space, for an aspect-preserving letterbox or an axis stretch. The
+ * inverse map is `(coord − offset) / scale` per axis (offset is 0 for stretch).
+ * Shared by {@link scalePoint} and `scaleBox` so the factors are derived once.
```
barhanc: The bottom line is that it computes the *factors*. As it is now it reads more like this function actually does some transformation.

packages/react-native-executorch/src/extensions/cv/ops/points.ts
```diff
@@ -9,6 +9,48 @@ export type Point = {
...
+ * @param resizeMode The resize mode the source was produced with.
+ * @returns The per-axis `scaleX`/`scaleY` and `offsetX`/`offsetY`.
+ */
+export function resizeFactors(
+  from: { readonly width: number; readonly height: number },
+  to: { readonly width: number; readonly height: number },
+  resizeMode: Exclude<ResizeMode, 'crop'>
+): { scaleX: number; scaleY: number; offsetX: number; offsetY: number } {
+  'worklet';
+  if (resizeMode === 'letterbox') {
```
barhanc: I would prefer explicit `switch` statement with the resize modes listed.

packages/react-native-executorch/src/extensions/cv/tasks/documentHelpers.ts
```diff
@@ -0,0 +1,179 @@
+import type { ImageBuffer } from '../image';
+import { FORMAT_CHANNELS } from '../ops/image';
+import { boundingBoxOf, type BoundingBox } from '../ops/boxes';
+import { clamp } from '../ops/points';
+import type { OCRDetection } from './ocr';
+
+// Crops an axis-aligned region out of an ImageBuffer (pure pixel slice, same
+// format). Used to feed a layout region (e.g. a table) to another model.
+export function cropImageBuffer(input: ImageBuffer, bbox: BoundingBox<'xyxy'>): ImageBuffer {
```
barhanc: There is a `restrictToBox` function added recently, perhaps it would be useful here. Also why `xyxy` and not generic?

packages/react-native-executorch/src/extensions/cv/tasks/documentHelpers.ts
```diff
@@ -0,0 +1,179 @@
...
+  const out = new Uint8Array(cw * chh * ch);
+  for (let y = 0; y < chh; y++) {
+    const srcStart = ((y0 + y) * width + x0) * ch;
+    out.set(data.subarray(srcStart, srcStart + cw * ch), y * cw * ch);
+  }
+  return { data: out, width: cw, height: chh, format, layout: input.layout };
+}
+
+// Rotates an ImageBuffer clockwise by 0/90/180/270 degrees (pure pixel reindex).
+export function rotateImageBuffer(input: ImageBuffer, degCW: 0 | 90 | 180 | 270): ImageBuffer {
```
barhanc: This seems like something that is 1) probably available in OpenCV and 2) should be added as native image op, instead of doing the for loop over every pixel in TS.

packages/react-native-executorch/src/extensions/cv/tasks/documentHelpers.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Index of the maximum value in `arr[offset, offset+len)` (single pass, no
+ * allocation). Shared by the layout / orientation / table decoders.
+ * @category Typescript API
+ * @param arr The array to scan.
+ * @param offset The start index of the window.
+ * @param len The window length.
+ * @returns The index (relative to `offset`) of the maximum value.
+ */
+export function argmaxRange(arr: ArrayLike<number>, offset: number, len: number): number {
```
barhanc: Using tensors this can be done right now using `copyTo` and `argmax`.

packages/react-native-executorch/src/extensions/cv/tasks/documentHelpers.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (v.length <= k) {
+    return v;
+  }
+  const cuts = v
+    .map((x, i) => ({ i, g: i ? x - v[i - 1]! : -1 }))
+    .slice(1)
+    .sort((a, b) => b.g - a.g)
+    .slice(0, k - 1)
+    .map((o) => o.i)
+    .sort((a, b) => a - b);
```
barhanc: Some comment would be in place, because it looks very cryptic as of now.

packages/react-native-executorch/src/extensions/cv/tasks/documentHelpers.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+  return b;
+}
+
+/**
+ * Reconstructs a table as filled HTML from the SLANet structure skeleton and the
+ * region's OCR lines. Grid size comes from the structure (`<tr>` count, max `<td>`
+ * per row); box centers are split into that many rows/cols at the largest gaps and
+ * each box placed in its nearest cell (global column centers keep columns aligned).
+ * Falls back to a document-order fill when there is no grid.
```
barhanc: Either add a link explaining this SLANet architecture or describe it in the jsdoc.

packages/react-native-executorch/src/extensions/cv/tasks/documentHelpers.ts
barhanc, line 1: Why is this file in `tasks/`?

packages/react-native-executorch/src/extensions/cv/tasks/documentOCR.ts
```diff
@@ -0,0 +1,263 @@
...
+  readonly dewarp?: boolean;
+};
+
+// Minimum orientation-classifier confidence (softmax of the argmax class) to act
+// on a non-zero rotation. Mirrors PaddleOCR's pipeline gate: out-of-distribution
+// inputs (photos, non-documents) produce low-confidence argmaxes that spuriously
+// flip the page, so below this we treat the page as already upright (0°). Set high
+// (0.85) — genuine documents score >0.95, leaving margin to reject OOD frames that
+// can still land ~0.74.
+const ORIENTATION_MIN_CONFIDENCE = 0.85;
```
barhanc: Shouldn't this be a configurable option passed to task constructor?

packages/react-native-executorch/src/extensions/cv/tasks/documentOCR.ts
```diff
@@ -0,0 +1,263 @@
...
+
+const isTableLabel = (label: unknown): boolean => {
+  'worklet';
+  return String(label) === 'table';
+};
```
barhanc: I don't think a single comparison justifies a separate function.

packages/react-native-executorch/src/extensions/cv/tasks/supporting.ts
barhanc, line 1: Please change a name of the file and inside it from "supporting" to something that actually conveys what this does. Also if it's only viable for a single model (SLANet) then make the name reflect that, if not then all the constants and references to the particular model should be removed.

packages/react-native-executorch/src/extensions/cv/tasks/ocrPipeline.ts
barhanc, line 1: Some general remarks, because I'm not going to mark every tiny thing as this whole file needs to be refactored:
1. Why is this in `tasks/`?
2. You cannot have model specific constants in here, as I noted in other files `models.ts` is a single source of truth for model specific configuration unless your pipeline is a single-model specific one like `kokoroTTS`.
3. Multiple exported public functions do not have proper jsdoc, just `//` comments.
4. You break multiple times the guideline of  not returning tensors in public functions (this moves the dispose responsibility to consumer).

packages/react-native-executorch/src/extensions/cv/tasks/ocrHelpers.ts
```diff
@@ -0,0 +1,542 @@
+import type { Quad } from '../ops/image';
+import { scalePoint, clamp, type Point } from '../ops/points';
+
+/**
+ * Worklet-safe millisecond clock for in-pipeline profiling. Prefers
+ * `performance.now()` (sub-ms) and falls back to `Date.now()`.
+ * @category Typescript API
+ * @returns The current time in milliseconds.
+ */
+export function nowMs(): number {
```
barhanc: This shouldn't be a public function!

packages/react-native-executorch/src/extensions/cv/tasks/ocrHelpers.ts
```diff
@@ -0,0 +1,542 @@
...
+// when their vertical extents overlap by at least this fraction of the shorter
+// box's height. Stable across models — not worth exposing.
+const READING_COL_GAP_FRAC = 0.06;
+const READING_ROW_OVERLAP_FRAC = 0.3;
+
+type ReadingBox = { xmin: number; ymin: number; xmax: number; ymax: number };
+
+// Axis-aligned bounds of a quad. Defined before readingOrderIndices so the worklet
+// plugin captures it (a referenced worklet must precede its caller in source order).
+function bboxOfQuad(quad: readonly Point[]): ReadingBox {
```
barhanc: Isn't this duplicated from `ops/`?

packages/react-native-executorch/src/extensions/cv/tasks/ocrHelpers.ts
```diff
@@ -0,0 +1,542 @@
...
+    if (p.y < ymin) ymin = p.y;
+    if (p.x > xmax) xmax = p.x;
+    if (p.y > ymax) ymax = p.y;
+  }
+  return { xmin, ymin, xmax, ymax };
+}
+
+// Horizontal sort key for a box: edge sum = 2× center-x (monotonic in center,
+// avoids the divide). Module-level so it isn't re-created per readingOrderIndices.
+function xSum(b: ReadingBox): number {
```
barhanc: Please remove this function.

packages/react-native-executorch/src/extensions/cv/tasks/ocrHelpers.ts
```diff
@@ -0,0 +1,542 @@
...
+  return out;
+}
+
+// Vertical-text grouping tuning. A box wider than this × its height is a
+// horizontal line, never a stacked-column glyph. A box joins a column when its
+// x-span overlaps the column's by VERTICAL_X_OVERLAP of the narrower width and the
+// y-gap is within VERTICAL_Y_GAP × its height (loose — signage spacing varies).
+const VERTICAL_GLYPH_ASPECT = 1.6;
+const VERTICAL_X_OVERLAP = 0.25;
+const VERTICAL_Y_GAP = 2.5;
```
barhanc: Where are these numbers coming from?

packages/react-native-executorch/src/index.ts
```diff
@@ -6,6 +6,10 @@ export * from './hooks/useInstanceSegmenter';
 export * from './hooks/useKeypointDetector';
 export * from './hooks/useObjectDetector';
 export * from './hooks/useTokenizer';
+export * from './hooks/useOCR';
+// Layout + supporting (orientation/dewarp/table) are internal to the document
+// pipeline — consumed by useDocumentOCR, not exposed as standalone hooks.
```
barhanc: Please remove this comment.

packages/react-native-executorch/src/models.ts
barhanc, line 1: Please follow the style of the file regarding comments, url definitions, constants' naming, etc. Also perhaps needing a multiple line comment before a config suggests the config is not descriptive enough.

packages/react-native-executorch/src/extensions/cv/tasks/ocr/charsets.ts
barhanc, line 1: This file shouldn't be in the root.

packages/react-native-executorch/src/extensions/cv/tasks/ocrHelpers.ts
```diff
@@ -0,0 +1,542 @@
...
+}
+
+/**
+ * The axis-aligned bounding quad (TL,TR,BR,BL) enclosing a set of quads. Returns a
+ * zero quad for empty input.
+ * @category Typescript API
+ * @param quads The quads to enclose.
+ * @returns The four corners of the enclosing box, ordered TL, TR, BR, BL.
+ */
+export function boundingQuadOf(quads: readonly (readonly Point[])[]): Point[] {
```
barhanc: Why is this here and not e.g. where the `boundingBoxOf` helper is defined, in general the helpers are all over the place.

packages/react-native-executorch/src/extensions/cv/tasks/ocr.ts
barhanc, line 1: Some general remarks because there are multiple problems:
- As in other files you specify model-specific constants and references inside here. The pattern `readonly detectorKind: 'craft' | 'dbnet' | 'custom';` is especially bad. The pipeline itself should be model agnositc, it should only validate a specific contract; from the point of view of the pipeline each model is `'custom'`. If the pipeline cannot be implemented as such because different models require very specific and non-shareable pre/post processing steps (that cannot be neatly factored out into common pipeline) then the pipeline should be single-model specific and its name should reflect that.
- As I already remarked in different file you break the guideline of never returning tensors from public functions.
- The `relaseMethodsWorklet` is unused in `useOCR` so why is it returned?
- Smaller nits: some of the variable names like `o` aren't very readable; some of the docs like inside the options objects are not needed, the fields should have descriptive-enough names

barhanc (changes requested):
- I only went over the library implementation, didn't test apps, as there are quite some changes to be made. 
- Why do we have both OCR and DocumentOCR, what are the use-cases where one would want the OCR without the additional processing and if there are such wouldn't it be better to have it configurable from a single `useOCR` hook?
- Since it seems that OCR is quite a large addition I think we can add `ocr/` directory to `cv/` that will host all the required pre/post processing helpers.
- Please fix the cspell warnings.
- Please fix the PR description.
- The added repo names on huggingface do not conform to snake-case convention.

barhanc:
@benITo47 fyi: we merged #1296 which adds utilities for JSI conversions that you needed

barhanc:
@benITo47 fyi: we merged #1292 which adds dynamic shape support

PR #1296, opened 2026-07-01
[RNE Rewrite] feat: add input validation and JSI conversion utilities

### LINKED ISSUE #1268: [RNE Rewrite] Recheck and streamline the core native implementation
We copied the core implementation of native `model.cpp` and `tensor.cpp` in the branch scaffold PR from the PoC, so that we have a base to work on, but we should recheck and review that implementation for any possible issues.

### LINKED ISSUE #1270: [RNE Rewrite] Add macros for input validation in native C++ implementation
Right now every function implemented in C++ that gets exposed to the TS requires substantial, often duplicated validation logic for tensors, models, etc. The scope of this issue is to add lightweight macros so that the boilerplate validation code can be minimized.

barhanc:
## Description

Refactors the C++ layer by pulling two families of scattered, copy-pasted utilities out of individual translation units and into purpose-built, reusable modules:

- `cpp/core/conversions.{h,cpp}`  JSI -> C++ type coercions
- `cpp/core/tensor_helpers.{h,cpp}`: tensor acquisition, locking, and shape/dtype validation

### Motivation

The old code had two recurring problems that became impossible to ignore as the extension surface grew:

#### JSI coercion was inlined everywhere
Every file that needed to read a number, string, or array out of a `jsi::Value` had its own copy of the same defensive checks, cast chains, and error-message construction. A bug or policy change (e.g. how `null` is treated, or which format an error message takes) had to be hunted down and patched in every translation unit independently.

#### Tensor validation was duplicated across all extension operations
Every operation in `box_ops`, `image_ops`, `operations`, and `tokenizer` that accepted a `Tensor` argument re-implemented the same sequence: unwrap the `jsi::Value`, cast to `TensorHostObject`, acquire the appropriate
shared/exclusive lock, and assert dtype + shape constraints. This led to subtle inconsistencies in error messages and locking strategies between operations. 

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Build all example apps on both Android and iOS and test that nothing is broken.

### Related issues

Closes #1270
Closes #1268 

### Checklist

### Additional notes

Since this PR changes quite a lot of stuff in `core/` it can also be used for #1268

### DIFF
--- a/packages/react-native-executorch/cpp/core/conversions.cpp
+++ b/packages/react-native-executorch/cpp/core/conversions.cpp
@@ -44,14 +44,13 @@ uint8_t asType<uint8_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Val
     return static_cast<uint8_t>(v);
 }
 
-template <>
-size_t asType<size_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
-    double v = asType<double>(rt, ctx, val);
-    if (std::isnan(v) || v < 0.0) {
-        throw jsi::JSError(rt, ctx + " must be a non-negative integer");
-    }
-    return static_cast<size_t>(v);
-}
+// Note: size_t is intentionally not specialised here.
+// On Linux x86_64 size_t and uint64_t are both `unsigned long`, so adding a
+// size_t specialisation would produce a duplicate-explicit-specialisation ODR
+// error on that platform. asType<size_t> routes through the uint64_t
+// specialisation on Linux (same type) and is a separate instantiation on
+// macOS (where uint64_t is unsigned long long). Either way the semantics —
+// non-negative double → cast — are identical.
 
 template <>
 bool asType<bool>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
--- a/packages/react-native-executorch/cpp/core/conversions.cpp
+++ b/packages/react-native-executorch/cpp/core/conversions.cpp
@@ -92,4 +92,28 @@ jsi::ArrayBuffer asType<jsi::ArrayBuffer>(jsi::Runtime &rt, const std::string &c
     return val.asObject(rt).getArrayBuffer(rt);
 }
 
+template <>
+jsi::Object asType<jsi::Object>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    if (!val.isObject()) {
+        throw jsi::JSError(rt, ctx + " must be an object");
+    }
+    return val.asObject(rt);
+}
+
+template <>
+jsi::Array asType<jsi::Array>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    if (!val.isObject() || !val.asObject(rt).isArray(rt)) {
+        throw jsi::JSError(rt, ctx + " must be an Array");
+    }
+    return val.asObject(rt).asArray(rt);
+}
+
+template <>
+jsi::Function asType<jsi::Function>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    if (!val.isObject() || !val.asObject(rt).isFunction(rt)) {
+        throw jsi::JSError(rt, ctx + " must be a function");
+    }
+    return val.asObject(rt).asFunction(rt);
+}
+
 } // namespace rnexecutorch::core::conversions
--- a/packages/react-native-executorch/cpp/core/conversions.cpp
+++ b/packages/react-native-executorch/cpp/core/conversions.cpp
@@ -1,0 +1,71 @@
+#include "conversions.h"
+#include <cmath>
+
+namespace rnexecutorch::core::conversions {
+
+template <>
+double asType<double>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    if (!val.isNumber()) {
+        throw jsi::JSError(rt, ctx + " must be a number");
+    }
+    return val.asNumber();
+}
+
+template <>
+float asType<float>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    return static_cast<float>(asType<double>(rt, ctx, val));
+}
+
+template <>
+int32_t asType<int32_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    return static_cast<int32_t>(asType<double>(rt, ctx, val));
+}
+
+template <>
+int64_t asType<int64_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    return static_cast<int64_t>(asType<double>(rt, ctx, val));
+}
+
+template <>
+uint64_t asType<uint64_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    double v = asType<double>(rt, ctx, val);
+    if (std::isnan(v) || v < 0.0) {
+        throw jsi::JSError(rt, ctx + " must be a non-negative integer");
+    }
+    return static_cast<uint64_t>(v);
+}
+
+template <>
+uint8_t asType<uint8_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    double v = asType<double>(rt, ctx, val);
+    if (std::isnan(v) || v < 0.0 || v > 255.0) {
+        throw jsi::JSError(rt, ctx + " must be between 0 and 255");
+    }
+    return static_cast<uint8_t>(v);
+}
+
+template <>
+size_t asType<size_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    double v = asType<double>(rt, ctx, val);
+    if (std::isnan(v) || v < 0.0) {
+        throw jsi::JSError(rt, ctx + " must be a non-negative integer");
+    }
+    return static_cast<size_t>(v);
+}
+
+template <>
+bool asType<bool>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    if (!val.isBool()) {
+        throw jsi::JSError(rt, ctx + " must be a boolean");
+    }
+    return val.asBool();
+}
+
+template <>
+std::string asType<std::string>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    if (!val.isString()) {
+        throw jsi::JSError(rt, ctx + " must be a string");
+    }
+    return val.asString(rt).utf8(rt);
+}
+} // namespace rnexecutorch::core::conversions
--- a/packages/react-native-executorch/cpp/core/conversions.cpp
+++ b/packages/react-native-executorch/cpp/core/conversions.cpp
@@ -1,7 +1,18 @@
 #include "conversions.h"
 #include <cmath>
+#include <limits>
 
 namespace rnexecutorch::core::conversions {
+
+constexpr double kMaxInt64Double = static_cast<double>(std::numeric_limits<int64_t>::max());
+constexpr double kMinInt64Double = static_cast<double>(std::numeric_limits<int64_t>::min());
+constexpr double kMaxUint64Double = static_cast<double>(std::numeric_limits<uint64_t>::max());
+
+constexpr double kMinInt32Double = static_cast<double>(std::numeric_limits<int32_t>::min());
+constexpr double kMaxInt32Double = static_cast<double>(std::numeric_limits<int32_t>::max());
+
+constexpr double kMinUint8Double = static_cast<double>(std::numeric_limits<uint8_t>::min());
+constexpr double kMaxUint8Double = static_cast<double>(std::numeric_limits<uint8_t>::max());
 
 template <>
 double asType<double>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
@@ -18,18 +29,26 @@ float asType<float>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &
 
 template <>
 int32_t asType<int32_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
-    return static_cast<int32_t>(asType<double>(rt, ctx, val));
+    double v = asType<double>(rt, ctx, val);
+    if (std::isnan(v) || std::isinf(v) || v != std::trunc(v) || v < kMinInt32Double || v > kMaxInt32Double) {
+        throw jsi::JSError(rt, ctx + " must be a 32-bit integer");
+    }
+    return static_cast<int32_t>(v);
 }
 
 template <>
 int64_t asType<int64_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
-    return static_cast<int64_t>(asType<double>(rt, ctx, val));
+    double v = asType<double>(rt, ctx, val);
+    if (std::isnan(v) || std::isinf(v) || v != std::trunc(v) || v < kMinInt64Double || v >= kMaxInt64Double) {
+        throw jsi::JSError(rt, ctx + " must be a 64-bit integer");
+    }
+    return static_cast<int64_t>(v);
 }
 
 template <>
 uint64_t asType<uint64_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
     double v = asType<double>(rt, ctx, val);
-    if (std::isnan(v) || v < 0.0) {
+    if (std::isnan(v) || std::isinf(v) || v != std::trunc(v) || v < 0.0 || v >= kMaxUint64Double) {
         throw jsi::JSError(rt, ctx + " must be a non-negative integer");
     }
     return static_cast<uint64_t>(v);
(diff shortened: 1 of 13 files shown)


barhanc:
@msluszniak This is still a draft and there are still quite some rough edges to polish, but could you take a look if this approach is something you would agree on and if there are any other aspects I missed. I was thinking that we could also use this PR for #1268 since adding the validation/conversion helpers modifies quite a lot of core code so it would be good to review and streamline also the other aspects.

packages/react-native-executorch/cpp/core/conversions.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+template <>
+uint64_t asType<uint64_t>(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    double v = asType<double>(rt, ctx, val);
+    if (std::isnan(v) || v < 0.0) {
+        throw jsi::JSError(rt, ctx + " must be a non-negative integer");
+    }
+    return static_cast<uint64_t>(v);
+}
```
msluszniak, line 55: there should also be check if double is in range of uint64_t, otherwise cast is undefined behaviour. This is the case for all analogical casts scenarios.

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -0,0 +1,99 @@
...
+    if (!tensor->data_) {
+        throw jsi::JSError(rt, std::format("{} tensor has been disposed", name));
+    }
+    return lock;
+}
+
+void checkNotSameTensor(jsi::Runtime &rt,
+                        const std::string &name1, const std::shared_ptr<TensorHostObject> &t1,
+                        const std::string &name2, const std::shared_ptr<TensorHostObject> &t2) {
+    if (t1.get() == t2.get()) {
```
msluszniak: Why not `t1 == t2`?

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -0,0 +1,99 @@
...
+                shapeStr += ", ";
+            }
+            const auto &dim = expectedShape->at(i);
+            if (std::holds_alternative<int32_t>(dim)) {
+                shapeStr += std::to_string(std::get<int32_t>(dim));
+            } else {
+                shapeStr += std::get<std::string>(dim);
+            }
+        }
+        shapeStr += "]";
```
msluszniak:
```suggestion
            const auto &dim = expectedShape->at(i);
            if (std::holds_alternative<int32_t>(dim)) {
                shapeStr += std::to_string(std::get<int32_t>(dim));
            } else {
                shapeStr += std::get<std::string>(dim);
            }
            shapeStr += ", ";
        }
        shapeStr.pop_back();
        shapeStr += "]";
```

msluszniak (commented):
I like this concept. I combines elegance of short macros with flexibility of manual code.  Just very small nits, will dive into it later today / tomorrow. I like usage of `std::format`, we previously tried use it in old codebase, but because it required some version of iOS we eventually dropped it, but I don't think it will be a case now.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -250,83 +243,54 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
+                const auto ctx = std::format("execute: {}", ident);
 
                 if (!tag.ok()) {
                     const std::string errorMsg = executorch::runtime::to_string(tag.error());
-                    throw jsi::JSError(rt, "execute: Failed to get input tag for inputs[" +
-                                               std::to_string(i) + "]: " + errorMsg);
+                    throw jsi::JSError(rt, std::format("{}: Failed to get input tag: {}", ctx, errorMsg));
                 }
 
                 switch (tag.get()) {
```
msluszniak: This input switch dropped the old `Tag::None` case (accept `null` → `EValue()`), so None-tagged/optional inputs now hit `default:` and throw "Unsupported input type". The output switch below still handles None. Was dropping it intentional?

packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -43,54 +51,22 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name)
...
+            auto srcLock = tryLockShared(rt, "copyTo: self", self);
+            auto dstLock = tryLockUnique(rt, "copyTo: dst", dst);
 
-                copyLen -= srcOffset;
-
-                if (optsObj.hasProperty(rt, "length")) {
-                    copyLen = static_cast<size_t>(optsObj.getProperty(rt, "length").asNumber());
-                }
-            }
+            const jsi::Object optsObj = (count == 2) ? args[1].asObject(rt) : jsi::Object(rt);
```
msluszniak: If `count == 2` but `args[1]` isn't an object, `asObject(rt)` throws a generic JSI error; the old code silently ignored a non-object opts arg. Minor, but worth a contextual check to match the rest of the PR.

packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -109,49 +85,16 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name)
...
-
-            if (dataObj.hasProperty(rt, "byteLength")) {
-                auto byteLengthValue = dataObj.getProperty(rt, "byteLength");
-                if (!byteLengthValue.isNumber()) {
-                    throw jsi::JSError(rt, "setData: Expected 'byteLength' to be a number");
-                }
-                byteLength = static_cast<size_t>(byteLengthValue.asNumber());
-            }
+            const auto dataObj = args[0].asObject(rt);
+            const auto buffer = dataObj.getProperty(rt, "buffer").asObject(rt).getArrayBuffer(rt);
```
msluszniak: getData/setData lost the old "Expected a TypedArray with a 'buffer' property" message — a non-TypedArray now yields a generic JSI error. Slightly ironic for a PR standardizing errors; consider an `asTypedArray` helper in `conversions`.

packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -227,10 +137,6 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name)
                 throw jsi::JSError(rt, "through: Usage: through(fn, ...args)");
             }
 
-            if (!args[0].isObject() || !args[0].asObject(rt).isFunction(rt)) {
-                throw jsi::JSError(rt, "through: First argument must be a function");
-            }
-
             auto fn = args[0].asObject(rt).asFunction(rt);
```
msluszniak: through/throughIf lost the explicit "must be a function" guard; a non-function arg now throws a generic `asFunction` error instead of the contextual one.

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -0,0 +1,99 @@
...
+    auto dtype = tensor->dtype_;
+    auto shape = tensor->shape_;
+
+    if (expectedDtype && dtype != *expectedDtype) {
+        throw jsi::JSError(rt, std::format("{} must be of type {}",
+                                           name, types::toString(*expectedDtype)));
+    }
+
+    if (expectedShape) {
+        std::string shapeStr = "[";
```
msluszniak: `shapeStr` is built unconditionally on every call, including the success path — an avoidable allocation for per-frame CV ops. Move it into the error branches (or a lazy lambda).

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -0,0 +1,99 @@
...
+            const auto &dim = expectedShape->at(i);
+            if (std::holds_alternative<int32_t>(dim)) {
+                if (shape[i] != std::get<int32_t>(dim)) {
+                    throw jsi::JSError(rt, std::format("{} must have shape {} (dim {} mismatch: expected {}, got {})",
+                                                       name, shapeStr, i, std::get<int32_t>(dim), shape[i]));
+                }
+            } else {
+                const auto &symbol = std::get<std::string>(dim);
+                if (symbolMap.contains(symbol) && shape[i] != symbolMap[symbol]) {
+                    throw jsi::JSError(rt, std::format("{} must have shape {} (dim '{}' mismatch: expected {}, got {})",
```
msluszniak: The message says `dim '{}'` but is fed `i`, so it prints the index (`dim '2'`) rather than the symbol name (`dim 'N'`). Pass `symbol` instead of `i`.

msluszniak (commented):
Another batch of comments.

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
barhanc: It should print the index, the quotes were a typo.

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
barhanc: Changed to `t1 == t2`.

barhanc:
The `core/` changes are ready for the next pass. Later I will also polish the code in `extensions/`.

packages/react-native-executorch/cpp/core/conversions.cpp
```diff
@@ -0,0 +1,119 @@
+#include "conversions.h"
+#include <cmath>
+#include <limits>
+
+namespace rnexecutorch::core::conversions {
+
+constexpr double kMaxInt64Double = static_cast<double>(std::numeric_limits<int64_t>::max());
```
msluszniak: Shouldn't clang tidy modernize use auto flag trigger here?

packages/react-native-executorch/cpp/core/conversions.h
```diff
@@ -0,0 +1,138 @@
...
+ * @return A std::vector containing the converted elements.
+ */
+template <typename T>
+std::vector<T> asVector(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
+    auto arr = asType<jsi::Array>(rt, ctx, val);
+    std::vector<T> vec;
+    const size_t len = arr.size(rt);
+    vec.reserve(len);
+    for (size_t i = 0; i < len; ++i) {
+        vec.push_back(asType<T>(rt, ctx + "[" + std::to_string(i) + "]", arr.getValueAtIndex(rt, i)));
```
msluszniak: Why not std::format here?

packages/react-native-executorch/cpp/core/conversions.h
```diff
@@ -0,0 +1,138 @@
...
+ * @param rt The JSI runtime instance.
+ * @param ctx Context description used for error messages.
+ * @param obj The JSI object containing the property.
+ * @param propName The name of the property to retrieve.
+ * @return The converted property value.
+ */
+template <typename T>
+T getRequiredProperty(jsi::Runtime &rt, const std::string &ctx, const jsi::Object &obj, const std::string &propName) {
+    if (!obj.hasProperty(rt, propName.c_str())) {
+        throw jsi::JSError(rt, ctx + ": option '" + propName + "' is required");
```
msluszniak: Why not std::format in these two?

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -1,21 +1,78 @@
...
 #include <executorch/runtime/core/tag.h>
 
+namespace {
+namespace jsi = facebook::jsi;
+namespace types = rnexecutorch::core::types;
+
+template <typename T>
+T unwrap(jsi::Runtime &rt, const std::string &ctx, executorch::runtime::Result<T> result) {
+    if (!result.ok()) {
+        throw jsi::JSError(rt, ctx + ": " + executorch::runtime::to_string(result.error()));
```
msluszniak: Std::format

packages/react-native-executorch/cpp/core/conversions.h
```diff
@@ -0,0 +1,138 @@
...
+template <typename T>
+std::optional<T> getOptionalProperty(jsi::Runtime &rt, const std::string &ctx, const jsi::Object &obj, const std::string &propName) {
+    if (!obj.hasProperty(rt, propName.c_str())) {
+        return std::nullopt;
+    }
+    auto val = obj.getProperty(rt, propName.c_str());
+    if (val.isUndefined() || val.isNull()) {
+        return std::nullopt;
+    }
+    return asType<T>(rt, ctx + ": option '" + propName + "'", val);
```
msluszniak: std::format

packages/react-native-executorch/cpp/core/tensor.h
```diff
@@ -1,43 +1,47 @@
...
+class TensorHostObject : public jsi::HostObject,
+                         public std::enable_shared_from_this<TensorHostObject> {
 public:
-    rnexecutorch::core::types::DType dtype_;
-    std::vector<std::int32_t> shape_;
-    size_t numel_;
+    const types::DType dtype_;
+    const std::vector<std::int32_t> shape_;
+    const size_t numel_;
+    const size_t size_;
```
msluszniak, line 32: I think that clang tidy checks does not work on every file. For example here there should be no const members, there is a clang tidy check for that: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html. And also all members are public.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
```diff
@@ -334,79 +241,33 @@ void install_toChannelsFirst(jsi::Runtime &rt, jsi::Object &module) {
...
-        std::vector<::cv::Mat> channels;
-        ::cv::split(srcMat, channels);
+            const ::cv::Mat srcMat(srcH, srcW, cvType, src->data_.get());
+            std::vector<::cv::Mat> channels;
+            ::cv::split(srcMat, channels);
 
-        const size_t hw = static_cast<size_t>(srcH) * static_cast<size_t>(srcW);
-        const size_t elemSize = rnexecutorch::core::types::elementSize(src->dtype_);
-        uint8_t *dstPtr = dst->data_.get();
+            const size_t hw = static_cast<size_t>(srcH) * static_cast<size_t>(srcW);
```
msluszniak, line 262: Same here, modernize use auto. We need to fix clang tidy checks in CI

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
```diff
@@ -20,7 +20,7 @@ class TokenizerHostObject : public facebook::jsi::HostObject,
     std::vector<facebook::jsi::PropNameID> getPropertyNames(facebook::jsi::Runtime &rt) override;
 
 private:
-    std::string tokenizerPath_;
+    const std::string tokenizerPath_;
```
msluszniak: again, do not use const in members.

packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -168,54 +117,21 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name)
...
             if (byteLength != self->size_) {
-                const std::string errorMsg = "getData: Data size mismatch: TypedArray is " + std::to_string(byteLength) +
-                                             " bytes, but Tensor requires " + std::to_string(self->size_) +
-                                             " bytes.";
-                throw jsi::JSError(rt, errorMsg);
+                throw jsi::JSError(rt, std::format("getData: Data size mismatch: TypedArray is {} bytes, but Tensor requires {} bytes.",
+                                                   byteLength, self->size_));
             }
 
             std::memcpy(buffer.data(rt) + byteOffset, self->data_.get(), byteLength);
```
msluszniak, line 143: `byteOffset` is read from the object but never bounds-checked against `buffer.size()` — only `byteLength == self->size_` is enforced. A caller passing `{buffer, byteOffset: <large>}` makes this memcpy write past the end of the ArrayBuffer (OOB write; setData at L106 is the symmetric OOB read). Worth a `byteOffset + byteLength > buffer.size(rt)` guard in both.

packages/react-native-executorch/cpp/core/tensor.cpp
```diff
@@ -43,61 +54,32 @@ jsi::Value TensorHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name)
...
-
-                copyLen -= srcOffset;
-
-                if (optsObj.hasProperty(rt, "length")) {
-                    copyLen = static_cast<size_t>(optsObj.getProperty(rt, "length").asNumber());
-                }
-            }
-
-            if (srcOffset + copyLen > self->numel_) {
+            if (offset + length > self->numel_) {
```
msluszniak: This bounds check is overflow-prone: with a large `offset`, `numel_ - offset` (L68) underflows and `offset + length` can wrap past `numel_`, so the check passes and the memcpy at L80 reads out of bounds. Safer as `offset > self->numel_ || length > self->numel_ - offset`.

msluszniak (commented):
Two more comments

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
barhanc: Reverted `const` here.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
barhanc: I think it doesn't trigger here since it's a mul expression not a static cast assignment. Checked it locally and clang-tidy properly catches e.g. `const double doubleVar = static_cast<double>(someVar)`.

packages/react-native-executorch/cpp/core/tensor.h
barhanc: The `TensorHostObject` is non-copyable and non-movable, thus the clang-tidy checks don't trigger for it. I've decided to leave these members as `const` since it's slightly more convenient to use as compared to adding getters (e.g. writing `->shape[dim]` vs `->shape()[dim]` and gives compiler-enforced immutability.

packages/react-native-executorch/cpp/core/conversions.h
barhanc: Fixed.

packages/react-native-executorch/cpp/core/conversions.h
barhanc: Fixed.

packages/react-native-executorch/cpp/core/conversions.cpp
barhanc: I've checked it locally and it seems that modernize-use-auto doesn't trigger on variables in global namespace scope. clang-tidy does trigger on such assignment if it were inside a function. Changed it manually to `auto`.

packages/react-native-executorch/cpp/extensions/cv/image_ops.cpp
msluszniak: Ok, good to know.

msluszniak:
No more comments from my side, I'm checking demo apps on physical android and ios simulator.

msluszniak:
One comment not related to the changes but spotted in demo app testing, can we fix the overlapping android nav bar with buttons in classifications section?

msluszniak:
Maybe a bit loose requirement on minimal confidence since 28% was accepted, do you remember what is the actual thershold? 
image

msluszniak (approved):
Two small comments which where strictly demo app related, so they might be addressed in a separate PR. 🚀

barhanc:
> Maybe a bit loose requirement on minimal confidence since 28% was accepted, do you remember what is the actual thershold? image

It is specified in the `models.ts` (advantage of the new implementation -- all model specific properties live in one file), assuming this is YOLO it's set to 0.25 --- definietly too low.

PR #1298, opened 2026-07-02
[RNE Rewrite] feat: add voice activity detection pipeline

### LINKED ISSUE #1249: [RNE Rewrite] Speech - add VAD pipeline implementation
Implement VAD pipeline using the new approach to achieve parity with current implementation.

msluszniak:
## Description

Adds a Voice Activity Detection (VAD) task pipeline and a corresponding `speech` example app. Chunked inference, segment postprocessing and streaming run in TypeScript on top of the core `model.execute` primitive. The per-frame feature extraction (framing, mean-removal, pre-emphasis, Hann window) is a native `speech.frameWaveform` C++ op: on device it dominated a `detect()` call (~86%, ~40&nbsp;ms of Hermes vs ~6&nbsp;ms for the model forward pass), so per the extension guidelines it lives in C++. It writes straight into the pre-allocated model-input tensor, fusing mean-removal + pre-emphasis + Hann into one dependency-free (vectorizable) pass; framing drops to ~3.4&nbsp;ms (~12×), below the model's own inference cost.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

- [ ] Build the `speech` app on iOS and Android
- [ ] Test the Voice Activity Detection screen on a physical device (mic + xnnpack; simulator can't record)
- [ ] Verify speech toggles SPEAKING/SILENT and logs begin/end events
- [ ] Check the HF repo: https://huggingface.co/software-mansion/react-native-executorch-fsmn-vad

### Related issues

Closes #1249

### Checklist

### Additional notes

- Depends on the per-method `get_dynamic_dims_forward` input validation from the embeddings PR (#1292): VAD feeds a variable-length `[frames, 512]` input tensor per chunk. Outputs are still validated exactly, so the output tensor is pre-allocated at the model-declared shape. Requires #1292 to land. The `fsmn-vad` model is re-exported (tag `v0.10.0`) with a `get_dynamic_dims_forward` method returning int32 `[rank, 3]` bounds per input.
- Segments are returned in seconds (the old native path returned raw sample indices).
- The FSMN output contract is assumed to be `[1, frames, classes]` with class 0 = non-speech (`speech = 1 - p0`), matching the current native implementation.


packages/react-native-executorch/src/models.ts
```diff
@@ -6,6 +6,7 @@ import type { KeypointDetectorModel } from './extensions/cv/tasks/keypointDetect
 import type { InstanceSegmenterModel } from './extensions/cv/tasks/instanceSegmentation';
 import type { ImageEmbedderModel } from './extensions/cv/tasks/imageEmbedding';
 import type { TextEmbedderModel } from './extensions/nlp/tasks/textEmbedding';
+import type { VADModel } from './extensions/speech/tasks/fsmnVad';
```
msluszniak: Probably should be named `fsmnVadModel`

msluszniak: Done — renamed `VADModel` → `FsmnVadModel` (0ee0fc1da).

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+            // Pass 2: fused mean-removal + pre-emphasis + Hann. Pre-emphasis of a
+            // mean-subtracted signal, `(raw[j]-mean) - c*(raw[j-1]-mean)`, equals
+            // `raw[j] - c*raw[j-1] - mean*(1-c)`, so every output reads only raw
+            // input samples — no serial dependency, and the loop vectorizes.
+            const float meanBias = mean * (1.0f - preemphasis);
+            base[0] = (start[0] - mean) * win[0];
+            for (int64_t j = 1; j < frameLength; ++j) {
+                base[j] = (start[j] - preemphasis * start[j - 1] - meanBias) * win[j];
+            }
```
msluszniak: Based on these, maybe there is a way to utilize std::span instead of raw pointers. How about the other places with raw pointers in this file? Also, maybe these comments are redundant / partially redundant?

msluszniak: Done (62aca1d60) — waveform/hann/dst are now `std::span`s with per-frame `subspan`s, zeroing via `std::ranges::fill` and the mean via a range-for. Dropped the redundant Pass 1/Pass 2 comments, kept the fused-identity note.

packages/react-native-executorch/src/extensions/speech/ops.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @returns The `dst` tensor, for convenience.
+ */
+export function frameWaveform(
+  waveform: Tensor,
+  hann: Tensor,
+  dst: Tensor,
+  startSample: number,
+  numFrames: number,
+  hopLength: number,
+  preemphasis: number
```
barhanc: I feel this would be cleaner to pass in TS and parse in C++ if it were an options object.

packages/react-native-executorch/src/extensions/speech/ops.ts
```diff
@@ -0,0 +1,40 @@
+import { rnexecutorchJsi } from '../../native/bridge';
+import { type Tensor } from '../../core/tensor';
+
+/**
+ * Native framing op: slices a mono `waveform` into `numFrames` overlapping
```
barhanc:
```suggestion
 * Slices a mono `waveform` into `numFrames` overlapping
```

packages/react-native-executorch/src/extensions/speech/ops.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param waveform Input audio samples, shape `[length]`.
+ * @param hann Precomputed Hann window, shape `[frameLength]`.
+ * @param dst Pre-allocated destination, shape `[frames, fftLength]`.
+ * @param startSample Index of the first sample of the first frame.
+ * @param numFrames Number of frames to write (must be `<= dst.shape[0]`).
+ * @param hopLength Samples between consecutive frames.
+ * @param preemphasis Pre-emphasis filter coefficient.
+ * @returns The `dst` tensor, for convenience.
+ */
+export function frameWaveform(
```
barhanc: The name is not great as it's not clear at all what it does from the name.

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,100 @@
...
+void install_frameWaveform(jsi::Runtime &rt, jsi::Object &module) {
+    const auto *name = "frameWaveform";
+    auto fnBody = [](jsi::Runtime &rt, const jsi::Value & /*thisVal*/, const jsi::Value *args, size_t count) -> jsi::Value {
+        if (count != 7) {
+            throw jsi::JSError(rt, "Usage: frameWaveform(waveform, hann, dst, startSample, numFrames, hopLength, preemphasis)");
+        }
+
+        auto waveform = tensor::fromJs(rt, "frameWaveform: waveform", args[0], DType::float32, std::nullopt);
+        auto hann = tensor::fromJs(rt, "frameWaveform: hann", args[1], DType::float32, std::nullopt);
+        auto dst = tensor::fromJs(rt, "frameWaveform: dst", args[2], DType::float32, std::nullopt);
```
barhanc: Why `std::nullopt` and not symbolic shape / some form of validation?

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        tensor::checkNotSameTensor(rt, "frameWaveform: waveform", waveform, "frameWaveform: dst", dst);
+        tensor::checkNotSameTensor(rt, "frameWaveform: hann", hann, "frameWaveform: dst", dst);
+        auto waveLock = tensor::tryLockShared(rt, "frameWaveform: waveform", waveform);
+        auto hannLock = tensor::tryLockShared(rt, "frameWaveform: hann", hann);
+        auto dstLock = tensor::tryLockUnique(rt, "frameWaveform: dst", dst);
+
+        if (dst->shape_.size() != 2) {
+            throw jsi::JSError(rt, "frameWaveform: dst must be 2D [frames, fftLength]");
+        }
```
barhanc: This should be done as a validation of symbolic shape i.e. 
```cpp
auto dst = tensor::fromJs(rt, "frameWaveform: dst", args[2], DType::float32, {"frames", "fftLength"});
```

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,100 @@
...
+            throw jsi::JSError(rt, "Usage: frameWaveform(waveform, hann, dst, startSample, numFrames, hopLength, preemphasis)");
+        }
+
+        auto waveform = tensor::fromJs(rt, "frameWaveform: waveform", args[0], DType::float32, std::nullopt);
+        auto hann = tensor::fromJs(rt, "frameWaveform: hann", args[1], DType::float32, std::nullopt);
+        auto dst = tensor::fromJs(rt, "frameWaveform: dst", args[2], DType::float32, std::nullopt);
+        auto startSample = conversions::asType<int64_t>(rt, "frameWaveform: startSample", args[3]);
+        auto numFrames = conversions::asType<int64_t>(rt, "frameWaveform: numFrames", args[4]);
+        auto hopLength = conversions::asType<int64_t>(rt, "frameWaveform: hopLength", args[5]);
+        auto preemphasis = static_cast<float>(conversions::asType<double>(rt, "frameWaveform: preemphasis", args[6]));
```
barhanc: Why not `conversions::asType<float>`?

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,100 @@
...
+        if (count != 7) {
+            throw jsi::JSError(rt, "Usage: frameWaveform(waveform, hann, dst, startSample, numFrames, hopLength, preemphasis)");
+        }
+
+        auto waveform = tensor::fromJs(rt, "frameWaveform: waveform", args[0], DType::float32, std::nullopt);
+        auto hann = tensor::fromJs(rt, "frameWaveform: hann", args[1], DType::float32, std::nullopt);
+        auto dst = tensor::fromJs(rt, "frameWaveform: dst", args[2], DType::float32, std::nullopt);
+        auto startSample = conversions::asType<int64_t>(rt, "frameWaveform: startSample", args[3]);
+        auto numFrames = conversions::asType<int64_t>(rt, "frameWaveform: numFrames", args[4]);
+        auto hopLength = conversions::asType<int64_t>(rt, "frameWaveform: hopLength", args[5]);
```
barhanc: These all sound  like non-negative numbers, perhaps `asType<uint64_t>` would be better to have automatic non-negativity check?

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,100 @@
...
+        auto preemphasis = static_cast<float>(conversions::asType<double>(rt, "frameWaveform: preemphasis", args[6]));
+
+        tensor::checkNotSameTensor(rt, "frameWaveform: waveform", waveform, "frameWaveform: dst", dst);
+        tensor::checkNotSameTensor(rt, "frameWaveform: hann", hann, "frameWaveform: dst", dst);
```
barhanc: Missing the third pairwise check.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+/**
+ * Model configuration required to instantiate a VAD task runner.
+ * @category Types
+ * @property {string} modelPath - Local path or remote URL of the `.pte` model.
+ * @property {VADFeatureConfig} featureConfig - Model-specific feature-extraction
+ * geometry.
+ * @property {VADOptions} [defaultOptions] - Detection thresholds tuned for this
+ * model; overridable per `detect` call. Falls back to the library defaults.
+ */
+export type FsmnVadModel = {
```
barhanc: Either this should be `FSMNVADModel` / `FSMNModel` so the convention is that we leave model  abbreviations in screaming case or other names should be `Vad...` conforming to standard camelcase. I honestly don't know what's better.

packages/react-native-executorch/src/extensions/speech/index.ts
```diff
@@ -0,0 +1 @@
+export * from './tasks/fsmnVad';
```
barhanc: Tasks shouldn't be exported as from root index as they are available from the package root. This should be the ops like in CV so we also expose them if the power user wants to use them.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+ * zero-padded up to the model minimum and its padding scores are discarded.
+ * @category Typescript API
+ * @param config VAD task configuration containing the model path and feature
+ * config.
+ * @param runtime Optional worklet runtime thread on which to run the model
+ * execution.
+ * @returns A promise resolving to an object containing detection and disposal
+ * controls.
+ */
+export async function createFsmnVad(
```
barhanc: The top-level exported entry points for a task should be more descriptive, so I think this should be `createFsmnVoiceActivityDetector` (or `...FSMN...` as discussed in other comment), and similarly for hook name and file names. The `...FsmnVad` is rather cryptic. The type names, etc. are ok with abbreviations.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+            fc.hopLength,
+            fc.preemphasis
+          );
+          model.execute('forward', [tInput], [tOutput]);
+          tOutput.getData(outBuffer);
+          for (let i = 0; i < realFrames; i++) {
+            scores[offset + i] = outBuffer[i * numClass]!;
+          }
+        } finally {
+          tInput.dispose();
```
barhanc: This should probably only enclose the smallest part that actually uses `tInput`.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+  const model = await wrapAsync(loadModel, runtime)(modelPath);
+
+  // Input: [frames, fftLength] with a dynamic frame count. Output: per-frame
+  // class probabilities, either [1, frames, classes] or [frames, classes]. Class
+  // 0 is the non-speech class.
+  const meta = validateModelSchema(
+    model,
+    'forward',
+    [SymbolicTensor('float32', ['N', fc.fftLength])],
+    [SymbolicTensor('float32', [1, 'F', 'C'], ['F', 'C'])]
```
barhanc: The comment reads as if `'N'` and `'F'` should be the same here. I'm also slightly confused about the `chunkCapacity` computation because of this.

Also this validation and pipeline in general reads like it could be used for a generic VAD model, but the name suggests it's only for FSMN. Is the world of VAD models similar to STT/TTS in that all of them are entirely different or could this pipeline actually be used not only for FSMN?

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+        }
+        offset += realFrames;
+      }
+    } finally {
+      tWaveform.dispose();
+    }
+
+    const raw = scoresToSegments(scores, opts, fc.hopLength, hopLengthMs);
+    const merged = mergeSegments(raw, opts.mergeGapMs * samplesPerMs);
+    return merged.map((s) => ({ start: s.start / fc.sampleRate, end: s.end / fc.sampleRate }));
```
barhanc: Maybe a single postprocess function would be cleaner here so that we do `return postprocess(scores, options)` and the segment logic is handled entirely somewhere else.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+    'worklet';
+    if (waveform.length < fc.frameLength) return [];
+
+    const opts: ResolvedOptions = { ...modelDefaults, ...options };
+    const numFrames = Math.floor((waveform.length - fc.frameLength) / fc.hopLength);
+    if (numFrames <= 0) return [];
+
+    const scores = new Float32Array(numFrames);
+    // Upload the waveform once; the native op frames slices of it per chunk.
+    const tWaveform = tensor('float32', [waveform.length], waveform);
```
barhanc: Perhaps it would be cleaner and easier in both TS and C++ if this was eagerly allocated inside the while loop and used the offset of `TypedArray` to `setData` correctly. We wouldn't need the padding/slicing, nested `try`'s, etc. Unless of course this has some non-negligible influence on performance.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+  const streamStop = (): void => {
+    if (!isStreaming) {
+      throw new Error('Streaming is not in progress');
+    }
+    isStreaming = false;
+    signal?.();
+    signal = null;
+  };
+
+  async function* stream(options?: VADStreamOptions): AsyncGenerator<VADStreamEvent> {
```
barhanc: I'm not sure if we actually want/need to implement streaming for VAD, since what it does can be done more easily by just running the `detectWorklet` inside `onAudioReady`, as VAD is extremely fast. I don't think we need this complex logic with inserting to stream, etc. like for more heavy models that need it for async concurrent processing; seems a bit overkill. What do you think?

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+ * @property {number} minFrames - Minimum frames the model accepts per forward pass (e.g. `100`).
+ */
+export type VADFeatureConfig = {
+  readonly sampleRate: number;
+  readonly frameLength: number;
+  readonly hopLength: number;
+  readonly fftLength: number;
+  readonly preemphasis: number;
+  readonly minFrames: number;
+};
```
barhanc: I'm wondering which options in a case like this with single-model-specific pipeline should be passed as config and defined in `models.ts` and which should be just `const CONSTANT`'s inside the file. I'm facing the same dilemma in supertonic TTS implementation and honestly don't have a good answer for this. For example here even the comment says that the values are model-specific, but this pipeline supports only a single model (which doesn't even have variants like Whisper) so idk. Unless of course this pipeline could actually be converted to generic as I mentioned in other comment.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+
+// Detection runs over a bounded window of the most recent audio (not the whole
+// growing session): model cost scales with window length, and 2.5s already
+// exceeds FSMN's receptive field (~200ms) and the min-speech duration (250ms).
+const DETECTION_WINDOW_SECONDS = 2.5;
+const DEFAULT_DETECTION_MARGIN_MS = 100;
+
+// A speech region measured in raw sample indices (internal to postprocessing).
+type SampleSegment = { start: number; end: number };
```
barhanc: This type duplicates `Segment`.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+ * transition.
+ * @category Types
+ * @property {boolean} isSpeaking - `true` when speech has just started, `false`
+ * when it has just ended.
+ */
+export type VADStreamEvent = {
+  readonly isSpeaking: boolean;
+};
+
+type ResolvedOptions = Required<VADOptions>;
```
barhanc: I would inline the `Required<VADOptions>` where it's needed, instead of defining this auxiliary type, it is more clear when reading the code imo.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+    window[i] = 0.5 * (1 - Math.cos((2 * Math.PI * i) / size));
+  }
+  return window;
+}
+
+// Converts per-frame non-speech probabilities into padded speech segments in
+// sample units. Mirrors `VoiceActivityDetection::postprocess` (excluding the
+// final merge step). `scores[i]` holds the non-speech probability of frame `i`,
+// so the speech probability is `1 - scores[i]`.
+function scoresToSegments(
```
barhanc: I would move this to some utils file and also export it as part of `speech.` api. The nested if's are also quite difficult to read, so if it could be refactored slightly it would be great.  Also I don't quite understand the split between `scoresToSegments` and `mergeSegments`, what is the reason behind it?

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
```diff
@@ -0,0 +1,430 @@
...
+  const streamInsert = (audioChunk: Float32Array): void => {
+    if (!isStreaming) {
+      throw new Error('Streaming is not in progress');
+    }
+    // Replace (never mutate) the buffer so a snapshot handed to a running
+    // detection stays valid, and drop audio beyond the detection window.
+    const next = new Float32Array(streamBuffer.length + audioChunk.length);
+    next.set(streamBuffer);
+    next.set(audioChunk, streamBuffer.length);
+    streamBuffer = next.length > windowSamples ? next.slice(next.length - windowSamples) : next;
```
barhanc: Slicing on insert is not a good pattern for heavy models, because it can lead to dropped frames if the model is too slow. Here it will work of course, since VAD is very light.

packages/react-native-executorch/src/extensions/speech/ops.ts
msluszniak: Applied.

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
msluszniak: Good catch — added the waveform/hann check (148a20060).

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
msluszniak: Done — `asType<float>` (148a20060).

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
msluszniak: Done — `startSample`/`numFrames`/`hopLength` are now `asType<uint64_t>`, which gives the non-negativity check for free and let me drop the manual `< 0` guards (148a20060).

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
msluszniak: No good reason — fixed. `waveform` is `{"length"}`, `hann` `{"frameLength"}`, `dst` `{"frames", "fftLength"}` (148a20060).

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
msluszniak: Done, exactly that — it also replaced the manual rank check (148a20060).

packages/react-native-executorch/src/extensions/speech/index.ts
msluszniak: Right — the barrel now exports `./ops` like cv; the task stays exported from the package root (148a20060).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Removed `SampleSegment`; the helpers now use `Segment` (rewrote the padding/merge to be non-mutating since `Segment` is readonly) (148a20060).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Done — inlined `Required<VADOptions>` (148a20060).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Done — the try now only wraps `frameWaveform` + `execute`; reading `tOutput` moved out (148a20060).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Agreed on the plumbing — dropped the generator/insert machinery for `push(chunk, opts) => 'speechStart' | 'speechEnd' | null` + `resetStream()`, driven straight from `onAudioReady` (c9b6112c9). One correction though: `detectWorklet` on a bare chunk can't work. A 100ms chunk is ~7 frames, but `minSpeechDurationMs: 250` = 25 hops, so the trigger can never fire and it silently returns `[]` forever. So `push` keeps a bounded rolling window internally — that part does need to live in the lib.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Right — and moot now: no more insert path, `push` slices the rolling window synchronously (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Went all-camel. `NmsOptions` is the existing precedent (NMS is just as much an acronym), and the `VAD*` types were the only names in the whole package with consecutive caps — so `VadOptions`/`VadStreamOptions`/`VadEvent`, `FsmnVadModel`. `models.vad.FSMN_VAD` stays screaming as a const key (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Done — `createFsmnVoiceActivityDetector` in `fsmnVoiceActivityDetection.ts`, hook `useVoiceActivityDetection` (mirroring `useSpeechToText`, which is model-agnostic despite taking `WhisperModel`) (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/ops.ts
msluszniak: Renamed to `extractFrames` — it does framing *and* mean-removal/pre-emphasis/windowing, so `frameWaveform` undersold it (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/ops.ts
msluszniak: Done — `extractFrames(waveform, hann, dst, { numFrames, hopLength, preemphasis })`, parsed with `getRequiredProperty` in C++ (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Good catch, the comment was misleading — `'N'`/`'F'` were never meant to be equal (symbols only bind within a single tensor, so input/output frames can't be cross-constrained). Renamed to `inFrames`/`outFrames` and spelled out that `chunkCapacity` is capped by both the max input frames and the output tensor's capacity. On generic-vs-FSMN: it's genuinely FSMN-specific — Silero, the main alternative, is stateful (LSTM h/c in and out, fixed 512-sample input, single sigmoid), so it can't run this contract. VAD is like STT/TTS here, so the FSMN naming stays.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Went with consts. These aren't tunable knobs — they're facts of the `.pte`, and a wrong value silently produces garbage, so `models.ts` advertised them misleadingly. Dropped `VadFeatureConfig`: `fftLength` now comes from model metadata, the rest are file consts, and the sample rate is exported as `FSMN_VAD_SAMPLE_RATE_HZ` (mirroring your `WHISPER_SAMPLE_RATE_HZ`). Only `VadOptions` stays configurable (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Done — the caller now passes the per-chunk waveform slice via `subarray`, so `startSample` is gone from the op and the nested try collapsed to one. Perf-neutral (same bytes copied). One note: the padding stays — the last-chunk pad to `MIN_FRAMES` is a frame-count constraint of the model, not a sample-offset thing (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Done — single `postprocess(scores, opts)` now, with the nested ifs flattened (c9b6112c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVad.ts
msluszniak: Folded into one `postprocess` and flattened. On the split you asked about: no good reason — it just mirrored the C++ (`VoiceActivityDetection::postprocess` + `utils::mergeSegments`), so it's gone. I'd hold off on moving it to a shared utils/`speech.` export until there's a second consumer, rather than commit to an API for one caller — happy to do it if you disagree (c9b6112c9).

packages/react-native-executorch/src/hooks/useVoiceActivityDetector.ts
barhanc, line 1: Should be named `...Detector` to be consistent with other hooks' naming.

packages/react-native-executorch/src/models.ts
```diff
@@ -796,6 +804,12 @@ export const models = {
       },
     },
   },
+  vad: {
```
barhanc: Here we should also use the full name `voiceActivityDetection`.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+   */
+  dispose: () => void;
+  /**
+   * Asynchronously detects speech segments within a mono waveform sampled at
+   * {@link FSMN_VAD_SAMPLE_RATE_HZ}.
+   * @param waveform The input audio samples.
+   * @param options Optional per-call overrides of the detection thresholds.
+   * @returns A promise resolving to the detected speech segments, in seconds.
+   */
+  detect: (waveform: Float32Array, options?: VadOptions) => Promise<Segment[]>;
```
barhanc: Maybe `detectVoice` would be more descriptive. In general the pattern in other tasks e.g. in CV is that the inference method name is rather specific for a given task.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
+import type { WorkletRuntime } from 'react-native-worklets';
+
+import { tensor } from '../../../core/tensor';
+import { loadModel } from '../../../core/model';
+import { validateModelSchema, SymbolicTensor } from '../../../core/modelSchema';
+import { wrapAsync } from '../../../core/runtime';
+import { extractFrames } from '../ops';
+
+/** Sample rate (Hz) the FSMN-VAD model expects its input waveform to be at. */
```
barhanc: This should be a proper jsdoc similar to the ones in `constants.ts` with category `@constants`.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  const [tHann] = tensors;
+
+  const dispose = () => {
+    tensors.forEach((t) => t.dispose());
+    model.dispose();
+  };
+
+  const detectWorklet = (waveform: Float32Array, options?: VadOptions): Segment[] => {
+    'worklet';
+    if (waveform.length < FRAME_LENGTH) return [];
```
barhanc: Isn't this redundant because of the check `if (numFrames <= 0) return [];` below?

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
```diff
@@ -0,0 +1,96 @@
...
+namespace jsi = facebook::jsi;
+namespace conversions = rnexecutorch::core::conversions;
+namespace tensor = rnexecutorch::core::tensor;
+using rnexecutorch::core::types::DType;
+
+// Slices a mono waveform into overlapping frames, applying per-frame
+// mean-removal, a pre-emphasis filter and a Hann window, writing each frame into
+// a zero-padded row of `dst`. Mirrors the reference FSMN-VAD feature extraction.
+// The whole per-frame inner loop dominates the VAD pipeline (~85% of a detect()
+// call on device), so it lives in native code per the extension guidelines.
```
barhanc: This comment is not needed imo, the documentation on the TS side is enough.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+   * Appends a live audio chunk to a bounded rolling window, runs detection over
+   * that window and reports a {@link VadEvent} when speech starts or stops,
+   * otherwise `null`. Designed to be driven straight from a recorder callback.
+   * Detection runs synchronously on the calling thread (a few ms). The rolling
+   * window is required: one short chunk is far too short to satisfy
+   * `minSpeechDurationMs` on its own.
+   * @param chunk The newly captured audio samples.
+   * @param options Optional overrides of the detection thresholds and margin.
+   */
+  push: (chunk: Float32Array, options?: VadStreamOptions) => VadEvent | null;
```
barhanc: Perhaps this could be more descriptive like `detectVoiceOnStream` or something like that. Also I'm not sure what's better `VadEvent | null` or `VadEvent | undefined` here?

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  return merged.map((s) => ({ start: s.start * hopSeconds, end: s.end * hopSeconds }));
+}
+
+/**
+ * Creates a Voice Activity Detection runner for the FSMN-VAD model.
+ *
+ * It loads the model, validates its input/output signature, pre-allocates the
+ * static output tensor and registers a disposal hook. The whole pipeline —
+ * feature extraction, chunked inference and segment postprocessing — runs in
+ * TypeScript on top of the core `model.execute` primitive.
```
barhanc:
```suggestion
 * static output tensor and registers a disposal hook.
```

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+   * or worklet thread.
+   */
+  detectWorklet: (waveform: Float32Array, options?: VadOptions) => Segment[];
+  /**
+   * Appends a live audio chunk to a bounded rolling window, runs detection over
+   * that window and reports a {@link VadEvent} when speech starts or stops,
+   * otherwise `null`. Designed to be driven straight from a recorder callback.
+   * Detection runs synchronously on the calling thread (a few ms). The rolling
+   * window is required: one short chunk is far too short to satisfy
+   * `minSpeechDurationMs` on its own.
```
barhanc:
```suggestion
   * Detection runs synchronously on the calling thread (a few ms).
```

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * It loads the model, validates its input/output signature, pre-allocates the
+ * static output tensor and registers a disposal hook. The whole pipeline —
+ * feature extraction, chunked inference and segment postprocessing — runs in
+ * TypeScript on top of the core `model.execute` primitive.
+ *
+ * The model exposes a dynamic frame dimension: each forward pass accepts
+ * `[frames, fftLength]` (up to the model-declared maximum) and returns per-frame
+ * class probabilities. Long inputs are split into chunks; a short final chunk is
+ * zero-padded up to the model minimum and its padding scores are discarded.
```
barhanc:
```suggestion
```
From the point of view of the API consumer this information doesn't say much, since they only see `Float32Array` audio input and `Segment` output.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+    window[i] = 0.5 * (1 - Math.cos((2 * Math.PI * i) / size));
+  }
+  return window;
+}
+
+// Turns per-frame non-speech probabilities into speech segments in seconds:
+// threshold with hysteresis, pad both ends, then merge near-adjacent regions.
+// Mirrors `VoiceActivityDetection::postprocess` + `utils::mergeSegments`.
+// `scores[i]` holds the non-speech probability of frame `i`, so the speech
+// probability is `1 - scores[i]`.
```
barhanc:
```suggestion
// `scores[i]` holds the non-speech probability of frame `i`, so the speech
// probability is `1 - scores[i]`.
```

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+};
+
+// `push` detects over a bounded window of the most recent audio (not the whole
+// growing session): model cost scales with window length, and 2.5s already
+// exceeds FSMN's receptive field (~200ms) and the min-speech duration (250ms).
+const DETECTION_WINDOW_SECONDS = 2.5;
+const DEFAULT_DETECTION_MARGIN_MS = 100;
+
+// Periodic Hann window used to reduce spectral leakage on each frame. Ported
+// from `dsp::hannWindow` (periodic definition, divides by `size`).
```
barhanc:
```suggestion
// Periodic Hann window used to reduce spectral leakage on each frame.
```
It would also perhaps be helpful to future maintainers if we added some link / longer comment explaining what Hann window is and how it is used in VAD.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  speechThreshold: 0.6,
+  minSpeechDurationMs: 250,
+  minSilenceDurationMs: 100,
+  speechPadMs: 30,
+  mergeGapMs: 0,
+};
+
+// `push` detects over a bounded window of the most recent audio (not the whole
+// growing session): model cost scales with window length, and 2.5s already
+// exceeds FSMN's receptive field (~200ms) and the min-speech duration (250ms).
```
barhanc:
```suggestion
```
Also perhaps we could group all constants together.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * A speech-activity transition reported by `push`.
+ * @category Types
+ */
+export type VadEvent = 'speechStart' | 'speechEnd';
+
+// Library fallback thresholds. A model may override any of these via
+// `FsmnVadModel.defaultOptions`, and callers via the `detect` options argument.
+const DEFAULT_OPTIONS: Required<VadOptions> = {
```
barhanc:
```suggestion
const DEFAULT_VAD_OPTIONS: Required<VadOptions> = {
```

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  // Input is [frames, fftLength] with a dynamic frame count; output is per-frame
+  // class probabilities, either [1, frames, classes] or [frames, classes], where
+  // class 0 is the non-speech class. The input and output frame symbols are
+  // deliberately distinct: a symbol only binds within a single tensor, so the
+  // two cannot be cross-constrained here.
+  const meta = validateModelSchema(
+    model,
+    'forward',
+    [SymbolicTensor('float32', ['inFrames', 'fftLength'])],
+    [SymbolicTensor('float32', [1, 'outFrames', 'classes'], ['outFrames', 'classes'])]
```
barhanc: The model only outputs `[1, 'F', 'C']` tensor so this second possibility is redundant here as well as the logic for `outFramesDim`. I think just explicitly getting the particular dimensions would be cleaner instead of constructing auxiliary shape using
```ts
      const chunkOutShape = outShape.slice();
      chunkOutShape[outFramesDim] = chunkFrames;
```

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  const { modelPath } = config;
+  const modelDefaults: Required<VadOptions> = { ...DEFAULT_OPTIONS, ...config.defaultOptions };
+
+  const model = await wrapAsync(loadModel, runtime)(modelPath);
+
+  // Input is [frames, fftLength] with a dynamic frame count; output is per-frame
+  // class probabilities, either [1, frames, classes] or [frames, classes], where
+  // class 0 is the non-speech class. The input and output frame symbols are
+  // deliberately distinct: a symbol only binds within a single tensor, so the
+  // two cannot be cross-constrained here.
```
barhanc: Hmm, I checked the exported model meta using inspection tool and it seems that the `'inFrames'` is the same as `'outFrames'` equal to 1000, which would also make sense if the model returns raw scores that given frame contains speech, so perhaps these actually should be the same in `validateModelSchema`? Also the comment about variable output tensor rank is misleading (see my comment below).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  const meta = validateModelSchema(
+    model,
+    'forward',
+    [SymbolicTensor('float32', ['inFrames', 'fftLength'])],
+    [SymbolicTensor('float32', [1, 'outFrames', 'classes'], ['outFrames', 'classes'])]
+  );
+  const chunkCapacity = meta.inputTensorMeta[0]!.shape[0]!;
```
barhanc: Maybe `maxFrames` would be a better name?

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+    const opts: Required<VadOptions> = { ...modelDefaults, ...options };
+    const numFrames = Math.floor((waveform.length - FRAME_LENGTH) / HOP_LENGTH);
+    if (numFrames <= 0) return [];
+
+    const scores = new Float32Array(numFrames);
+    let offset = 0;
+    while (offset < numFrames) {
+      const realFrames = Math.min(numFrames - offset, chunkCapacity);
+      // The model needs at least MIN_FRAMES rows, so a short final chunk is
+      // zero-padded up to it and its padding scores are discarded below.
```
barhanc, line 256: I'm not sure if zero padding is the best strategy in here. In Whisper I just discard the trailing chunk if it has less then required min number of samples and it is what we also did in the legacy implementation.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+      }
+      offset += realFrames;
+    }
+
+    return postprocess(scores, opts);
+  };
+
+  const detect = wrapAsync(detectWorklet, runtime);
+
+  const windowSamples = DETECTION_WINDOW_SECONDS * FSMN_VAD_SAMPLE_RATE_HZ;
```
barhanc: Maybe this could be moved as a constant to `const WINDOW_SAMPLES`?

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,352 @@
...
+  let window = new Float32Array(0);
+  let isSpeaking = false;
```
barhanc: The naming with `isSpeaking` and `speaking` variables is slightly confusing. Maybe rename `speaking` -> `isSpeaking` and `isSpeaking` -> `wasSpeaking`?

packages/react-native-executorch/src/hooks/useVoiceActivityDetection.ts
```diff
@@ -0,0 +1,48 @@
...
+} from '../extensions/speech/tasks/fsmnVoiceActivityDetection';
+
+/**
+ * React hook to load and run a Voice Activity Detection model.
+ *
+ * This hook manages downloading (if it's a remote URL) and loading the model
+ * file, compiling it, tracking download progress and compilation errors, and
+ * cleaning up native model memory when the component unmounts or configuration
+ * changes. The rolling-window state driven by `push` lives inside the task (see
+ * {@link createFsmnVoiceActivityDetector}), so it is exposed here directly.
```
barhanc:
```suggestion
 * changes.
```

packages/react-native-executorch/src/extensions/speech/utils/vadUtils.ts
barhanc, line 1: Perhaps we could name it `vadUtils.ts` and place it under `utils/`. In TTS PR I will also be adding some utils files (like `supertonicUtils.ts`, `textPartitioner.ts`, etc.) so I thought we could group them.

barhanc (commented):
Tested example app on Android and it works great. I don't have an iOS device to test unfortunately.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: I'm in favour of adding links explaining these topics.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — proper jsdoc with `@category Constants` (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Yes, fully redundant — `length < FRAME_LENGTH` makes `numFrames` negative, which `numFrames <= 0` already returns on. Removed (ed8f6f8c9).

packages/react-native-executorch/cpp/extensions/speech/operations.cpp
msluszniak: Removed (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — and the sentence was stale anyway (we no longer pre-allocate a static output tensor, since the fix to allocate it per chunk). Now reads '…validates its input/output signature and registers a disposal hook.' (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Applied (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Removed that paragraph from the public jsdoc (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Kept those two lines as-is — they already match your suggestion (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Trimmed and added a Hann-function link for future maintainers (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Renamed to `DEFAULT_VAD_OPTIONS` (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Renamed to `maxFrames` (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — `WINDOW_SAMPLES` module const (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — `speaking`→`isSpeaking`, `isSpeaking`→`wasSpeaking` (ed8f6f8c9).

packages/react-native-executorch/src/hooks/useVoiceActivityDetection.ts
msluszniak: Applied (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Grouped the numeric geometry/tuning consts into one block up top. `DEFAULT_VAD_OPTIONS` I left right under the `VadOptions` type it defaults (it depends on the type) rather than in the numeric block — happy to move it too if you'd prefer everything in one place (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — output is always `[1, F, C]`, so I dropped the 2D alternative and the `outFramesDim` slicing and build `[1, chunkFrames, numClass]` directly (ed8f6f8c9).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Fixed the misleading comment. On the two points:

1. **Output is genuinely dynamic**, not static 1000. The meta reporting `inFrames == outFrames == 1000` is just the **static upper bound** — that's all ExecuTorch serializes for a dynamic dim. At runtime it's `[1, N, C]`: feeding a 247-frame window made `execute` throw `outputTensors[0] must have shape [1, 247, 248] … got 1000` on device, which is exactly the bug fixed in b08811b9a (allocate the output per chunk). So they're equal to each other but both variable.
2. **Same-vs-different symbol doesn't matter for enforcement** — `matchShape` builds a fresh `symbolMap` per tensor and there's a `TODO: Implement cross-tensor symbol validation` in modelSchema.ts, so a shared `'frames'` is *not* cross-checked. I used the same name `'frames'` for both now since they *are* the same quantity and it reads better, but it's cosmetic — the real per-call output shape is built explicitly in code.

packages/react-native-executorch/src/hooks/useVoiceActivityDetector.ts
msluszniak: Done — `useVoiceActivityDetector` (agent-noun, matching `createFsmnVoiceActivityDetector` and the other `use*` hooks) (55de22876).

packages/react-native-executorch/src/models.ts
msluszniak: Done — `models.voiceActivityDetection.FSMN_VAD` (55de22876). Left the on-demand `vad` feature flag in download-libs.js as-is since that's a separate build-artifact key, not the model registry — say the word if you want that renamed too.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — `detectVoice`/`detectVoiceWorklet`, matching the CV convention (`classify`, `detectObjects`, `transferStyle`…) (55de22876).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
barhanc: Yeah, I know they are dynamic, my comment was regarding exactly the cosmetic change, that otherwise could be misleading. Now it is clear from looking at the code that the input and output dimension `'frames'` is the same, even if it is not strictly validated (yet) in the `validateModelSchema` function.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
```diff
@@ -0,0 +1,335 @@
...
+
+// Library fallback thresholds. A model may override any of these via
+// `FsmnVadModel.defaultOptions`, and callers via the `detectVoice` options argument.
+const DEFAULT_VAD_OPTIONS: Required<VadOptions> = {
+  speechThreshold: 0.6,
+  minSpeechDurationMs: 250,
+  minSilenceDurationMs: 100,
+  speechPadMs: 30,
+  mergeGapMs: 0,
+};
```
barhanc: One more thing, I'm wondering if these should be inside this file or as `defaultOptions` in `models.ts`.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — `detectVoiceOnStream` (pairs with `detectVoice`), and switched the no-transition return to `VadEvent | undefined` to match the codebase's undefined-leaning style (the hooks already return `undefined` when not ready, composes with `?.`) (9d4e5c735).

packages/react-native-executorch/src/extensions/speech/utils/vadUtils.ts
msluszniak: Done — moved to `extensions/speech/utils/vadUtils.ts` (9d4e5c735). I also codified the split in the add-native-extension skill so TTS lands the same way: domain-general ops (like cv's `box_ops`/`image_ops`) stay in `ops.ts`, model-/task-specific ops go under `extensions/<domain>/utils/<name>.ts`.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: I'd keep the padding here, and not for legacy reasons — the concrete benefit is that it's the only way to run detection on a window shorter than the model's ~1s minimum (`MIN_FRAMES=100`). That case isn't rare: `detectVoiceOnStream`'s rolling window starts empty and grows ~1600 samples/call, so for roughly the first second of every session the window is <100 frames. With padding it detects from ~0.4s in; discarding would return `[]` until the window fills (~1s), delaying onset detection ~0.6s. Same reason one-shot `detectVoice()` on a <1s clip works with padding but would return `[]` under discard. The padded frames are zeros (= trailing silence) and their scores are discarded (we read only `realFrames`), and VAD forward is a few ms, so the downside is negligible. Whisper discards because a sub-min chunk can't be transcribed and its chunks are expensive — neither holds for VAD. Happy to revisit if you'd rather not support sub-1s windows at all.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: I'd keep them here as the library fallback. These thresholds are standard, model-agnostic VAD tuning (unlike the feature geometry, which is FSMN-specific) and are user-overridable per `detectVoice` call — so a task-level `DEFAULT_VAD_OPTIONS` is their natural home, and it keeps `models.ts` to just `modelPath`. The `FsmnVadModel.defaultOptions` field is still there for a future model/variant that needs different defaults (`{...DEFAULT_VAD_OPTIONS, ...config.defaultOptions}`), we just don't populate it for the one model today. Moving them to `models.ts` would duplicate the same five values per entry and lose the 'library default' notion. Open to flipping it if you'd prefer defaults declared per-model.

barhanc: Hmm, these options actually seem closer to what we do for models like YOLO or others from CV pipelines, where there are sane defaults that are overridable from inference method call, but they are defined in `models.ts` so that users don't have to search for the defaults in the implementation file.

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Sorry, I misread this the first time — you were pointing at the `// Mirrors `VoiceActivityDetection::postprocess` + `utils::mergeSegments`` line, not the `scores[i]` lines. Removed that provenance line now, kept the rest (b-latest).

packages/react-native-executorch/src/extensions/speech/tasks/fsmnVoiceActivityDetection.ts
msluszniak: Done — moved the defaults onto the `FSMN_VAD` entry in `models.ts` and made `defaultOptions` required on `FsmnVadModel` (mirroring CV's required `opts`).

Two naming points I'd like your call on:
- CV distinguishes `defaultIouThreshold` (on the model) from `iouThreshold` (per-call), whereas VAD reuses the same `VadOptions` keys for both, disambiguated by nesting under `defaultOptions`. I kept that — prefixing every key with `default*` felt redundant when the whole object is already the defaults. OK with you?
- CV names the field `opts`, VAD names it `defaultOptions`. Unify? My take: CV's `opts` is a broader bag (labels, preprocessing, box format *plus* the default thresholds) so the generic name fits, while VAD's object is *only* the defaults for `VadOptions`, so `defaultOptions` reads more precisely. Happy to rename to `opts` for cross-domain consistency if you'd prefer.

barhanc (approved):

Issue #1194, opened 2026-05-27
Support useTextToSpeech with  Supertonic, a lightning-fast, on-device multilingual text-to-speech


StevePhuc:
### Problem description

https://github.com/supertone-inc/supertonic 

Some quick info about the modal

⚡ Blazingly Fast — Low-latency, real-time synthesis across desktop, browser, mobile, and edge — fast enough to turn an entire webpage into audio in under a second
🌍 31-Language Multilingual 
🪶 99M-Parameter Open-Weight Model  
📱 Edge-Device Ready — Runs locally on desktop, mobile, browsers, and resource-constrained hardware like Raspberry Pi or e-readers, with zero network dependency, complete privacy, and no GPU required
 🛠️ Multi-Runtime SDKs — Ready-to-use examples through ONNX Runtime across Python, Node.js, Browser (WebGPU), Java, C++, C#, Go, Swift, iOS, Rust, and Flutter

### Proposed solution

add to support modal in https://docs.swmansion.com/react-native-executorch/docs/hooks/natural-language-processing/useTextToSpeech#supported-models

### Benefits to React Native ExecuTorch

support more language than current tts solution

msluszniak:
Hi @StevePhuc, thank you for submitting an issue. Supertonic looks like a great addition which will cover more than 20 new languages in the library. We will check it out, try to export to ExecuTorch format and reach out to you with feedback if we can add this model. 😄

msluszniak:
I see one pretty big blocker here. There is no PyTorch model of supertonic 3, only ONNX format. Huggingface model card of the Supertonic 3 states that explicitly:
>This model was trained using PyTorch, which is licensed under the BSD 3-Clause License but is not redistributed with this project. See the [PyTorch license](https://docs.pytorch.org/FBGEMM/general/License.html) for details.

So as long as we don't have any format suitable for export to `*.pte` format, we won't be able to export it and support in our library :/

Marking it blocked. If you'll see something like this: https://github.com/chantysothy/supertonic-pytorch but for Supertonic 3, please reach out to us.

IgorSwat:
Apparently converting model weights to PyTorch format was not too difficult and we just managed to do that.

There is already an [open PR](https://github.com/software-mansion/react-native-executorch/pull/1300) with Supertonic support. Once it gets in, it will be available to be used it in a similar way as our previous TTS - Kokoro.

PR #1300, opened 2026-07-06
feat: add Supertonic 3 text-to-speech model

IgorSwat:
## Description

This PR adds a full pipeline and API for Supertonic 3 model. It also refactors the native TTS code to make it better organized across 2 different text-to-speech pipelines.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Run demo app in apps/speech.

### Related issues

- #1194 

### Checklist

### Additional notes

Needs to be tested on iOS. Also, the typescript API is in a somewhat PoC state and will most likely need some changes.


packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
```diff
@@ -119,8 +132,7 @@ std::vector<float> Kokoro::generate(std::u32string input, float speed,
 
   // Divide the phonemes string into substrings, minimizing the amount of
   // breaks.
-  auto partition = partitioner_.partition(phonemes, context_.inputTokensLimit,
-                                          Partitioner::Mode::MIN_BREAKS);
```
barhanc, line 123: The new TextPartitioner only supports a single partitioning logic (equivalent to Kokoro's old `MIN_LATENCY` mode), the offline/batch generate call originally used `Partitioner::Mode::MIN_BREAKS` to split the text. Is this deliberate or accidental regression?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+         vocoder_.getMemoryLowerBound() + textProcessor_.getMemoryLowerBound() +
+         sizeof(voice_);
+}
+
+void Supertonic::unload() noexcept {
+  durationPredictor_.unload();
+  textEncoder_.unload();
+  vectorEstimator_.unload();
+  vocoder_.unload();
+}
```
barhanc, line 303: `unload()` does not wait for or join the detached thread, and instead immediately frees the models, the background thread may still be running `synthesize(...)` or processing the loop. When the background thread invokes the next ExecuTorch model run, it will access freed/unloaded memory and crash the application instantly.

barhanc (commented):
Tested example app and it works fine.

docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
```diff
@@ -58,9 +91,10 @@ const handleSpeech = async (text: string) => {
...
-- [`phonemizerConfig`](../../06-api-reference/interfaces/TextToSpeechModelConfig.md#phonemizerconfig) of type [`TextToSpeechPhonemizerConfig`](../../06-api-reference/interfaces/TextToSpeechPhonemizerConfig.md) - configuration of the phonemizer.
+- `model` of type [`TextToSpeechModelSources`](../../06-api-reference/type-aliases/TextToSpeechModelSources.md) — model configuration.
+- [`voiceSource`](../../06-api-reference/interfaces/TextToSpeechModelConfig.md#voicesource) of type [`ResourceSource`](../../06-api-reference/type-aliases/ResourceSource.md) — the voice tensor used for synthesis.
+- [`phonemizerConfig`](../../06-api-reference/interfaces/TextToSpeechModelConfig.md#phonemizerconfig) of type [`TextToSpeechPhonemizerConfig`](../../06-api-reference/interfaces/TextToSpeechPhonemizerConfig.md) — Kokoro only: phonemizer configuration. Unused by Supertonic.
+- [`lang`](../../06-api-reference/interfaces/TextToSpeechModelConfig.md#lang) of type [`TextToSpeechSupertonicLanguage`](../../06-api-reference/type-aliases/TextToSpeechSupertonicLanguage.md) — Supertonic only: default language token (e.g. `'en'`, `'na'`). Unused by Kokoro.
```
msluszniak, line 97: Language which is typed as sth strictly related to supertonic, generally incorrect abstraction. If adding this  in a correct way causes huge rewrite please ignore it since we will introduce the model in rewritten flow in v0.10.0 either way.

docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
```diff
@@ -79,21 +113,28 @@ You need more details? Check the following resources:
 
 ## Running the model
 
-The module provides two ways to generate speech using either raw text or pre-generated phonemes:
+The module provides two ways to generate speech. The available parameters differ by model family:
+
+| Parameter    | Kokoro | Supertonic |
+| ------------ | ------ | ---------- |
+| `speed`      | ✓      |            |
+| `phonemize`  | ✓      |            |
```
msluszniak, line 121: Same as above.

docs/docs/04-typescript-api/01-natural-language-processing/TextToSpeechModule.md
```diff
@@ -61,22 +83,42 @@ Since `forward` processes the entire input at once, it might take a significant
...
-  models.text_to_speech.kokoro.en_us.heart()
+  models.text_to_speech.supertonic.m1(),
+  (progress) => console.log(progress)
 );
-const audioContext = new AudioContext({ sampleRate: 24000 });
+// Kokoro — language-specific voice bundle:
+// const tts = await TextToSpeechModule.fromModelName(
+//   models.text_to_speech.kokoro.en_us.heart(),
+//   (progress) => console.log(progress)
+// );
```
msluszniak, line 101: Do we really want to keep this one for Kokoro as comment? Maybe drop it. Same for all such cases in this file.

docs/docs/04-typescript-api/01-natural-language-processing/TextToSpeechModule.md
```diff
@@ -61,22 +83,42 @@ Since `forward` processes the entire input at once, it might take a significant
...
+// Kokoro: sampleRate: 24000
 
 try {
-  const waveform = await tts.forward('Hello from ExecuTorch!', 1.0);
+  const waveform = await tts.forward(
+    'Hello from ExecuTorch!',
+    1.0,
+    true,
+    8,
+    'en'
```
msluszniak: maybe adding args names as inline commets next to the passed values might be more readable. Same for all such cases in this file.

packages/react-native-executorch/common/rnexecutorch/host_objects/ModelHostObject.h
```diff
@@ -236,6 +236,24 @@ template <typename Model> class ModelHostObject : public JsiHostObject {
...
+          ModelHostObject<Model>, synchronousHostFunction<&Model::streamStop>,
+          "streamStop"));
+      addFunctions(JSI_EXPORT_FUNCTION(
+          ModelHostObject<Model>, synchronousHostFunction<&Model::streamInsert>,
+          "streamInsert"));
+      addFunctions(JSI_EXPORT_FUNCTION(
+          ModelHostObject<Model>, synchronousHostFunction<&Model::streamFlush>,
+          "streamFlush"));
+    }
+
```
msluszniak: These are exactly the same as in kokoro, you can just modify the if condition.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/common/TextPartitioner.h
```diff
@@ -0,0 +1,94 @@
...
+   * The content is stored as logical views to avoid copying. Segments
+   * defines ranges of the content views for smaller segments.
+   */
+  struct Partition {
+    std::u32string_view content;
+    std::vector<std::pair<size_t, size_t>>
+        segments; // Pairs of {offset, length} for each segment.
+  };
+
+  TextPartitioner(const TextPartitionerConfig &config);
```
msluszniak: Why not explicit

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Utils.cpp
```diff
@@ -104,4 +43,4 @@ std::vector<Token> tokenize(std::u32string_view phonemes,
   return tokens;
 }
 
-} // namespace rnexecutorch::models::text_to_speech::kokoro::utils
+} // namespace rnexecutorch::models::text_to_speech::kokoro::utils
```
msluszniak: Please keep newlines at the end of the files.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/DurationPredictor.h
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Wraps the duration_predictor submodule.
+ *
+ * Receives token ids, an attention mask, and the duration-predictor
+ * style vector; returns the predicted utterance length in seconds.
+ */
+class DurationPredictor : public BaseModel {
+public:
+  DurationPredictor(const std::string &modelSource,
```
msluszniak: Again, probably we can mark constructor explicit.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+namespace rnexecutorch::models::text_to_speech::supertonic {
+
+namespace common_constants = ::rnexecutorch::models::text_to_speech::constants;
+namespace common_utils = ::rnexecutorch::models::text_to_speech::utils;
+
+namespace {
+
+// Returns the last non-space character of a view (for pause selection).
+char32_t lastMeaningfulChar(std::u32string_view s) {
+  for (auto it = s.rbegin(); it != s.rend(); ++it) {
```
msluszniak: We can use const iterators.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+    throw RnExecutorchError(
+        RnExecutorchErrorCode::FileReadFailed,
+        "[Supertonic::loadVoice] voice file too small: need " +
+            std::to_string(expectedFloats * sizeof(float)) + " bytes, got " +
+            std::to_string(fileSize));
+  }
+  in.read(reinterpret_cast<char *>(voice_.ttl.data()),
+          constants::kStyleTtlSize * sizeof(float));
+  in.read(reinterpret_cast<char *>(voice_.dp.data()),
+          constants::kStyleDpSize * sizeof(float));
```
msluszniak, line 86: casting float array to char*, not the best pattern, can't we implement this one somehow different.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+  } else if (tok.ids.size() < constants::kMinTokens) {
+    tok.ids.resize(constants::kMinTokens, constants::kPadToken);
+    tok.mask.resize(constants::kMinTokens, 0.0F);
+  }
+  const auto tLen = static_cast<int32_t>(tok.ids.size());
+
+  // 2. Duration predictor.
+  float durationSec =
+      durationPredictor_.generate(tok.ids, tok.mask, voice_.dp, speed);
+  if (!(durationSec > 0.0F)) {
```
msluszniak: just reverse comparison sign here.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+            static_cast<size_t>(pauseMsFor(lastMeaningfulChar(sub))) *
+            constants::kSamplesPerMillisecond;
+        audioPart.resize(audioPart.size() + pause, 0.0F);
+        nativeCallback(std::move(audioPart));
+      }
+    }
+
+    if (isStreaming_) {
+      std::this_thread::sleep_for(
+          std::chrono::milliseconds(params::kStreamPause));
```
msluszniak, line 262: why not conditional variable here?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+    inputTextBuffer_.clear();
+    isStreaming_ = false;
+    flushPending_ = false;
+  }
+}
+
+void Supertonic::streamInsert(std::u32string chunk) noexcept {
+  std::scoped_lock<std::mutex> lock(inputTextBufferMutex_);
+  inputTextBuffer_.append(chunk);
+}
```
msluszniak: std::u32string::append is no `noexcept` so I wouldn't mark this function noexcept (there is also RAII here which might fail).

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
```diff
@@ -0,0 +1,308 @@
...
+  std::span<const float> audio(wav.data(),
+                               cropLen > 0 ? cropLen : totalSamples);
+  audio = common_utils::stripAudio(
+      audio, paddingMs * constants::kSamplesPerMillisecond,
+      params::cropping::kAudioCroppingSteps,
+      params::cropping::kAudioSilenceThreshold);
+  return {audio.begin(), audio.end()};
+}
+
+std::vector<float> Supertonic::generate(std::u32string input, float speed,
```
msluszniak, line 158: Isn't this one const method?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/TextEncoder.h
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Wraps the text_encoder submodule.
+ *
+ * Encodes token ids together with the text-to-latent style vector into
+ * a text embedding. The output is copied to host memory so it remains
+ * valid across the vector-estimator flow loop.
+ */
+class TextEncoder : public BaseModel {
+public:
+  TextEncoder(const std::string &modelSource,
```
msluszniak: explicit

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/TextProcessor.cpp
```diff
@@ -0,0 +1,215 @@
...
+    return;
+  }
+  size_t pos = 0;
+  while ((pos = s.find(from, pos)) != std::u32string::npos) {
+    s.replace(pos, from.size(), to);
+    pos += to.size();
+  }
+}
+
+bool isAsciiSpace(char32_t c) {
```
msluszniak: noexcept

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/TextProcessor.cpp
```diff
@@ -0,0 +1,215 @@
...
+    bool prevSpace = false;
+    for (char32_t c : s) {
+      if (isAsciiSpace(c)) {
+        if (!prevSpace) {
+          cleaned.push_back(U' ');
+        }
+        prevSpace = true;
+      } else {
+        cleaned.push_back(c);
+        prevSpace = false;
```
msluszniak:
```suggestion
    for (char32_t c : s) {
      prevSpace = isAsciiSpace(c);
      if (prevSpace) {
        if (!prevSpace) {
          cleaned.push_back(U' ');
        }
      } else {
        cleaned.push_back(c);
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/VectorEstimator.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+  // Only the frames that actually produce audio samples are valid;
+  // the remainder is zero-padded for the static model shape.
+  const size_t latentValid =
+      std::min<size_t>(static_cast<size_t>((wavLen + chunk - 1) / chunk),
+                       static_cast<size_t>(L));
+
+  // Mask to silence invalid latent frames during the flow loop.
+  std::vector<float> latentMask(static_cast<size_t>(L), 0.0F);
+  std::fill(latentMask.begin(), latentMask.begin() + latentValid, 1.0F);
```
msluszniak, line 56: use ranges here

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp
IgorSwat: Deliberate, I just came to a conclusion that the old MIN_BREAKS mode is not that much useful and the code can be simplified by removing it.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
IgorSwat: It's a common issue across the current codebase - can be fixed globally in the upcoming RNE refactor.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
IgorSwat: Could you explain? We are not waiting here for streaming to end, it's just fixed pause which is there to slow down the streaming loop when nothing new is being inserted as input. And I think this is much simpler than applying conditional variable.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
IgorSwat: In theory it is, but changing it to const would require making things more ugly elsewhere - the voice array (a Supertonic class member) is being passed via non-const span to one of the submodules - to avoid casting it right back with const_cast (since ExecuTorch requires non-const data). Therefore if we want to avoid these weird const_casts. it should be non-const.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/TextProcessor.cpp
IgorSwat:
```
if (prevSpace) {
  if (!prevSpace) {
    cleaned.push_back(U' ');
  }
}
```

What is that 🤔?

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/VectorEstimator.cpp
IgorSwat: > use ranges here

Is it actually worthed? Note that we do not apply the fill for the entire range - but rather for a subrange starting from the beginning.

docs/docs/04-typescript-api/01-natural-language-processing/TextToSpeechModule.md
IgorSwat: The thing is, Kokoro still has some valid use cases over Supertonic, so I am not sure dropping it entirely from the docs sample usages is a good idea.

docs/docs/03-hooks/01-natural-language-processing/useTextToSpeech.md
IgorSwat: Yeah, a few things about this design (particularly the common abstraction between the 2 models) are definitely dubious, but as I was playing with TTS implementation in the new architecture, I am pretty sure it can be easily fixed in the upcoming RNE rewrite -  easier than in the current codebase.

docs/docs/04-typescript-api/01-natural-language-processing/TextToSpeechModule.md
msluszniak: I know that Kokoro is still used, by the way it is used is exactly the same as supertonic here. So there is no specific thing for Kokoro in this commented code and that's the reason why I think it is redundant.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
msluszniak: With cv you can potentially wait for notify, and do not sleep in the loop. But I think we can indeed use sleep here eventually.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
msluszniak: This is a good material for a comment, since it is not obvious.

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/TextProcessor.cpp
msluszniak: Acgually I meant sth like this:

```cpp
const bool space = isAsciiSpace(c);
  if (space && prevSpace) {
    continue;
  }
  cleaned.push_back(space ? U' ' : c);
  prevSpace = space;
```

packages/react-native-executorch/common/rnexecutorch/models/text_to_speech/supertonic/Supertonic.cpp
IgorSwat: Would need to be quite an invasive change 😔

msluszniak:
Hugging face page does not comply with the format of the model repository. Also if you are publishing it under `v0.10.0` it needs to comply with the rewrite requirement that is: "For every method that supports dynamic input, you need to have a function named: `get_dynamic_dims_<methodName>`". Please check this: https://github.com/software-mansion/react-native-executorch/pull/1292 for a reference. If output is static there is no need for additional method.

msluszniak (approved):

PR #1302, opened 2026-07-07
[RNE Rewrite] feat: add Whisper STT pipeline

### LINKED ISSUE #1208: [RNE Rewrite] - Aggregated issue

### LINKED ISSUE #1251: [RNE Rewrite] Speech - add Whisper STT pipeline implementation
Implement STT pipeline using the new approach to achieve parity with current implementation.

barhanc:
## Description

Adds a full Whisper speech-to-text task pipeline, React hook, model registry entries, and an interactive example screen.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Build speech example app on both Android and iOS.
- [ ] Test the newly added speech-to-text feature.

### Related issues

Closes #1208 

### Checklist

### Additional notes

Needs to wait for #1309 before merging so that the paths in `models.ts` can be pointed at SWM HF.


packages/react-native-executorch/src/models.ts
```diff
@@ -817,6 +902,44 @@ export const models = {
       XNNPACK_FP32: FSMN_VAD_XNNPACK_FP32,
     },
   },
+  speechToText: {
+    WHISPER: {
+      ...WHISPER_TINY_XNNPACK_FP32,
+      TINY: {
+        ...WHISPER_TINY_XNNPACK_FP32,
+        XNNPACK_FP32: WHISPER_TINY_XNNPACK_FP32,
+        COREML_FP16: WHISPER_TINY_COREML_FP16,
```
msluszniak, line 961: Just to remember, we should wait for Igor MLX delegate, and update this section accordingly by dropping coreml completely and replacing by MLX

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -0,0 +1,357 @@
...
+
+/**
+ * Options for the live-streaming transcription API.
+ * Extends {@link WhisperSttOptions} with optional VAD tuning.
+ * @category Types
+ * @property vadOptions - Fine-tuning knobs forwarded to the FSMN voice-activity
+ * detector. Omit to use the detector's built-in defaults.
+ */
+export type WhisperStreamOptions<L extends WhisperLanguage = WhisperLanguage> =
+  WhisperSttOptions<L> & { readonly vadOptions?: VadStreamOptions };
```
msluszniak, line 74: I'm wondering if we should couple in such way one specific model with another. Right now, when we have only one VAD model. But I'm wondering if one day we will have a better one. Won't it be a breaking change to add a better one? Probably we should have an interface here that will accept any VAD.

packages/react-native-executorch/src/models.ts
```diff
@@ -605,6 +609,87 @@ const FSMN_VAD_XNNPACK_FP32: FsmnVadModel = {
   },
 };
 
+// =============================================================================
+// Speech-To-Text
+// =============================================================================
+const WHISPER_TINY_EN_XNNPACK_FP32: WhisperSttModel<'en'> = {
+  modelPath: 'https://huggingface.co/bhanc/scratch/resolve/main/whisper_tiny_en_xnnpack_fp32.pte',
+  tokenizerPath: `${BASE_URL}-whisper-tiny.en/${VERSION_TAG}/tokenizer.json`,
```
msluszniak: Naming is inconsistent across the English variants: tiny uses `whisper-tiny.en` (dot) while base/small use `-whisper-base-en`/`-whisper-small-en` (dash). Let's settle on one unified convention and align all six repos (rest of `models.ts` is all-dashes) as during the rewrite we should unify such cases.

packages/react-native-executorch/src/models.ts
```diff
@@ -605,6 +609,87 @@ const FSMN_VAD_XNNPACK_FP32: FsmnVadModel = {
   },
 };
 
+// =============================================================================
+// Speech-To-Text
+// =============================================================================
+const WHISPER_TINY_EN_XNNPACK_FP32: WhisperSttModel<'en'> = {
+  modelPath: 'https://huggingface.co/bhanc/scratch/resolve/main/whisper_tiny_en_xnnpack_fp32.pte',
```
msluszniak: Reminder: needs repointing to SWM HF (per #1309) before merge.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -0,0 +1,357 @@
...
+/**
+ * Sample rate (Hz) Whisper models expect their input waveform to be at.
+ * @category Constants
+ */
+export const WHISPER_SAMPLE_RATE_HZ = 16000;
+
+const MAX_SEQ_LEN = 128; // maximum decoder output tokens per chunk
+const MIN_CHUNK_SIZE = 201; // shortest audio slice (samples) the model was exported for
+const CHUNK_LENGTH_SECONDS = 29; // Whisper's fixed context window length
+const STRIDE_SIZE = 1 * WHISPER_SAMPLE_RATE_HZ; // overlap between consecutive chunks (1s)
```
msluszniak: Documented as chunk overlap, but offline `transcribeWorklet` advances by `offset += BUFFER_SIZE` with no overlap; `STRIDE_SIZE` is actually the stream process-threshold/retain window. Either apply the overlap offline or fix the comment.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -0,0 +1,357 @@
...
+    audio: Float32Array,
+    options: WhisperSttOptions<L>,
+    onToken?: (token: string) => void
+  ): string => {
+    'worklet';
+
+    const promptTokenStrings = isEnglishOnly
+      ? ['<|startoftranscript|>', '<|notimestamps|>']
+      : ['<|startoftranscript|>', `<|${options.language}|>`, '<|transcribe|>', '<|notimestamps|>'];
+    const promptTokens = promptTokenStrings.map((token) => tokenizer.tokenToId(token)!);
```
msluszniak: If `options.language` isn't in `supportedLanguages`, `tokenToId(...)!` returns `undefined` and `BigInt(undefined)` throws opaquely. Worth an explicit check with a clear message.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -0,0 +1,357 @@
...
+    return text.trim();
+  };
+
+  const transcribe = wrapAsync(transcribeWorklet, runtime);
+
+  let isStreaming = false;
+  let audioBuffer = new Float32Array(0);
+  let signal: (() => void) | null = null;
+
+  const streamInsert = (audioChunk: Float32Array): void => {
```
msluszniak, line 280: Each insert reallocates and copies the whole growing buffer (up to ~BUFFER_SIZE) on every audio callback — O(n^2) over a segment. Did you measure the impact of this copying, or would a ring buffer be worth it here?

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -0,0 +1,357 @@
...
+          yield { committed: committedText, nonCommitted: currentText };
+        } else {
+          const retainSamples = Math.min(audioBuffer.length, STRIDE_SIZE);
+          audioBuffer = audioBuffer.slice(audioBuffer.length - retainSamples);
+          processedLength = audioBuffer.length;
+          yield { committed: committedText, nonCommitted: '' };
+        }
+      }
+
+      if (isSpeaking && audioBuffer.length >= MIN_CHUNK_SIZE) {
```
msluszniak, line 359: Teardown race: on unmount `useModel`'s cleanup calls `dispose()` (freeing tensors/model) while `streamStop()` only signals the generator, which may then resume and run this final-flush `transcribe` on already-disposed tensors. Worth guarding with a disposed flag.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -0,0 +1,357 @@
...
+const STRIDE_SIZE = 1 * WHISPER_SAMPLE_RATE_HZ; // overlap between consecutive chunks (1s)
+const BUFFER_SIZE = CHUNK_LENGTH_SECONDS * WHISPER_SAMPLE_RATE_HZ; // samples per full chunk
+
+/**
+ * BCP-47 language codes supported by Whisper multilingual models. English-only
+ * model variants only accept `'en'`.
+ * @category Constants
+ */
+// prettier-ignore
+export const WHISPER_LANGUAGES = [
```
msluszniak, line 35: Minor: these are Whisper's own language codes rather than strict BCP-47 (`jw`, `yue`, `haw`), so the doc comment is slightly off.

barhanc: Fixed comments.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: The allocation is about 50x faster than the streaming inference, so imo no need to optimize.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: Fixed comment.

packages/react-native-executorch/src/models.ts
barhanc: These were typos in the urls. Fixed them and added all models to example app.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: Added explicit check.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: The coupling is not that tight actually. If we add more VADs then we can add type parameter and switch statement for `createVAD` and appropriate VAD options type for given VAD type. I thought about passing explicitly the VAD model instance to stream, but this is an unnecessary complication for API users because then they would need to manage two models to use STT.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: I've added the `transcribeStop` to stop the in-flight transcription on a long audio file. Regarding the disposal behavior I would leave it as it is. Clanker gave a pretty compelling argument (attached below) that this is not something we should handle at the task pipeline level but rather at the react lifecycle integration level---meaning either we let users deal with this and document the behavior, especially since it doesn't lead to any memory leaks, issues, etc. or we refactor the `useModel` hook somehow---right now I'm in favour of the first option.

> Standard Low-Level Resource Behavior 
> In any software engineering context, if you close/dispose of a resource (like a database connection, a file handle, or a WebGL context) while an operation is in-flight, the caller expects that operation to fail with an error. If the task pipeline silently swallowed the error and returned an empty string "" on disposal, it would be masking the fact that the operation failed to complete. It is much more correct for the API to throw, and let the caller decide how to handle the failure.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: I've renamed the user-facing options to more generic `vad...` so that it will be easy to integrate other vad pipelines in the future.

packages/react-native-executorch/src/models.ts
barhanc: Updated the models on HF and updated the urls.

packages/react-native-executorch/src/models.ts
barhanc: Added the MLX variant.

barhanc:
@msluszniak I've updated the models on HF and added the MLX variant. I can't unfortunately test them since I don't have iOS device available.

msluszniak:
@barhanc Ok, I will test mlx delegates tomorrow.

apps/speech/app/audio-file-transcription/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+    name: 'Small Multilingual (CPU)',
+    config: models.speechToText.WHISPER.SMALL.XNNPACK_FP32,
+  },
+  {
+    name: 'Small Multilingual (CoreML)',
+    config: models.speechToText.WHISPER.SMALL.COREML_FP16,
+    disabled: Platform.OS !== 'ios',
+  },
+  {
+    name: 'Small Multilingual (MLX)',
```
msluszniak, line 100: When I spoke to it in polish, it transcribed correctly but in english (?). I don't know if when we run native inference, shouldn't we pass the language or sth to it?

EDIT, claude suggests this is indeed problem with language argument:

```
That confirms it — it's not a model problem, it's the example app. Here's the mechanism:

At [whisperSpeechToText.ts:226](vscode-webview://1den94tu8hqi0qi2gqm2m4tdov78vp2um8u6i1gktvdssupbm5in/packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts#L226), the multilingual decoder is force-prompted with the language token:


['<|startoftranscript|>', `<|${options.language}|>`, '<|transcribe|>', '<|notimestamps|>']
Both example screens pass { language: 'en' } hardcoded, so that becomes <|en|>. That token tells Whisper the audio is English, overriding its own language detection. When you speak Polish, the multilingual small/base models dutifully decode against the English language prompt — so they phonetically force your Polish into English words. (The task token is <|transcribe|>, not <|translate|>, so it's not even translating — it's just decoding-as-English.)

Why the .en models don't show this: they hit the isEnglishOnly branch (line ~225) which omits the language token entirely, since they're English-only anyway.

The fix / how to get Polish: pass language: 'pl' → <|pl|>. The API already supports it (and validates it at line 228-229). The problem is purely that the example hardcodes 'en' in both screens and exposes no language picker, so the multilingual models can never be driven in any other language from the UI.
```

msluszniak (commented):
MLX delegates works, just one comment about strange behaviour mentioned in the inline comment.

apps/speech/app/microphone-transcription/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+    if (!isSttReady || !stream || !streamInsert || isRecording) return;
+    setRunError(null);
+    setCommittedText('');
+    setNonCommittedText('');
+    setStatus('Streaming...');
+    setIsRecording(true);
+
+    (async () => {
+      try {
+        const textStream = stream({ language: 'en' });
```
IgorSwat: If we allow multilingual models, then the `language` should not be hardcoded. I think a separate selection bar (active only if we use a multilingual whisper) would be a better idea.

apps/speech/app/audio-file-transcription/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+    setIsTranscribing(true);
+
+    try {
+      const audioBuffer = await decodeAudioData(url, WHISPER_SAMPLE_RATE_HZ);
+      const samples = audioBuffer.getChannelData(0);
+
+      setStatus('Transcribing...');
+
+      let currentText = '';
+      const text = await transcribe(samples, { language: 'en' }, (token) => {
```
IgorSwat: Same issue as in microphone-transcription app screen.

IgorSwat:
Okay, one big concern I have is the audio buffer management. 

I see that you dropped most of the audio buffer cleanup heuristics - including the one with saving full sentence timestamps and leaving the last sentence in the buffer in concrete cases.
This requires extensive testing, particularly for situations where VAD is not active or user performs a continuous speech which at some point exceeds the 30s buffer length. Lack of some of these heuristics might lead to missing some words and errors.

But If you tested it extensively and it does look fine - then we can leave it as it is.

IgorSwat (changes requested):

apps/speech/app/microphone-transcription/index.tsx
barhanc: Fixed.

apps/speech/app/audio-file-transcription/index.tsx
barhanc: Fixed.

barhanc:
> Okay, one big concern I have is the audio buffer management.
> 
> I see that you dropped most of the audio buffer cleanup heuristics - including the one with saving full sentence timestamps and leaving the last sentence in the buffer in concrete cases. This requires extensive testing, particularly for situations where VAD is not active or user performs a continuous speech which at some point exceeds the 30s buffer length. Lack of some of these heuristics might lead to missing some words and errors.
> 
> But If you tested it extensively and it does look fine - then we can leave it as it is.

Yeah, I've tested it on multiple youtube shorts with continuous speech where the VAD almost never triggered end of speech until the very end and the transcription almost never missed words. In the legacy implementation it happened more often actually.

apps/speech/app/audio-file-transcription/index.tsx
barhanc: Added the language selector to example screens.

msluszniak (approved):

IgorSwat (approved):

PR #1305, opened 2026-07-13
[RNE Rewrite] feat: add LLM chat session implementation

### LINKED ISSUE #429: Improve tool calling
### Problem description

Currently we do not really support tool calling, we rely on models declaring what tools to use and then for developer to provide the tools response. This is currenlty impossible to do properly with `useLlm` hook as we would need to provide different message role `tool` and also this should happen automatically. This makes it impossible to write proper multistep function calling even if given model supports it.

### Proposed solution

At the very least rewrite `useLlm` hook to allow for multistep tool calling. This includes adding automated tool execution upon model request and providing the output inside the generate loop, not allowing user to send message between tool call and execution. Also possibly add some kind of formatting, something alike <thinking> mode in qwen.
Typescript API change are TBD

### Alternative solutions

_No response_

Adds proper tool calling functionalities to rn-executorch, also enables more powerful LLM capabiliies

### Additional context

_No response_

### LINKED ISSUE #1226: Add option to not encode every context data on forward call
Currently, we call reset on every forward call in llms. This causes, especially in multimodal llms (Gemma 4 with multiple audio inputs), that we need to encode everything from the context from scratch. This makes TTFT grows on every call. We need to handle smart strategy to cache already encoded info.

### LINKED ISSUE #1245: [RNE Rewrite] Add LLM chat pipeline implementation
Implement LLM chat pipeline using the new approach to achieve parity with current implementation. It should support:
- [x] text-only LLMs,
- [x] multimodal LLMs,
-  [x] tool calling,
- ~structured output~ (see: https://github.com/pytorch/executorch/issues/19215).

For models like Gemma which require custom runners there should be a separate runner implementation. Also see: #429, #1226, #1225, #1220.

For PoC implementation and discussion see: https://github.com/barhanc/rnet-poc/pull/3.

barhanc:
## Description

This PR implements the native ExecuTorch LLM extension and TypeScript chat session pipeline on top of it. It brings low-level runner bindings for text and multimodal LLMs, high-level multi-turn chat sessions with incremental KV cache prefilling, automated tool calling loops, model definitions for 9 model families (v0.10.0), and a full demo screen in apps/nlp.

####  C++ Native Layer (llm_runner.cpp, llm_runner.h, JSI host objects)

* LLMRunnerHostObject: Created JSI wrapper around ExecuTorch's `TextLLMRunner` and `MultimodalRunner`. Polymorphic dispatch is handled through `IRunner`.

* KV Cache Access & Rewinding (Friend Injection Hack):
   - ExecuTorch's TextLLMRunner and TextDecoderRunner keep critical state private (pos_,
     max_seq_len_, managed_kv_cache_, stop condition pointers) without public getters/setters.
   - Instead of patching upstream headers or modifying submodules, we used template-based
      explicit instantiation friend injection (member-pointer accessors) to gain access to
      private members.
    - This powers getKVCacheState() (live token position, max sequence length, remaining
      capacity, usage ratio) and reset(targetPos) (rewinding the KV cache back to a specific
      token index for multi-turn prefilling and multi-step tool iterations).
* Generation & Thread Safety: Implemented streaming token callbacks with tryLockUnique concurrency guards and abort handling via stop().

#### TypeScript Pipeline & Chat Session (packages/react-native-executorch)

* createChatPreprocessor:
   - Compiles Jinja templates via @huggingface/jinja.
   - Enforces strict multi-turn prefix prefilling (curr.startsWith(prev)). If a template retroactively modifies past turns, it catches and rejects it before sending invalid token slices to ExecuTorch.
    - Multimodal Sentinel Token Resolution: Handles vision inputs by inserting model-specific
      sentinel tokens (e.g.<|image_start|>...<|image_end|>) and matching image
      dimensions based on preprocessorConfig.
* createLLMChatSession / useLLMChatSession:
    - Stateful multi-turn chat orchestrator. Prefills new turns incrementally without reencoding existing conversation history (when flag `resetOnTurn` is set to false, closes #1226).
    - Multi-Turn Tool Calling: When the model outputs tool calls, the session pauses
      generation, executes the tool callbacks, rewinds the KV cache to posAtEndOfUser (runner.
      reset(posAtEndOfUser)), and prefills the tool responses formatted cleanly into the history before
      triggering the next continuation turn (up to maxToolTurns).
    - Automatic memory management with dispose().

#### Model Registry (models.ts)

* Added configurations for 9 model families on NEXT_VERSION_TAG (resolve/v0.10.0):
   - LFM 2.5 (350M, 1.2B, VL 450M, VL 1.6B)
   - Llama 3.2 (1B, 3B - BF16 & SpinQuant)
   - Qwen 2.5 (0.5B, 1.5B, 3B)
   - Qwen 3 (0.6B, 1.7B, 4B)
   - Gemma 4 (E2B)
   - Hammer 2.1 (0.5B, 1.5B, 3B)
   - SmolLM2 (135M, 360M, 1.7B)
   - Phi-4-Mini & Bielik v3.0

The only one missing is the Gemma 4 Multimodal, as stock ExecuTorch multimodal runner doesn't support passing additional PLE encodings. Requires input as what should be the course of action for this model.

#### HuggingFace Chat Template Fixes (v0.10.0)

Audited all 17 templates across our 11 HF repos to guarantee KV cache monotonicity and tool
calling stability:

* Monotonicity (Qwen 3, Qwen 3.5, LFM 2.5): Removed upstream backward lookbacks (messages[::-1] and loop.index0 != ns.last_assistant_index thinking strippers) that retroactively mutated previous assistant turns when new messages arrived.
* Llama 3.2 Tool Crash: Defaulted tools_in_user_message: false. Upstream template crashed with "Cannot put tools in the first user message when there's no first user message!" when initializing a session with just a system prompt. Tool definitions now live in the system header.
* Gemma 4 Single-Turn Structure: Kept Google's continuous <|turn|>model ... <turn|> turn structure intact across tool calls without inserting duplicate generation headers.
* Undefined Content Fallbacks: Added (message.content | default('')) and is iterable and is not string guards across Bielik, SmolLM2, Phi-4-Mini, Hammer, and LFM to prevent Jinja runtime errors when assistant messages only contain tool calls.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

- [ ] Build the NLP app and test models on the newly added LLM screen. The multimodal input and tool calling are preconfigured for suppported models and the support is clearly indicated on the model card.
- [ ] Verify modified chat templates listed above in PR description.

### Related issues

Closes #1245 #1226 #429

### Checklist

### Additional notes

- The one model missing is the Gemma 4 Multimodal, as stock ExecuTorch multimodal runner doesn't support passing additional PLE encodings. Requires input as what should be the course of action for this model.


barhanc:
cc: @kfaracik @NorbertKlockiewicz
You might want to take a look at TS implementation (mostly `llmRunner.ts` and `llmChatSession.ts` files) if you plan to migrate Private Mind to new flow rn-executorch implementation in the foreseeable future.

msluszniak:
> * KV Cache Access & Rewinding (Friend Injection Hack):
>   * ExecuTorch's TextLLMRunner and TextDecoderRunner keep critical state private (pos_,
max_seq_len_, managed_kv_cache_, stop condition pointers) without public getters/setters.
>   * Instead of patching upstream headers or modifying submodules, we used template-based
explicit instantiation friend injection (member-pointer accessors) to gain access to
private members.
>   * This powers getKVCacheState() (live token position, max sequence length, remaining
capacity, usage ratio) and reset(targetPos) (rewinding the KV cache back to a specific
token index for multi-turn prefilling and multi-step tool iterations).

We have to file an issue on upstream executorch to add these getters / setters. Could you create such issue? @barhanc

EDIT: I see there is an RFC for that. https://github.com/pytorch/executorch/issues/20431

msluszniak:
How we would like to deal with this one:

> The one model missing is the Gemma 4 Multimodal, as stock ExecuTorch multimodal runner doesn't support passing additional PLE encodings. Requires input as what should be the course of action for this model.

Do we want to implement it by ourselves first and create a PR for an upstream ET?

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
```diff
@@ -0,0 +1,518 @@
...
+    } else {
+        runner_ = executorch::extension::llm::create_multimodal_runner(modelPath, std::move(tokenizer));
+    }
+
+    if (!runner_) {
+        throw error::LoadFailed("LLMRunner: Failed to create llm runner");
+    }
+
+    auto loadError = runner_->load();
+    if (loadError != executorch::runtime::Error::Ok) {
```
msluszniak: I would rather name it as a `loadStatus` or sth like that

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
```diff
@@ -0,0 +1,518 @@
...
+        return it->second;
+    }
+    return 0;
+}
+
+void setRunnerPos(executorch::extension::llm::IRunner *runner, bool isMultimodal, int64_t targetPos) {
+    if (runner == nullptr) {
+        return;
+    }
+    if (!isMultimodal) {
```
msluszniak: maybe use no-negated value and swap if else blocks

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    auto loadError = runner_->load();
+    if (loadError != executorch::runtime::Error::Ok) {
+        std::string errorMsg = executorch::runtime::to_string(loadError);
+        throw error::LoadFailed(std::format("LLMRunner: Failed to load model: {}", errorMsg), loadError);
+    }
+}
+
+std::unique_lock<std::mutex> LLMRunnerHostObject::tryLockUnique(std::string_view ctx) {
```
msluszniak, line 274: why std::string here is passed as value?

barhanc:
We already maintain a fork of ExecuTorch with our fixes so I guess we could do it there but in a way that doesn't break the contract with the upstream runner API, since if our fork diverges too much from the upstream we might have a problem in the future. Also I'm not sure they will want a change in runner for this specific model as it seems their runner targets a different LMM architecture in general.

msluszniak:
>We already maintain a fork of ExecuTorch with our fixes so I guess we could do it there but in a way that doesn't break the contract with the upstream runner API, since if our fork diverges too much from the upstream we might have a problem in the future. Also I'm not sure they will want a change in runner for this specific model as it seems their runner targets a different LMM architecture in general.

Yeah, it might be the not aligned with the model architecture they target, but I feel like we need to have support for these kinds of models. Otherwise, we drop support for SOTA on-device models.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
```diff
@@ -0,0 +1,518 @@
...
+                finalStats->inference_start_ms = stats.inference_start_ms;
+                finalStats->inference_end_ms = stats.inference_end_ms;
+                finalStats->model_load_start_ms = stats.model_load_start_ms;
+                finalStats->model_load_end_ms = stats.model_load_end_ms;
+                finalStats->aggregate_sampling_time_ms = stats.aggregate_sampling_time_ms;
+            };
+
+            auto lock = self->tryLockUnique("LLMRunner.generate");
+
+            auto genError = executorch::runtime::Error::Ok;
```
msluszniak: assignment not needed here.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.h
```diff
@@ -0,0 +1,83 @@
...
+     * @param modelPath Absolute file system path to the `.pte` LLM model
+     * binary.
+     * @param tokenizerPath Absolute file system path to the local tokenizer
+     * configuration file (e.g. `tokenizer.json`).
+     * @param modalities Vector of supported input modality names (e.g.
+     * `{"image"}`).
+     * @throws core::error::RnExecuTorchException with code LoadFailed if
+     * loading the model or tokenizer fails.
+     */
+    LLMRunnerHostObject(const std::string &modelPath,
```
msluszniak: explicit

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
```diff
@@ -0,0 +1,332 @@
...
+  config: LLMModel,
+  options: LLMChatSessionOptions = {},
+  runtime?: WorkletRuntime
+): Promise<LLMChatSession> {
+  const {
+    generationConfig: defaultGenerationConfig,
+    initialMessages = [],
+    stopRegex,
+    toolOpts,
+    resetOnTurn = false,
```
msluszniak, line 197: Are we sure we don't want to reset it by default?

barhanc: Yes. Resetting prefills entire history on every chat turn, causing growing TTFT. Keeping it false is better and will work the same as resetting if template is monotonic (if it's not the chat preprocessor throws an error).

packages/react-native-executorch/cpp/extensions/llm/llm_runner.h
barhanc: Done.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
barhanc: Do you mean it would be better to do `executorch::runtime::Error genError{};` instead?

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
barhanc: It's passed as `std::string_view`, not `std::string`.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
barhanc: Changed.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
barhanc: Renamed.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
msluszniak: Oh, my bad

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
msluszniak: You know what, I meant something different we cannot apply here, so maybe rename it similarly as genStatus or similar? We can also alias executorch::runtime::Error as Status but it's up to you. For me Status::Ok reads better than Error::Ok heh

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
msluszniak: Understood

.cspell-wordlist.txt
```diff
@@ -314,5 +316,8 @@ Partitioner
 denoised
 ttfa
 TTFA
-phonemized
+agenticphonemized
```
msluszniak: `phonemized` got merged into `agentic` here, so the list now has neither word. `phonemized` is still used in `kokoroTextToSpeech.ts`, `phonemizer.ts` and `models.ts`, and `agentic` appears in `models.ts`, so cspell CI fails on both. The file also lost its trailing newline.

msluszniak (commented):
Review of the LLM extension. Two CI/config blockers, then correctness items in the chat session, the demo screen, and the runner. The cross-runtime host object and host function hops check out: worklets' `makeSerializableFunction` handles JSI host functions, so `wrapAsync(runner.prefill, runtime)` is fine.

apps/nlp/package.json
```diff
@@ -3,6 +3,12 @@
   "version": "1.0.0",
   "main": "expo-router/entry",
   "react-native-executorch": {
+    "backends": [
+      "xnnpack",
+      "coreml",
+      "mlx",
+      "vulkan"
+    ],
```
msluszniak, line 11: The backends are covered, but `libs` is never declared and none of the listed features expand to one, so `download-libs.js` writes `enableOpencv: false` and `extensions::cv::install` is skipped. Sending an image to any of the three LFM2.5-VL entries then dies with a TypeError on undefined `cv` inside the preprocessor (`resize`). Adding `multimodalLLM` to `features` (or `libs: ["opencv"]`) fixes it. Note this only bites when install runs with INIT_CWD at `apps/nlp`, which is why a root `yarn install` hides it.

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
```diff
@@ -0,0 +1,338 @@
...
+    const posAtEndOfUser = runner.getKVCacheState().pos;
+    committed = history.length;
+
+    for (let currentTurn = 0; currentTurn < maxToolTurns; ++currentTurn) {
+      const uncommitted = history.length - committed;
+      const prompt = chatPreprocessor.process(history, uncommitted, { addGenPrompt: true });
+
+      const prefillDurationMs = Date.now() - prefillStartMs;
+
+      const { response, stats } = await generateChatTurn(runner, prompt, generationOpts);
```
msluszniak: No error handling around the generation loop. If `generateChatTurn` or a `prefill` rejects, `runner.reset(posAtEndOfUser)` and `chatPreprocessor.clear()` are skipped while `committed` has already been advanced, so the KV cache keeps orphan tokens and desyncs from `history` for every later turn (and the image tensors leak). A try/catch that rewinds to `posAtEndOfUser`, restores `committed`, and truncates `history` back to `turnStartIdx` would keep the session usable after a failure.

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
```diff
@@ -0,0 +1,338 @@
...
+      chatPreprocessor.clear();
+      generationStatsList.push({ ...stats, firstTokenMs: stats.firstTokenMs + prefillDurationMs });
```
msluszniak: `stats.firstTokenMs` is ET's `Stats::first_token_ms`, an absolute timestamp on the same clock as `inference_start_ms`, not a duration. Adding `prefillDurationMs` shifts it forward, so the demo's `ttft` is inflated and `decodeMs = inferenceEndMs - firstTokenMs` shrinks or goes negative, showing 0.0 tok/s. Prefill time should be reported as its own field instead of folded into a timestamp.

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
```diff
@@ -0,0 +1,338 @@
...
+        history.push({
+          role: 'tool',
+          toolCallId: toolCall.id,
+          name: toolCall.function.name,
+          content: toolContent,
+        });
+      }
+
+      prefillStartMs = Date.now();
+    }
```
msluszniak: If the loop runs out of `maxToolTurns`, `sendMessage` resolves normally with history ending on a tool result and no assistant answer, and the caller gets no signal that the turn was cut short. Worth returning a flag or throwing so the UI can say something.

apps/nlp/app/llm/index.tsx
```diff
@@ -0,0 +1,682 @@
...
+    setAttachedImage(null);
+    setInput('');
+    setStreamingResponse('');
+
+    setTurns((prev) => [
+      ...prev,
+      { role: 'user', content: textMessage, imageUri: currentImage?.uri },
+    ]);
+
+    try {
```
msluszniak, line 192: `try`/`finally` with no `catch`: a rejected `sendMessage` only surfaces as an unhandled rejection and the user sees nothing after their message. The gallery picker right above already alerts on failure, so the same treatment here would be consistent.

apps/nlp/app/llm/index.tsx
```diff
@@ -0,0 +1,682 @@
...
+
+      const generatedTurns: Turn[] = result.messages
+        .filter(
+          (m): m is Extract<typeof m, { role: 'assistant' | 'tool' }> =>
+            m.role === 'assistant' || m.role === 'tool'
+        )
+        .map((m, idx) => ({
+          role: m.role,
+          content: typeof m.content === 'string' ? m.content : '',
+          stats: result.stats[idx],
```
msluszniak: `result.stats` has one entry per generation, but `idx` counts filtered messages. With a tool call the filtered list is [assistant(toolCalls), tool, assistant(final)] against 2 stats entries, so the tool bubble shows the final answer's numbers and the final answer gets `undefined`. The stats need to be keyed to assistant messages only.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
```diff
@@ -0,0 +1,518 @@
...
+    } else {
+        auto *r = dynamic_cast<TextLLMRunner *>(runner);
+        if (r != nullptr) {
+            meta = &(r->*getPrivateMember(TextRunnerMetadataTag{}));
+        }
+    }
+    if (meta == nullptr) {
+        return 0;
+    }
+    auto it = meta->find(executorch::extension::llm::kMaxSeqLen);
```
msluszniak: This prefers `kMaxSeqLen` and falls back to `kMaxContextLen`. Upstream bounds generation by the context length, so when the two differ, `remainingTokens` and `usageRatio` overstate what is left and any trimming built on them overruns the real window. Preferring `kMaxContextLen` looks safer.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
```diff
@@ -0,0 +1,518 @@
...
+        auto fnBody = [self](jsi::Runtime &rt, const jsi::Value & /*thisVal*/, const jsi::Value *args, size_t count) -> jsi::Value {
+            if (count > 1) {
+                throw error::InvalidArgument("LLMRunner.reset: Usage: reset(targetPos?)");
+            }
+
+            auto lock = self->tryLockUnique("LLMRunner.reset");
+            if (count == 0 || args[0].isUndefined()) {
+                (*self->runner_).reset();
+            } else {
+                auto targetPos = static_cast<int64_t>(conversions::asType<uint64_t>(rt, "LLMRunner.reset", args[0]));
```
msluszniak, line 448: `targetPos` goes straight into the private `pos_` with no upper bound, so a value past `max_context_len` leaves the KV write head outside the preallocated cache. Since this is public API, clamping to `[0, currentPos]` (or throwing) would be worth it.

barhanc: Added validation inside reset.

packages/react-native-executorch/cpp/extensions/llm/llm_runner.cpp
barhanc: Switched `kMaxSeqLen` with `kMaxContextLen`.

apps/nlp/app/llm/index.tsx
barhanc: Fixed.

apps/nlp/app/llm/index.tsx
barhanc: Added alert on send message failure.

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
barhanc: Added field to the turn result type which signals the reason generation finished.

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
barhanc: Added `prefillDurationMs` and modified total TTFT calculation in example app.

packages/react-native-executorch/src/extensions/llm/tasks/llmChatSession.ts
barhanc: Added `try { ... } catch { ... }` with explicit rollback on failure to pre-turn state.

apps/nlp/package.json
barhanc: Fixed.

.cspell-wordlist.txt
barhanc: Fixed.

msluszniak (approved):
Tested the app and it worked corretcly.

msluszniak:
One last thing I missed. I exported lfm2.5-vl 1.6 to vulkan. Could you please check it and wire if working correctly? MR with export script is in our internal repository.

barhanc:
> One last thing I missed. I exported lfm2.5-vl 1.6 to vulkan. Could you please check it and wire if working correctly? MR with export script is in our internal repository.

Added vulkan variants and tested them on few simple prompts---they seem to work fine.

barhanc:
@kfaracik We are merging this PR to progress with other tasks. If you find something that would require our attention regarding this new implementation in the future, please open an issue.

Issue #939, opened 2026-03-06
Export and quantize better text-to-image model


barhanc:
Our current diffusion model is quite bad and it's not worth to quantize it right now. In the future we can export better models e.g. https://huggingface.co/IDKiro/sdxs-512-dreamshaper and then return to quantizing it.

barhanc:
Closed as completed in #1307

PR #1307, opened 2026-07-14
[RNE Rewrite] feat: add SDXS text-to-image pipeline

### LINKED ISSUE #1244: [RNE Rewrite] CV - add text-to-image pipeline implementation
Implement text-to-image pipeline using the new approach to achieve parity with current implementation. Also see: #939

msluszniak:
## Description

Adds the **SDXS-512-DreamShaper** text-to-image pipeline to the rewrite CV extension (issue #1244 and #939). SDXS is a distilled **single-step** diffusion model, so the pipeline is a single multi-method `.pte` (`encode` / `denoise` / `decode`) driven by a lightweight TypeScript scheduler + CLIP tokenizer.

- `sdxsTextToImage` task (`createSdxsTextToImage`): tokenize → `encode` → `denoise` → exact single-step DEIS scheduler (linear in latents+eps) → TAESD `decode` → in-memory RGBA `ImageBuffer`.
- `useTextToImage` hook: two-resource download (combined `.pte` + `tokenizer.json`).
- `models.textToImage.SDXS_512_DREAMSHAPER`: `XNNPACK_FP32` (default, universal) and `COREML_FP16` (iOS).
- `math.mulberry32` / `math.randomNormal`: seeded uniform generator + Box–Muller normal draw.
- `apps/computer-vision` Text-to-Image demo screen.

Since the pipeline targets exactly one model, values fixed by the export (shapes, scheduler and decoder scalars) are constants in the task file rather than options — all variants differ only in which `.pte` they load. The convention is written up in `.agents/skills`.

Models exported and published to `software-mansion/react-native-executorch-sdxs-512-dreamshaper` (`v0.10.0`). The exported pipeline reproduces the reference diffusers output pixel-for-pixel (fp32, mean pixel err ~1.8e-5); CoreML fp16 is ~99.7% delegated at mean pixel err ~0.002.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. Run `apps/computer-vision` on a physical device.
2. Open **Text to Image** and pick a variant (`CoreML FP16` is offered on physical iOS devices only).
3. Enter a prompt and tap **Generate**; a 512×512 image renders.

Verified on a physical iPhone 17 Pro (`COREML_FP16`) and a Galaxy S24 Ultra (`XNNPACK_FP32`).

### Related issues

#1244, #939

### Checklist

### Additional notes

**SDXS is highly quantization-sensitive.** Quantizing the UNet convolutions produces noise, so an `8da4w` variant was dropped (poorer quality, and no speed win over fp32 since both are conv-bound).

**No MLX variant.** The MLX backend does not support `aten.native_group_norm`, which a diffusion UNet uses in every ResBlock, so `denoise` fragments into 28 subgraphs and thrashes the GPU/CPU boundary — measured ~3x slower than CoreML on device, with worse quality. The split is structural, not quantization-related: an unquantized fp32 export fragments identically, so no dtype fixes it. Worth revisiting once MLX supports `native_group_norm`.

**The TS scheduler stays in TypeScript.** It is two single passes over 16,384 floats, once per image (~0.34 ms measured, ~2.4 ms at a conservative Hermes penalty) against a UNet forward that dominates by orders of magnitude, so a C++ port is not worth it.


packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,251 @@
...
+  /** Local path to the combined `encode`/`denoise`/`decode` `.pte` program. */
+  readonly modelPath: string;
+  /** Local path to the CLIP `tokenizer.json`. */
+  readonly tokenizerPath: string;
+  /** Model-specific pipeline configuration. */
+  readonly opts: SdxsOptions;
+};
+
+// Encodes a prompt into a fixed-length token id buffer (BOS/EOS added by the
+// tokenizer.json post-processor), truncating or right-padding to CLIP_MAX_TOKENS.
```
msluszniak: Make sure this comment is needed and if not please remove it.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,251 @@
...
+  const tokens = new BigInt64Array(CLIP_MAX_TOKENS);
+  for (let i = 0; i < CLIP_MAX_TOKENS; i++) {
+    tokens[i] = BigInt(i < ids.length ? ids[i]! : CLIP_PAD_TOKEN_ID);
+  }
+  return tokens;
+}
+
+// Deterministic seeded standard-normal noise (mulberry32 + Box–Muller), so the
+// same (prompt, seed) reproduces the same image without relying on Math.random
+// being available on the worklet runtime.
```
msluszniak: Make sure this comment is needed and if not please remove it.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,251 @@
...
+    const mag = Math.sqrt(-2.0 * Math.log(u1));
+    out[i] = mag * Math.cos(2.0 * Math.PI * u2);
+    if (i + 1 < size) out[i + 1] = mag * Math.sin(2.0 * Math.PI * u2);
+  }
+  return out;
+}
+
+// Single-step scheduler update: turns the UNet output into the clean latent that
+// is fed to the decoder. For the distilled single step the DEIS update is exactly
+// linear in the latents and the model output.
```
msluszniak: Make sure this comment is needed and if not please remove it.

packages/react-native-executorch/src/models.ts
```diff
@@ -5,6 +5,7 @@ import type { SemanticSegmentationModel } from './extensions/cv/tasks/semanticSe
 import type { KeypointDetectorModel } from './extensions/cv/tasks/keypointDetection';
 import type { InstanceSegmenterModel } from './extensions/cv/tasks/instanceSegmentation';
 import type { ImageEmbedderModel } from './extensions/cv/tasks/imageEmbedding';
+import type { SdxsModel } from './extensions/cv/tasks/sdxsTextToImage';
```
msluszniak: Probably should be named SdxsTextToImageModel

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,251 @@
...
+  /** The single training timestep fed to the UNet for the distilled step. */
+  readonly timestep: number;
+  /**
+   * Scheduler-step coefficients. For the distilled single step the DEIS update
+   * is exactly linear in the latents and the UNet output:
+   * `clean = sampleCoeff * latents + noiseCoeff * modelOutput`. Both are pinned
+   * from the reference scheduler during export.
+   */
+  readonly sampleCoeff: number;
+  readonly noiseCoeff: number;
```
msluszniak: Why is this member not described?

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Removed — the function name and the documented pad-token constant already cover it.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Kept but tightened to one line; the mulberry32+Box–Muller determinism rationale is non-obvious from the bit math.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Removed — that detail now lives in the sampleCoeff JSDoc and the name is descriptive.

packages/react-native-executorch/src/models.ts
msluszniak: Done, renamed to SdxsTextToImageModel.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Added a description and split the shared coefficient JSDoc so each field is documented.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+  const tokens = new BigInt64Array(CLIP_MAX_TOKENS);
+  for (let i = 0; i < CLIP_MAX_TOKENS; i++) {
+    tokens[i] = BigInt(i < ids.length ? ids[i]! : CLIP_PAD_TOKEN_ID);
+  }
+  return tokens;
+}
+
+// Deterministic seeded standard-normal noise (mulberry32 + Box–Muller): a fixed
+// seed must reproduce the same image, and Math.random cannot be seeded.
+function seededGaussian(size: number, seed: number): Float32Array {
```
barhanc: Perhaps this should be moved to math and maybe separate the seed RNG from the Box-Muller.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+// fixed length when the tokenizer.json does not enforce a padding strategy.
+const CLIP_PAD_TOKEN_ID = 49407;
+
+/**
+ * Model-specific configuration for the SDXS text-to-image pipeline. These
+ * constants describe how the exported program behaves and are pinned during the
+ * `.pte` export by numerically matching the reference diffusers output.
+ * @category Types
+ */
+export type SdxsOptions = {
```
barhanc: Since this is a single-model-specific pipeline, I'm wondering (the same thing as in VAD) which params should be provided as options and defined in `models.ts` and which should just be `const CONSTANT`'s inside the file. Here for example there is a mix of both.

There is also a dilemma with the model schema validation, namely wether to hard code the specific shapes, not validate at all since the .pte file must be very specific, or do a similar validation as pipelines that support multiple models (like the CV pipelines).

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+    const u1 = Math.max(next(), 1e-7);
+    const u2 = next();
+    const mag = Math.sqrt(-2.0 * Math.log(u1));
+    out[i] = mag * Math.cos(2.0 * Math.PI * u2);
+    if (i + 1 < size) out[i + 1] = mag * Math.sin(2.0 * Math.PI * u2);
+  }
+  return out;
+}
+
+function toCleanLatents(
```
barhanc: I think this function could be inlined in the generation loop.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+  };
+
+  const generateWorklet = (prompt: string, seed: number): ImageBuffer => {
+    'worklet';
+
+    tTokens.setData(encodePrompt(tokenizer.encode(prompt)));
+    model.execute('encode', [tTokens], [tEmbeddings]);
+
+    const noise = seededGaussian(latentNumel, seed);
+    for (let i = 0; i < latentNumel; i++) noise[i]! *= opts.initNoiseSigma;
```
barhanc: Perhaps the random normal generator could take mean and std parameters.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+    [SymbolicTensor('float32', [1, latentChannels, latentSize, latentSize])]
+  );
+  validateModelSchema(
+    model,
+    'decode',
+    [SymbolicTensor('float32', [1, latentChannels, latentSize, latentSize])],
+    [SymbolicTensor('float32', [1, 3, imageSize, imageSize])]
+  );
+
+  const latentNumel = latentChannels * latentSize * latentSize;
```
barhanc: Why not just use `tLatents.numel`?

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+export type SdxsTextToImageModel = {
+  /** Local path to the combined `encode`/`denoise`/`decode` `.pte` program. */
+  readonly modelPath: string;
+  /** Local path to the CLIP `tokenizer.json`. */
+  readonly tokenizerPath: string;
+  /** Model-specific pipeline configuration. */
+  readonly opts: SdxsOptions;
+};
+
+function encodePrompt(ids: number[]): BigInt64Array {
```
barhanc: The name doesn't really reflect what this does. Maybe we could just inline this.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    model.execute('decode', [tLatents], [tDecoded]);
+
+    const data = new Uint8Array(imageSize * imageSize * 4);
+    tDecoded
+      .copyTo(tReshape)
+      .through(normalize, tUint8, { alpha: opts.outAlpha, beta: opts.outBeta })
+      .through(toChannelsLast, tChanLast)
+      .through(cvtColor, tRgba, 'RGB2RGBA')
+      .getData(data);
```
barhanc:
```suggestion
    const data = tDecoded
      .copyTo(tReshape)
      .through(normalize, tUint8, { alpha: opts.outAlpha, beta: opts.outBeta })
      .through(toChannelsLast, tChanLast)
      .through(cvtColor, tRgba, 'RGB2RGBA')
      .getData(new Uint8Array(imageSize * imageSize * 4));
```

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,246 @@
...
+    tEmbeddings,
+    tTimestep,
+    tLatents,
+    tNoisePred,
+    tDecoded,
+    tReshape,
+    tUint8,
+    tChanLast,
+    tRgba,
+  ] = tensors;
```
barhanc, line 122:
```suggestion
  // prettier-ignore
  const [
    tTokens, tEmbeddings, tTimestep, tLatents, tNoisePred,
    tDecoded, tReshape, tUint8, tChanLast, tRgba
  ] = tensors;
```

barhanc (commented):
Checked only lib implementation for now. In general the implementation is incredibly clean. I mostly marked some small nits. The one more important issue worth discussing is how we should handle configuration options for single-model-specific pipelines, something I also commented about in the VAD PR.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Done — moved to `extensions/math` as `mulberry32` (seeded uniform) + `randomNormal` (Box–Muller over any uniform generator), so the RNG and the transform are independent.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Done, inlined into the denoise loop (now writes in place, dropping an allocation).

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Done — `randomNormal(size, uniform, mean, std)`. Passing `initNoiseSigma` as `std` folds away the separate scaling pass.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Good catch, done — using `tLatents.numel`/`tNoisePred.numel` and dropped the recomputed `latentNumel`.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Agreed — it padded rather than encoded (the tokenizer does the encoding). Inlined it.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Applied.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Applied.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Went all-constants. The deciding fact: all three variants shared an identical `opts` object, so nothing here was ever per-model config — and every field, if changed, breaks the pipeline (shapes are static in the `.pte`; the scalars are pinned by the export). An option with one valid value is just a footgun, so `SdxsTextToImageModel` is now `{ modelPath, tokenizerPath }`.

Proposed rule for the general case (VAD/TTS too): **varies across shipped variants → `models.ts` opts; fixed by the architecture/export → `const`; per-call user choice → `generate()` arg.**

Also dropped `numInferenceSteps` and the loop — the linear DEIS coefficients only hold for the single distilled step at t=999, so looping them was silently wrong, not just suboptimal.

Kept `validateModelSchema`, now asserting the constants: it's load-time only, and "the .pte must be very specific" argues *for* asserting — it turns a stale/wrong export into a clear error instead of a native crash (which already bit us once on a bad CoreML export).

packages/react-native-executorch/src/hooks/useTextToImage.ts
```diff
@@ -0,0 +1,45 @@
...
+    isResourcesReady
+      ? { ...config, modelPath: localModelPath!, tokenizerPath: localTokenizerPath! }
+      : null,
+    [localModelPath, localTokenizerPath]
+  );
+
+  return {
+    isReady: !!model,
+    error: modelResource.downloadError || tokenizerResource.downloadError || error,
+    downloadProgress: (modelResource.downloadProgress + tokenizerResource.downloadProgress) / 2,
```
barhanc: This should just be `modelResource.downloadProgress` since the file size of the tokenizer is much smaller.

packages/react-native-executorch/src/models.ts
```diff
@@ -590,6 +591,26 @@ const CLIP_VIT_BASE_PATCH32_IMAGE_XNNPACK_INT8: ImageEmbedderModel = {
   opts: CLIP_IMAGE_EMBEDDINGS_OPTS,
 };
 
+// =============================================================================
+// Text to Image
+// =============================================================================
+// These variants differ only in which `.pte` they load. Everything else about
+// the pipeline (shapes, scheduler and decoder scalars) is fixed by the export
+// and is pinned as constants in `sdxsTextToImage.ts`.
```
```suggestion
```

packages/react-native-executorch/src/models.ts
```diff
@@ -590,6 +591,26 @@ const CLIP_VIT_BASE_PATCH32_IMAGE_XNNPACK_INT8: ImageEmbedderModel = {
...
+  tokenizerPath: SDXS_512_DREAMSHAPER_TOKENIZER,
+};
+const SDXS_512_DREAMSHAPER_COREML_FP16: SdxsTextToImageModel = {
+  modelPath: `${BASE_URL}-sdxs-512-dreamshaper/${NEXT_VERSION_TAG}/coreml/sdxs_512_dreamshaper_coreml_fp16.pte`,
+  tokenizerPath: SDXS_512_DREAMSHAPER_TOKENIZER,
+};
+// No MLX variant: the MLX backend does not support `aten.native_group_norm`,
+// which a diffusion UNet uses in every ResBlock, so `denoise` fragments into 28
+// subgraphs and thrashes the GPU/CPU boundary. Measured ~3x slower than CoreML
+// on device, at every dtype (the split is structural, not quantization-related).
```
```suggestion
```

packages/react-native-executorch/src/extensions/math.ts
```diff
@@ -61,3 +61,56 @@ export function threshold(src: Tensor, dst: Tensor, thresholdVal: number): Tenso
...
+ * Draws normally distributed values using the Box–Muller transform over a
+ * caller-supplied uniform generator.
+ * @category Typescript API
+ * @param size The number of values to draw.
+ * @param uniform A generator of uniform values in `[0, 1)`, e.g. {@link mulberry32}.
+ * @param mean The mean of the distribution. Defaults to 0.
+ * @param std The standard deviation of the distribution. Defaults to 1.
+ * @returns A `Float32Array` of `size` normally distributed values.
+ */
+export function randomNormal(
```
barhanc, line 101: Perhaps the signature `randomNormal(size, { mean?, std?, seed? })` would be more intuitive and similar to what other frameworks do.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+): Promise<{
+  /** Releases all allocated native resources. */
+  dispose: () => void;
+  /**
+   * Generates an image from a text prompt.
+   * @param prompt The text prompt describing the desired image.
+   * @param seed Seed for the initial latent noise (same seed → same image).
+   * @returns A promise resolving to the generated RGBA image buffer.
+   */
+  generate: (prompt: string, seed: number) => Promise<ImageBuffer>;
```
barhanc: The seed should be optional imo, so that API users don't have to use weird tricks with `Date.now`, etc.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+export type SdxsTextToImageModel = {
+  /** Local path to the combined `encode`/`denoise`/`decode` `.pte` program. */
+  readonly modelPath: string;
+  /** Local path to the CLIP `tokenizer.json`. */
+  readonly tokenizerPath: string;
+};
+
+/**
+ * Creates an SDXS text-to-image runner backed by a single multi-method `.pte`
+ * program (`encode` / `denoise` / `decode`) and a CLIP tokenizer.
```
barhanc: The comment about .pte program is imo unnecessary in public API doc, it's rather an implementation detail.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Model configuration required to instantiate the SDXS text-to-image runner.
+ *
+ * The pipeline ships as a single `.pte` program exporting three methods —
+ * `encode` (text encoder), `denoise` (UNet) and `decode` (TAESD decoder) — plus
+ * a CLIP `tokenizer.json`.
+ * @category Types
+ */
+export type SdxsTextToImageModel = {
+  /** Local path to the combined `encode`/`denoise`/`decode` `.pte` program. */
```
barhanc: I'm wondering when should we add docs to properties like that and when they are redundant. Right now in most of the CV models there is no description of properties, but I guess they should be there? Also what is the difference between adding docs like this vs with the `@property` symbol?

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+//
+// Every value below is fixed by the exported program rather than being a knob:
+// the shapes are static in the `.pte`, and the scheduler/decoder scalars are
+// pinned during export by numerically matching the reference diffusers output.
+// They are therefore constants here instead of per-model options — all shipped
+// variants (XNNPACK / CoreML / MLX) differ only by which `.pte` they load.
+const CLIP_MAX_TOKENS = 77;
+const CLIP_HIDDEN_SIZE = 768;
+// CLIP pad/eot token id (`<|endoftext|>`). Used to pad the token sequence to a
+// fixed length when the tokenizer.json does not enforce a padding strategy.
```
barhanc:
```suggestion
```
I think the name is clear enough.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+const TIMESTEP = 999;
+// At the distilled single step the DEIS update collapses to an exactly linear
+// combination of the latents and the UNet output:
+// `clean = SAMPLE_COEFF * latents + NOISE_COEFF * modelOutput`. These hold only
+// for one step at TIMESTEP — re-applying them would not be a valid schedule.
+const SAMPLE_COEFF = 14.642591;
+const NOISE_COEFF = -14.579279;
+// The exported `decode` emits RGB in [0,1], so mapping to [0..255] is a plain scale.
+const OUT_ALPHA = 255.0;
+const OUT_BETA = 0.0;
```
barhanc: I think we don't need these. We can just write
```ts
.through(normalize, tUint8, { alpha: 255.0 })
```
directly, as this is fairly standard convention for all CV models. We do similar thing in other CV pipelines.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+// encoder (77 tokens, 768 hidden), a UNet operating on 4-channel latents at
+// 1/8 resolution, and a TAESD tiny decoder. Unlike the classic pipeline it is a
+// distilled, single-step model run without classifier-free guidance, so there
+// is no unconditional branch and no denoising loop.
+//
+// Every value below is fixed by the exported program rather than being a knob:
+// the shapes are static in the `.pte`, and the scheduler/decoder scalars are
+// pinned during export by numerically matching the reference diffusers output.
+// They are therefore constants here instead of per-model options — all shipped
+// variants (XNNPACK / CoreML / MLX) differ only by which `.pte` they load.
```
barhanc:
```suggestion
```

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
```diff
@@ -0,0 +1,182 @@
...
+// Scheduler `init_noise_sigma`: the initial latents are standard normal.
+const INIT_NOISE_SIGMA = 1.0;
+// The single training timestep the model was distilled for.
+const TIMESTEP = 999;
+// At the distilled single step the DEIS update collapses to an exactly linear
+// combination of the latents and the UNet output:
+// `clean = SAMPLE_COEFF * latents + NOISE_COEFF * modelOutput`. These hold only
+// for one step at TIMESTEP — re-applying them would not be a valid schedule.
+const SAMPLE_COEFF = 14.642591;
+const NOISE_COEFF = -14.579279;
```
barhanc:
```suggestion

// At the distilled single step the DEIS update collapses to an exactly linear
// combination of the latents and the UNet output:
// `clean = SAMPLE_COEFF * latents + NOISE_COEFF * modelOutput`. These hold only
// for one step at TIMESTEP — re-applying them would not be a valid schedule.
const TIMESTEP = 999;
const SAMPLE_COEFF = 14.642591;
const NOISE_COEFF = -14.579279;
```

barhanc (commented):
I've tested the example app and it works great. The difference compared to our previous example is staggering. Added only some small nits. In general great work on this one.

packages/react-native-executorch/src/models.ts
msluszniak: This comment should stay. It is non-trivial observation that MLX is 4 times slower than CoreML and that there is no MLX export in the HF/repository. Otherwise, one might try the export in the future to cover more backends. Such information in non-internal gitlab repository is imo helpful.

packages/react-native-executorch/src/extensions/math.ts
msluszniak: Done — `randomNormal(size, { mean?, std?, seed? })`, seeding mulberry32 internally. Verified bit-identical to the prior signature.

packages/react-native-executorch/src/hooks/useTextToImage.ts
msluszniak: Done, using `modelResource.downloadProgress` directly.

packages/react-native-executorch/src/models.ts
msluszniak: Removed.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Done — seed is optional now, defaulting to Date.now() (Date is in the worklet global allowlist).

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Removed the .pte detail from the doc.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Removed.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Done — inlined as { alpha: 255.0 } and dropped the constants.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Applied — grouped TIMESTEP with the coefficients under the shared comment.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: Removed completely.

packages/react-native-executorch/src/extensions/cv/tasks/sdxsTextToImage.ts
msluszniak: TypeDoc documentation suggests that inline comments are preferred and @property should be used only for namespaces and interfaces: https://typedoc.org/documents/Tags._property.html. So it sounds like we should add these to CV models.

packages/react-native-executorch/src/extensions/math.ts
```diff
@@ -61,3 +61,58 @@ export function threshold(src: Tensor, dst: Tensor, thresholdVal: number): Tenso
...
+
+/**
+ * Draws normally distributed values using the Box–Muller transform, seeded via
+ * {@link mulberry32} so a fixed `seed` reproduces the same sequence.
+ * @category Typescript API
+ * @param size The number of values to draw.
+ * @param options Distribution parameters.
+ * @param options.mean The mean of the distribution. Defaults to 0.
+ * @param options.std The standard deviation of the distribution. Defaults to 1.
+ * @param options.seed The seed for the underlying generator. Defaults to 0.
```
barhanc: I feel like not providing seed should default to random results (so like the `Date.now` trick).

apps/computer-vision/app/textToImage/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    generate,
+    generateWorklet,
+  } = useTextToImage(selectedModel);
+
+  const runGenerate = async (sync: boolean) => {
+    if (!generate || !generateWorklet || !prompt.trim()) return;
+    if (!sync) setIsProcessing(true);
+    setError(null);
+    try {
+      const seed = Date.now() % 0x80000000;
```
barhanc: This shouldn't be needed for optional seed, i.e. not passing seed == random results.

packages/react-native-executorch/src/models.ts
barhanc: Ok, I see your point, but that changes our default approach to missing export variants in models.ts from "if it isn't in models.ts then probably it couldn't be exported to that backend" to "if it isn't in models.ts then there must be explicit reason for why it's missing, otherwise assume someone just forgot". Since for most of the models we don't cover all the backends that would require adding quite some comments.

msluszniak: I think after analysing all cases that if we'll try to export all models to mlx where we already have coreml degelation, then removing comment will be fine. In the future, there might be some cases where vulkan will be slower than xnnpack, so then we need to think how to  communicate this fact. For now, I will create an issue for mentioned iOS delegations and remove the comment. Because removing it without any further action won't cover some unhandled cases.

packages/react-native-executorch/src/extensions/math.ts
msluszniak: Done — randomNormal now seeds from Date.now() when no seed is passed, so omitting it gives random results.

apps/computer-vision/app/textToImage/index.tsx
msluszniak: Done — the demo calls generate(prompt) with no seed now.

packages/react-native-executorch/src/models.ts
msluszniak: Created #1318 to decide the convention, and removed the inline comment here as discussed.

barhanc (approved):
🚀

PR #1312, opened 2026-07-16
feat: added system and bundleId to telemetry to version 0.8

### LINKED ISSUE #1313: Add system and bundleId to model downloads telemetry
### Problem description

Currently we don't know on what system the models are used and what apps are using our lib.

### Proposed solution

Add budleId and system to model downloads data sent and displayed on telemetry website for better usage metrics tracking

### Alternative solutions

_No response_

Better usage metrics

### Additional context

_No response_

mkopcins:
## Description

Added two new fields to model downloads telemetry: system and bundleId.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Run any new model (to trigger a download event) and check if new event with that bundle id is registered on the website https://ai.swmansion.com/telemetry/admin/downloads

### Checklist

### DIFF
--- a/apps/llm/package.json
+++ b/apps/llm/package.json
@@ -26,7 +26,7 @@
     "metro-config": "^0.81.5",
     "react": "19.1.0",
     "react-native": "0.81.5",
-    "react-native-audio-api": "^0.8.2",
+    "react-native-audio-api": "^0.12.2",
     "react-native-device-info": "^15.0.2",
     "react-native-executorch": "workspace:*",
     "react-native-executorch-expo-resource-fetcher": "workspace:*",
--- a/packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts
+++ b/packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts
@@ -1,12 +1,7 @@
 import { Platform } from 'react-native';
 import { ResourceSource } from '..';
 import { getModelNameForUrl } from '../constants/modelUrls';
-import {
-  DOWNLOAD_EVENT_ENDPOINT,
-  LIB_VERSION,
-} from '../constants/resourceFetcher';
-import DeviceInfo from 'react-native-device-info';
-
+import { DOWNLOAD_EVENT_ENDPOINT } from '../constants/resourceFetcher';
 /**
  * Http status codes
  * @category Types
@@ -215,6 +210,10 @@ export namespace ResourceFetcherUtils {
     return global.__rne_isEmulator;
   }
 
+  export function getBundleId(): string {
+    return global.__rne_bundleId;
+  }
+
   function getModelNameFromUri(uri: string): string {
     const knownName = getModelNameForUrl(uri);
     if (knownName) {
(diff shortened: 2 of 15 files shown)


apps/llm/package.json
```diff
@@ -26,7 +26,7 @@
     "metro-config": "^0.81.5",
     "react": "19.1.0",
     "react-native": "0.81.5",
-    "react-native-audio-api": "^0.8.2",
+    "react-native-audio-api": "^0.12.2",
```
msluszniak, line 29: Why do we need changes to demo app in `release*` branch?

mkopcins: I had troubles testing it with demo app without those changes

packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts
```diff
@@ -1,9 +1,11 @@
+import { Platform } from 'react-native';
 import { ResourceSource } from '..';
 import { getModelNameForUrl } from '../constants/modelUrls';
 import {
   DOWNLOAD_EVENT_ENDPOINT,
   LIB_VERSION,
 } from '../constants/resourceFetcher';
+import DeviceInfo from 'react-native-device-info';
```
msluszniak: `react-native-device-info` isn't in this package's `dependencies` or `peerDependencies`. It worked for you only because it is hoisted in our demo apps and you probably tested on them. Please add it to dependencies (a peer dep won't work as Metro fails on unresolved imports even inside a try, and it'd force a manual install and re-pod on a patch upgrade). Also, prefer a permissive range like >=13.

Alternative worth considering: inject the bundle id as a JSI global like `__rne_isEmulator`, quick fix. For Android, passing `context.getPackageName()` down from the Kotlin module, roughly 20 lines. This way we avoid imposing new dep.

packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts
```diff
@@ -233,10 +235,12 @@ export namespace ResourceFetcherUtils {
...
-          modelName: getModelNameFromUri(uri),
+          bundleId: DeviceInfo.getBundleId(),
           countryCode: getCountryCode(),
           isEmulator: isEmulator(),
           libVersion: LIB_VERSION,
+          modelName: getModelNameFromUri(uri),
+          system: Platform.OS,
         }),
       });
     } catch (e) {}
```
msluszniak: This try/catch never fires: fetch() rejects asynchronously, so a synchronous catch can't see it. Every telemetry network failure is an unhandled promise rejection in consumer apps. Needs .catch(() => {}) on the promise itself.

packages/react-native-executorch/src/utils/ResourceFetcherUtils.ts
msluszniak: Heads up, unrelated to this PR but blocking it in practice: the cert for ai.swmansion.com expired on Jul 20 (notAfter=Jul 20 21:14:44 2026, Let's Encrypt R12). All download events are currently failing TLS — combined with the comment above, that means an unhandled rejection on every model download. Needs someone with deployment access.

msluszniak (commented):
There's no PrivacyInfo.xcprivacy in the package. Sending bundle id + derived country off-device is the kind of collection Apple's privacy-manifest rules cover — pre-existing gap, but this PR makes it more relevant. Same for docs: telemetry has no prose page and no opt-out, and bundleId identifies the app shipping RNE.

msluszniak:
`ios/PrivacyInfo.xcprivacy` doesn't exist in the package, the commit adds the resource_bundles reference but not the file itself. The `PrivacyInfo.xcprivacy` file needs to be committed here, with the actual `NSPrivacyCollectedDataTypes` entries for the data being sent. Also add `ios/PrivacyInfo.xcprivacy` to the files array in `package.json` so it's included in the npm tarball.

msluszniak (approved):

msluszniak:
Ok, I'm merging this one and now I'll test the 0.9 version of patch. Then will proceed with both patches to v0.8 and v0.9.

PR #1315, opened 2026-07-17
[RNE Rewrite] Use `std::span` instead of raw pointers in native C++ ops + tokenizer cosmetic refactor

### LINKED ISSUE #1308: [RNE Rewrite] Use `std::span` instead of raw pointers in native C++ ops
In the FSMN-VAD framing op (`cpp/extensions/speech/operations.cpp`, #1298) we replaced raw `float*` pointers and manual pointer arithmetic over tensor buffers with `std::span`: wrapping `reinterpret_cast<T*>(tensor->data_.get())` in a `span`, taking a `subspan` per row/window, and using range-based algorithms (`std::ranges::fill`, range-`for`). This is more expressive and less error-prone than raw pointer math.

The scope of this issue is to apply the same treatment to the other native ops that still index tensor buffers through raw pointers:

- `cpp/extensions/cv/image_ops.cpp`
- `cpp/extensions/math/operations.cpp`
- `cpp/extensions/cv/box_ops.cpp`

Core files (`cpp/core/{tensor,model}.cpp`) also use raw buffer access, but per the core guidelines they should be touched with more care — out of scope here unless it is a clear, low-risk win.

msluszniak:
## Description

Applies the treatment from the FSMN-VAD framing op (#1298) to the remaining native ops that indexed tensor buffers through raw pointers: wrap the buffer in a `std::span`, take a `subspan` per row/lane, and use range-based algorithms instead of manual pointer arithmetic.

- **`math/operations.cpp`** — `sigmoid`/`threshold` become `std::ranges::transform`; `softmax`/`argmax` take a per-lane `subspan` instead of re-deriving offsets from a raw pointer on each access.
- **`cv/image_ops.cpp`** — channel-plane and colormap-pixel access go through subspans; the `toChannelsFirst` `memcpy` becomes `std::ranges::copy`.
- **`cv/box_ops.cpp`** — `decodeToXyxy` takes a `std::span<const float, 4>` row rather than four loose floats, removing the `boxesPtr[idx * 4 + 0..3]` unpacking at both call sites.

Byte-level spans are bounded by `size_` (bytes) and element spans by `numel_`, so every `subspan` is bounds-checked against the real allocation.

`cv::Mat` constructions that take `data_.get()` are left as-is: they hand the whole buffer to OpenCV rather than doing pointer math, so a span adds nothing. Core files are untouched, per the issue.

Also drops the five repeated `if (!result.ok())` blocks in `nlp/tokenizer.cpp` in favour of a local `unwrap` helper, mirroring the one in `core/model.cpp` (that one is file-local and typed to `executorch::runtime::Result`, so it cannot be reused for `tokenizers::Result`).

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

Refactor is behaviour-preserving, and verification so far is compile-only — not yet run on a device.

All four translation units compile clean under the project's strict warning set from `.clangd` (`-Wall -Wextra -Wpedantic -Wconversion -Wsign-conversion -Wold-style-cast -Wcast-align -Wshadow -Wdouble-promotion`), with no new warnings:

```
cd packages/react-native-executorch
for f in cpp/extensions/{math/operations.cpp,cv/image_ops.cpp,cv/box_ops.cpp,nlp/tokenizer.cpp}; do
  clang++ -fsyntax-only $(tr '\n' ' ' < compile_flags.txt) -Wall -Wextra -Wpedantic -Wconversion -Wsign-conversion "$f"
done
```

To exercise at runtime: run `apps/computer-vision` (classification/segmentation/detection cover `softmax`, `argmax`, `applyColormap`, `resize`, `normalize`, `toChannelsFirst`, and `nms`) and the tokenizer screen in `apps/nlp`.

### Related issues

Closes #1308

### Checklist

### Additional notes

### DIFF
--- a/packages/react-native-executorch/cpp/extensions/cv/box_ops.cpp
+++ b/packages/react-native-executorch/cpp/extensions/cv/box_ops.cpp
@@ -3,10 +3,12 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
+#include <cstddef>
 #include <format>
 #include <numeric>
 #include <optional>
 #include <shared_mutex>
+#include <span>
 #include <stdexcept>
 #include <utility>
 #include <vector>
@@ -60,9 +62,13 @@ NmsType parseNmsType(const std::string &s) {
     throw std::invalid_argument("unsupported nmsType '" + s + "'");
 }
 
-std::array<float, 4> decodeToXyxy(
-    float a, float b, float c, float d,
-    BoxFormat format) {
+constexpr size_t kBoxCoords = 4;
+
+std::array<float, kBoxCoords> decodeToXyxy(std::span<const float, kBoxCoords> box, BoxFormat format) {
+    const float a = box[0];
+    const float b = box[1];
+    const float c = box[2];
+    const float d = box[3];
     switch (format) {
     case BoxFormat::XYXY:
         return {a, b, c, d};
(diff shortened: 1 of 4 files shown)


msluszniak:
Split into two scoped PRs: #1319 (std::span refactor, closes #1308) and #1320 (tokenizer unwrap cleanup, part of #1316).

PR #1317, opened 2026-07-17
[RNE Rewrite] feat: add supertonic TTS pipeline

### LINKED ISSUE #1250: [RNE Rewrite] Speech - add Kokoro/Supertonic TTS pipeline implementation
Implement Kokoro TTS pipeline using the new approach to achieve parity with current implementation.

barhanc:
## Description

Add supertonic TTS pipeline and example app screen to test it.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Run the Speech example app on both iOS and Android and test the newly added TTS functionality.

### Related issues

Part of #1250 

### Checklist

### Additional notes

Should wait for #1328 so that the hook implementation can be refactored to a more concise and elegant one.

### DIFF
--- a/apps/speech/app/text-to-speech/index.tsx
+++ b/apps/speech/app/text-to-speech/index.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { View, Text, StyleSheet, ScrollView, TextInput } from 'react-native';
 import {
-  useSupertonicTextToSpeech,
+  useTextToSpeech,
   models,
   SUPERTONIC_SAMPLE_RATE,
   SUPERTONIC_SUPPORTED_LANGUAGES,
@@ -74,8 +74,9 @@ function TTSContent() {
   const queueSourceRef = useRef<AudioBufferQueueSourceNode | null>(null);
   const isPlayingRef = useRef(false);
 
-  const { isReady, downloadProgress, error, synthesize, synthesizeStop } =
-    useSupertonicTextToSpeech(models.textToSpeech.SUPERTONIC[selectedModel]);
+  const { isReady, downloadProgress, error, synthesize, synthesizeStop } = useTextToSpeech(
+    models.textToSpeech.SUPERTONIC[selectedModel]
+  );
 
   const getAudioContext = useCallback(async () => {
     if (!audioCtxRef.current || audioCtxRef.current.state === 'closed') {
--- a/apps/speech/app/text-to-speech/index.tsx
+++ b/apps/speech/app/text-to-speech/index.tsx
@@ -52,8 +52,14 @@ const STEPS_OPTIONS = [
   { label: '12', value: 12 },
 ];
 
+const MODEL_OPTIONS = [
+  { label: 'XNNPACK (CPU)', value: 'XNNPACK_FP32' as const },
+  { label: 'MLX (Apple Silicon)', value: 'MLX_FP32' as const },
+];
+
 function TTSContent() {
   const [text, setText] = useState(SAMPLE_TEXT);
+  const [selectedModel, setSelectedModel] = useState<'XNNPACK_FP32' | 'MLX_FP32'>('XNNPACK_FP32');
   const [selectedVoice, setSelectedVoice] = useState<constants.SupertonicDefaultVoiceName>('F1');
   const [selectedLang, setSelectedLang] = useState<SupertonicLanguage>('en');
   const [speed, setSpeed] = useState(1.05);
@@ -69,10 +75,10 @@ function TTSContent() {
   const isPlayingRef = useRef(false);
 
   const { isReady, downloadProgress, error, synthesize, synthesizeStop } =
-    useSupertonicTextToSpeech(models.textToSpeech.SUPERTONIC);
+    useSupertonicTextToSpeech(models.textToSpeech.SUPERTONIC[selectedModel]);
 
   const getAudioContext = useCallback(async () => {
-    if (!audioCtxRef.current) {
+    if (!audioCtxRef.current || audioCtxRef.current.state === 'closed') {
       audioCtxRef.current = new AudioContext({ sampleRate: SUPERTONIC_SAMPLE_RATE });
     }
     if (audioCtxRef.current.state === 'suspended') {
@@ -95,7 +101,10 @@ function TTSContent() {
     return () => {
       synthesizeStop?.();
       stopAudioQueue();
-      audioCtxRef.current?.close().catch(() => {});
+      if (audioCtxRef.current) {
+        audioCtxRef.current.close().catch(() => {});
+        audioCtxRef.current = null;
+      }
     };
   }, [stopAudioQueue, synthesizeStop]);
 
@@ -208,8 +217,17 @@ function TTSContent() {
         </View>
       )}
 
-      {/* Voice & Language Picker */}
+      {/* Model & Voice & Language Picker */}
       <View style={styles.card}>
+        <ModelPicker
+          label="Backend"
+          options={MODEL_OPTIONS.map((m) => ({
+            ...m,
+            disabled: isBusy,
+          }))}
+          selectedValue={selectedModel}
+          onValueChange={setSelectedModel}
+        />
         <ModelPicker
           label="Voice"
           options={VOICE_OPTIONS.map((v) => ({
--- a/apps/speech/app/text-to-speech/index.tsx
+++ b/apps/speech/app/text-to-speech/index.tsx
@@ -35,14 +35,12 @@ const LANGUAGE_OPTIONS = SUPERTONIC_SUPPORTED_LANGUAGES.map((lang) => ({
 }));
 
 const SPEED_OPTIONS = [
-  { label: '0.7x', value: 0.7 },
+  { label: '0.8x', value: 0.8 },
   { label: '0.85x', value: 0.85 },
   { label: '1.0x', value: 1.0 },
   { label: '1.05x', value: 1.05 },
   { label: '1.10x', value: 1.1 },
   { label: '1.2x', value: 1.2 },
-  { label: '1.5x', value: 1.5 },
-  { label: '2.0x', value: 2.0 },
 ];
 
 const STEPS_OPTIONS = [
--- a/apps/speech/app/text-to-speech/index.tsx
+++ b/apps/speech/app/text-to-speech/index.tsx
@@ -1,0 +1,416 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import { View, Text, StyleSheet, ScrollView, TextInput } from 'react-native';
+import {
+  useSupertonicTextToSpeech,
+  models,
+  SUPERTONIC_SAMPLE_RATE,
+  SUPERTONIC_SUPPORTED_LANGUAGES,
+  constants,
+  type SupertonicLanguage,
+} from 'react-native-executorch';
+import { AudioContext, type AudioBufferQueueSourceNode } from 'react-native-audio-api';
+
+import ScreenWrapper from '../../components/ScreenWrapper';
+import { ModelPicker } from '../../components/ModelPicker';
+import { ModelStatus } from '../../components/ModelStatus';
+import { Button } from '../../components/Button';
+import { theme } from '../../theme';
+
+const SAMPLE_TEXT =
+  'The SuperTonic text-to-speech model generates natural sounding speech entirely on-device. ' +
+  'It supports over thirty languages and ten distinct voice styles, from deep male voices to bright female ones. ' +
+  'Unlike cloud-based solutions, SuperTonic runs completely offline with no internet connection required. ' +
+  'The model uses a flow-matching architecture that iteratively denoises random latent representations into coherent speech waveforms. ' +
+  'Each voice style is encoded as a compact embedding that captures the unique timbre, pitch, and speaking patterns of the target speaker. ' +
+  'This makes it ideal for accessibility applications, voice assistants, and content creation tools that need high-quality speech synthesis without sending data to external servers.';
+
+const VOICE_OPTIONS = constants.SUPERTONIC_DEFAULT_VOICE_NAMES.map((name) => ({
+  label: name,
+  value: name as constants.SupertonicDefaultVoiceName,
+}));
+
+const LANGUAGE_OPTIONS = SUPERTONIC_SUPPORTED_LANGUAGES.map((lang) => ({
+  label: lang,
+  value: lang,
+}));
+
+const SPEED_OPTIONS = [
+  { label: '0.7x', value: 0.7 },
+  { label: '0.85x', value: 0.85 },
+  { label: '1.0x', value: 1.0 },
+  { label: '1.05x', value: 1.05 },
+  { label: '1.10x', value: 1.1 },
+  { label: '1.2x', value: 1.2 },
+  { label: '1.5x', value: 1.5 },
+  { label: '2.0x', value: 2.0 },
+];
+
+const STEPS_OPTIONS = [
+  { label: '6', value: 6 },
+  { label: '8 (default)', value: 8 },
+  { label: '10', value: 10 },
+  { label: '12', value: 12 },
+];
+
+function TTSContent() {
+  const [text, setText] = useState(SAMPLE_TEXT);
+  const [selectedVoice, setSelectedVoice] = useState<constants.SupertonicDefaultVoiceName>('F1');
+  const [selectedLang, setSelectedLang] = useState<SupertonicLanguage>('en');
+  const [speed, setSpeed] = useState(1.05);
+  const [totalSteps, setTotalSteps] = useState(8);
+  const [isSynthesizing, setIsSynthesizing] = useState(false);
+  const [isPlaying, setIsPlaying] = useState(false);
+  const [chunkProgress, setChunkProgress] = useState<string | null>(null);
+  const [runError, setRunError] = useState<string | null>(null);
+  const [totalDuration, setTotalDuration] = useState<number | null>(null);
+
+  const audioCtxRef = useRef<AudioContext | null>(null);
+  const queueSourceRef = useRef<AudioBufferQueueSourceNode | null>(null);
+  const isPlayingRef = useRef(false);
+
+  const { isReady, downloadProgress, error, synthesize, synthesizeStop } =
+    useSupertonicTextToSpeech(models.textToSpeech.SUPERTONIC);
+
+  const getAudioContext = useCallback(async () => {
+    if (!audioCtxRef.current) {
+      audioCtxRef.current = new AudioContext({ sampleRate: SUPERTONIC_SAMPLE_RATE });
+    }
+    if (audioCtxRef.current.state === 'suspended') {
+      await audioCtxRef.current.resume();
+    }
+    return audioCtxRef.current;
+  }, []);
+
+  const stopAudioQueue = useCallback(() => {
+    if (queueSourceRef.current) {
+      queueSourceRef.current.clearBuffers();
+      queueSourceRef.current.stop();
+      queueSourceRef.current = null;
+    }
+    isPlayingRef.current = false;
+    setIsPlaying(false);
+  }, []);
+
+  useEffect(() => {
+    return () => {
+      synthesizeStop?.();
+      stopAudioQueue();
+      audioCtxRef.current?.close().catch(() => {});
+    };
+  }, [stopAudioQueue, synthesizeStop]);
+
+  const preparePlaybackSource = useCallback(async () => {
+    stopAudioQueue();
+    const ctx = await getAudioContext();
+
+    const source = ctx.createBufferQueueSource();
+    source.connect(ctx.destination);
+    source.onBufferEnded = (event) => {
+      if (event.isLastBufferInQueue) {
+        isPlayingRef.current = false;
+        setIsPlaying(false);
+      }
+    };
+    queueSourceRef.current = source;
+    return { ctx, source };
+  }, [getAudioContext, stopAudioQueue]);
+
+  const enqueueChunk = useCallback(
+    (ctx: AudioContext, source: AudioBufferQueueSourceNode, audio: Float32Array) => {
+      const buffer = ctx.createBuffer(1, audio.length, SUPERTONIC_SAMPLE_RATE);
+      buffer.copyToChannel(audio as Float32Array<ArrayBuffer>, 0);
+      source.enqueueBuffer(buffer);
+    },
+    []
+  );
+
+  const handleSynthesize = async () => {
+    if (!synthesize || isSynthesizing || !text.trim()) return;
+
+    setRunError(null);
+    setChunkProgress(null);
+    setTotalDuration(null);
+    setIsSynthesizing(true);
+
+    try {
+      const { ctx, source } = await preparePlaybackSource();
+      let durationSum = 0;
+      let started = false;
+      let ttfa: number | null = null;
+      const t0 = performance.now();
+
+      for await (const chunk of synthesize(text, {
+        voiceStyle: selectedVoice,
+        speed,
+        lang: selectedLang,
+        totalSteps,
+      })) {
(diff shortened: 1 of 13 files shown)


packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+const DEFAULT_SPEED = 1.05;
+const MIN_SPEED = 0.7;
+const MAX_SPEED = 2.0;
+
+function getDefaultMaxChunkLength(lang?: SupertonicLanguage): number {
+  switch (lang) {
+    case 'ko':
+      return 120;
+    default:
+      return 300;
```
msluszniak: shouldn't we do the same for Japanese?

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+   * @param options.voiceStyle Voice style key or {@link SupertonicVoiceStyle} object.
+   * @param options.speed Speech speed factor (range: 0.7 to 2.0). Defaults to 1.05.
+   * @param options.totalSteps Number of flow-matching denoising steps. Defaults to 8.
+   * @param options.lang Language ISO code. Defaults to 'na'.
+   * @param options.maxChunkLength Maximum character limit per text chunk. Defaults to 300 (120 for 'ko').
+   * @returns An AsyncGenerator yielding {@link SupertonicTtsChunk} audio buffers.
+   */
+  synthesize: (
+    text: string,
+    options: SupertonicTtsOptions<K>
```
msluszniak, line 150: What is our convention right now, when do we use `opts` and when `options`?

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+
+    // Collect dynamic execution tensors for cleanup in a single try/finally block
+    const auxTensors: Tensor[] = [];
+
+    try {
+      // 1. Predict duration
+      const tTextIds = tensor('int64', [1, textLen], textIdsData);
+      const tTextMask = tensor('float32', [1, 1, textLen], new Float32Array(textLen).fill(1.0));
+      auxTensors.push(tTextIds, tTextMask);
+
```
msluszniak, line 292: Don't we want to catch here some errors on running model inferences?

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+const LATENT_DIM = 24 * CHUNK_COMPRESS_FACTOR; // 144 — latent channels
+const TEXT_EMB_DIM = 256;
+const STYLE_DP_SHAPE = [1, 8, 16] as const; // duration predictor style: 8 prototypes × 16 dims
+const STYLE_TTL_SHAPE = [1, 50, 256] as const; // text-to-latent style: 50 tokens × 256 dims
+
+// Synthesis Execution Defaults & Boundaries
+const DEFAULT_TOTAL_STEPS = 8;
+const DEFAULT_SPEED = 1.05;
+const MIN_SPEED = 0.7;
+const MAX_SPEED = 2.0;
```
msluszniak: Are these values taken from supertonic repo or based on experiments?

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
```diff
@@ -0,0 +1,130 @@
...
+      }
+    }
+  }
+
+  if (minCost[breakpoints.length - 1] === Infinity) {
+    throw new Error(`partition: text cannot be divided into chunks of length <= ${limit}`);
+  }
+
+  const cuts: number[] = [];
+  let i = breakpoints.length - 1;
```
msluszniak, line 192: Backtracking always starts at the last breakpoint, so we force a cut there even when the whole text fits under the limit:

```js
partition('Hello world', 300)  // ['Hello', 'world']
partition('Hello world.', 300) // ['Hello world.']  ok
```

Since `preprocessText` appends the trailing `.` *after* partitioning, any input not ending in punctuation gets its last word synthesized as its own chunk (extra 4 model runs plus an audible seam). Could we model the end of the text as a virtual terminal node rather than forcing a cut at the last breakpoint?

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
```diff
@@ -0,0 +1,130 @@
...
+    return [];
+  }
+
+  if (limit < MIN_PARTITION_LIMIT) {
+    throw new Error(`partition: limit ${limit} is below minimum ${MIN_PARTITION_LIMIT}`);
+  }
+
+  const breakpoints: { idx: number; tag: Tag }[] = [];
+  let charIdx = 0;
+  for (const char of text) {
```
msluszniak, line 117: `for...of` iterates code points but `sliceAtCuts` uses `slice`, which is code unit based. A single non-BMP char shifts every cut after it:

```js
partition('Ok. Fine 🙂 yes, this text has an emoji before a comma and needs splitting properly.', 40)
// ['Ok. Fine 🙂 yes', ', this text has an emoji before a', 'comma and needs splitting properly', '.']
```

Note the rejoined output no longer matches the input. This is reachable in practice because emoji stripping lives in `preprocessText`, which runs after partitioning. `charIdx += char.length` should fix it.

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  const cuts: number[] = [];
+  let i = breakpoints.length - 1;
+  while (i >= 0) {
+    cuts.push(breakpoints[i]!.idx);
+    i = predecessor[i]!;
+  }
+
+  return sliceAtCuts(text, cuts.reverse());
```
msluszniak, line 198: The DP bounds segments between breakpoints, but the tail after the last breakpoint never gets checked against `limit`:

```js
partition('Short one. ' + 'x'.repeat(350), 300) // second chunk is 350 chars
```

Looks like the leftover case from the "enforce limit in all code paths" commit.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+    const maxChunkLength = options.maxChunkLength ?? getDefaultMaxChunkLength(options.lang);
+
+    const voiceStyle =
+      typeof options.voiceStyle === 'object'
+        ? options.voiceStyle
+        : parsedVoiceStyles[options.voiceStyle];
+
+    isCancelled.setBlocking(false);
+
+    const textChunks = partition(text, maxChunkLength);
```
msluszniak: We partition the raw text and preprocess each chunk afterwards, but `preprocessText` grows the chunk: NFKD splits Hangul syllables into jamo (roughly 3x), plus the `<lang></lang>` wrapper and the trailing dot. So `maxChunkLength` isn't actually bounding what reaches the model. Any reason not to preprocess first and then partition?

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+    if (!text || !text.trim()) {
+      throw new Error('synthesize: Input text cannot be empty.');
+    }
+
+    const speed = options.speed ?? DEFAULT_SPEED;
+    if (speed < MIN_SPEED || speed > MAX_SPEED) {
+      throw new Error(`synthesize: speed must be between ${MIN_SPEED} and ${MAX_SPEED}.`);
+    }
+
+    const totalSteps = options.totalSteps ?? DEFAULT_TOTAL_STEPS;
```
msluszniak, line 367: `speed` gets a range check but `totalSteps` doesn't. `totalSteps: 0` skips the denoising loop entirely and sends raw gaussian noise to the vocoder.

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
```diff
@@ -0,0 +1,99 @@
...
+  const dpRes = useResourceDownload(config.modelPaths.durationPredictor, options?.preventLoad);
+  const teRes = useResourceDownload(config.modelPaths.textEncoder, options?.preventLoad);
+  const veRes = useResourceDownload(config.modelPaths.vectorEstimator, options?.preventLoad);
+  const vocRes = useResourceDownload(config.modelPaths.vocoder, options?.preventLoad);
+  const indexerRes = useResourceDownload(config.unicodeIndexerPath, options?.preventLoad);
+
+  const voiceKeys = Object.keys(config.voiceStyles) as K[];
+  const voiceResList = voiceKeys.map((key) =>
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useResourceDownload(config.voiceStyles[key], options?.preventLoad)
```
msluszniak: Hooks in a loop. This works today only because both registry entries happen to have 10 voices, but `voiceStyles` is `Record<K, string>` in the public API, so anyone passing a different number of voices (or changing it at runtime) breaks hook ordering. `useSpeechToText` keeps the call count fixed for this reason. Could we have a hook variant that takes an array of sources, or move the voice style fetching into the task itself?

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
```diff
@@ -0,0 +1,99 @@
...
+          vectorEstimator: veRes.localPath!,
+          vocoder: vocRes.localPath!,
+        },
+        unicodeIndexerPath: indexerRes.localPath!,
+        voiceStyles: localVoiceStyles!,
+      }
+    : null;
+
+  const configKey = localConfig
+    ? JSON.stringify(localConfig.modelPaths) + localConfig.unicodeIndexerPath
```
msluszniak: `configKey` leaves out the voice style paths. They get parsed once into `parsedVoiceStyles` at create time, so swapping `voiceStyles` while keeping the same `.pte` paths silently keeps the old voices loaded.

msluszniak (commented):
Main thing is `partition()`, I found three cases where it misbehaves (comments inline). The first one fires on pretty much any input that doesn't end in punctuation, so I'd want that sorted before merge.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+        );
+        tDenoisedLatent.copyTo(tNoisyLatent);
+      }
+
+      // 4. Vocoder waveform generation
+      models.vocoder.execute('forward', [tNoisyLatent], [tWav]);
+
+      const audio = tWav.getData(new Float32Array(tWav.numel));
+
+      return { audio, sampleRate: SUPERTONIC_SAMPLE_RATE, duration: durationSec };
```
msluszniak: `duration` here is the predicted `durationSec`, but the buffer we actually emit is `CHUNK_SIZE * latentLen` samples, which is the ceil'd value. So the two drift apart. The example screen sums `chunk.duration` for both the RTF log and the "Generated Xs of audio" label, so both are a bit off. `audio.length / SUPERTONIC_SAMPLE_RATE` gives the exact figure.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+    ),
+  });
+
+  // Parse unicode indexer JSON
+  const indexStr = await RNFS.readFile(config.unicodeIndexerPath, 'utf8');
+  const indexer: readonly number[] = JSON.parse(indexStr);
+
+  // Pre-parse voice styles map into memory
+  const parsedVoiceStyles = {} as Record<K, SupertonicVoiceStyle>;
+  for (const [key, path] of Object.entries(config.voiceStyles) as [K, string][]) {
```
msluszniak, line 250: We download and `JSON.parse` all 10 voice styles up front (roughly 2.9 MB of JSON, ~292 KB each) even though a call uses one. That's a chunk of blocking parse work on every pipeline creation, including when the user just switches backend. Could we parse lazily on first use of a given key, or ship the style embeddings as binary instead of JSON?

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
```diff
@@ -0,0 +1,375 @@
...
+      auxTensors.forEach((t) => t.dispose());
+    }
+  };
+
+  const synthesizeChunk = wrapAsync(synthesizeChunkWorklet, runtime);
+
+  const isCancelled = createSynchronizable(false);
+  const synthesizeStop = () => isCancelled.setBlocking(true);
+
+  async function* synthesize(
```
msluszniak, line 350: `synthesize` is effectively single flight but nothing says so. Two overlapping iterations write to the same `tStyleDp`/`tStyleTtl`/`tStep` between inference calls, and the second one resets `isCancelled` while the first is still running. The example screen happens to guard this with `isSynthesizing`, but a library consumer has no way to know.

Guarding it properly means either a mutex or per call tensors, which feels like more than this PR needs. Could we just note on the JSDoc that concurrent `synthesize` calls aren't supported?

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: In all task pipeline implementations we use `options` right now for user-facing API. In helpers we mostly use `opts` but there are some inconsistencies there. I think we should make them all use one or the other (which one would you prefer?), but this can be done in a follow-up style PR.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: What errors do you have in mind? We generally don't catch anything in any other task implementation.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: These are taken from @IgorSwat C++ implementation on main branch.

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
barhanc: This implementation is just temporary as per PR description. I want to wait with this PR until the resource-fetcher #1328 lands because then the hook will be exactly the same as for all the other tasks.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: Yes, good catch.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: Refactored the loop a bit to use the same pattern as Whisper streaming with `isSynthesizing` guard (the worklet synchronizable was redundant there as the main loop wasn't running on a different thread, only the `synthesizeChunk` was).

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
msluszniak: I'm in favour of options everywhere. `Opts` suffixes in config object field and types are fine I guess. Also when we add PR with refactor we should add note in `core-guidelines/SKILL.md`.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
msluszniak: Nvm, my bad.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
msluszniak: Ok

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: Added check for `totalSteps`.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: This is by design. Namely there is an option to pass any custom voice style vector directly to `synthesize` skipping the config. The `voiceStyles` property in model config is there exactly to have preloaded voice style vectors with type-safe names. Also the lightest model is already >300MB of memory so these 3MB of voice vectors is <1%.

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
barhanc: I refactored the `preprocessText` into two helpers `cleanText` and `formatChunk` and fixed the partitioning algorithm. This fixes all the issues tagged regarding the partition problems.

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
barhanc: Fixed.

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
barhanc: Fixed. It now correctly throws.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: We now preprocess first and I decreased the `maxChunkLength` per language slightly to be sure it fits in the model.

packages/react-native-executorch/src/extensions/speech/tasks/supertonicTextToSpeech.ts
barhanc: Fixed.

apps/speech/app/text-to-speech/index.tsx
msluszniak, line 1: Demo app doesn't protect from choosing MLX on Android.

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
msluszniak: Still repros after the two phase split, though my original example doesn't. `cleanText` strips 🙂 before partitioning now, so that specific case is genuinely fixed.

What still gets through are the non-BMP chars `EMOJI_PATTERN` doesn't cover: `1F000-1F0FF` (mahjong, playing cards), `1F200-1F2FF`, `1D100-1D1FF`, and CJK Ext B. That last one is the one that matters, since those show up in Chinese and Japanese proper names and we advertise `ja` and `zh`.

```js
const t = cleanText('Ok. Fine 𠮷 yes, this text has some more words, and then a final clause to force a split.');
partition(t, 40);
// ['Ok. Fine 𠮷 yes', ', this text has some more words, and', 'then a final clause to force a split.']
partition(t, 40).join(' ') === t; // false
```

Second chunk picks up the stray comma. `charIdx` counts code points but `sliceAtCuts` slices by code units, and the terminal breakpoint you added uses `text.length - 1`, which is code units, so the array now mixes both conventions. `charIdx += char.length` on line 54 fixes it and makes the two consistent.

The invariant worth asserting: rejoining the chunks reproduces the cleaned input. It holds for any ASCII text, which is why this isn't obvious from the demo app.

msluszniak:
When I set the slowest speed aka 0.7 I got in half of the cases strange artefacts with repetitions of some parts of sentences. Exact parameters for reproduction:
device: Android
backend: Xnnpack
voice: M1
language: pl
speed: 0.7
denoising steps: 8

Sentence: "Cześć co tam słychać? Testuję funkcje mowy w naszej demo aplikacji."

EDIT: When I tested the same configuration but with speed 2.0, the results were even worse :((

apps/speech/app/text-to-speech/index.tsx
barhanc: Fixed.

barhanc:
> When I set the slowest speed aka 0.7 I got in half of the cases strange artefacts with repetitions of some parts of sentences. Exact parameters for reproduction: device: Android backend: Xnnpack voice: M1 language: pl speed: 0.7 denoising steps: 8
> 
> Sentence: "Cześć co tam słychać? Testuję funkcje mowy w naszej demo aplikacji."
> 
> EDIT: When I tested the same configuration but with speed 2.0, the results were even worse :((

Yeah, from my experiments the extreme values for speed are often hit-or-miss depending on the voice and text to synthesize. I checked the Transformers.js implementation of Supertonic (https://huggingface.co/spaces/webml-community/Supertonic-TTS-WebGPU) and they only allow speeds in [0.8; 1.2]. This is a limitation of the model I believe so we can narrow the available speed range.

msluszniak:
Yeah, I'm in favour of limiting speed min and max.

packages/react-native-executorch/src/extensions/speech/utils/textPartitioner.ts
barhanc: Added `charIdx += char.length` fix. Now should be fine.

msluszniak (approved):
LGTM! If this PR is fine for you @IgorSwat ,we just need to rebase it with resource fetcher PR and ship :))

IgorSwat (changes requested):
Text partitioner in this form is unacceptable. It's not prioritizing low latency (time to first audio chunk) as the old one does. And it will hurt the performance (particularly for Kokoro) very seriously.

I would just add a mechanic (like a flag or smth) that if set, makes the algorithm prioritize splitting heavily at the beginning and slowly disable it as we process more and more of text.

msluszniak:
For Kokoro, there will be a separate flow I guess, since these models are too specific to support them both under same functionality.

IgorSwat:
> For Kokoro, there will be a separate flow I guess, since these models are too specific to support them both under same functionality.

In current codebase same TextPartitioner is used for both. And works fine.

barhanc:
> Text partitioner in this form is unacceptable. It's not prioritizing low latency (time to first audio chunk) as the old one does. And it will hurt the performance (particularly for Kokoro) very seriously.
> 
> I would just add a mechanic (like a flag or smth) that if set, makes the algorithm prioritize splitting heavily at the beginning and slowly disable it as we process more and more of text.

I've added the configurable mechanism to prioritize splitting at the beginning and also made other options of the partitioner configurable.

msluszniak:
@barhanc Would we share partitioner part across supertonic and kokoro?

barhanc:
Yes, right now it is very similar to the main branch one, and configurable options make it adjustable for different models.

msluszniak:
Thank you @IgorSwat for this on point observation. Please check if now everything is fine. Btw, maybe you want to add Kokoro to the new flow as a flow-up? ;)

IgorSwat:
`Thank you @IgorSwat for this on point observation. Please check if now everything is fine. Btw, maybe you want to add Kokoro to the new flow as a flow-up? ;)`

I will adjust my Kokoro PR after this one is merged.

IgorSwat (approved):

Issue #1323, opened 2026-07-20
[RNE Rewrite] Proper way to handle dynamic input shapes


barhanc:
cc: @msluszniak @benITo47 

This issue starts a discussion about the proper way we should handle dynamic input shapes. The current solution is a temporary one that fills the missing ExecuTorch runtime feature with adhoc solution, but seeing the changes required for OCR (#1322) makes it clear to me that we need a much more robust solution. I quickly came up with the following idea and would like your input on this:

1. The `SymbolicShape` utility should be mirrored in both C++ and TS.
2. On both sides the validation should allow for checking:
   - static vs range dim (both symbolic and concrete, so validating e.g. shape `{1, "H", RangeDim{.min=0, .max="W", .step=4}}` (C++ does it but only for concrete RangeDim, TS doesn't have RangeDims)
   - validating an enumeration of shapes, so in C++ it should be possible to write `fromJs(..., {{1, "H", 64}, {1, "H", 128}}` (TS has this)
   - (optionally) it should also allow cartesian product of dims as a shortcut to explicit enumeration, so e.g. `{1, {64,128}, {64, 128}}` <=> `{{1, 64, 64}, {1, 64, 128}, ...}` but this might introduce very severe difficulties for companion method export and symbolic -> concrete shape decay, so I wouldn't actually implement that
3. We need a differentiation between `SymbolicShape` and `ConcreteShape` (in both TS and C++).
    ```ts
    type Symbol = number | string;
    type SymbolicRangeDim = readonly {min: Symbol, max: Symbol, step?: Symbol}
    type SymbolicShape = readonly (Symbol | SymbolicRangeDim)[]

    type RangeDim = readonly {min: number, max: number, step?: number}
    type Shape = readonly (number | RangeDim)[]
    ```
    and similar distinction on C++ side.
    - C++ `fromJs` would take an enumeration of `SymbolicShape` or `ConcreteShape` to validate the input tensor.
    - C++  in model.cpp the `dynamicInputShapes` would contain only `ConcreteShape` for each dynamic method (or even perhaps we could build map `inputShapes` for all methods (not only dynamic) to unify the C++ API)
    - `getMethodMeta` would return `ConcreteShape` for each method input tensor meta (I don't know what about the output tensor meta in the case of dynamic outputs)
[...]

PR #1324, opened 2026-07-21
@bh/model-validation

barhanc:
### Summary Proof of Concept

Introduces a typed dimension model (`Static` / `Dynamic` / `Constant`) on the JS side and a companion-method parser on the C++ side so models can declare **runtime-enforced** dynamic shape constraints (min/max/step ranges *or* enumerated shape lists) that ExecuTorch's `.pte` metadata does not otherwise expose.

### Changes
- **`modelSchema.ts`** — Replaces loose `number | string` shape dims with a tagged union (`StaticDim | DynamicDim | ConstantDim`) and `Static()` / `Dynamic()` / `Constant()` constructors. `SymbolicTensor` normalizes plain numbers/strings to the typed form internally, so existing call sites keep working. `Dynamic('L')` additionally asserts the `.pte` exports a matching companion method.
- **`model_companion_parser.{h,cpp}`** — Extracts `parseDynamicInputShapes` (formerly inlined in `model.cpp`) and adds `parseEnumeratedInputShapes` for the `get_enumerated_dims_<method>` variant. Same tensor-shape contract (`[rank, 3]` int32 for ranges; `[num_shapes, rank]` int32 for enums) — kept identical to avoid a breaking change to the Python export convention.
- **`tensor_helpers.h`** — Introduces `ShapeConstraint = variant<SymbolicShape, vector<SymbolicShape>>` and a routing `fromJs` overload so a single constraints map can hold either range or enum shapes without leaking the distinction into `Model`.
- **`model.{h,cpp}`** — Renames `dynamicInputShapes_` → `inputShapeConstraints_`; rejects methods that define both companion flavors as mutually exclusive.
- **Extensions** — `textEmbedding` and `fsmnVoiceActivityDetection` adopt `Dynamic(...)` to declare the symbolic dims that already required companion methods in practice.

### Design rationale
- **Tagged union over plain primitives** on the JS side makes the *intent* of a dimension (named wildcard vs. companion-method-backed dynamic vs. literal) plain at the call site, while `toTypedShape` keeps the public API source-compatible.
- **Companion-method approach (vs. parsing `.pte` metadata)** is the only way to recover min/max/step ranges today — ExecuTorch serializes only the static upper bound. Reusing it for enumerated shapes keeps a single extension mechanism.
- **`ShapeConstraint` variant + routing overload** centralizes range/enum dispatch in `tensor_helpers` rather than `model.cpp`, so `Model` stays agnostic of the constraint flavor and future constraint types stay pluggable.
- **Mutual-exclusion guard** prevents ambiguous exports (both `get_dynamic_dims_*` and `get_enumerated_dims_*`) at load time rather than at validation time.

### Load-time (JS) vs. runtime (C++)

This split is intentional because the `.pte` metadata only serializes the **static upper bounds** of dynamic dimensions — the actual `[min, max, step]` ranges and enumerated shape lists are only retrievable by executing the companion method, and the C++ layer has to run them anyway.

- **Load-time (`validateModelSchema`):** pure metadata check run once when a schema is bound to a model in `loadModel`. Verifies tag counts, dtype, static shape signature, and — new in this branch — asserts that any input shape using `Dynamic(...)` has a corresponding companion method present in `model.getMethodNames()`. Surfaces misconfigured schemas (`Dynamic` without a companion, or non-matching shapes) as a hard error **before** any forward call instead of silently degrading to static-upper-bound-only behavior.
- **Runtime (`Model` ctor + per-call `tensor::fromJs`):** the `Model` constructor executes each companion method **once**, caches the parsed `ShapeConstraint`s into `inputShapeConstraints_`, and rejects methods that declare both `get_dynamic_dims_*` and `get_enumerated_dims_*`. On every `forward`, `fromJs` resolves the per-input constraint and validates the **concrete** incoming tensor against the actual dynamic range or enumerated list — this is where the real numeric bounds get enforced, not in JS.

### Why validation isn't duplicated at load-time

The companion methods *could* be executed from JS at validate-time (the module is loaded by then, and the cost is negligible), but doing so duplicates work the runtime already does:

- **JS validation is declarative.** `validateModelSchema` is intentionally a pure function of `(modelMetadata, schema)`. `Dynamic(...)` is a *promise* the schema makes about the model export ("this dim is genuinely dynamic and a companion exists"), not a specification of numeric values the schema enforces. The schema layer only needs to confirm the promise is honored — which a `methodNames.includes(...)` check already does — and let the runtime enforce the actual numbers.
- **No duplicated work.** C++ `Model` must parse the companion output into `inputShapeConstraints_` regardless, since `fromJs` enforces per-input bounds on every forward. Executing and parsing the same companions again from JS would establish a second representation of the same constraint with no mechanism to keep them in sync across rebinds/reloads.
- **Clear runtime errors.** C++ runtime validation already produces actionable error messages (range violations, dtype mismatch, shape mismatch) before the forward hits ExecuTorch. There is no debugging-information gap that running the companions at validate-time would close.

### Known limitation: cross-tensor constraints

Per-tensor runtime validation is *not* complete — `fromJs` validates each input tensor against its own `ShapeConstraint` independently and cannot enforce consistency of a shared symbolic dim across multiple inputs. Concretely, passing `tokens` and `mask` of mismatched `L` to `textEmbedding` bypasses our validators and surfaces as an opaque ExecuTorch internal error deep in the forward.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

- [ ] Build and run all example apps to check that there are no regressions.

### Related issues

#1323 

### Checklist

### DIFF
--- a/packages/react-native-executorch/cpp/core/model_companion_parser.cpp
+++ b/packages/react-native-executorch/cpp/core/model_companion_parser.cpp
@@ -1,0 +1,207 @@
+#include "model_companion_parser.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <format>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include <executorch/runtime/core/tag.h>
+
+namespace {
+
+template <typename T>
+T unwrap(const std::string &ctx, executorch::runtime::Result<T> result) {
+    if (!result.ok()) {
+        throw std::runtime_error(
+            std::format("{}: {}", ctx, executorch::runtime::to_string(result.error())));
+    }
+    return std::move(result.get());
+}
+
+} // namespace
+
+namespace rnexecutorch::core::model {
+
+std::optional<std::vector<tensor::ShapeConstraint>>
+parseDynamicInputShapes(executorch::extension::Module &module, const std::string &methodName) {
+    using executorch::aten::ScalarType;
+
+    const auto getDynamicShapesMethodName = std::format("get_dynamic_dims_{}", methodName);
+    const auto ctx = getDynamicShapesMethodName + ": ";
+
+    auto methodNames = unwrap(ctx + "failed to get method names", module.method_names());
+    if (methodName == getDynamicShapesMethodName ||
+        !methodNames.contains(getDynamicShapesMethodName)) {
+        return std::nullopt;
+    }
+
+    auto methodMeta = unwrap(std::format("{}failed to get meta for method '{}'", ctx, methodName),
+                             module.method_meta(methodName));
+
+    size_t expectedTensorInputs = 0;
+    for (size_t i = 0; i < methodMeta.num_inputs(); ++i) {
+        auto tag = unwrap(std::format("{}failed to get tag for input [{}]", ctx, i), methodMeta.input_tag(i));
+        if (tag == executorch::runtime::Tag::Tensor) {
+            expectedTensorInputs++;
+        }
+    }
+
+    auto result = unwrap(ctx + "failed to execute", module.execute(getDynamicShapesMethodName));
+    if (result.size() != expectedTensorInputs) {
+        throw std::runtime_error(std::format("{}number of outputs returned ({}) does not match the number of "
+                                             "tensor inputs declared by method '{}' ({})",
+                                             ctx, result.size(), methodName, expectedTensorInputs));
+    }
+
+    std::vector<tensor::ShapeConstraint> dynamicShapes;
+    dynamicShapes.reserve(methodMeta.num_inputs());
+    size_t tensorIndex = 0;
+
+    for (size_t i = 0; i < methodMeta.num_inputs(); ++i) {
+        auto tag = unwrap(std::format("{}failed to get tag for input [{}]", ctx, i),
+                          methodMeta.input_tag(i));
+
+        if (tag != executorch::runtime::Tag::Tensor) {
+            dynamicShapes.emplace_back();
+            continue;
+        }
+
+        const auto &out = result.at(tensorIndex);
+        if (!out.isTensor()) {
+            throw std::runtime_error(std::format("{}output[{}] is not a tensor", ctx, tensorIndex));
+        }
+
+        auto inputMeta = unwrap(std::format("{}failed to get tensor meta for input [{}]", ctx, i),
+                                methodMeta.input_tensor_meta(i));
+        const auto rank = inputMeta.sizes().size();
+        const auto shapeTensor = out.toTensor();
+
+        if (shapeTensor.dim() != 2 || shapeTensor.size(1) != 3 ||
+            shapeTensor.size(0) != static_cast<ssize_t>(rank) ||
+            shapeTensor.scalar_type() != ScalarType::Int) {
+            throw std::runtime_error(std::format("{}output[{}] expected to be a 2D int32_t tensor of shape [{}, 3]",
+                                                 ctx, tensorIndex, rank));
+        }
+
+        const auto *shape = shapeTensor.const_data_ptr<int32_t>();
+        tensor::SymbolicShape symbolicShape;
+        symbolicShape.reserve(rank);
+
+        for (size_t axis = 0; axis < rank; ++axis) {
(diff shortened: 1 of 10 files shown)


msluszniak:
I guess that cross-tensor validation is important and ultimately we want it, but for now it's ok to skip it. Just make sure we have some follow-up issue for this kind of check.

packages/react-native-executorch/cpp/core/model_companion_parser.cpp
```diff
@@ -0,0 +1,207 @@
...
+            if (minDim < 0 || maxDim < minDim || step < 1) {
+                throw std::runtime_error(std::format("{}output[{}], axis {} is invalid: "
+                                                     "expected 0 <= min <= max and step >= 1 but got [{}, {}, {}]",
+                                                     ctx, tensorIndex, axis, minDim, maxDim, step));
+            }
+            if (maxDim > inputMeta.sizes()[axis]) {
+                throw std::runtime_error(std::format("{}output[{}], axis {} max dimension ({}) "
+                                                     "exceeds model metadata upper limit ({})",
+                                                     ctx, tensorIndex, axis, maxDim, inputMeta.sizes()[axis]));
+            }
```
msluszniak, line 106: How about the analogical check for minDim?

packages/react-native-executorch/cpp/core/model_companion_parser.cpp
```diff
@@ -0,0 +1,207 @@
...
+            expectedTensorInputs++;
+        }
+    }
+
+    auto result = unwrap(ctx + "failed to execute", module.execute(getEnumeratedShapesMethodName));
+    if (result.size() != expectedTensorInputs) {
+        throw std::runtime_error(std::format("{}number of outputs returned ({}) does not match the number of "
+                                             "tensor inputs declared by method '{}' ({})",
+                                             ctx, result.size(), methodName, expectedTensorInputs));
+    }
```
msluszniak, line 148: Maybe we should abstract common part of these two into function in anonymous helper?

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -64,6 +64,45 @@ std::string shapeToString(const SymbolicShape &shape) {
...
+    if (shape.size() != expectedShape.size()) {
+        return std::format("expected {} dimensions, got {}", expectedShape.size(), shape.size());
+    }
+
+    std::unordered_map<std::string, int32_t> symbolToConcrete;
+
+    for (size_t i = 0; i < expectedShape.size(); ++i) {
+        const auto &dim = expectedShape.at(i);
+
+        if (std::holds_alternative<int32_t>(dim)) {
```
msluszniak, line 78: Maybe use std::visit here instead of if else checks?

packages/react-native-executorch/cpp/core/tensor_helpers.h
```diff
@@ -116,7 +129,23 @@ fromJs(jsi::Runtime &rt, const std::string &name, const jsi::Value &value,
...
+/**
+ * @overload
+ *
+ * Convenience wrapper that accepts a variant ShapeConstraint (either a single
+ * shape or a list of alternative shapes) and routes it to the correct overload.
+ */
+inline std::shared_ptr<TensorHostObject>
+fromJs(jsi::Runtime &rt, const std::string &name, const jsi::Value &value,
+       std::optional<DType> expectedDtype, const ShapeConstraint &shapeConstraint) {
+    if (std::holds_alternative<SymbolicShape>(shapeConstraint)) {
```
msluszniak, line 145: Consider std::visit here, definitely not as useful as in tge previous case

msluszniak (commented):
Only cpp side for now

barhanc:
@msluszniak I'm going to close this PR because I came to a conclusion (based on exhaustive experiments with different API designs) that this is not an ideal way of solving the validation problem. I think we should actually export static method in .pte files which would return JSON string with full info about model schema. I would like to discuss the details whenever you're free.

msluszniak:
Let's discuss it today! :))

PR #1325, opened 2026-07-21
fix(android): forward ExecuTorch consumer ProGuard rules

injunchoi98:
## Description

Fixes an Android release crash caused by R8 removing ExecuTorch classes that are accessed during JNI initialization.

at minified release builds without the consumer rules, the app crashes during `ETInstaller` initialization:

```text
java.lang.ClassNotFoundException: org.pytorch.executorch.Module
at java.lang.System.loadLibrary(...)
at com.swmansion.rnexecutorch.ETInstaller.<init>(...)
```
This PR ports the upstream rules and registers them through `consumerProguardFiles`

The first two rules ared covered with RN's own consumer rules, but are intentionally retained to keep the upstream ExecuTorch rule block intact. The rule that fixes the observed `Module` crash is:

```proguard
-keepclasseswithmembers class org.pytorch.executorch.** {
    native <methods>;
}
```

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. Enable R8/ProGuard for the release build.
2. Build the minified release APK
3. Verify that JNI-accessed classes retain their names in mapping.txt

### Checklist

### Additional notes

Android-only; iOS does not use R8. tested on a android physical device(galaxy s20 FE)

The upstream ExecuTorch rules are copied without semantic changes.

### DIFF
--- a/packages/react-native-executorch/android/consumer-proguard-rules.pro
+++ b/packages/react-native-executorch/android/consumer-proguard-rules.pro
@@ -1,0 +1,51 @@
+# Ported from ExecuTorch's official Android consumer rules:
+# https://github.com/pytorch/executorch/blob/904c667007aed15e9a3bb8278aa271b479450c85/extension/android/executorch_android/consumer-proguard-rules.pro
+
+# ExecuTorch Android AAR — Consumer ProGuard/R8 Rules
+#
+# These rules are automatically applied to any app that depends on the
+# ExecuTorch AAR. They prevent R8/ProGuard from stripping classes and
+# methods that are called from native (JNI) code.
+
+# Keep ExecuTorch classes and members annotated with @DoNotStrip.
+# Scoped to org.pytorch.executorch to avoid affecting unrelated libraries.
+-keep @com.facebook.jni.annotations.DoNotStrip class org.pytorch.executorch.** { *; }
+-keepclassmembers class org.pytorch.executorch.** {
+    @com.facebook.jni.annotations.DoNotStrip *;
+}
+
+# Keep all native methods across ExecuTorch packages.
+# Use -keepclasseswithmembers (not -keepclasseswithmembernames) to prevent
+# both shrinking and obfuscation of JNI entry points.
+-keepclasseswithmembers class org.pytorch.executorch.** {
+    native <methods>;
+}
+
+# Keep HybridData fields (accessed by fbjni via reflection).
+-keepclassmembers class org.pytorch.executorch.** {
+    com.facebook.jni.HybridData *;
+}
+
+# Keep ExecutorchRuntimeException and its factory/subclasses.
+# These are instantiated from JNI via jni_helper.cpp.
+-keep class org.pytorch.executorch.ExecutorchRuntimeException { *; }
+-keep class org.pytorch.executorch.ExecutorchRuntimeException$* { *; }
+
+# Keep EValue (fields and type codes accessed from native code).
+-keep class org.pytorch.executorch.EValue { *; }
+
+# Keep Tensor and its inner classes. The Tensor class has methods and fields
+# accessed from JNI (dtypeJniCode, shape, getRawDataBuffer, nativeNewTensor).
+-keep class org.pytorch.executorch.Tensor { *; }
+-keep class org.pytorch.executorch.Tensor$* { *; }
+
+# Keep LlmCallback interface methods (invoked from native code).
+-keep interface org.pytorch.executorch.extension.llm.LlmCallback { *; }
+
+# Keep AsrCallback interface methods (invoked from native code).
+-keep interface org.pytorch.executorch.extension.asr.AsrCallback { *; }
+
+# --- react-native-executorch rules (not from upstream) ---
+
+# RNE's own JNI entry point, resolved by literal descriptor from C++.
+-keep class com.swmansion.rnexecutorch.ETInstaller { *; }
--- a/packages/react-native-executorch/android/build.gradle
+++ b/packages/react-native-executorch/android/build.gradle
@@ -115,6 +115,7 @@ android {
   defaultConfig {
     minSdkVersion getExtOrIntegerDefault("minSdkVersion")
     targetSdkVersion getExtOrIntegerDefault("targetSdkVersion")
+    consumerProguardFiles "consumer-proguard-rules.pro"
     externalNativeBuild {
         cmake {
           abiFilters (*reactNativeArchitectures())


packages/react-native-executorch/android/consumer-proguard-rules.pro
```diff
@@ -0,0 +1,46 @@
+# Ported from ExecuTorch's official Android consumer rules:
+# https://github.com/pytorch/executorch/blob/main/extension/android/executorch_android/consumer-proguard-rules.pro
```
msluszniak: Please pin the upstream commit SHA (or the ExecuTorch version `libs/classes.jar` was built from) instead of `/main/` — gives the next jar bump a clear re-sync point.

packages/react-native-executorch/android/consumer-proguard-rules.pro
```diff
@@ -0,0 +1,46 @@
...
+# These rules are automatically applied to any app that depends on the
+# ExecuTorch AAR. They prevent R8/ProGuard from stripping classes and
+# methods that are called from native (JNI) code.
+
+# Keep ExecuTorch classes and members annotated with @DoNotStrip.
+# Scoped to org.pytorch.executorch to avoid affecting unrelated libraries.
+-keep @com.facebook.jni.annotations.DoNotStrip class org.pytorch.executorch.** { *; }
+-keepclassmembers class org.pytorch.executorch.** {
+    @com.facebook.jni.annotations.DoNotStrip *;
+}
```
msluszniak, line 15: FYI, these two are already covered globally by RN's own consumer rules (`ReactAndroid/proguard-rules.pro:35-37` keeps `@com.facebook.jni.annotations.DoNotStrip` classes and members for `class *`). Harmless to keep for standalone correctness, just not where the fix comes from.

The rule that actually earns its keep is the next one: RN line 45 is `-keepclassmembers,includedescriptorclasses class * { native <methods>; }`, which preserves native *members* but still lets R8 rename the class. `-keepclasseswithmembers` pins the class name for JNI `FindClass`. Worth noting in the PR description.

packages/react-native-executorch/android/consumer-proguard-rules.pro
```diff
@@ -0,0 +1,46 @@
...
+# Keep Tensor and its inner classes. The Tensor class has methods and fields
+# accessed from JNI (dtypeJniCode, shape, getRawDataBuffer, nativeNewTensor).
+-keep class org.pytorch.executorch.Tensor { *; }
+-keep class org.pytorch.executorch.Tensor$* { *; }
+
+# Keep LlmCallback interface methods (invoked from native code).
+-keep interface org.pytorch.executorch.extension.llm.LlmCallback { *; }
+
+# Keep AsrCallback interface methods (invoked from native code).
+-keep interface org.pytorch.executorch.extension.asr.AsrCallback { *; }
```
msluszniak, line 46: Missing RNE's own JNI entry point. `ETInstallerModule.cpp:50-52` does `javaClassStatic()` + `GetStaticMethodID(cls, "fetchByteDataFromUrl", ...)` against the literal descriptor in `ETInstallerModule.h:18`. RN's rules keep the annotated method name but allow the class to be renamed/repackaged, so this still breaks under R8.

```proguard
# --- react-native-executorch rules (not from upstream) ---

# RNE's own JNI entry point — resolved by literal descriptor from ETInstallerModule.cpp.
-keep class com.swmansion.rnexecutorch.ETInstaller { *; }
```

Keeping it under a separator preserves the upstream block as a clean re-syncable copy.

msluszniak (commented):
Thanks for contributing! The PR looks solid.

One blocking gap: the rules are scoped to `org.pytorch.executorch.**`, but RNE resolves its *own* class by literal descriptor from C++ (`ETInstallerModule.h:18` → `Lcom/swmansion/rnexecutorch/ETInstaller;`). `ETInstaller` has `@DoNotStrip` on the method, not the class, so R8 can still rename the class — which is the failure shape your stack trace shows (`ETInstaller.<init>`). See the inline comment.

Also: could you paste the real logcat? `System.loadLibrary` doesn't throw `ClassNotFoundException`, and `ETInstaller.init` catches only `UnsatisfiedLinkError`, so the trace in the description doesn't match the diagnosis. Helps confirm which rule is load-bearing.

packages/react-native-executorch/android/consumer-proguard-rules.pro
injunchoi98: Thanks for the clarification. I updated the PR description to explain which rules preserve the class names required for JNI lookup!

packages/react-native-executorch/android/consumer-proguard-rules.pro
injunchoi98: Thanks for pointing this out. I only checked the release build and its R8 mapping.. I’ve added the RNE-specific keep rule to the PR.

packages/react-native-executorch/android/consumer-proguard-rules.pro
injunchoi98: I hadn’t considered that. Thanks for pointing it out! I’ve updated the link to pin the upstream commit SHA as suggested.

injunchoi98:
Thank you very much for the thorough review and detailed explanations. I’ve applied the requested changes!

We’re using React Native ExecuTorch successfully in our app, so thank you as well for building and maintaining this library.

Here is the raw logcat from a minified release build without the consumer rules, reproduced on a physical Galaxy S20 FE:

```text
07-22 12:51:50.508 20405 20679 D nativeloader: Load /data/app/~~IBCYjbXes9KZ7UOlbtmEmA==/com.earlyriser42.classic-2SuLRCS9eEf6EjiAWQzJyA==/base.apk!/lib/arm64-v8a/libexecutorch.so using class loader ns clns-4 (caller=/data/app/~~IBCYjbXes9KZ7UOlbtmEmA==/com.earlyriser42.classic-2SuLRCS9eEf6EjiAWQzJyA==/base.apk!classes4.dex): ok
07-22 12:51:50.544 20405 20679 E log     : error java.lang.ClassNotFoundException: org.pytorch.executorch.Module
07-22 12:51:50.544 20405 20679 E log     :      at java.lang.Runtime.nativeLoad(Native Method)
07-22 12:51:50.544 20405 20679 E log     :      at java.lang.Runtime.loadLibrary0(Runtime.java:1109)
07-22 12:51:50.544 20405 20679 E log     :      at java.lang.Runtime.loadLibrary0(Runtime.java:1033)
07-22 12:51:50.544 20405 20679 E log     :      at java.lang.System.loadLibrary(System.java:1765)
07-22 12:51:50.544 20405 20679 E log     :      at com.swmansion.rnexecutorch.ETInstaller.<init>(SourceFile:11)

--------- beginning of crash
07-22 12:51:50.550 20405 20680 E AndroidRuntime: FATAL EXCEPTION: mqt_v_native
07-22 12:51:50.550 20405 20680 E AndroidRuntime: Process: com.earlyriser42.classic, PID: 20405
07-22 12:51:50.550 20405 20680 E AndroidRuntime: com.facebook.react.common.JavascriptException: Error: Exception in HostObject::get for prop 'ETInstaller': java.lang.ClassNotFoundException: org.pytorch.executorch.Module
```

msluszniak:
Thanks for the logcat — it makes the mechanism clear. Worth noting for the record: `System.loadLibrary` *can* surface a `ClassNotFoundException` indirectly. Loading `libexecutorch.so` runs its `JNI_OnLoad`, which does `FindClass("org/pytorch/executorch/Module")`; when R8 has renamed `Module`, that lookup fails and the exception propagates up through `nativeLoad` → `loadLibrary`. Since it's a `ClassNotFoundException` (not an `UnsatisfiedLinkError`), `ETInstaller.init`'s catch doesn't swallow it, so it reaches JS as the fatal you saw. That confirms the load-bearing rule here is the upstream `native <methods>` keep pinning `Module`'s name.

msluszniak (approved):
I tested the fix on minified build with our llm demo app, and the issue reproduced clearly. Without the fix, I got the same crash. After applying the fix, it worked correctly. Btw, can you disclose the name of the app you are working on? Our team is very curious to see how React Native ExecuTorch is used in apps :))

injunchoi98:
Thank you so much for reviewing and testing the fix! Sure thing — our app is Lita Book. We are using kokoro at book reader. it’s one of several projects I’m building with my teammates. We’re continuing to improve it and gradually grow the user base :)

msluszniak:
Thank you for your response, we'll look at Lita Book for sure :))

msluszniak:
@barhanc we should apply this one on `rne-rewrite` as well if I'm correct.

PR #1327, opened 2026-07-23
[RNE Rewrite] refactor!: add better model schema contract and validation logic

### LINKED ISSUE #1323: [RNE Rewrite] Proper way to handle dynamic input shapes

barhanc:
## Description

Intoduces a new powerful and extensible mechanism for validating model contracts both during the runtime execution (mitigating e.g. internal error crashes when two input tensors should have the same dimension during runtime) and pipeline construction (matching model contract against a specified interface). 

- Introduces consistent, clean, simple and extensible contract for custom schema specification as an optional constant companion method `get_model_schema` returning string-encoded JSON. No need for constructing auxiliary tensors for specifying dynamic dimensions, multiple companions for different kinds, etc.
- Introduces distinction between symbolic dims and concrete dims which represent dimension domains the model can require (includes constant, enumerated and range dims).
- Adds extensible mechanism of specifying runtime constraints (such as dynamic dim equality, linearity) which is distinct from dim domain constraints.
- Adds ergonomic helper functions which make the model spec specification in pipeline implementation concise yet clear.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS (Simulator only, re-exported MLX and CoreML haven't been tested)
- [x] Android

### Testing instructions

Models that use the companion method to get the JSON schema have not yet been updated on the SWM HF as they wait for the review of this PR. For testing you can use the models at https://huggingface.co/bhanc/scratch/tree/main/ptes.

- [ ] Build all example apps on both Android and iOS and check that all models still work

### Related issues

Closes #1323 

### Checklist

### Additional notes

- The updated export scripts are on the internal gitlab on the `@bh/schema-workflow` branch.
- @msluszniak I haven't been able to re-export the LFM2.5 embeddings from the `@nk/universal-exporter` branch. I would be grateful if you could take a look.
- @benITo47 Please take a look if you can find some use-case which this mechanism cannot express.


msluszniak:
@barhanc I updated the export script so know it should work correctly (at least I can export them my myself from Norbert's branch). The problem was in the newest version of transformers. Please give it a try.

.agents/skills/model-schema-validation/SKILL.md
```diff
@@ -1,162 +1,208 @@
...
-  expectedOutputs
-);
+import {
+  validateSpec,
+  method,
+  f32,
+  i64,
+  i32,
+  DynamicDim as Dyn,
+  constr,
```
msluszniak, line 33: These are not used in the code snippet, but described below. I wonder if we don't want to add example that uses them or will it be too long?

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -339,43 +118,39 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
-                                                   inputsArray.size(rt), methodMeta.num_inputs()));
+            if (inputsArray.size(rt) != methodSpec.inputs.size()) {
+                throw jsi::JSError(rt, std::format("execute: Incorrect size for inputs"));
             }
 
-            std::vector<executorch::runtime::EValue> inputs(methodMeta.num_inputs());
+            std::vector<executorch::runtime::EValue> inputs(methodSpec.inputs.size());
             std::vector<std::unique_lock<std::shared_mutex>> tensorLocks;
             std::unordered_set<TensorHostObject *> lockedTensors;
+            std::vector<std::vector<int32_t>> inputShapes;
```
msluszniak, line 138: Is there any reason why we want tensor shapes to be int32_t and not uint32_t or similar?

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+ParamSpec tensorMetaToParamSpec(const executorch::runtime::TensorInfo &tensorMeta) {
+    ParamSpec p{.tag = Tag::Tensor, .dtype = types::fromScalarType(tensorMeta.scalar_type())};
+    const auto sizes = tensorMeta.sizes();
+    p.shape.reserve(sizes.size());
+    for (const auto size : sizes) {
+        p.shape.emplace_back(size);
+    }
+    return p;
+}
```
msluszniak, line 242: type deduction for concerete dims here is unambigious and will resolve with int32_t variant.
```suggestion
 ParamSpec tensorMetaToParamSpec(const executorch::runtime::TensorInfo &tensorMeta) {
    const auto sizes = tensorMeta.sizes();
    return ParamSpec{
        .tag = Tag::Tensor,
        .dtype = types::fromScalarType(tensorMeta.scalar_type()),
        .shape = std::vector<ConcreteDim>(sizes.begin(), sizes.end()),
    };
}
```

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+namespace {
+
+template <typename T>
+T unwrap(const std::string &ctx, executorch::runtime::Result<T> result) {
+    if (!result.ok()) {
+        throw std::runtime_error(std::format("{}: {}", ctx, executorch::runtime::to_string(result.error())));
+    }
+    return std::move(result.get());
+}
```
msluszniak, line 77: We have exact duplicate of this one in `cpp/core/model.cpp:26`

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+        if (e->choices.empty()) {
+            throw std::runtime_error(std::format("{}: enum must have at least one choice", ctx));
+        }
+        for (const auto &choice : e->choices) {
+            if (choice <= 0) {
+                throw std::runtime_error(std::format("{}: enum choices must be positive", ctx));
+            }
+        }
+    }
+}
```
msluszniak, line 325: Either:
Add this:
```cpp
/// Visitor helper for std::visit: combines lambdas into one overload set.
template <class... Ts>
struct overloaded : Ts... {
    using Ts::operator()...;
};
```
to `cpp/core/utils.h`, add headers and then use the suggestion from below. This is better because if we add new constraint it will automatically fail to compile, and it removes necessity of checking each branch as we do now using three sequential ifs.
```suggestion
void validateConcreteDim(const ConcreteDim &dim, const std::string &ctx) {
    std::visit(overloaded{
        [&ctx](int32_t c) {
            if (c <= 0) {
                throw std::runtime_error(std::format("{}: constant dim must be positive", ctx));
            }
        },
        [&ctx](const RangeDim &r) {
            if (r.min < 0) {
                throw std::runtime_error(std::format("{}: range min must be non-negative", ctx));
            }
            if (r.max < r.min) {
                throw std::runtime_error(std::format("{}: range max must be >= min", ctx));
            }
            if (r.step <= 0) {
                throw std::runtime_error(std::format("{}: range step must be positive", ctx));
            }
        },
        [&ctx](const EnumDim &e) {
            if (e.choices.empty()) {
                throw std::runtime_error(std::format("{}: enum must have at least one choice", ctx));
            }
            for (const auto choice : e.choices) {
                if (choice <= 0) {
                    throw std::runtime_error(std::format("{}: enum choices must be positive", ctx));
                }
            }
        },
    }, dim);
}
```
Or: Use `else if` instead of plain three `if`.

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+    auto metaShape = tensorMeta.sizes();
+    if (param.shape.size() != metaShape.size()) {
+        throw std::runtime_error(std::format("{}: rank mismatch", ctx));
+    }
+
+    for (size_t d = 0; d < param.shape.size(); ++d) {
+        if (std::holds_alternative<int32_t>(param.shape[d]) &&
+            std::get<int32_t>(param.shape[d]) != metaShape[d]) {
+            throw std::runtime_error(std::format("{}: shape[{}] mismatch", ctx, d));
+        }
```
msluszniak: Above you used different way to handle the variant access:

```cpp
    if (const auto *c = std::get_if<int32_t>(&dim)) {
        if (*c <= 0) {
            throw std::runtime_error(std::format("{}: constant dim must be positive", ctx));
        }
    }
``` 

so decide on one style.

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+            }
+            for (const auto &dim : eq->dims) {
+                validateDimRef(dim, inputRanks, outputRanks, cctx);
+            }
+        }
+
+        if (const auto *lin = std::get_if<LinearConstraint>(&constraint)) {
+            validateDimRef(lin->dimLhs, inputRanks, outputRanks, cctx);
+            validateDimRef(lin->dimRhs, inputRanks, outputRanks, cctx);
+        }
```
msluszniak: same as above, either std::visit with overloaded or else if

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+                lin->dimRhs.paramSide == ParamSide::output) {
+                continue;
+            }
+            int32_t lhs = getInputDimValue(lin->dimLhs, inputShapes);
+            int32_t rhs = getInputDimValue(lin->dimRhs, inputShapes);
+            if (lhs != lin->coefficients[0] * rhs + lin->coefficients[1]) {
+                throw jsi::JSError(rt, std::format("{}: linear constraint violated", cctx));
+            }
+        }
+    }
```
msluszniak, line 532: again, std::visit with overload or else if

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -1,86 +1,99 @@
...
+            s += *str;
         }
-        s += ", ";
+        if (const auto *val = std::get_if<int32_t>(&dim)) {
+            s += std::to_string(*val);
+        }
+        if (const auto *range = std::get_if<schema::RangeDim>(&dim)) {
+            s += std::format("[{}..{}:{}]", range->min, range->max, range->step);
+        }
+        if (const auto *enumeration = std::get_if<schema::EnumDim>(&dim)) {
```
msluszniak: again, std::visit with overloaded or else ifs

packages/react-native-executorch/cpp/core/tensor_helpers.cpp
```diff
@@ -89,40 +102,45 @@ fromJs(jsi::Runtime &rt, const std::string &name, const jsi::Value &value,
...
+                                                   ctx, shapeToString(*expectedShape), i, shape[i], range->max));
             }
-            if (rangeDim.step && (shape[i] - rangeDim.min) % *rangeDim.step != 0) {
+            if ((shape[i] - range->min) % range->step != 0) {
                 throw jsi::JSError(rt, std::format("{} must have shape {} (dim {} must be min({}) + k*step({}), got {})",
-                                                   name, shapeToString(*expectedShape), i, rangeDim.min, *rangeDim.step, shape[i]));
+                                                   ctx, shapeToString(*expectedShape), i, range->min, range->step, shape[i]));
+            }
+        }
+        if (const auto *enumeration = std::get_if<schema::EnumDim>(&dim)) {
```
msluszniak: std::visit and overloaded or else if

packages/react-native-executorch/src/core/schema.ts
```diff
@@ -0,0 +1,813 @@
...
+
+function rangesEqual(r1: Range, r2: Range): boolean {
+  return r1.min === r2.min && r1.max === r2.max && r1.step === r2.step;
+}
+
+function choicesEqual(c1: readonly number[], c2: readonly number[]): boolean {
+  const s1 = new Set(c1);
+  const s2 = new Set(c2);
+  return s1.size === s2.size && [...s1].every((elem) => s2.has(elem));
+}
```
msluszniak, line 373: Why do we want do remove duplications here by using sets?
```suggestion
function choicesEqual(c1: readonly number[], c2: readonly number[]): boolean {
  const s1 = new Set(c1);
  const s2 = new Set(c2);
  return s1.size === s2.size && c1.every((elem) => s2.has(elem));
}
```

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+        throw std::runtime_error(std::format("{}: dtype mismatch", ctx));
+    }
+
+    auto metaShape = tensorMeta.sizes();
+    if (param.shape.size() != metaShape.size()) {
+        throw std::runtime_error(std::format("{}: rank mismatch", ctx));
+    }
+
+    for (size_t d = 0; d < param.shape.size(); ++d) {
+        if (std::holds_alternative<int32_t>(param.shape[d]) &&
```
msluszniak: `validateTensorParam` only checks `constant` dims against `MethodMeta`; `range`/`enum` dims aren't bounded by the static upper bound. The old `parseDynamicInputShapes` guarded this (`maxDim > inputMeta.sizes()[axis]`). A wrong/stale `get_model_schema` (e.g. `range` max 1024 vs. compiled 512) now passes load-time validation and crashes opaquely at `execute`. Re-add the "dynamic max ≤ MethodMeta bound" check.

packages/react-native-executorch/src/core/schema.ts
```diff
@@ -0,0 +1,813 @@
...
+  if (min < 0 || !Number.isInteger(min)) {
+    throw new Error(`Invalid range min (${min}): must be a non-negative integer.`);
+  }
+  if (max < min) {
+    throw new Error(`Invalid range [${min}, ${max}]: max cannot be less than min.`);
+  }
+  if (!Number.isInteger(max)) {
+    throw new Error(`Invalid range max (${max}): must be a non-negative integer.`);
+  }
+  if (step && (step <= 0 || !Number.isInteger(step))) {
```
msluszniak: `if (step && ...)` skips the guard when `step === 0`, so `RangeDim(1, 10, 0)` yields `step: 0` (caught only later in `validateDimDomains`). Use `if (step !== undefined && (step <= 0 || !Number.isInteger(step)))`.

packages/react-native-executorch/cpp/core/schema.cpp
```diff
@@ -0,0 +1,491 @@
...
+    }
+    if (const auto *lin = std::get_if<LinearConstraint>(&c)) {
+        j = json::object({{"kind", "linear"},
+                          {"dimLhs", lin->dimLhs},
+                          {"dimRhs", lin->dimRhs},
+                          {"coefficients", lin->coefficients}});
+    }
+}
+
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(MethodSpec, inputs, outputs, runtimeConstraints)
```
msluszniak, line 160: This macro parses with `at()`, so every method object in the companion JSON must physically include `runtimeConstraints` (plus `inputs`/`outputs`); omitting it on a constraint-free method throws an opaque `key 'runtimeConstraints' not found` at load. How is this handled on the export side — do the `@bh/schema-workflow` scripts always emit `"runtimeConstraints": []`? If we want to tolerate omission, `NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT` would default it to `[]` (note: that also makes `inputs`/`outputs` optional, deferring the "missing inputs" error to `validateSpec`).

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -339,43 +118,39 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
+            const auto &methodSpec = self->spec_.at(methodName);
 
             auto inputsArray = conversions::asType<jsi::Array>(rt, "execute: inputs", args[1]);
             auto outputTensorsArray = conversions::asType<jsi::Array>(rt, "execute: outputTensors", args[2]);
 
-            if (inputsArray.size(rt) != methodMeta.num_inputs()) {
-                throw jsi::JSError(rt, std::format("execute: Incorrect size for inputs: got {}, expected {}",
-                                                   inputsArray.size(rt), methodMeta.num_inputs()));
+            if (inputsArray.size(rt) != methodSpec.inputs.size()) {
+                throw jsi::JSError(rt, std::format("execute: Incorrect size for inputs"));
```
msluszniak: `std::format` with no args here, and the message dropped the `got {}, expected {}` counts the old code reported. Restore the detail, e.g. `std::format("execute: Incorrect size for inputs: got {}, expected {}", inputsArray.size(rt), methodSpec.inputs.size())`.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
```diff
@@ -166,32 +166,28 @@ export async function createWhisperSpeechToText<L extends WhisperLanguage = Whis
...
-      SymbolicTensor('int64', ['Tokens']),
-      SymbolicTensor('float32', [1, encSeqLen, encStateDim]),
-    ],
-    [SymbolicTensor('float32', [1, 'Tokens', 'Vocab'])]
-  );
+  const { dims } = validateSpec(model.schema, {
+    default: {
+      ...method(
+        'encode', // prettier-ignore
+        [f32(Dyn('T_audio'))],
```
msluszniak: Under then hood, the time audio that is consumed in whisper is fixed 30 sec and model itself pads it to 30 sec, so for end user it is dynamic but in fact the consumed input isn't. Don't we want to add comment here?

packages/react-native-executorch/cpp/core/schema.cpp
barhanc: Fixed.

packages/react-native-executorch/cpp/core/model.cpp
barhanc: This is because ExecuTorch uses `int32_t` to represent dimensions, so we want to match that.

packages/react-native-executorch/cpp/core/schema.cpp
barhanc: Fixed to use `std::visit` pattern everywhere.

packages/react-native-executorch/src/core/schema.ts
barhanc: Choices is basically a set but we use array for it to be easily compatible with JSON, therefore for equality checking we remove duplicates.

packages/react-native-executorch/cpp/core/schema.cpp
barhanc: Added upper bound checks for both range.max and enum choices.

packages/react-native-executorch/cpp/core/model.cpp
barhanc: Restored the previous detailed error message.

packages/react-native-executorch/cpp/core/schema.cpp
barhanc: The model's schema.json should basically be a 1-to-1 representation of the `ModelSpec<ConcreteDim>` type which doesn't have optional properties, so we always explicitly set `"runtimeConstraints": []` when there are no constraints.

msluszniak: Ok, understood.

packages/react-native-executorch/src/extensions/speech/tasks/whisperSpeechToText.ts
barhanc: That is a model export detail but I guess we can add a comment so that there is no surprise why the encode takes the same amount of time regardless of input size. Added short inline comment there.

packages/react-native-executorch/src/core/schema.ts
barhanc: Fixed.

packages/react-native-executorch/cpp/core/schema.cpp
barhanc: I basically treat the `unwrap`s as quick, dirty ad hoc helpers for getting the actual value from ExecuTorch's `Result` type. We could move all these `unwrap`s to a single location but I'm not sure if it's worth it (similar case as with the `overloaded` helper for `std::visit`). What do you think?

msluszniak: Yeah, I'm okay with that, that was more an observation rather than some concrete suggestion ;)

.agents/skills/model-schema-validation/SKILL.md
barhanc: Changed the example snippet to include dynamic dim and constraint specification.

msluszniak:
For text embeddinga, I got two errors: 
Screenshot_20260730_123847_nlp
Screenshot_20260730_123834_nlp

The rest of the text embedding models worked correctly.

msluszniak:
Regarding vad and whisper, I only had problems with coreml variants of whisper. For all that I was able to run on Iphone SE, I get the following error:
image

barhanc:
The models should be fixed now. I've updated the `.pte` files on the https://huggingface.co/bhanc/scratch/tree/main repo for now. After you test that everything is fine I will upload them to the SWM HF under `v0.10.0` tag.

msluszniak (approved):
Now everything worked correctly, @benITo47 are you fine with these changes?

benITo47:
@msluszniak


From what I've been reading, changes seem sane. One real gap for OCR: enumerated shapes are a per-dimension cross product. tensor_helpers.cpp fromJs validates each dim against its own EnumDim.choices independently, and dims.enum only exposes per-dim arrays. There's no whole-shape domain, and runtimeConstraints (eq/linear) are dim-relations, so they can't carve out a sparse set either.

For example, our CoreML detectors ship sets like {640², 960², 1280², 1280×640} for PaddleOCR. To describe that in the new schema I can only give H:{640,960,1280}, W:{640,960,1280} - whose cross product also admits 640×1280, which the PTE never compiled. fromJs green-lights it, then CoreML rejects it at execute. 

While resizing to match input, im picking which shape to use (snap up or snap down), in that process I can actively select that illegal shape. 

To make the per-dim schema match the compiled model I'd have to either compile the full cross product (more ANE specializations, more coreML overhead, worse performance) or drop to square-only (lose the portrait bucket). The workaround is to hardcode the whole-shape list in the task and snap to it - but then the model doesn't self-describe its contract, code breaks when hardcoded list drifts from the export, and inspectModel shows a wrong over-permissive shape set. So at minimum this should be expressible in schema.ts - a whole-shape / tuple-enum domain (or constraint kind), whatever fits the model best.

barhanc:
> @msluszniak
> 
> From what I've been reading, changes seem sane. One real gap for OCR: enumerated shapes are a per-dimension cross product. tensor_helpers.cpp fromJs validates each dim against its own EnumDim.choices independently, and dims.enum only exposes per-dim arrays. There's no whole-shape domain, and runtimeConstraints (eq/linear) are dim-relations, so they can't carve out a sparse set either.
> 
> For example, our CoreML detectors ship sets like {640², 960², 1280², 1280×640} for PaddleOCR. To describe that in the new schema I can only give H:{640,960,1280}, W:{640,960,1280} - whose cross product also admits 640×1280, which the PTE never compiled. fromJs green-lights it, then CoreML rejects it at execute.
> 
> While resizing to match input, im picking which shape to use (snap up or snap down), in that process I can actively select that illegal shape.
> 
> To make the per-dim schema match the compiled model I'd have to either compile the full cross product (more ANE specializations, more coreML overhead, worse performance) or drop to square-only (lose the portrait bucket). The workaround is to hardcode the whole-shape list in the task and snap to it - but then the model doesn't self-describe its contract, code breaks when hardcoded list drifts from the export, and inspectModel shows a wrong over-permissive shape set. So at minimum this should be expressible in schema.ts - a whole-shape / tuple-enum domain (or constraint kind), whatever fits the model best.

For such cases the runtime constraint mechanism is intentionally extensible without breaking backward compatibility. If you need to express a constraint as the one in the example you could just add a new runtime constraint type (I don't know the details of what exactly you would need that to look like so perhaps it will be better to add it in the OCR PR). We don't support specifying a list of shapes in the schema because that leads to an NP-hard problem when matching two schemas. Additionally the CoreML enumerated shapes are somewhat of an exception specific to that backend, so in general we want to treat it the same way as range dims in which case the per-dim domains + runtime constraints is the way to go inspired by how [PyTorch's `export` works](https://docs.pytorch.org/docs/2.13/user_guide/torch_compiler/export.html#expressing-dynamism).

benITo47:
Alrighty, just wanted to flag that such cases exist.
I don't have anything more to flag.  
Godspeed.

barhanc:
@msluszniak  I've rebased, updated the models on SWM HF and tested them again (the ones I could on a simulator). If there is a green light I will merge this PR.

msluszniak:
There is a green light indeed, today I tested both coreml and mlx variants of uploaded files.

PR #1328, opened 2026-07-24
[RNE Rewrite] Integrate resource fetcher (react-native-blob-util)

### LINKED ISSUE #1253: [RNE Rewrite] Integrate ResourceFetcher with refactor

msluszniak:
## Description

Adds resource fetching to the rewrite as a single `react-native-blob-util`-backed fetcher inside `react-native-executorch`. Replaces the temporary `react-native-fs` hook.

`download(source, { onProgress, signal })` is a single generic entry point: it takes any nested structure of plain objects and arrays, downloads every string leaf that is an `http(s)` URL, and returns the value with those URLs replaced by local paths. Everything else passes through untouched, so the result is structurally identical to the input. `download<T>(source: T): Promise<T>` types correctly for every task config without per-task overloads:

```ts
const model = await download(models.classification.EFFICIENTNET_V2_S.XNNPACK_FP32);
const { classify, dispose } = await createClassifier(model);
```

Backend is split by platform: Android uses the system DownloadManager (reliable for files >2 GB, background), iOS streams via NSURLSession with `.partial`/HTTP-Range resume.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

1. Run the `computer-vision` example app.
2. Open Classification, pick an XNNPACK model — it downloads with progress, then loads and runs.
3. Reopen the screen — the model is served from cache instantly (no re-download).
4. Open Speech-to-Text — model, tokenizer and the nested VAD model resolve in one pass, with progress weighted across all three.
5. Imperative: `createClassifier(await download(models.classification.EFFICIENTNET_V2_S.XNNPACK_FP32))`.

### Related issues

Closes #1253

### Checklist

### Additional notes

- Swaps the `react-native-fs` peer dependency for `react-native-blob-util`.
- Android must use DownloadManager.
- `isEmulator` is now installed on the `__rnexecutorch_jsi__` module by the native layer.
- File management (list/delete downloaded models) is intentionally out of scope; the fetcher only downloads and returns paths.

### DIFF
--- a/packages/react-native-executorch/cpp/core/utils.cpp
+++ b/packages/react-native-executorch/cpp/core/utils.cpp
@@ -5,8 +5,53 @@
 #include <executorch/runtime/backend/interface.h>
 #include <executorch/runtime/core/error.h>
 
+#if defined(__ANDROID__)
+#include <sys/system_properties.h>
+#elif defined(__APPLE__)
+#include <TargetConditionals.h>
+#endif
+
 namespace rnexecutorch::core::utils {
 namespace jsi = facebook::jsi;
+
+namespace {
+// Detects an Android emulator / iOS simulator. On Android the build
+// fingerprint and hardware name identify the AVD images (goldfish and ranchu
+// are the QEMU-based emulator kernels); on Apple platforms the simulator is
+// known at compile time.
+bool isEmulator() {
+#if defined(__ANDROID__)
+    auto readProp = [](const char *key) -> std::string {
+#if __ANDROID_API__ >= 26
+        const prop_info *pi = __system_property_find(key);
+        if (pi == nullptr) {
+            return "";
+        }
+        std::string result;
+        __system_property_read_callback(
+            pi,
+            [](void *cookie, const char * /*name*/, const char *value, uint32_t /*serial*/) {
+                *static_cast<std::string *>(cookie) = value;
+            },
+            &result);
+        return result;
+#else
+        char value[PROP_VALUE_MAX] = {0};
+        __system_property_get(key, value);
+        return {value};
+#endif
+    };
+
+    const std::string fingerprint = readProp("ro.build.fingerprint");
+    const std::string hardware = readProp("ro.hardware");
+    return fingerprint.rfind("generic", 0) == 0 || hardware == "goldfish" || hardware == "ranchu";
+#elif defined(__APPLE__) && TARGET_OS_SIMULATOR
+    return true;
+#else
+    return false;
+#endif
+}
+} // namespace
 
 void install_getExecuTorchRegisteredBackends(jsi::Runtime &rt, jsi::Object &module) {
     const auto *name = "getExecuTorchRegisteredBackends";
@@ -30,5 +75,9 @@ void install_getExecuTorchRegisteredBackends(jsi::Runtime &rt, jsi::Object &modu
     auto fn = jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, name), 0, fnBody);
 
     module.setProperty(rt, name, fn);
+}
+
+void install_isEmulator(jsi::Runtime &rt, jsi::Object &module) {
+    module.setProperty(rt, "isEmulator", jsi::Value(isEmulator()));
 }
 } // namespace rnexecutorch::core::utils
--- a/packages/react-native-executorch/cpp/core/utils.cpp
+++ b/packages/react-native-executorch/cpp/core/utils.cpp
@@ -15,10 +15,13 @@ namespace rnexecutorch::core::utils {
 namespace jsi = facebook::jsi;
 
 namespace {
-// Detects an Android emulator / iOS simulator. On Android the build
-// fingerprint and hardware name identify the AVD images (goldfish and ranchu
-// are the QEMU-based emulator kernels); on Apple platforms the simulator is
-// known at compile time.
+// Detects an Android emulator / iOS simulator. On Android no single property
+// covers every image, so we check three: the build fingerprint (`generic...`
+// for AOSP images), the hardware name (`goldfish`/`ranchu` are the QEMU
+// emulator kernels, `cutf`/`vsoc` prefixes are Cuttlefish virtual devices), and
+// the product model (Play-store and SDK images report `sdk_gphone`,
+// `google_sdk`, `Emulator` or `Cuttlefish`). On Apple platforms the simulator
+// is known at compile time.
 bool isEmulator() {
 #if defined(__ANDROID__)
     auto readProp = [](const char *key) -> std::string {
@@ -42,9 +45,27 @@ bool isEmulator() {
 #endif
     };
 
+    const auto startsWith = [](const std::string &value, const char *prefix) {
+        return value.rfind(prefix, 0) == 0;
+    };
+
     const std::string fingerprint = readProp("ro.build.fingerprint");
+    if (startsWith(fingerprint, "generic") || startsWith(fingerprint, "unknown")) {
+        return true;
+    }
+
     const std::string hardware = readProp("ro.hardware");
-    return fingerprint.rfind("generic", 0) == 0 || hardware == "goldfish" || hardware == "ranchu";
+    if (hardware == "goldfish" || hardware == "ranchu" || startsWith(hardware, "cutf") ||
+        startsWith(hardware, "vsoc")) {
+        return true;
+    }
+
+    const std::string model = readProp("ro.product.model");
+    return model.find("sdk_gphone") != std::string::npos ||
+           model.find("google_sdk") != std::string::npos ||
+           model.find("Emulator") != std::string::npos ||
+           model.find("Android SDK built for") != std::string::npos ||
+           model.find("Cuttlefish") != std::string::npos;
 #elif defined(__APPLE__) && TARGET_OS_SIMULATOR
     return true;
 #else
--- a/packages/react-native-executorch/src/fetcher/ResourceFetcher.ts
+++ b/packages/react-native-executorch/src/fetcher/ResourceFetcher.ts
@@ -1,11 +1,20 @@
 /* eslint-disable no-bitwise */
+import { Platform } from 'react-native';
 import RNBlobUtil from 'react-native-blob-util';
 import { triggerDownloadEvent, triggerHuggingFaceDownloadCounter } from './telemetry';
 
-// Persistent, per-app directory where downloaded model assets are cached. We use
-// DocumentDir rather than CacheDir so the OS won't evict large models between
-// runs, forcing a costly re-download.
-const RNE_DIRECTORY = `${RNBlobUtil.fs.dirs.DocumentDir}/react-native-executorch`;
+const IS_ANDROID = Platform.OS === 'android';
+
+// Persistent, per-app directory where downloaded model assets are cached.
+//   iOS: internal DocumentDir (not CacheDir) so the OS won't evict large models
+//        between runs and force a costly re-download.
+//   Android: the app-private EXTERNAL files dir (getExternalFilesDir), so the
+//        system DownloadManager can write there and same-volume moves stay cheap
+//        even for multi-GB files. Falls back to DocumentDir if unmounted.
+const ANDROID_DIRECTORY = RNBlobUtil.fs.dirs.SDCardDir || RNBlobUtil.fs.dirs.DocumentDir;
+const RNE_DIRECTORY = IS_ANDROID
+  ? `${ANDROID_DIRECTORY}/react-native-executorch`
+  : `${RNBlobUtil.fs.dirs.DocumentDir}/react-native-executorch`;
 
 /**
  * Progress callback reporting overall completion as a fraction in `[0, 1]`.
@@ -21,8 +30,8 @@ export interface DownloadOptions {
   /** Called with overall progress in `[0, 1]` as bytes arrive. */
   onProgress?: DownloadProgressCallback;
   /**
-   * Aborts the download. Bytes fetched so far are kept on disk so a later
-   * {@link download} of the same source resumes instead of restarting.
+   * Aborts the download. On iOS the bytes fetched so far are kept on disk so a
+   * later {@link download} of the same source resumes instead of restarting.
    */
   signal?: AbortSignal;
 }
@@ -79,15 +88,10 @@ interface DownloadOneCallbacks {
 }
 
 /**
- * Downloads a single remote file into the cache with HTTP-Range resume support.
- * Returns the local path. `canResume` is set to `false` on an internal retry to
- * avoid recursing forever when partial-file assembly fails.
+ * Downloads a single remote file into the cache, dispatching to the
+ * platform-appropriate backend. Returns the local path.
  */
-async function downloadOne(
-  url: string,
-  cb: DownloadOneCallbacks,
-  canResume = true
-): Promise<string> {
+async function downloadOne(url: string, cb: DownloadOneCallbacks): Promise<string> {
   const dest = cachePathFor(url);
 
   // Cache hit — nothing to download.
@@ -97,14 +101,83 @@ async function downloadOne(
     return dest;
   }
 
-  // Count this actual (non-cached) fetch once — not on the internal retry.
-  if (canResume) {
-    triggerHuggingFaceDownloadCounter(url);
-    triggerDownloadEvent(url);
-  }
+  // Count this actual (non-cached) fetch once.
+  triggerHuggingFaceDownloadCounter(url);
+  triggerDownloadEvent(url);
 
   await RNBlobUtil.fs.mkdir(RNE_DIRECTORY).catch(() => {});
 
+  return IS_ANDROID ? downloadViaDownloadManager(url, dest, cb) : downloadViaStream(url, dest, cb);
+}
+
+/**
+ * Android backend: the system DownloadManager streams to app-private external
+ * storage. Unlike blob-util's in-process reader it handles files larger than
+ * 2 GB, keeps downloading while the app is backgrounded or killed, and resumes
+ * across transient network drops on its own — so no manual Range logic here.
+ */
+async function downloadViaDownloadManager(
+  url: string,
+  dest: string,
+  cb: DownloadOneCallbacks
+): Promise<string> {
+  const tmp = `${dest}.downloading`;
+  await RNBlobUtil.fs.unlink(tmp).catch(() => {});
+
+  if (cb.signal?.aborted) throw abortError();
+
+  const task = RNBlobUtil.config({
+    addAndroidDownloads: {
+      useDownloadManager: true,
+      path: tmp,
+      notification: false,
+      mediaScannable: false,
+      mime: 'application/octet-stream',
+    },
+  }).fetch('GET', url);
+
+  const onAbort = () => task.cancel();
+  cb.signal?.addEventListener('abort', onAbort);
+
+  // DownloadManager reports total as -1 until the size is known; forward the
+  // received byte count regardless so byte-weighted progress still advances.
+  task.progress({ count: 100 }, (received, total) => {
+    const recv = Number(received);
+    const tot = Number(total);
+    cb.onBytes?.(recv, tot > 0 ? tot : recv);
+  });
+
+  try {
+    await task;
+  } catch (e) {
+    await RNBlobUtil.fs.unlink(tmp).catch(() => {});
+    throw cb.signal?.aborted ? abortError() : e;
+  } finally {
+    cb.signal?.removeEventListener('abort', onAbort);
+  }
+
+  // DownloadManager doesn't surface an HTTP status; an empty file means failure.
+  const size = await fileSize(tmp);
+  if (size <= 0) {
+    await RNBlobUtil.fs.unlink(tmp).catch(() => {});
+    throw new Error(`Download of ${url} failed (empty response).`);
+  }
+  await RNBlobUtil.fs.mv(tmp, dest);
+  return dest;
+}
+
+/**
+ * iOS backend: blob-util streams via NSURLSession straight to disk. Interrupted
+ * downloads resume from a `.partial` file via an HTTP Range request. `canResume`
+ * is set to `false` on an internal retry to avoid recursing forever if
+ * partial-file assembly ever fails.
+ */
+async function downloadViaStream(
+  url: string,
+  dest: string,
+  cb: DownloadOneCallbacks,
+  canResume = true
+): Promise<string> {
   const part = `${dest}.partial`;
   if (!canResume) await RNBlobUtil.fs.unlink(part).catch(() => {});
   const offset = canResume ? await fileSize(part) : 0;
@@ -169,7 +242,7 @@ async function downloadOne(
     // once as a plain full download so correctness never depends on resume.
     await RNBlobUtil.fs.unlink(part).catch(() => {});
     await RNBlobUtil.fs.unlink(target).catch(() => {});
-    if (canResume) return downloadOne(url, cb, false);
+    if (canResume) return downloadViaStream(url, dest, cb, false);
     throw assemblyErr;
   }
 }
@@ -179,9 +252,10 @@ async function downloadOne(
  * resolves with their local file paths.
  *
  * Local paths (anything not starting with `http`) are returned unchanged.
- * Remote files are streamed to disk with resume support: an interrupted
- * download continues from where it stopped on the next call rather than
- * restarting. When several sources are passed, overall progress is weighted by
+ * Remote files are downloaded to a persistent cache: on Android via the system
+ * DownloadManager (handles multi-GB files and continues in the background), on
+ * iOS via a streaming request that resumes an interrupted download from where
+ * it stopped. When several sources are passed, overall progress is weighted by
  * their byte sizes so a large model isn't reported the same as a tiny
  * tokenizer.
  * @category Fetching
--- a/packages/react-native-executorch/src/fetcher/ResourceFetcher.ts
+++ b/packages/react-native-executorch/src/fetcher/ResourceFetcher.ts
@@ -87,10 +87,8 @@ interface DownloadOneCallbacks {
   signal?: AbortSignal;
 }
 
-/**
- * Downloads a single remote file into the cache, dispatching to the
- * platform-appropriate backend. Returns the local path.
- */
+// Downloads a single remote file into the cache, dispatching to the
+// platform-appropriate backend. Returns the local path.
 async function downloadOne(url: string, cb: DownloadOneCallbacks): Promise<string> {
   const dest = cachePathFor(url);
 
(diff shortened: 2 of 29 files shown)


packages/react-native-executorch/src/utils.ts
```diff
@@ -35,8 +34,10 @@ export async function inspectModel(source: string): Promise<{
   let downloaded = false;
 
   if (source.startsWith('http')) {
-    localPath = `${RNFS.TemporaryDirectoryPath}/inspect_model_${Date.now()}.pte`;
-    await RNFS.downloadFile({ fromUrl: source, toFile: localPath }).promise;
+    // Throwaway download to a temp path — inspection shouldn't populate the
+    // persistent resource cache, so we don't go through `download()`.
```
barhanc, line 46: Since this will now become part of the standard API we should add an explanation of this throwaway behavior to the jsdoc.

packages/react-native-executorch/src/fetcher/telemetry.ts
```diff
@@ -0,0 +1,94 @@
...
+    const region = locale.split('-').pop();
+    if (region && region.length === 2) return region.toUpperCase();
+  } catch {}
+  return 'UNKNOWN';
+}
+
+// Set by the native layer when running on a simulator/emulator, so dev traffic
+// can be filtered out server-side. Absent (⇒ false) until then.
+function isEmulator(): boolean {
+  return (globalThis as { __rne_isEmulator?: boolean }).__rne_isEmulator === true;
```
barhanc: How does this work? Where is the property `__rne_isEmulator` set?

packages/react-native-executorch/src/fetcher/fetcher.ts
barhanc, line 1: Let's rename this file to `fetcher.ts` so that all files have consistent naming.

packages/react-native-executorch/src/fetcher/ResourceFetcher.ts
```diff
@@ -0,0 +1,317 @@
...
+ * matching the shape of `source`.
+ */
+export function download(source: string, options?: DownloadOptions): Promise<string>;
+// `const T` preserves tuple length so `download([a, b])` resolves to a
+// fixed-length `[string, string]` rather than a `(string | undefined)[]`.
+export function download<const T extends readonly string[]>(
+  sources: T,
+  options?: DownloadOptions
+): Promise<{ [K in keyof T]: string }>;
+export async function download(
```
barhanc: I'm not sure that the pattern of adding `download(modelPath)` to every task implementation is the best way to achieve that. I think we should separate the downloading from creating the pipeline, so imo the pipelines should stay as they were and the pattern should rather be something like
```ts
const model = download(models.classification.EFFICIENTNET_V2_S);
const { classify, dispose } = createClassifier(model);
```
where `download` would work in the following way: it takes an arbitrary nested object and for each string leaf attribute it checks if it's a valid url and if so downloads it and substitutes the url for the local path. This way we wouldn't need two different mechanisms for downloading in `create<Task>` and `use<Task>` (where the path substitution is left as it was) and we wouldn't have to actually write the path substitution by hand in each hook. The only problem I see is that I'm not sure how difficult would it be to implement something like this in a way that download also returns a proper type but is generic at the same time (so it works for all model types out of the box). My gut feeling is that it should be possible as TS type system is quite powerful but I don't have a concrete idea right now.

packages/react-native-executorch/src/fetcher/telemetry.ts
msluszniak: It wasn't set. Now it is.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: Done.

packages/react-native-executorch/src/fetcher/ResourceFetcher.ts
msluszniak: As in the updated PR description:

`download(source, { onProgress, signal })` is a single generic entry point: it takes any nested structure of plain objects and arrays, downloads every string leaf that is an http(s) URL, and returns the value with those URLs replaced by local paths. Everything else passes through untouched, so the result is structurally identical to the input. `download<T>(source: T): Promise<T>` types correctly for every task config without per-task overloads:

```ts
const model = await download(models.classification.EFFICIENTNET_V2_S.XNNPACK_FP32);
const { classify, dispose } = await createClassifier(model);
```

packages/react-native-executorch/src/fetcher/telemetry.ts
```diff
@@ -0,0 +1,95 @@
...
+ * @param uri The URI of the file being downloaded.
+ */
+export function triggerHuggingFaceDownloadCounter(uri: string): void {
+  try {
+    const url = new URL(uri);
+    if (!isSwmHuggingFaceRepo(url)) return;
+    const base = `${url.protocol}//${url.host}${url.pathname.split('resolve')[0]}`;
+    // Fire-and-forget; its success is irrelevant to the download itself.
+    fetch(`${base}resolve/main/config.json`, { method: 'HEAD' }).catch(() => {});
+  } catch {}
```
barhanc: Instead of empty `catch` here and in other functions maybe do something like
```ts
} catch (e) {
   if (__DEV__) console.warn('[RNE Telemetry] Failed to ...:', e);
}
```

packages/react-native-executorch/src/fetcher/telemetry.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function getCountryCode(): string {
```
barhanc, line 67: This might return language code which is not a valid country code, e.g. if `locale` is `'en'`.

packages/react-native-executorch/src/fetcher/telemetry.ts
```diff
@@ -0,0 +1,95 @@
...
+    const locale = Intl.DateTimeFormat().resolvedOptions().locale;
+    const region = locale.split('-').pop();
+    if (region && region.length === 2) return region.toUpperCase();
+  } catch {}
+  return 'UNKNOWN';
+}
+
+// Reported by the native installer (Android build props / TARGET_OS_SIMULATOR)
+// so development traffic can be filtered out server-side.
+function isEmulator(): boolean {
```
barhanc: Why not just use the `rnexecutorchJsi.isEmulator` property directly?

packages/react-native-executorch/src/hooks/useClassifier.ts
```diff
@@ -19,21 +19,17 @@ import { createClassifier, type ClassifierModel } from '../extensions/cv/tasks/c
...
-    [localPath]
-  );
+  const { model, error } = useModel(createClassifier<L>, resource ?? null, [resource]);
 
   return {
     isReady: !!model,
     error: downloadError || error,
     downloadProgress,
-    localPath,
+    localPath: resource?.modelPath,
```
barhanc: I'm wondering if we shouldn't actually return the full resolved resource here instead of `localPath` and the same for other models, so that it is easy for end-users to manage the files. What do you think?

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+const ANDROID_DIRECTORY = RNBlobUtil.fs.dirs.SDCardDir || RNBlobUtil.fs.dirs.DocumentDir;
+const RNE_DIRECTORY = IS_ANDROID
+  ? `${ANDROID_DIRECTORY}/react-native-executorch`
+  : `${RNBlobUtil.fs.dirs.DocumentDir}/react-native-executorch`;
+
+/**
+ * Progress callback reporting overall completion as a fraction in `[0, 1]`.
+ * @category Types
+ */
+export type DownloadProgressCallback = (progress: number) => void;
```
barhanc: Not sure if this type is necessary.

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+
+/**
+ * Options controlling a {@link download} call.
+ * @category Types
+ */
+export interface DownloadOptions {
```
barhanc, line 23: It would be good to have like a `forceDownload` option or something like that to redownload cached resource if necessary.

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+  } catch {
+    return 0;
+  }
+}
+
+function abortError(): Error {
+  const err = new Error('The download was aborted.');
+  err.name = 'AbortError';
+  return err;
+}
```
barhanc: Maybe use
```ts
class AbortError extends Error {
...
}
```

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+interface DownloadOneCallbacks {
+  // Reports absolute received/total bytes for this file, including any bytes
+  // already present from a previous partial download.
+  onBytes?: (received: number, total: number) => void;
+  signal?: AbortSignal;
+}
+
+// Downloads a single remote file into the cache, dispatching to the
+// platform-appropriate backend. Returns the local path.
+async function downloadOne(url: string, cb: DownloadOneCallbacks): Promise<string> {
```
barhanc: Maybe rename to `downloadUrl` and `DownloadUrlCallbacks`.

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+  // Cache hit — nothing to download.
+  if (await RNBlobUtil.fs.exists(dest)) {
+    const size = await fileSize(dest);
+    cb.onBytes?.(size, size);
+    return dest;
+  }
+
+  // Count this actual (non-cached) fetch once.
+  triggerHuggingFaceDownloadCounter(url);
+  triggerDownloadEvent(url);
```
barhanc: Maybe use
```ts
import * as telemetry from './telemetry'
...
  telemetry.triggerHuggingFaceDownloadCounter(url);
  telemetry.triggerDownloadEvent(url);
```

packages/react-native-executorch/src/fetcher/fetcher.ts
barhanc: There is no in-flight request deduplication or locking in `downloadOne`. If two hooks or components call download() concurrently for the same URL, both resolve to the same destination path and attempt to manage the same temporary file.

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+  await RNBlobUtil.fs.mkdir(RNE_DIRECTORY).catch(() => {});
+
+  return IS_ANDROID ? downloadViaDownloadManager(url, dest, cb) : downloadViaStream(url, dest, cb);
+}
+
+// Android backend: the system DownloadManager streams to app-private external
+// storage. Unlike blob-util's in-process reader it handles files larger than
+// 2 GB, keeps downloading while the app is in the background or killed, and
+// resumes across transient network drops on its own — so no manual Range logic.
+async function downloadViaDownloadManager(
```
barhanc: Since these are platform specific maybe the names should reflect that, e.g. `downloadUrlViaAndroidDownloadManager` or something shorter.

Also why do we actually need/want different implementations on Android and iOS?

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+
+const IS_ANDROID = Platform.OS === 'android';
+
+// Persistent, per-app directory where downloaded model assets are cached.
+//   iOS: internal DocumentDir (not CacheDir) so the OS won't evict large models
+//        between runs and force a costly re-download.
+//   Android: the app-private EXTERNAL files dir (getExternalFilesDir), so the
+//        system DownloadManager can write there and same-volume moves stay cheap
+//        even for multi-GB files. Falls back to DocumentDir if unmounted.
+const ANDROID_DIRECTORY = RNBlobUtil.fs.dirs.SDCardDir || RNBlobUtil.fs.dirs.DocumentDir;
```
barhanc, line 14: According to clanker `RNBlobUtil.fs.dirs.SDCardDir` resolves to the root external storage directory (/storage/emulated/0), not an app-private external files directory. On Android 10+ (API 29+), Scoped Storage blocks direct writes to /storage/emulated/0/react-native-executorch without broad WRITE_EXTERNAL_STORAGE or MANAGE_EXTERNAL_STORAGE permissions. Android DownloadManager and RNBlobUtil.fs.mkdir/mv operations will throw EACCES (Permission denied).

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+  const sizes = await Promise.all(urls.map(remoteSize));
+  const haveAllSizes = sizes.every((size) => size > 0);
+  const total = haveAllSizes ? sizes.reduce((a, b) => a + b, 0) : urls.length;
+
+  const received = new Array<number>(urls.length).fill(0);
+  const report = () => {
+    if (!options.onProgress) return;
+    const sum = received.reduce((a, b) => a + b, 0);
+    options.onProgress(total > 0 ? Math.min(sum / total, 1) : 0);
+  };
```
barhanc, line 442: This seems slightly convoluted. Why do we need an array `received`, wouldn't just a single number that we update directly in `onBytes` be enough?

packages/react-native-executorch/src/fetcher/fetcher.ts
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/**
+ * Walks a value and collects every distinct remote URL sitting on a string
+ * leaf. Deduplicating here means a URL repeated across fields is fetched once.
+ * @param node The value to walk.
+ * @param out Accumulator, so recursive calls share one set.
+ * @returns The set of remote URLs referenced by `node`.
+ */
+export function collectRemoteSources(node: unknown, out = new Set<string>()): Set<string> {
```
barhanc: Instead of defining `out = ...` as default param, it's a better pattern to use
```ts
export function collectRemoteSources(node: unknown): Set<string> {
  const out = new Set<string>();
  const visit = (curr: unknown) => {
     ...
  };
  visit(node);
  return out;
}
```

packages/react-native-executorch/cpp/core/utils.cpp
```diff
@@ -5,9 +5,54 @@
...
+#else
+        char value[PROP_VALUE_MAX] = {0};
+        __system_property_get(key, value);
+        return {value};
+#endif
+    };
+
+    const std::string fingerprint = readProp("ro.build.fingerprint");
+    const std::string hardware = readProp("ro.hardware");
+    return fingerprint.rfind("generic", 0) == 0 || hardware == "goldfish" || hardware == "ranchu";
```
barhanc: Clanker says that modern Android emulators (Google Play x86_64 / gphone64, Cuttlefish, and ARM virtual devices) use also model names like google_sdk or cuttlefish.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
```diff
@@ -1,110 +1,80 @@
...
   const [downloadError, setDownloadError] = useState<Error | null>(null);
 
+  // Downloading is keyed purely on the URLs referenced by `config`, so passing
+  // an inline config object doesn't restart the download on every render.
+  const sourcesKey = useMemo(() => [...collectRemoteSources(config)].sort().join('\n'), [config]);
+
+  // Read at substitution time only, so unrelated config edits don't re-download.
+  const configRef = useRef(config);
+  configRef.current = config;
+
```
barhanc: Couldn't we just do something like this. We don't need to expose the `collectRemoteSources` and `substituteRemoteSources` this way and since the `download` caches downloads internally there shouldn't be any problems with re-renders.
```ts
export function useResourceDownload<T>(config: T, preventLoad?: boolean) {
  const [resource, setResource] = useState<T>();
  const [downloadError, setDownloadError] = useState<Error | null>(null);

  useEffect(() => {
    setResource(undefined);
    setDownloadProgress(0);
    setDownloadError(null);

    if (preventLoad) return;

    let isMounted = true;
    const controller = new AbortController();

    download(config, {
      signal: controller.signal,
      onProgress: (progress) => {
        if (isMounted) setDownloadProgress(progress * 100);
      },
    })
      .then((resolved) => {
        if (!isMounted) return;
        setResource(resolved);
        setDownloadProgress(100);
      })
      .catch((e) => {
        if (!isMounted || e instanceof AbortError) return;
        setDownloadError(e instanceof Error ? e : new Error(String(e)));
      });

    return () => {
      isMounted = false;
      controller.abort();
    };
  }, [config, preventLoad]);

  return { resource, downloadProgress, downloadError };
}
```

barhanc (commented):
We should also delete the resource fetcher packages. It can be in a follow-up PR though as not to clutter this one.

packages/react-native-executorch/src/fetcher/telemetry.ts
msluszniak: Yes, I know. But this is how it has been implemented since now. Do you know if there are any better ways to get country code?

packages/react-native-executorch/src/fetcher/telemetry.ts
msluszniak: Yeah, we can simplify this one.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
msluszniak: Ok, I'll test it.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: Yeah, will drop type annotation.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: Agreed, I will add it.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: I think we should consider how we want to deal with RNE errors in general. Do we want to have one aggregated error type for all errors that come from RNE functionalities like before? Or have some groups of errors like FetcherError, RNEError etc? Something different?

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: We need to use different mechanism on Android to support in-background download of bigger files. So this change is forced by how background download works on both platforms.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: I'll check that.

packages/react-native-executorch/src/hooks/useClassifier.ts
msluszniak: Yeap, agreed.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: I think this claim is genuinely wrong. With current version (0.24.10) of blob-util, `SDCardDir` is `getExternalFilesDirPath(ctx, null)` → `ctx.getExternalFilesDir(null)`. This results in `/storage/emulated/0/Android/data/<pkg>/files` directory which is cool. The legacy and incorrect one is: `LegacySDCardDir` (`Environment.getExternalStorageDirectory()`). I already tested fetcher on Android, and if it would be a problem, this should have probably appeared.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: Probably not, a single accumulator would be wrong. `onBytes` reports absolute bytes-so-far per file, not deltas. `DownloadManager` gives cumulative and the iOS path reports offset + recv to account for resume. += would double-count every progress tick.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
msluszniak: Deps `[config, preventLoad]` are fine for every current call site, but an inline config: `useClassifier({ modelPath, modelOpts })` loops forever. Effect runs → `setResource(resolved)` → re-render → new object identity → effect re-runs. So this really only prevents re-downloads. Each cycle still tears down and reloads the native model via `useModel`. So we need to decide: indirection and tolerate inline configs or your variant, document that configs must be referentially stable. I think your approach is a cleaner one, but want to make sure wa are on the same page.

packages/react-native-executorch/cpp/core/utils.cpp
msluszniak: Addressed

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: Done.

packages/react-native-executorch/src/fetcher/fetcher.ts
msluszniak: Resolved, this was a real issue.

msluszniak:
> We should also delete the resource fetcher packages. It can be in a follow-up PR though as not to clutter this one.

What do you mean by deleting fetcher packages? We cannot delete them as such, because we need them for versions <= 0.9.x. I guess the first moment when we could remove them will be release of version 2.0.0. In the current implementation we don't ship any resource fetchers.

packages/react-native-executorch/src/fetcher/telemetry.ts
barhanc: I don't so I guess it can stay the way it is. Maybe just reporting the language code would be more accurate but that would require changing what the telemetry endpoint expects.

packages/react-native-executorch/src/fetcher/fetcher.ts
barhanc: I don't have a definite answer to the error question but regardless, the `AbortError` is not exposed to the public API so this change can be applied independently.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
barhanc: One other approach would be to pass only `JSON.stringify(config)` instead of `config` in both here and in `useModel` (where we wouldn't need to pass `deps` array directly then). I honestly don't know which approach is better.

Also looking at the hooks right now, I think we need to refactor them. The `use<Task>` hooks are now pure boilerplate copy-pasted multiple times, so defining a single `useTask` (or changing `useModel` to internally download the config as well) which handles both the downloading, loading task pipeline and exposing it would allow for clean one line definitions of `useClassifier`, etc. in single file (e.g. directly in `index.ts` in `hooks/`).

barhanc:
> > We should also delete the resource fetcher packages. It can be in a follow-up PR though as not to clutter this one.
> 
> What do you mean by deleting fetcher packages? We cannot delete them as such, because we need them for versions <= 0.9.x. I guess the first moment when we could remove them will be release of version 2.0.0. In the current implementation we don't ship any resource fetchers.

I meant deleting the directories in the repo, not the npm packages. Do we need to keep the directories in repo?

msluszniak:
>I meant deleting the directories in the repo, not the npm packages. Do we need to keep the directories in repo?

Oh, I forgot we still have them. Sure, we need to delete them, they are not needed anymore.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
msluszniak: Went with `JSON.stringify(config)`. Keying on the URL set was actually wrong: `create<Task>` bakes modelOpts into the pipeline, hence opts changes never propagates. `useModel` now derives its own key too, so the `deps` param is gone and `collectRemoteSources/substituteRemoteSources` are no longer exported.

packages/react-native-executorch/src/fetcher/telemetry.ts
msluszniak: I wouldn't add language code as a replacement since it will e.g. mix up all english speakers. However, I Found a concrete bug while looking at this: `locale.split('-').pop()` returns the language subtag when the locale has no region, so de became DE, uk became UK and sv became El Salvador. Now walks the subtags and only accepts a real BCP-47 region, reporting UNKNOWN otherwise.

packages/react-native-executorch/src/hooks/useTextToImage.ts
```diff
@@ -20,25 +20,16 @@ import {
...
+    options?.preventLoad
   );
+  const { model, error } = useModel(createSdxsTextToImage, resource ?? null);
 
   return {
     isReady: !!model,
-    error: modelResource.downloadError || tokenizerResource.downloadError || error,
-    downloadProgress: modelResource.downloadProgress,
+    error: downloadError || error,
+    downloadProgress,
```
barhanc, line 28: Missing `resource` return in here.

packages/react-native-executorch/src/hooks/useModel.ts
```diff
@@ -56,8 +56,10 @@ export function useModel<TConfig, TModel extends { dispose: () => void }>(
       isMounted = false;
       instance?.dispose();
     };
+    // `createModel` is a module-level factory, so `config` alone decides when
+    // the instance has to be rebuilt.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, deps);
+  }, [config]);
```
barhanc: Shouldn't this also be a `JSON.stringify` key?

packages/react-native-executorch/src/hooks/useResourceDownload.ts
```diff
@@ -1,110 +1,73 @@
...
+ * instead of restarting.
  *
- * If the source is a remote URL starting with `http`, the hook checks the local
- * filesystem cache for a matching file. If cached, it immediately returns the
- * local file path. If not cached, it starts downloading the file to the
- * application cache directory, reporting download progress (0-100) and any
- * network/disk errors.
+ * Work is keyed on the *value* of `config` rather than its identity, so passing
+ * an inline object is safe. Any change to the config resolves again, which is
+ * cheap for the files themselves (already-cached URLs are a no-op) but does
```
barhanc, line 34: It should be possible ot also force-download from the hook.

packages/react-native-executorch/src/hooks/useTextToImage.ts
msluszniak: Added.

packages/react-native-executorch/src/hooks/useResourceDownload.ts
msluszniak: I implemented it via `ResourceOptions`, please check if this works for you.

packages/react-native-executorch/src/hooks/useModel.ts
msluszniak: Corrected.

barhanc (approved):

PR #1331, opened 2026-07-27
[RNE Rewrite] Extract tokenizer try-lock helper

### LINKED ISSUE #1316: [RNE Rewrite] Clean up the tokenizer host object: locking, error unwrapping and native default arg
While reviewing `cpp/extensions/nlp/tokenizer.cpp` alongside the `std::span` refactor (#1308 / #1315), three related deviations surfaced. They all sit in the same file and overlap enough that they are probably best resolved together rather than piecemeal. Splitting them out of #1315 kept that PR scoped to the span change.

### 1. `TokenizerHostObject` uses `std::mutex`, while `TensorHostObject` uses `shared_mutex`

The three host objects don't split the way the mutex choice implies:

- `TensorHostObject` uses `std::shared_mutex` because it has genuine read-only concurrency: many ops read the same `src` tensor at once via `tryLockShared`, and only writers need exclusivity.
- `ModelHostObject` uses `std::mutex` for a real reason: `Module::execute()` mutates internal ExecuTorch state and is not `const`, so every method must be exclusive.
- `TokenizerHostObject` uses `std::mutex` — but **every one of its operations is `const`** upstream: `encode`, `decode`, `id_to_piece`, `piece_to_id` and `vocab_size` are all declared `const` in `pytorch/tokenizers/tokenizer.h` and `hf_tokenizer.h`. Only `load()` mutates, and that runs in the constructor before the object reaches JS.

Structurally the tokenizer therefore resembles `Tensor` (read-mostly, with a single exclusive `dispose`) rather than `Model`, and looks to have inherited `std::mutex` from the `ModelHostObject` skeleton.

This has a user-visible consequence. Because the lock is taken with `std::try_to_lock`, two concurrent `encode` calls from different threads do not queue — one **throws** `"encode: Tokenizer is currently in use"`. A pure read operation fails spuriously. Under a `shared_mutex` both would succeed, with `dispose` taking the unique lock.

**Caveat that must be settled before acting:** `const` does not imply thread-safety. Grepping `hf_tokenizer.h`, `bpe_tokenizer_base.h`, `tokenizer.h` and the regex headers turns up no `mutable` state, but `third-party/include` ships headers only, so the implementation could not be inspected for internal scratch state (regex match contexts are a classic offender — `re2` is thread-safe, `pcre2` match data is not). Switching to `shared_mutex` on the strength of `const` alone risks a data race, so this needs the upstream `tokenizers` sources checked for reentrancy first.

### 2. The try-lock + disposed-check boilerplate is duplicated 8×

This block is repeated verbatim (modulo the context string) in 5 places in `tokenizer.cpp` — `encode`, `decode`, `getVocabSize`, `idToToken`, `tokenToId`:

```cpp
std::unique_lock<std::mutex> lock(self->mutex_, std::try_to_lock);
if (!lock.owns_lock()) {
    throw jsi::JSError(rt, "encode: Tokenizer is currently in use");
}
if (!self->tokenizer_) {
    throw jsi::JSError(rt, "encode: Tokenizer has been disposed");
}
```

It is **not** only a tokenizer problem: `ModelHostObject` has the same duplication at `core/model.cpp:231`, `:263` and `:332`. So a tokenizer-local helper would fix 5 of 8 sites and leave the pattern duplicated in core.

The existing `tensor::tryLockShared` / `tryLockUnique` (`core/tensor_helpers.cpp:12-34`) cannot be reused as-is: they are non-template functions fixed on the parameter type (`TensorHostObject`), the mutex type (`std::shared_mutex`, written literally in the signature and body), the liveness check (`data_`) and the error wording ("tensor"). `TokenizerHostObject` differs on every axis, and its members are private where `TensorHostObject`'s are public.

Resolving this properly means a generic helper in core (templated over the host object and a liveness accessor) shared by tensor, model and tokenizer. That edits core, which `core-guidelines` says to leave alone — hence a deliberate decision rather than a drive-by. Note this interacts with (1): the right helper signature depends on whether the tokenizer keeps `std::mutex` or moves to `shared_mutex`.

### Suggested order

1. Settle the thread-safety question upstream, which decides (1).
2. Do (2) with the helper shape that (1) implies, covering `model.cpp` as well.

msluszniak:
## Description

Resolves the locking + boilerplate items of #1316. Extracts the try-lock + liveness-check block that was duplicated across the five `TokenizerHostObject` ops into a single class-scoped `tryLockUnique` member, so callsites read `self->tryLockUnique(rt, ctx)` — no lambdas or template parameters. Error messages are preserved verbatim. The helper is scoped to the tokenizer rather than shared across host objects, to keep callsites simple and avoid coupling unrelated types.

`TokenizerHostObject` keeps its exclusive `std::mutex`.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

Pure refactor with no behavior change: the extracted helper reproduces the exact same lock discipline and error strings as the inlined code.

### Related issues

Resolves the locking/deduplication parts of #1316.

### Checklist

### Additional notes

### DIFF
--- a/packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
+++ b/packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
@@ -3,6 +3,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include <jsi/jsi.h>
@@ -20,6 +21,13 @@ public:
     std::vector<facebook::jsi::PropNameID> getPropertyNames(facebook::jsi::Runtime &rt) override;
 
 private:
+    // Non-blocking try-lock on mutex_ that also checks the tokenizer is still
+    // alive, throwing a facebook::jsi::JSError prefixed with `context` on
+    // contention ("... is currently in use") or after dispose ("... has been
+    // disposed").
+    [[nodiscard]] std::unique_lock<std::mutex> tryLockUnique(facebook::jsi::Runtime &rt,
+                                                             std::string_view context);
+
     std::string tokenizerPath_;
     std::unique_ptr<tokenizers::HFTokenizer> tokenizer_;
     std::mutex mutex_;
--- a/packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
+++ b/packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
@@ -21,10 +21,6 @@ public:
     std::vector<facebook::jsi::PropNameID> getPropertyNames(facebook::jsi::Runtime &rt) override;
 
 private:
-    // Non-blocking try-lock on mutex_ that also checks the tokenizer is still
-    // alive, throwing a facebook::jsi::JSError prefixed with `context` on
-    // contention ("... is currently in use") or after dispose ("... has been
-    // disposed").
     [[nodiscard]] std::unique_lock<std::mutex> tryLockUnique(facebook::jsi::Runtime &rt,
                                                              std::string_view context);
 
--- a/packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
+++ b/packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
@@ -67,6 +67,18 @@ TokenizerHostObject::TokenizerHostObject(std::string tokenizerPath)
     }
 }
 
+std::unique_lock<std::mutex> TokenizerHostObject::tryLockUnique(jsi::Runtime &rt,
+                                                               std::string_view context) {
+    std::unique_lock<std::mutex> lock(mutex_, std::try_to_lock);
+    if (!lock.owns_lock()) {
+        throw jsi::JSError(rt, std::format("{} is currently in use", context));
+    }
+    if (!tokenizer_) {
+        throw jsi::JSError(rt, std::format("{} has been disposed", context));
+    }
+    return lock;
+}
+
 jsi::Value TokenizerHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
     auto nameStr = name.utf8(rt);
 
@@ -81,14 +93,7 @@ jsi::Value TokenizerHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &nam
                 throw jsi::JSError(rt, "encode: Usage: encode(text)");
             }
 
-            std::unique_lock<std::mutex> lock(self->mutex_, std::try_to_lock);
-            if (!lock.owns_lock()) {
-                throw jsi::JSError(rt, "encode: Tokenizer is currently in use");
-            }
-
-            if (!self->tokenizer_) {
-                throw jsi::JSError(rt, "encode: Tokenizer has been disposed");
-            }
+            auto lock = self->tryLockUnique(rt, "encode: Tokenizer");
 
             auto text = conversions::asType<std::string>(rt, "encode: text", args[0]);
             auto tokens = unwrap(rt, "encode: Failed to encode input",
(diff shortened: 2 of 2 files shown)


packages/react-native-executorch/cpp/core/host_object_lock.h
barhanc, line 1: I think it would be better to scope the locking helpers to specific classes instead of trying to fit them all into a generic helper. This creates cleaner ergonomics and simpler looking callsites (e.g. `self->tryLockUnique(rt, ctx)` in tokenizer) that don't require specifying lambdas and template parameters (which end up looking quite complex). Additionally generic helper creates a coupling between unrelated host objects that might become a problem if we wanted to modify the implementation.

packages/react-native-executorch/cpp/core/model.cpp
barhanc, line 1: The `model.cpp` doesn't require helpers as after schema refactor PR (#1327) there are no duplicated locking callsites.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
```diff
@@ -22,6 +22,17 @@ class TokenizerHostObject : public facebook::jsi::HostObject,
...
+    // Exclusive std::mutex, NOT std::shared_mutex, even though every tokenizer
+    // op (encode/decode/id_to_piece/piece_to_id/vocab_size) is const upstream.
+    // const does not imply thread-safe here: HFTokenizer encodes via a PCRE2
+    // fallback (RE2 rejects the lookahead in HF pre-tokenizer regexes), and the
+    // vendored Pcre2Regex shares a single pcre2_match_data member across const
+    // find_all() calls, which pcre2_match() writes into, so concurrent encodes
+    // on one tokenizer would race.
+    // TODO(#1330): upstream allocates match data per call as of
+    // pytorch-labs/tokenizers c321bca3d; once the vendored lib includes that
+    // fix, switch this to shared_mutex so pure reads run concurrently.
```
barhanc: I think we can just leave the `std::mutex` here as it is a safe option and we don't need to worry in the future about checking implementation details of libraries the tokenizer depends on to see if it is safe to run concurrent calls.

packages/react-native-executorch/cpp/core/host_object_lock.h
msluszniak: Done — dropped the generic helper for a class-scoped `tryLockUnique` member; callsites are now `self->tryLockUnique(rt, ctx)`.

packages/react-native-executorch/cpp/core/model.cpp
msluszniak: Reverted `model.cpp`; the dedup is tokenizer-only now.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
msluszniak: Kept the plain `std::mutex` and removed the shared-mutex comment/TODO.

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
```diff
@@ -20,6 +21,13 @@ class TokenizerHostObject : public facebook::jsi::HostObject,
     std::vector<facebook::jsi::PropNameID> getPropertyNames(facebook::jsi::Runtime &rt) override;
 
 private:
+    // Non-blocking try-lock on mutex_ that also checks the tokenizer is still
+    // alive, throwing a facebook::jsi::JSError prefixed with `context` on
+    // contention ("... is currently in use") or after dispose ("... has been
+    // disposed").
```
msluszniak:
```suggestion
```

barhanc (approved):

PR #1332, opened 2026-07-27
[RNE Rewrite] perf(nlp): return tokenizer ids as Int32Array instead of number[]

msluszniak:
## Description

`encode()` now returns an ArrayBuffer-backed `Int32Array` (and `decode()` takes one) instead of `number[]`. When the tokenizer runs on a background worklet runtime, react-native-worklets clones the result to the caller thread with a single `ArrayBuffer` memcpy rather than serializing each token individually — the per-element cost flagged in the review of #1274, which is significant for long texts.

Adds `conversions::toJsiInt32Array` / `fromJsiInt32Array` and migrates the internal consumers (whisper decode call sites; the text-embedding and SDXS paths already use array-like access).

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

Run the `tokenizer` screen in `apps/nlp` — it auto-runs an encode/decode round-trip and asserts the results (logged under `[TokenizerTest]`).

### Related issues

Part of #1208; follows up review feedback on #1274.

### Checklist

### DIFF
--- a/packages/react-native-executorch/cpp/core/conversions.h
+++ b/packages/react-native-executorch/cpp/core/conversions.h
@@ -2,6 +2,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <format>
 #include <optional>
 #include <string>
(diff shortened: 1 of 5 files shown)


barhanc (commented):
I'm not sure if this fix is needed. In fact right now there is no place in the code (except in the example tokenizer screen in NLP app) where the input to one of tokenizer's `encode` / `decode` methods crosses the runtime boundary and as I noted in the [review comment](https://github.com/software-mansion/react-native-executorch/pull/1274#pullrequestreview-4552834718) the mentioned issue could only be a real problem if we had to call `encode`/`decode` asynchronously on a different thread. Additionally using `number[]` is actually imo more ergonomic in the pipeline implementations like embeddings, whisper, sdxs. Did you measure how both implementation scale and where is the crossing point (I would imagine that for small number of tokens serializing `number[]` is actually cheaper) and at what point does serialization of `number[]` actually break? Without this data and/or actual use-case where current implementation breaks, this fix seems redundant.

msluszniak:
Yeah, I'll benchmark it and reach our to you. If this would give as diminishing returns, then we can close it and resolve in aggregated issue :)

msluszniak:
@barhanc measured on a Galaxy S26 Ultra (Hermes). Timed `runOnRuntimeAsync` round-trips returning an N-int array, ms/call (scalar baseline ~0.2–0.3ms is common to both):

| N (tokens) | `number[]` | `Int32Array` | ratio |
|---:|---:|---:|---:|
| 1 | 0.27 | 0.36 | 0.76× (number[] wins) |
| 16 | 0.61 | 0.40 | 1.5× |
| 64 | 0.84 | 0.38 | 2.2× |
| 256 | 1.77 | 0.33 | 5.4× |
| 1,024 | 5.26 | 0.40 | 13× |
| 4,096 | 14.5 | 0.41 | 35× |
| 16,384 | 49.6 | 0.71 | 70× |
| 65,536 | 221 | 3.2 | 69× |
| 262,144 | 809 | 13.1 | 62× |
| 1,048,576 | 3,148 | 21.4 | 147× |

- **Crossover ≈ 10 tokens** — `number[]` only wins at N=1.
- `number[]` is linear at ~**3µs/token** and never plateaus: 16ms (a dropped frame) by ~5k tokens, ~150ms at 50k, **3.1s at 1M**. Reconstruction runs on the JS thread, so that's a UI freeze, not background work. `Int32Array` stays ~flat (single-memcpy clone).

Right now our API indeed doesn't use any async call to raw encode / decode. But it might be useful in token counting, RAG / long-document handling, custom generation. So we need to balance what is more important: the support for current flow or benefits that comes from single memcpy call.

packages/react-native-executorch/src/extensions/nlp/tokenizer.ts
```diff
@@ -13,19 +13,22 @@ export type Tokenizer = {
 
   /**
    * Encodes a string into token ids (special tokens are added according to the
-   * tokenizer.json post_processor).
+   * tokenizer.json post_processor). Token ids are returned as an `Int32Array`
+   * so that, when the tokenizer runs on a background worklet runtime, the
+   * result is transferred to the caller thread as a single ArrayBuffer copy
+   * rather than serialized element by element (which is costly for long texts).
```
barhanc: The comment about serialisation cost is an implementation detail, not belonging to public API docs.

packages/react-native-executorch/cpp/core/conversions.h
barhanc, line 1: I'm not sure about this implicit casting behaviour, seems very error prone if someone confuses which typeparam means what. Additionally it doesn't actually prevent users from interpreting TypedArray's bytes in a wrong way e.g. when someone passes `Float32Array` in TS and in C++ does `fromJsiTypedArray<in32_t>(...)` the code will happily interpret fp32 bytes as i32, while the signature and docs are a bit misleading in suggesting that this function prevents such cases. I would keep it simple and reuse the pattern that we have in `tensor.cpp` where we only check that pure bytes offset/length match and return these bytes interpreted as concrete `std::vector<T>`. If we need to then reinterpret each element as something else, then e.g. in tokenizer we can just add `std::vector<int32_t>(tokens.begin(), tokens.end())` / `std::vector<uint64_t>(tokens.begin(), tokens.end())`

packages/react-native-executorch/cpp/core/conversions.h
```diff
@@ -112,6 +112,49 @@ std::vector<T> asVector(jsi::Runtime &rt, const std::string &ctx, const jsi::Val
...
+    }
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast): reading the typed array's raw storage
+    const auto *src = reinterpret_cast<const Storage *>(buffer.data(rt) + byteOffset);
+    std::vector<T> vec;
+    vec.reserve(length);
+    for (size_t i = 0; i < length; ++i) {
+        vec.push_back(static_cast<T>(src[i]));
+    }
+    return vec;
+}
```
barhanc, line 150:
```suggestion
template <typename T>
std::vector<T> fromJsiTypedArray(jsi::Runtime &rt, const std::string &ctx, const jsi::Value &val) {
    static_assert(std::is_arithmetic_v<T>, "fromJsiTypedArray requires an arithmetic storage type");

    auto obj = asType<jsi::Object>(rt, ctx, val);
    auto buffer = getRequiredProperty<jsi::ArrayBuffer>(rt, ctx, obj, "buffer");

    const size_t byteOffset = getOptionalProperty<uint64_t>(rt, ctx, obj, "byteOffset").value_or(0);
    const size_t byteLength = getOptionalProperty<uint64_t>(rt, ctx, obj, "byteLength").value_or(buffer.size(rt));

    if (byteOffset > buffer.size(rt) || byteLength > buffer.size(rt) - byteOffset) {
        throw jsi::JSError(rt, ctx + " out of bounds bytesoffset/byteslength for ArrayBuffer");
    }
    if (byteLength % sizeof(T) != 0) {
        throw jsi::JSError(rt, std::format("{}: byteLength is not a multiple of sizeof(T)={}",
                                           ctx, std::to_string(sizeof(T))));
    }

    auto numel = byteLength / sizeof(T);
    std::vector<T> vec(numel);

    std::memcpy(vec.data(), buffer.data(rt) + byteOffset, byteLength);

    return vec;
}
```

packages/react-native-executorch/cpp/core/conversions.h
```diff
@@ -136,4 +179,70 @@ jsi::Array toJsiArray(jsi::Runtime &rt, const std::vector<T> &vec) {
...
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast): writing directly into the typed array's storage
+    auto *dst = reinterpret_cast<Storage *>(arrayBuffer.getArrayBuffer(rt).data(rt));
+    for (size_t i = 0; i < vec.size(); ++i) {
+        dst[i] = static_cast<Storage>(vec[i]);
+    }
+    return rt.global()
+        .getPropertyAsFunction(rt, jsiTypedArrayName<Storage>())
+        .callAsConstructor(rt, arrayBuffer)
+        .asObject(rt);
+}
```
barhanc, line 238:
```suggestion
template <typename T>
jsi::Object toJsiTypedArray(jsi::Runtime &rt, const std::vector<T> &vec) {
    static_assert(std::is_arithmetic_v<T>, "toJsiTypedArray requires an arithmetic type");
    const size_t byteLength = vec.size() * sizeof(T);
    auto arrayBuffer = rt.global()
                           .getPropertyAsFunction(rt, "ArrayBuffer")
                           .callAsConstructor(rt, static_cast<double>(byteLength))
                           .asObject(rt);

    if (!vec.empty()) {
        auto *bufferData = arrayBuffer.getArrayBuffer(rt).data(rt);
        std::memcpy(bufferData, vec.data(), byteLength);
    }

    return rt.global()
        .getPropertyAsFunction(rt, jsiTypedArrayName<T>())
        .callAsConstructor(rt, arrayBuffer)
        .asObject(rt);
}
```

packages/react-native-executorch/cpp/core/conversions.h
msluszniak: Yeah, I added two types only because this mismatch in tokenizer (int32_t vs uint64_t). I would mimic the way we have functions implemented in tensor.cpp in very similar way to what you proposed in review diffs.

barhanc (approved):

PR #1337, opened 2026-07-30
[RNE Rewrite] docs(cpp): add missing C++ documentation in header files

barhanc:
## Description

Adds missing Doxygen documentation across C++ header files in `packages/react-native-executorch/cpp/`.

- [ ] Yes
- [x]  No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

N/A

### Checklist

### DIFF
--- a/packages/react-native-executorch/cpp/core/dtype.h
+++ b/packages/react-native-executorch/cpp/core/dtype.h
@@ -5,6 +5,10 @@
 #include <string>
 
 namespace rnexecutorch::core::types {
+
+/**
+ * Supported tensor data types across the native runtime and JavaScript interface.
+ */
 enum class DType {
     uint8,
     int32,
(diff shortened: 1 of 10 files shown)


packages/react-native-executorch/cpp/core/dtype.h
msluszniak, line 1: Now when I read the names of the functions I'm wondering why we have such name convention. Str -> parseDtype, but reverse is named toString. For the next two the convention is unified, but we only know that we convert sth to scalar and we get sth from scalar. Maybe we should name it xFromY and xToY for both conversions? Especially, because we have `dtypeToCvDepth` and similar.

barhanc: Yeah, you are right. Let's wait with this PR until #1327 lands so we can change the names.

barhanc: Fixed.

msluszniak (approved):

Issue #1288, opened 2026-06-26
[RNE Rewrite] Streamline apps to show idiomatic usage of the library


barhanc:
We need to refactor the example apps implementations so that the implemetnation of appropriate screens showcases the idiomatic usage of the library and doesn't focus on everything around it. This should be achieved mainly by introducing common helper functions and components so that the focus point of the screen implementation is the idiomatic usage of our hook API, what it takes and returns and how to use it downstream. We can also do a visual overhaul of the apps as part of this issue.

barhanc:
Closed as completed in #1388

PR #1354, opened 2026-08-07
[RNE Rewrite] feat: coded, worklet-safe error handling

### LINKED ISSUE #1353: [RNE Rewrite] Setup new error handling schema
We need to setup new way how to handling errors in rewrite branch.

msluszniak:
## Description

This branch had no error module: everything threw a plain `Error` with a message, and `wrapAsync` flattened anything down to `e?.message`. `main` has a code-based contract, but that contract cannot survive in worklets, etc.

This keeps the machine-readable `code` contract, adopts this branch's message discipline, and makes both survive worklets and JSI.

- Error codes: 46 down to 13. ExecuTorch runtime errors travel in a separate `etCode` field so RNE and ET errors stay independent. Drops `Ok`, the `UnknownError`/`Internal` overlap, and `ModuleNotLoaded` (not applicable anymore).
- `RnExecutorchError` sets `name`, keeps `instanceof` working through subclasses via `new.target`, no `cause` anymore.
- Worklets cannot carry class identity across runtimes, so `rnExecutorchError()` throws a plain data inside them and `wrapAsync` rebuilds the error. `isRnExecutorchError()` works in both and is the recommended check.
- Native: `CodedError` plus `guarded()` on all host functions, and raw `jsi::JSError` / `std::runtime_error` / `std::invalid_argument` throws converted, so a code raised deep in the native stack reaches the app's catch block. This also covers the sync worklet path used by VisionCamera frame processors, which previously had no normalization at all.
- Demo apps: Corrected all places including error handling.
- Rewires error codegen and the CI drift check.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

1. `yarn install && yarn workspace react-native-executorch prepare`
2. `yarn typecheck && yarn lint`, both clean.
3. `yarn codegen:errors` then `git diff --exit-code` on `src/errors/codes.ts` and `cpp/core/error_codes.h`, no output.
4. Run the cv app, classification screen. Tap "Run" repeatedly while a run is still in flight: the extra attempt is dropped silently instead of an error, on both the async and sync button.
5. Kill networking before a model has been cached, then open any screen: the status banner shows the download guidance rather than a raw HTTP string.

### Related issues

#1208

### Checklist

### Additional notes

`jsi::JSError` deliberately passes through `guard` untouched, so an error thrown by an app's own callback is never rewritten with our name and code.


packages/react-native-executorch/cpp/core/conversions.h
```diff
@@ -138,11 +139,11 @@ std::vector<T> fromJsiTypedArray(jsi::Runtime &rt, const std::string &ctx, const
...
 
     if (byteOffset > buffer.size(rt) || byteLength > buffer.size(rt) - byteOffset) {
-        throw jsi::JSError(rt, std::format("{}: out-of-bounds byteOffset ({}) or byteLength ({}) for ArrayBuffer of size {}",
-                                           ctx, byteOffset, byteLength, buffer.size(rt)));
+        throw core::error::CodedError(core::error::ErrorCode::InvalidArgument, std::format("{}: out-of-bounds byteOffset ({}) or byteLength ({}) for ArrayBuffer of size {}",
+                                                                                           ctx, byteOffset, byteLength, buffer.size(rt)));
     }
     if (byteLength % sizeof(T) != 0) {
-        throw jsi::JSError(rt, std::format("{}: byteLength is not a multiple of sizeof(T)={}", ctx, sizeof(T)));
+        throw core::error::CodedError(core::error::ErrorCode::InvalidArgument, std::format("{}: byteLength is not a multiple of sizeof(T)={}", ctx, sizeof(T)));
```
msluszniak: Can't we use scoping of this: `core::error::CodedError(core::error::ErrorCode::InvalidArgument` to shorten this a bit?

packages/react-native-executorch/cpp/core/error.h
```diff
@@ -0,0 +1,116 @@
...
+ * worker threads). `guard` turns it into a coded JavaScript Error at the JSI
+ * boundary, so the code survives all the way to the application's catch block.
+ */
+class CodedError : public std::runtime_error {
+public:
+    CodedError(ErrorCode code, const std::string &message)
+        : std::runtime_error(message), code(code) {}
+
+    CodedError(ErrorCode code, const std::string &message, executorch::runtime::Error etError)
+        : std::runtime_error(message), code(code), etCode(static_cast<int32_t>(etError)) {}
```
msluszniak: Are these classes not explicit because of overload for `char*` message?

packages/react-native-executorch/cpp/core/error.h
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @param code The RNE classification to report the failure as.
+ * @param ctx A short prefix naming the operation, e.g. "loadModel".
+ */
+template <typename T>
+T unwrapEt(ErrorCode code, const std::string &ctx, executorch::runtime::Result<T> result) {
+    if (!result.ok()) {
+        throw CodedError(code,
+                         ctx + ": " + executorch::runtime::to_string(result.error()),
+                         result.error());
```
msluszniak: Wouldn't be `std::format` better here?

packages/react-native-executorch/cpp/core/conversions.h
msluszniak: Yes. `conversions.h` is in `rnexecutorch::core::conversions`, so unqualified `error` already resolves to the sibling `rnexecutorch::core::error`. Dropped the redundant `core::` on all three sites.

packages/react-native-executorch/cpp/core/error.h
msluszniak: No, `char*` is not a factor: `explicit` does not affect argument conversion, only construction of the class itself. It was an oversight. Both are `explicit` now and nothing relied on copy-list-initialization.

packages/react-native-executorch/cpp/core/error.h
msluszniak: Agreed, switched to `std::format` to match the rest of the codebase.

apps/computer-vision/app/classification/index.tsx
barhanc, line 1: I think we should leave the example apps as they were. They are mostly a testing ground for us and we should see the full low-level errors (e.g. even to test this new error handling). Streamlining apps is a scope of another issue #1288 and I'm even thinking if these user-facing example apps showcasing the library shouldn't actually be in a separate repository like `react-native-executorch-examples` which installs rn-executorch from npm, etc. (so it is a full app-dev experience instead of the slightly contrived one we have with the `apps/` directory). Regardless, I think these apps should be mainly for us to test and all errors should be surfaced exactly.

packages/react-native-executorch/cpp/core/error.h
barhanc, line 1: I really like the idea of using a `guarded` wrapper. There are however some rough edges in this implementation. 

1. First of all using `guarded` we should *never* throw `jsi::JSError` anywhere in the C++ code therefore I don't see the purpose of the `throwJs` function here.
2. I feel like to make it even clearer that now we should only use C++ exceptions the `makeJs` should be replaced with something like `inline void throwJsRnExecuTorchError(jsi::Runtime &rt, const RnExecuTorchException &e)`  and the `guard` should look more like 
  ```c++
  try {
      return std::forward<Fn>(fn)();
  } catch (const RnExecuTorchException &e) {
      errors::throwJsiRnExecuTorchError(rt, e);
  } catch (const jsi::JSError &) {
      // Already a JavaScript value, and possibly one thrown by user code
      // called back into. Pass it through untouched.
      throw;
  } catch (const std::exception &e) {
      errors::throwJsiRnExecuTorchError(rt, errors::RnExecuTorchException(errors::RnExecuTorchErrorCode::Unknown, e.what()));
  } catch (...) {
      errors::throwJsiRnExecuTorchError(rt, errors::RnExecuTorchException(errors::RnExecuTorchErrorCode::Unknown, "Unknown native exception occurred"))
  }
  ```
   so there is no way of throwing the error to JS without constructing the C++ exception first.

3. The naming could be a bit clearer. I would prefer `RnExecuTorchException` instead of `CodedError` as the second one doesn't make it immediately clear that this is our custom error and the same for `ErrorCode`. I would also rename `etCode` to something more descriptive e.g. `executorchRuntimeErrorCode` or some variation on that if it is too long. `guarded` is also a bit too generic for my liking, maybe something like `withJsiErrorHandling(fnBody)` or `jsiErrorGuarded(fnBody)` would be better, but I won't push too hard on this one---`error::guarded(fnBody)` has a nice ring to it ;). 
4. I would leave the `unwrap`s as a quick, dirty helper functions instead of putting them under `errors` namespace. They should, however, only be available now with one signature without the `rt` JSI runtime.

packages/react-native-executorch/cpp/core/error_codes.h
barhanc, line 1: I'm really against having the error codes auto-generated from a script. It feels like a solution that goes against the general way we implemented the TS-JSI interface up till now. Throughout the whole library we have clear mirroring of TS and C++ files that implement common interface which is shared implicitly between these two layers. We don't, for example, auto-generate dtypes, schema types, function signatures or host object interfaces from some spec/config file. The pattern is that TS contains the definitions and C++ JSI layer conforms to what TS has. Following this in `error.{h, cpp}` files would mean implementing the C++ equivalent of TS like so (simplifying a bit ofc)
```c++
enum class RnExecuTorchErrorCode {
    LoadFailed,
    ExecutionFailed,
    SchemaMismatch,
    InvalidArgument,
    ResourceUnavailable,
    Unknown
};

constexpr const char* errorCodeToString(RnExecuTorchErrorCode code) {
    switch (code) {
        case RnExecuTorchErrorCode::LoadFailed:          return "LOAD_FAILED";
        case RnExecuTorchErrorCode::ExecutionFailed:     return "EXECUTION_FAILED";
        case RnExecuTorchErrorCode::SchemaMismatch:      return "SCHEMA_MISMATCH";
        case RnExecuTorchErrorCode::InvalidArgument:     return "INVALID_ARGUMENT";
        case RnExecuTorchErrorCode::ResourceUnavailable: return "RESOURCE_UNAVAILABLE";
        default:                                         return "UNKNOWN";
    }
}

class RnExecuTorchException : public std::runtime_error {
   explicit RnExecuTorchException(RnExecuTorchErrorCode code, const std::string &message)
        : std::runtime_error(message),
          code_(code) {}

    explicit RnExecuTorchException(ErrorCode code, const std::string &message, executorch::runtime::Error etError)
        : std::runtime_error(message), 
          code_(code), 
          etRuntimeErrorCode_(static_cast<int32_t>(etError)) {}

    RnExecuTorchErrorCode code_;
    std::optional<int32_t> etRuntimeErrorCode_;
};

inline void throwJsiRnExecuTorchError(jsi::Runtime &rt, const RnExecuTorchException &e) {
    auto errorCtor = rt.global().getPropertyAsFunction(rt, "Error");
    auto errObj = errorCtor.call(rt, jsi::String::createFromUtf8(rt, e.what())).asObject(rt);

    errObj.setProperty(rt, "name", jsi::String::createFromUtf8(rt, "RnExecuTorchError"));
    errObj.setProperty(rt, "code", jsi::String::createFromUtf8(rt, errorCodeToString(e.code_)));

    if (e.etRuntimeError_.has_value()) {
        errObj.setProperty(rt, "etRuntimeErrorCode", *e.etRuntimeErrorCode_);
    }

    throw jsi::JSError(rt, std::move(errObj));
}
```

packages/react-native-executorch/src/errors/codes.ts
barhanc, line 1: Continuing the comment from `error_codes.h`, we should just have `error.ts` inside `core/` that would define the error codes and the custom error class. I think using type union is more idiomatic than enum here, so I would rather see something like (screaming case or pascal case)
```ts
const VALID_ERROR_CODES = [
  'LOAD_FAILED',
  'EXECUTION_FAILED',
  'SCHEMA_MISMATCH',
  'INVALID_ARGUMENT',
  'INVALID_STATE',
  'RESOURCE_DISPOSED',
  'RESOURCE_BUSY',
  'DOWNLOAD_FAILED',
  'DOWNLOAD_ABORTED',
  'UNKNOWN',
] as const;

export type RnExecuTorchErrorCode = (typeof VALID_ERROR_CODES)[number];
```
Also I believe some of the error codes are too specific (e.g. the tokenizer or not supported errors). Imo the specific errors should only be for cases where the user genuinely can take some action upon fail to recover (good examples are download errors or resource errors). Other errors can  just be general categories for tools like Sentry to group them (e.g. execution or load errors are generally non-recoverable so subdividing them doesn't provide anything on top of the error message). But this can be discussed later as these are details.

packages/react-native-executorch/src/errors/error.ts
barhanc, line 1: This design with both class `RnExecuTorchError` and `rnExecuTorchError()` is flaky. Since we cannot really use classes due to how worklets work, I think we must do something like this
```ts
export type RnExecuTorchError<C extends RnExecuTorchErrorCode = RnExecuTorchErrorCode> = Error & {
  name: 'RnExecuTorchError';
  code: C;
  etRuntimeErrorCode?: number;
};

export function RnExecuTorchError<C extends RnExecuTorchErrorCode>(
  code: C,
  message: string,
  etRuntimeErrorCode?: number
): RnExecuTorchError<C> {
  'worklet';
  const err = new Error(message) as RnExecuTorchError<C>;
  err.name = 'RnExecuTorchError';
  err.code = code;

  if (etRuntimeErrorCode !== undefined) {
    err.etRuntimeErrorCode = etRuntimeErrorCode;
  }

  return err;
}

export function isRnExecuTorchError<C extends RnExecuTorchErrorCode = RnExecuTorchErrorCode>(
  err: unknown,
  code?: C
): err is RnExecuTorchError<C> {
  'worklet';
  if (err === null || typeof err !== 'object') {
    return false;
  }
  if (!('name' in err) || err.name !== 'RnExecuTorchError') {
    return false;
  }
  if (!('code' in err) || typeof err.code !== 'string') {
    return false;
  }
  if (!(VALID_ERROR_CODES as readonly string[]).includes(err.code)) {
    return false;
  }
  return code === undefined || err.code === code;
}
```
and in `runtime.ts`
```ts
export function wrapAsync<Args extends any[], R>(
  fn: (...args: Args) => R,
  runtime: WorkletRuntime = defaultWorkletRuntime
) {
  return async (...args: Args): Promise<R> => {
    const result = await runOnRuntimeAsync(
      runtime,
      (argsArray) => {
        'worklet';
        try {
          return { ok: true, value: fn(...argsArray) };
        } catch (e: any) {
          let error;
          if (isRnExecuTorchError(e)) {
            error = {
              name: 'RnExecuTorchError',
              code: e.code,
              message: e.message,
              etRuntimeErrorCode: e.etRuntimeErrorCode,
            };
          } else {
            error = e?.message ?? String(e);
          }
          return { ok: false, error };
        }
      },
      args
    );

    if (!result.ok) {
      if (isRnExecuTorchError(result.error)) {
        throw RnExecuTorchError(
          result.error.code,
          result.error.message,
          result.error.etRuntimeErrorCode
        );
      }
      throw new Error(result.error);
    }
    return result.value!;
  };
}
```
This way we can just throw `throw RnExecuTorchError(code, message)` everywhere.

barhanc (commented):
Only top level general comments for now as I want to first discuss the overall premises of the implementation before diving into details.

apps/computer-vision/app/classification/index.tsx
msluszniak: Agreed. I changed them mainly to test if the new flow works. But I like the idea of spliting examples into two categories and making it in one specialized PR.

packages/react-native-executorch/cpp/core/error.h
msluszniak: Ok, agreed with all.

packages/react-native-executorch/cpp/core/error_codes.h
msluszniak: Indeed, this would be the only one codegened part in the repository. I took it from the old flow where errorcodes were also generated.

packages/react-native-executorch/src/errors/codes.ts
msluszniak: Regarding nunber of error codes I had a dillema. Now there are 13 of them, but indeed I though to make this group much smaller, e.g. 5. And your argumentation about being too specific without real gains works for me. Regarding type union vs enum let's move with TU then.

packages/react-native-executorch/src/errors/error.ts
msluszniak: Yeah, I remembered that in worklets, we cannot use regular Errors and asked clanker to propose some solution that bypass this, but I agree that your solution is better

msluszniak:
Agreed on all changes; I'll implement them after the weekend ;)

barhanc (commented):
Looks good. Some header files have outdated docs regarding what they throw, so this needs to be fixed (e.g. `tensor_helpers.h`). Other than that one additional comment regarding C++ error syntax sugar. I've tested the error handling on a simple example screen. It's on branch [@bh/rewrite-error-handling-test](https://github.com/software-mansion/react-native-executorch/tree/%40bh/rewrite-error-handling-test) if you'd need it.

barhanc (commented):
Looks good. The call sites are a bit long though, so I was thinking about maybe adding factory methods so we could do `RnExecuTorchException::InvalidArgument(msg, etCode)` (also I'm thinking whether `RnExecuTorchException` or `...Error` is better) or even `error::InvalidArgument(msg, etCode)`. Doing this cleanly would require some x-macro magic though I'm afraid, so I leave it to you to decide if it's worth it or should we stay with explicit callsites. The x-macro magic in question:
```c++
// NOLINTNEXTLINE(cppcoreguidelines-macro-usage): X-macro is used for concise error code enum, string mapping, and factory method definitions
#define FOR_ALL_RNEXECUTORCH_ERROR_CODES(V)  \
    V(LoadFailed, "LOAD_FAILED")             \
    V(ExecutionFailed, "EXECUTION_FAILED")   \
    V(SchemaMismatch, "SCHEMA_MISMATCH")     \
    V(InvalidArgument, "INVALID_ARGUMENT")   \
    V(InvalidState, "INVALID_STATE")         \
    V(ResourceDisposed, "RESOURCE_DISPOSED") \
    V(ResourceBusy, "RESOURCE_BUSY")         \
    V(DownloadFailed, "DOWNLOAD_FAILED")     \
    V(DownloadAborted, "DOWNLOAD_ABORTED")   \
    V(Unknown, "UNKNOWN")

enum class RnExecuTorchErrorCode {
// NOLINTNEXTLINE(cppcoreguidelines-macro-usage): helper macro for X-macro expansion
#define DEFINE_ENUM(name, str) name,
    FOR_ALL_RNEXECUTORCH_ERROR_CODES(DEFINE_ENUM)
#undef DEFINE_ENUM
};

constexpr const char *errorCodeToString(RnExecuTorchErrorCode code) {
    switch (code) {
// NOLINTNEXTLINE(cppcoreguidelines-macro-usage): helper macro for X-macro expansion
#define DEFINE_CASE(name, str)        \
    case RnExecuTorchErrorCode::name: \
        return str;
        FOR_ALL_RNEXECUTORCH_ERROR_CODES(DEFINE_CASE)
#undef DEFINE_CASE
    }
    return "UNKNOWN";
}

// Alternative 1: Static Factory Methods on Exception
class RnExecuTorchException : public std::runtime_error {
// ...
// NOLINTNEXTLINE(cppcoreguidelines-macro-usage): helper macro for X-macro expansion
#define DEFINE_FACTORY(name, str)                                                                         \
    static RnExecuTorchException name(const std::string &message,                                         \
                                      std::optional<executorch::runtime::Error> etError = std::nullopt) { \
        return etError ? RnExecuTorchException(RnExecuTorchErrorCode::name, message, *etError)            \
                       : RnExecuTorchException(RnExecuTorchErrorCode::name, message);                     \
    }
    FOR_ALL_RNEXECUTORCH_ERROR_CODES(DEFINE_FACTORY)
#undef DEFINE_FACTORY
// ...
}

// Alternative 2: Namespace-Level Helper Functions
// NOLINTNEXTLINE(cppcoreguidelines-macro-usage): helper macro for X-macro expansion
#define DEFINE_NAMESPACE_FACTORY(name, str)                                                               \
    inline RnExecuTorchException name(const std::string &message,                                         \
                                      std::optional<executorch::runtime::Error> etError = std::nullopt) { \
        return etError ? RnExecuTorchException(RnExecuTorchErrorCode::name, message, *etError)            \
                       : RnExecuTorchException(RnExecuTorchErrorCode::name, message);                     \
    }
FOR_ALL_RNEXECUTORCH_ERROR_CODES(DEFINE_NAMESPACE_FACTORY)
#undef DEFINE_NAMESPACE_FACTORY
```

msluszniak:
@barhanc 

Enum and errorCodeToString were two hand-written lists with a default: between them, so a missing case silently reached JS as UNKNOWN. The X-macro makes that impossible.

Went with namespace-level factories: shorter call sites, and they drop the two using declarations, leaving just the namespace error = ... alias needed for error::guarded.

I prefer Exception because it's the correct term and distinguishes it from the JS error.

I fixed all stale header comments. That sweep also found two unreachable catch (const std::invalid_argument &) blocks in box_ops.cpp.

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -106,30 +103,30 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
 
             if (!self->etModule_) {
-                throw jsi::JSError(rt, "execute: Model has been disposed");
+                throw error::ResourceDisposed("execute: Model has been disposed");
             }
 
             auto methodName = conversions::asType<std::string>(rt, "execute: methodName", args[0]);
             if (!self->spec_.contains(methodName)) {
-                throw jsi::JSError(rt, std::format("execute: Unknown method '{}'", methodName));
+                throw error::SchemaMismatch(std::format("execute: Unknown method '{}'", methodName));
```
barhanc: Shouldn't this be `InvalidArgument`?

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -106,30 +103,30 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
...
             }
             const auto &methodSpec = self->spec_.at(methodName);
 
             auto inputsArray = conversions::asType<jsi::Array>(rt, "execute: inputs", args[1]);
             auto outputTensorsArray = conversions::asType<jsi::Array>(rt, "execute: outputTensors", args[2]);
 
             if (inputsArray.size(rt) != methodSpec.inputs.size()) {
-                throw jsi::JSError(rt, std::format("execute: Incorrect size for inputs of method '{}': got {}, expected {}",
-                                                   methodName, inputsArray.size(rt), methodSpec.inputs.size()));
+                throw error::SchemaMismatch(std::format("execute: Incorrect size for inputs of method '{}': got {}, expected {}",
```
barhanc: `InvalidArgument`?

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -169,8 +166,8 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
                     inputs[i] = executorch::runtime::EValue();
                     break;
                 default:
-                    throw jsi::JSError(rt, std::format("{}: Unsupported input type: {}",
-                                                       ctx, executorch::runtime::tag_to_string(tag)));
+                    throw error::SchemaMismatch(std::format("{}: Unsupported input type: {}",
```
barhanc: `InvalidArgument`?

packages/react-native-executorch/cpp/core/model.cpp
```diff
@@ -262,36 +260,36 @@ jsi::Value ModelHostObject::get(jsi::Runtime &rt, const jsi::PropNameID &name) {
                     jsOutputArray.setValueAtIndex(rt, outputIdx, jsi::String::createFromUtf8(rt, std::string(output.toString())));
                     break;
                 default:
-                    throw jsi::JSError(rt, std::format("execute: Unsupported return type: {}",
-                                                       executorch::runtime::tag_to_string(output.tag)));
+                    throw error::SchemaMismatch(std::format("execute: Unsupported return type: {}",
```
barhanc: `ExecutionFailed`?

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.cpp
```diff
@@ -62,19 +67,19 @@ TokenizerHostObject::TokenizerHostObject(std::string tokenizerPath)
       tokenizer_(std::make_unique<tokenizers::HFTokenizer>()) {
     auto error = tokenizer_->load(tokenizerPath_);
     if (error != tokenizers::Error::Ok) {
-        throw std::runtime_error(std::format("Failed to load tokenizer from '{}': {}",
-                                             tokenizerPath_, toString(error)));
+        throw error::ExecutionFailed(std::format("Failed to load tokenizer from '{}': {}",
```
barhanc: `LoadFailed`?

packages/react-native-executorch/cpp/extensions/nlp/tokenizer.h
```diff
@@ -25,7 +25,8 @@ class TokenizerHostObject : public facebook::jsi::HostObject,
      * Constructs a TokenizerHostObject by loading a HuggingFace tokenizer configuration file.
      *
      * @param tokenizerPath File system path to the tokenizer configuration file.
-     * @throws std::runtime_error If loading the tokenizer fails.
+     * @throws core::error::RnExecuTorchException with code ExecutionFailed if
```
barhanc:
```suggestion
     * @throws core::error::RnExecuTorchException with code LoadFailed if
```

packages/react-native-executorch/src/core/schema.ts
```diff
@@ -780,17 +854,23 @@ export function validateSpec<const T extends Record<string, ModelSpec<SymbolicDi
       const dimFn = (name: string, kind?: string): any => {
         const dim = bindings.get(name);
         if (!dim) {
-          throw new Error(`Symbol '${name}' not found in bindings.`);
+          throw RnExecuTorchError('SCHEMA_MISMATCH', `Symbol '${name}' not found in bindings.`);
```
barhanc: I think these should be `'INVALID_ARGUMENT'` throughout this `dimFn` function.

barhanc (approved):
Just some small nits regarding thrown error codes. Other than that ready to ship 🚀

PR #1364, opened 2026-08-17
[RNE Rewrite] feat: add Kokoro TTS pipeline implementation

### LINKED ISSUE #1250: [RNE Rewrite] Speech - add Kokoro/Supertonic TTS pipeline implementation

IgorSwat:
## Description

What this PR adds:
- Implementation of the Kokoro (text-to-speech) pipeline for one-way streaming (single input, streamed output)
- Implementation for boolean DType
- Demo app for Kokoro model

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

Run the dedicated demo app.

### Checklist

### Additional notes

The implementation of the Kokoro pipeline follows strictly the one already done for Supertonic. For this purpose, it got simplified a bit, with the following changes:
- Duration predictor no longer has 3 specialized input-size methods: instead of 'forward_32', 'forward_64' and 'forward_128' we use one 'forward' for the entire available token range. Costs a little bit worse performance, but gains reduced memory usage.
- Text partitioning and phonemization are applied in a simple, sequential order. In old Kokoro pipeline they would interleave themselves. In most practical cases it is a meaningless change, although if user passes a very long text in a language where Phonemis does not provide a lexicon (currently only English pipeline does), it can be significant for latency.


packages/react-native-executorch/cpp/core/dtype.cpp
```diff
@@ -17,7 +17,10 @@ DType dtypeFromString(const std::string &s) {
     if (s == "float32") {
         return DType::float32;
     }
-    throw error::InvalidArgument("Unsupported dtype: '" + s + "'. Expected 'uint8', 'int32', 'int64', or 'float32'");
+    if (s == "bool") {
+        return DType::boolean;
+    }
+    throw error::InvalidArgument("Unsupported dtype: '" + s + "'. Expected 'uint8', 'int32', 'int64', 'float32' or 'bool'");
```
msluszniak: I know this was already there, but please use `std::format` here instead.

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,113 @@
...
+    auto nameStr = name.utf8(rt);
+
+    if (nameStr == "phonemize") {
+        auto self = shared_from_this();
+        auto fnBody = [self](jsi::Runtime &rt, const jsi::Value &,
+                             const jsi::Value *args,
+                             size_t count) -> jsi::Value {
+            if (count < 1 || !args[0].isString()) {
+                throw jsi::JSError(rt, "Phonemizer.phonemize: Usage: phonemize(text)");
+            }
```
msluszniak: I don't think throwing `jsi::JSError` is correct anymore since we introduced our own error in the PR merged few days ago. Please change all places where applicable.

packages/react-native-executorch/cpp/extensions/speech/phonemizer.h
```diff
@@ -0,0 +1,29 @@
...
+#include <jsi/jsi.h>
+#include <memory>
+#include <phonemis/base/pipeline.h>
+
+namespace rnexecutorch::extensions::speech {
+
+class PhonemizerHostObject : public facebook::jsi::HostObject,
+                             public std::enable_shared_from_this<PhonemizerHostObject> {
+public:
+    PhonemizerHostObject(const std::string &lang,
```
msluszniak: Why not explicit?

packages/react-native-executorch/src/core/schema.ts
```diff
@@ -345,6 +345,7 @@ export const f32 = (...shape: SymbolicShape) => SymbolicTensor('float32', shape)
 export const i64 = (...shape: SymbolicShape) => SymbolicTensor('int64', shape);
 export const i32 = (...shape: SymbolicShape) => SymbolicTensor('int32', shape);
 export const ui8 = (...shape: SymbolicShape) => SymbolicTensor('uint8', shape);
+export const bl = (...shape: SymbolicShape) => SymbolicTensor('bool', shape);
```
msluszniak: Maybe `bool` instead of `bl`, wdyt @barhanc?

IgorSwat: Changed to `bool`.

packages/react-native-executorch/android/CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ file(GLOB MATH_SOURCES   ${CPP_DIR}/extensions/math/*.cpp)
 file(GLOB NLP_SOURCES    ${CPP_DIR}/extensions/nlp/*.cpp)
 file(GLOB SPEECH_SOURCES ${CPP_DIR}/extensions/speech/*.cpp)
 file(GLOB OPENCV_SOURCES ${CPP_DIR}/extensions/cv/*.cpp)
+set(PHONEMIS_SOURCES ${CPP_DIR}/extensions/speech/phonemizer.cpp)
+
+if(NOT RNE_ENABLE_PHONEMIS)
+  list(REMOVE_ITEM SPEECH_SOURCES ${PHONEMIS_SOURCES})
+endif()
```
barhanc: I'm not sure how I feel about not allowing any speech functionalities if phonemis is not available, since it's only required for Kokoro (and the `#ifdef` guard is only local). Maybe there is some way to add more granularity so that other speech functionalities could be still available without phonemis? cc: @msluszniak

packages/react-native-executorch/cpp/core/dtype.cpp
```diff
@@ -64,6 +75,7 @@ DType dtypeFromScalarType(executorch::aten::ScalarType st) {
 size_t elementSize(DType dtype) {
     switch (dtype) {
     case DType::uint8:
+    case DType::boolean:
```
barhanc, line 78: Let's add explicit duplicated `case` like for `int32` / `float32`.

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+              .lang = lang,
+              .lexicon_filepath = lexiconPath.empty()
+                                      ? std::nullopt
+                                      : std::make_optional(lexiconPath),
+              .nn_model_filepath = neuralModelPath.empty()
+                                       ? std::nullopt
+                                       : std::make_optional(neuralModelPath),
+          }})) {}
+
+jsi::Value PhonemizerHostObject::get(jsi::Runtime &rt,
```
barhanc, line 50: Any reason this host object is not thread safe?

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+void install_createPhonemizer(jsi::Runtime &rt, jsi::Object &module) {
+    const auto *name = "createPhonemizer";
+    auto fnBody = [](jsi::Runtime &rt, const jsi::Value &,
+                     const jsi::Value *args, size_t count) -> jsi::Value {
+        if (count != 4) {
+            throw error::InvalidArgument("createPhonemizer: Usage: createPhonemizer(lang, "
+                                         "taggerPath, lexiconPath, neuralPath)");
+        }
```
barhanc: Maybe use an `options` object instead of inlined arguments for `createPhonemizer`, seems more appropriate in this case.

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+                lang, taggerPath, lexiconPath, neuralPath);
+            return jsi::Object::createFromHostObject(rt, instance);
+        } catch (const std::exception &e) {
+            throw error::LoadFailed(std::format("createPhonemizer: {}", e.what()));
+        }
+    };
+
+    module.setProperty(rt, name,
+                       jsi::Function::createFromHostFunction(
+                           rt, jsi::PropNameID::forAscii(rt, name), 4, error::guarded(fnBody)));
```
barhanc:
```suggestion
    module.setProperty(rt, name, jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, name), 4, error::guarded(fnBody)));
```

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        auto lang = conversions::asType<std::string>(rt, "createPhonemizer: lang", args[0]);
+        auto taggerPath = conversions::asType<std::string>(rt, "createPhonemizer: taggerPath", args[1]);
+        auto lexiconPath = conversions::asType<std::string>(rt, "createPhonemizer: lexiconPath", args[2]);
+        auto neuralPath = conversions::asType<std::string>(rt, "createPhonemizer: neuralPath", args[3]);
+
+        try {
+            auto instance = std::make_shared<PhonemizerHostObject>(
+                lang, taggerPath, lexiconPath, neuralPath);
```
barhanc:
```suggestion
            auto instance = std::make_shared<PhonemizerHostObject>(lang, taggerPath, lexiconPath, neuralPath);
```

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+jsi::Value PhonemizerHostObject::get(jsi::Runtime &rt,
+                                     const jsi::PropNameID &name) {
+    auto nameStr = name.utf8(rt);
+
+    if (nameStr == "phonemize") {
+        auto self = shared_from_this();
+        auto fnBody = [self](jsi::Runtime &rt, const jsi::Value &,
+                             const jsi::Value *args,
+                             size_t count) -> jsi::Value {
```
barhanc:
```suggestion
        auto fnBody = [self](jsi::Runtime &rt, const jsi::Value &, const jsi::Value *args, size_t count) -> jsi::Value {
```

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+        };
+        return jsi::Function::createFromHostFunction(
+            rt, jsi::PropNameID::forAscii(rt, "phonemize"), 1, error::guarded(fnBody));
+    }
+
+    if (nameStr == "dispose") {
+        auto self = shared_from_this();
+        auto fnBody = [self](jsi::Runtime & /*rt*/, const jsi::Value &,
+                             const jsi::Value * /*args*/,
+                             size_t count) -> jsi::Value {
```
barhanc:
```suggestion
        auto fnBody = [self](jsi::Runtime & /*rt*/, const jsi::Value &, const jsi::Value * /*args*/, size_t count) -> jsi::Value {
```

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
```diff
@@ -0,0 +1,120 @@
...
+            if (count != 0) {
+                throw error::InvalidArgument("dispose: Usage: dispose()");
+            }
+            self->pipeline_.reset();
+            return jsi::Value::undefined();
+        };
+        return jsi::Function::createFromHostFunction(
+            rt, jsi::PropNameID::forAscii(rt, "dispose"), 0, error::guarded(fnBody));
```
barhanc:
```suggestion
        return jsi::Function::createFromHostFunction(rt, jsi::PropNameID::forAscii(rt, "dispose"), 0, error::guarded(fnBody));
```

packages/react-native-executorch/cpp/extensions/speech/phonemizer.h
```diff
@@ -0,0 +1,29 @@
...
+public:
+    explicit PhonemizerHostObject(const std::string &lang,
+                                  const std::string &taggerPath,
+                                  const std::string &lexiconPath,
+                                  const std::string &neuralModelPath);
+
+    facebook::jsi::Value get(facebook::jsi::Runtime &rt,
+                             const facebook::jsi::PropNameID &name) override;
+    std::vector<facebook::jsi::PropNameID> getPropertyNames(
+        facebook::jsi::Runtime &rt) override;
```
barhanc:
```suggestion
    facebook::jsi::Value get(facebook::jsi::Runtime &rt, const facebook::jsi::PropNameID &name) override;
    std::vector<facebook::jsi::PropNameID> getPropertyNames(facebook::jsi::Runtime &rt) override;
```

packages/react-native-executorch/cpp/extensions/speech/phonemizer.h
```diff
@@ -0,0 +1,29 @@
...
+
+private:
+    std::unique_ptr<phonemis::Pipeline> pipeline_;
+};
+
+void install_createPhonemizer(facebook::jsi::Runtime &rt,
+                              facebook::jsi::Object &module);
```
barhanc:
```suggestion
void install_createPhonemizer(facebook::jsi::Runtime &rt, facebook::jsi::Object &module);
```

packages/react-native-executorch/src/extensions/speech/utils/phonemizer.ts
```diff
@@ -0,0 +1,59 @@
+import { rnexecutorchJsi } from '../../../native/bridge';
+import { RnExecuTorchError } from '../../../core/error';
+
+declare const phonemizerBrand: unique symbol;
+
+/**
+ * List of all (currently) supported languages.
+ */
+export type Language = 'en-us' | 'en-gb' | 'fr' | 'es' | 'it' | 'pt' | 'de' | 'pl' | 'hi';
```
barhanc: Rename this to `KokoroLanguage` or even better use the pattern with defining both a constant array of supported languages and a type derived from it as `(typeof KOKORO_SUPPORTED_LANGUAGES)[number] `.

packages/react-native-executorch/src/extensions/speech/utils/phonemizer.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export type Phonemizer = {
+  /**
+   * A standard G2P (grapheme to phoneme) utility.
+   * @param text Input text to be phonemized.
+   */
+  phonemize(text: string): string;
+
+  dispose(): void;
```
barhanc, line 30: Missing jsdoc.

packages/react-native-executorch/src/extensions/speech/utils/phonemizer.ts
```diff
@@ -0,0 +1,59 @@
...
+      'INVALID_STATE',
+      "createPhonemizer: The native build has no phonemizer. Add the 'textToSpeech' feature (or " +
+        "the 'phonemis' lib) to the app's react-native-executorch config and rebuild."
+    );
+  }
+  return rnexecutorchJsi.speech.createPhonemizer(
+    config.lang,
+    config.taggerSource ?? '',
+    config.lexiconSource ?? '',
+    config.neuralModelSource ?? ''
```
barhanc: You should pass the config object directly to JSI and use `getOptionalProperty` inside C++.

packages/react-native-executorch/src/extensions/speech/utils/phonemizer.ts
```diff
@@ -0,0 +1,59 @@
+import { rnexecutorchJsi } from '../../../native/bridge';
+import { RnExecuTorchError } from '../../../core/error';
+
+declare const phonemizerBrand: unique symbol;
+
+/**
+ * List of all (currently) supported languages.
```
barhanc:
```suggestion
 * List of all (currently) supported languages by Kokoro TTS.
```

packages/react-native-executorch/src/models.ts
```diff
@@ -1427,5 +1510,22 @@ export const models = {
...
+    KOKORO: {
+      EN_US: KOKORO_EN_US,
+      EN_GB: KOKORO_EN_GB,
+      ES: KOKORO_ES,
+      FR: KOKORO_FR,
+      IT: KOKORO_IT,
+      PT: KOKORO_PT,
+      HI: KOKORO_HI,
+      PL: KOKORO_PL,
+      DE: KOKORO_DE,
```
barhanc: Add backend nesting, even if only single backend is supported.

packages/react-native-executorch/src/models.ts
```diff
@@ -770,6 +771,88 @@ const SUPERTONIC_3_MLX_FP32: SupertonicTtsModel<SupertonicDefaultVoiceName> = {
...
+// English relies on a part-of-speech tagger and a pronunciation lexicon; the
+// remaining languages are phonemized by a neural grapheme-to-phoneme model.
+const kokoroEnglishPhonemizer = (lang: 'en-us' | 'en-gb') => ({
+  lang,
+  taggerSource: `${KOKORO_PHONEMIZER_ROOT}/${lang}/tags.json`,
+  lexiconSource: `${KOKORO_PHONEMIZER_ROOT}/${lang}/lexicon.json`,
+  neuralModelSource: `${KOKORO_PHONEMIZER_ROOT}/${lang}/phonemizer_${lang.replace('-', '_')}.pte`,
+});
+
+const kokoroNeuralPhonemizer = <const L extends 'fr' | 'es' | 'it' | 'pt' | 'hi' | 'pl' | 'de'>(
```
barhanc: Instead of listing languages use `Exclude<KokoroLanguage, ...>`

packages/react-native-executorch/src/hooks/useKokoroTextToSpeech.ts
barhanc, line 1: Instead of defining new hook, let's modify the `useTextToSpeech` so that it takes additional option, like `...: 'kokoro' | 'suprtonic'` and add overloaded definitions
```ts
useTextToSpeech<K extends PropertyKey>(config: KokoroTtsModel<K>, ...: 'kokoro', options?: ResourceOptions);
useTextToSpeech<K extends PropertyKey>(config: SuperonicTtsModel<K>, ...;

```

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
```diff
@@ -0,0 +1,411 @@
...
+
+  const [predictorTokens] = predictorSpec.dims.range('T');
+  const [synthesizerTokens, durations] = synthesizerSpec.dims.range('T', 'D');
+
+  if (predictorTokens.max !== synthesizerTokens.max) {
+    throw RnExecuTorchError(
+      'SCHEMA_MISMATCH',
+      'createKokoroTextToSpeech: incompatible duration predictor and synthesizer token limits.'
+    );
+  }
```
barhanc: Not sure this is needed, it only checks the upper bound of the range, which is neither the necessary nor sufficient condition for two range dims to be compatible. If these were not compatible C++ runtime will throw descriptive error when validating shapes of input tensors against the synthesizer's schema.

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
```diff
@@ -0,0 +1,411 @@
...
+  };
+
+  const synthesizeChunkWorklet = (
+    chunkPhonemes: string,
+    chunkOpts: { voice: K; speed: number }
+  ): { audio: Float32Array; sampleRate: number; duration: number } => {
+    'worklet';
+
+    const phonemes = Array.from(chunkPhonemes.trim());
+    const voice = parsedVoices[chunkOpts.voice]!;
```
barhanc: The `!` shouldn't be needed.

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
```diff
@@ -0,0 +1,411 @@
...
+} from '../utils/kokoroUtils';
+
+/**
+ * Kokoro audio sampling rate in Hz (24000).
+ * @category Constants
+ */
+export const KOKORO_SAMPLE_RATE = 24000;
+
+const SAMPLES_PER_MS = KOKORO_SAMPLE_RATE / 1000;
+const VOICE_REF_HALF_SIZE = VOICE_REF_SIZE / 2;
```
barhanc, line 38: Why can't we inline this constant?

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
```diff
@@ -0,0 +1,411 @@
...
+    } finally {
+      auxTensors.forEach((t) => t.dispose());
+    }
+  };
+
+  const synthesizeChunk = wrapAsync(synthesizeChunkWorklet, runtime);
+  const phonemize = wrapAsync((text: string) => {
+    'worklet';
+    return phonemizer.phonemize(text);
+  }, runtime);
```
barhanc: How long does phonemization take? Note that there is a considerable overhead for sending data between JS runtimes boundary.

packages/react-native-executorch/src/extensions/speech/utils/kokoroUtils.ts
```diff
@@ -0,0 +1,233 @@
...
+ * Checks whether a phoneme is a letter (as opposed to punctuation or a
+ * suprasegmental marker).
+ * @category Utils
+ * @param phoneme The phoneme to test.
+ * @returns True when the phoneme is a unicode letter.
+ */
+export function isLetter(phoneme: string): boolean {
+  'worklet';
+  return LETTER_PATTERN.test(phoneme);
+}
```
barhanc: Both the function and the constant can be inlined inside that one line where they are used.

packages/react-native-executorch/src/extensions/speech/utils/kokoroUtils.ts
```diff
@@ -0,0 +1,233 @@
...
+ * Returns the length of the silence to append after a chunk ending with the
+ * given phoneme.
+ * @category Utils
+ * @param phoneme The last phoneme of a chunk.
+ * @returns The pause length in milliseconds.
+ */
+export function pauseAfter(phoneme: string): number {
+  'worklet';
+  return PAUSE_MS[phoneme] ?? 0;
+}
```
barhanc: Not sure we need a function for this, can just be inlined.

packages/react-native-executorch/src/extensions/speech/utils/kokoroUtils.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+}
+
+/**
+ * Repeats each token index as many times as its predicted duration.
+ * @category Utils
+ * @param durations The per-token durations.
+ * @returns Token indices ready to be written into an `int64` tensor.
+ */
+export function repeatInterleave(durations: Int32Array): BigInt64Array {
```
barhanc: This is a pretty general helper, perhaps should be in `math.ts` with more generic argument names?

packages/react-native-executorch/android/CMakeLists.txt
msluszniak: Yeap, agreed.

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
IgorSwat: Because Phonemis is not thread-safe itself.

packages/react-native-executorch/src/extensions/speech/utils/phonemizer.ts
IgorSwat: I renamed it to `PhonemizerLanguage`, because it's not strictly the Kokoro language, as Phonemis can be used with any future TTS model which requires g2p.

packages/react-native-executorch/cpp/extensions/speech/phonemizer.cpp
barhanc: All the more so, there should be a mutex preventing concurrent calls to the `phonemize` method. If Phonemis phonemizer was actually thread-safe then locking wouldn't be needed.

msluszniak:
Example app looks solid, everything worked correctly. I will do a second pass on this one and I will be good with this PR :))

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    const maxChunkLength = Math.min(options.maxChunkLength ?? maxTokens - 2, maxTokens - 2);
+
+    // Phonemize once up front, then partition the phonemes — every chunk is
+    // then guaranteed to fit the models' token limit.
+    const phonemes = options.phonemize === false ? text : await phonemize(text);
+    const chunks = partition(phonemes, maxChunkLength, { prioritizeInitialTtfa: true });
+
+    isSynthesizing = true;
```
msluszniak: `isSynthesizing` is checked at the top but only set here, after `await phonemize(text)`. So a `synthesizeStop()` during phonemization gets dropped, and two `synthesize()` calls in the same tick both pass the guard, then the first one to finish flips the flag and cuts the other off. Setting it before the await fixes both.

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
```diff
@@ -0,0 +1,411 @@
...
+      'SCHEMA_MISMATCH',
+      'createKokoroTextToSpeech: incompatible duration predictor and synthesizer token limits.'
+    );
+  }
+
+  const minTokens = Math.max(predictorTokens.min, synthesizerTokens.min);
+  const maxTokens = predictorTokens.max;
+  const maxDurationTicks = durations.max;
+
+  const phonemizer = await wrapAsync(createPhonemizer, runtime)(config.phonemizer);
```
msluszniak: Both `.pte` models are already loaded at this point, so if `createPhonemizer` or a voice read throws we leak them. Worth wrapping the rest of the setup in a try/catch that disposes them before rethrowing, especially since the missing-phonemizer path is a documented failure mode.

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
IgorSwat: It depends. Usually under 10 ms for typical inputs, but if the text is long and language does not have built-in lexicons, then it can take much longer.

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
IgorSwat: We can, but exposing them like that is more elegant - at least that's what I think.

packages/react-native-executorch/cpp/core/dtype.cpp
msluszniak: @IgorSwat you missed adding similar line as this one: `// NOLINTNEXTLINE(bugprone-branch-clone): int32 and float32 are both 4 bytes; the identical branches are intentional.` Other way is to explicitly disable this check, I'm wondering what is better here.

packages/react-native-executorch/src/extensions/speech/tasks/kokoroTextToSpeech.ts
barhanc: Ok, then let's leave it async, just change it to 
```ts
const phonemize = wrapAsync(phonemizer.phonemize, runtime);
```
With the way we implement native host object methods (the `shared_from_this` trick) the auxiliary lambda shouldn't be needed

packages/react-native-executorch/src/extensions/math.ts
```diff
@@ -117,3 +118,44 @@ export function randomNormal(
...
+    throw RnExecuTorchError(
+      'INVALID_ARGUMENT',
+      `repeatInterleave: repeats length (${repeats.length}) must match values length (${values.length}).`
+    );
+  }
+
+  let total = 0;
+  for (let i = 0; i < repeats.length; i++) total += Math.max(0, repeats[i]!);
+
+  const out = new (values.constructor as { new (length: number): T })(total);
```
barhanc:
```suggestion
  const Ctor = values.constructor as new (length: number) => T;
  const out = new Ctor(total);
  const target = out as Record<number, number | bigint>;
```

packages/react-native-executorch/src/extensions/math.ts
```diff
@@ -117,3 +118,44 @@ export function randomNormal(
...
+  let next = 0;
+  for (let i = 0; i < values.length; i++) {
+    const value = values[i]!;
+    for (let j = 0; j < repeats[i]!; j++) (out as { [index: number]: unknown })[next++] = value;
```
barhanc:
```suggestion
    const value = values[i]!;
    const count = Math.max(0, repeats[i]!);
    
    for (let j = 0; j < count; j++) target[next++] = value;
```

packages/react-native-executorch/src/extensions/speech/utils/phonemizer.ts
```diff
@@ -0,0 +1,55 @@
+import { rnexecutorchJsi } from '../../../native/bridge';
+import { RnExecuTorchError } from '../../../core/error';
+
+declare const phonemizerBrand: unique symbol;
+
+/**
+ * List of all (currently) supported languages in our G2P pipeline.
```
barhanc:
```suggestion
 * Union of all (currently) supported languages in our G2P pipeline.
```

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
```diff
@@ -1,10 +1,52 @@
...
   createSupertonicTextToSpeech,
   type SupertonicTtsModel,
 } from '../extensions/speech/tasks/supertonicTextToSpeech';
 
+type KokoroTts<K extends PropertyKey> = Awaited<ReturnType<typeof createKokoroTextToSpeech<K>>>;
+type SupertonicTts<K extends PropertyKey> = Awaited<
+  ReturnType<typeof createSupertonicTextToSpeech<K>>
+>;
+
+type TtsHookResult<C, P extends { synthesize: unknown; synthesizeStop: unknown }> = {
```
barhanc:
```suggestion
export type TtsHookResult<
  C,
  P extends {
    synthesize: (...args: any[]) => any;
    synthesizeStop: (...args: any[]) => any;
  },
> = {
```

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
```diff
@@ -1,10 +1,52 @@
...
+} from '../extensions/speech/tasks/kokoroTextToSpeech';
 import {
   createSupertonicTextToSpeech,
   type SupertonicTtsModel,
 } from '../extensions/speech/tasks/supertonicTextToSpeech';
 
+type KokoroTts<K extends PropertyKey> = Awaited<ReturnType<typeof createKokoroTextToSpeech<K>>>;
+type SupertonicTts<K extends PropertyKey> = Awaited<
+  ReturnType<typeof createSupertonicTextToSpeech<K>>
+>;
```
barhanc:
```suggestion
type KokoroTts<K extends PropertyKey> = Awaited<
  // prettier-ignore
  ReturnType<typeof createKokoroTextToSpeech<K>>
>;
type SupertonicTts<K extends PropertyKey> = Awaited<
  // prettier-ignore
  ReturnType<typeof createSupertonicTextToSpeech<K>>
>;
```

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
```diff
@@ -21,9 +63,21 @@ import {
...
+export function useTextToSpeech<K extends PropertyKey>(
+  config: KokoroTtsModel<K> | SupertonicTtsModel<K>,
+  options?: ResourceOptions
 ) {
+  // The two configs are structurally disjoint, so the pipeline is picked from
+  // the config's own shape — `voices` is Kokoro-only, `voiceStyles` Supertonic-only.
+  // The overloads above give callers back the precise type.
+  const create = ('voices' in config ? createKokoroTextToSpeech : createSupertonicTextToSpeech) as (
+    ttsConfig: KokoroTtsModel<K> | SupertonicTtsModel<K>
+  ) => Promise<KokoroTts<K> | SupertonicTts<K>>;
```
barhanc, line 88: Checking `'voices' in config` to distinguish between two pipelines is fragile. In TS we unfortunately have to add an explicit parameter to function so that the information about what pipeline we use is preserved after types are stripped. That's why we need something like this
```ts
export function useTextToSpeech<K extends PropertyKey>(
  config: KokoroTtsModel<K>,
  tts: 'Kokoro',
  options?: ResourceOptions
): TtsHookResult<KokoroTtsModel<K>, KokoroTts<K>>;

export function useTextToSpeech<K extends PropertyKey>(
  config: SupertonicTtsModel<K>,
  tts: 'Supertonic',
  options?: ResourceOptions
): TtsHookResult<SupertonicTtsModel<K>, SupertonicTts<K>>;

export function useTextToSpeech<K extends PropertyKey>(
  config: KokoroTtsModel<K> | SupertonicTtsModel<K>,
  tts: 'Kokoro' | 'Supertonic',
  options?: ResourceOptions
) {
  const { resource, downloadProgress, downloadError } = useResourceDownload(config, options);

  let factory;
  switch (tts) {
    case 'Kokoro':
      factory = createKokoroTextToSpeech;
      break;
    case 'Supertonic':
      factory = createSupertonicTextToSpeech;
      break;
  }

  const { model, error } = useModel(
    factory as (
      c: KokoroTtsModel<K> | SupertonicTtsModel<K>
    ) => Promise<KokoroTts<K> | SupertonicTts<K>>,
    resource ?? null
  );

  return {
    isReady: !!model,
    error: downloadError || error,
    downloadProgress,
    resource,
    synthesize: model?.synthesize,
    synthesizeStop: model?.synthesizeStop,
  };
}
```

barhanc (commented):
Second pass. Tested example app and it works great. After the pending comments are resolved, I think we can merge this.

packages/react-native-executorch/src/hooks/useTextToSpeech.ts
IgorSwat: Changed. Now there is a `name` field, but bundled into model configs such that user does not need to explicitly specify it alongside the config itself.

barhanc (approved):

msluszniak (approved):

PR #1365, opened 2026-08-17
[RNE Rewrite] fix(fetcher): make iOS resume actually append, and never cache a short file

msluszniak:
## Description

This PR fixes the way how models are redownloaded after interruption by switching them in the middle of downloading etc. Previously, it caused errors, now it clearly re-start download from the point where it stopped.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

1. Run the computer-vision demo app, pick a large model (FastSAM x, 276 MiB), and let the download start.
2. Switch to another model mid-download, then switch back: it resumes and completes in one pass. Before this change it reached 100%, restarted at 0% and downloaded everything again, and before that it cached a truncated file.
3. Repeat but interrupt the *resumed* download too (SDXS CoreML, 840 MiB, is a good case). Coming back continues from where it stopped instead of falling back to the earlier offset.
4. Kill wifi mid-transfer: the download fails with `Download of ... is incomplete (N of M bytes)` rather than caching a short file, and retrying resumes.
5. Confirm no cached file is smaller than its published size on HF.

Verified on device across the CV app's models. Android is unverified: DownloadManager exposes no status and cancel-then-resume goes through a different code path, so it needs its own pass before merge.

### Related issues

Follow-ups split out into #1366.

### Checklist

### DIFF
--- a/packages/react-native-executorch/src/fetcher/fetcher.ts
+++ b/packages/react-native-executorch/src/fetcher/fetcher.ts
@@ -75,9 +75,49 @@ async function remoteSize(url: string): Promise<number> {
   }
 }
 
+// The file's full length, or 0 when it can't be established. `remoteSize`
+// follows redirects (Hugging Face answers a model URL with a 302 to a CDN), so
+// it reports the length of the file itself. `download()` already computes this
+// for progress weighting and passes it in, so the HEAD is not repeated.
+async function expectedBytesFor(url: string, known: number | undefined): Promise<number> {
+  if (known && known > 0) return known;
+  return remoteSize(url);
+}
+
+// Raised when a transfer reported success but the bytes on disk don't match
+// what the server advertised.
+const incompleteError = (url: string, got: number, want: number) =>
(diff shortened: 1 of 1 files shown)


barhanc:
Two additional things I noticed when working with LLMs, that maybe you could take a look at:
- the progress bar behaves weird sometimes, e.g. being stuck at a low value for very long and then suddenly jumping to 80-90%
- I don't have concrete measurements, but it feels like the download takes noticeably longer (especially for very large files, e.g. 3GB LLM models) than it did previously (for small models I didn't notice any difference)

msluszniak:
>I don't have concrete measurements, but it feels like the download takes noticeably longer (especially for very large files, e.g. 3GB LLM models) than it did previously (for small models I didn't notice any difference)

I had that too, but I thought that was because of my bad internet connection. 😅

barhanc (approved):

Issue #1366, opened 2026-08-17
[RNE Rewrite] Resource fetcher: misleading progress, slow large-file downloads, no iOS background download


msluszniak:
Follow-ups on the resource fetcher (`packages/react-native-executorch/src/fetcher/fetcher.ts`), split out of #1365. First two reported by @barhanc while working with LLMs.

## 1. Progress reporting is misleading

Progress sits at a low value for a long time, then jumps to 80-90%.

Suspects, unverified:
- `download()` weights progress by HEAD sizes only when **every** HEAD succeeds. A single failure flips all files to equal weighting, so a tokenizer counts the same as a 3 GB model.
- Under equal weighting, a cache hit reports `onBytes(size, size)` immediately, so a cached tokenizer alongside an uncached model puts the bar at 50% before any bytes arrive.
- iOS asks for 20 progress callbacks per file (`task.progress({ count: 20 })`) against 100 on Android, so a 3 GB download advances in 150 MB steps.

## 2. Large files download noticeably slower

Multi-GB LLM models feel slower than before the fetcher rewrite (#1253). Small models feel unchanged. No measurements yet, so step one is a before/after benchmark on a 3 GB model for both platforms.

Suspects, unverified: the iOS path streams in-process through blob-util rather than a system session, and every URL is preceded by a HEAD.

## 3. iOS downloads do not continue in the background

Android runs on DownloadManager, which keeps downloading while the app is backgrounded or killed. iOS streams in-process, so suspending the app stops the transfer. For 1-3 GB models this is the difference between a download finishing and never finishing, and a suspended transfer is one of the ways a partial file appears.

Wanted: an iOS background-capable session (`URLSessionConfiguration.background`), so behaviour matches Android.

Note that #1365 makes an interrupted transfer safe (it is detected and resumed rather than cached truncated), but it does not stop the interruption from happening.

msluszniak:
Resolved by Stack #1371

PR #1369, opened 2026-08-18
[RNE Rewrite] feat(fetcher): keep iOS downloads running while the app is backgrounded

msluszniak:
## Description

This PR adds background download functionality for iOS. Previously, only Android had out-of-the-box download functionality.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

Point `download()` at a few hundred MB served from a machine that logs what it actually sends, so the transfer can be observed independently of the app. Then:

1. **Survives suspension.** Start the download, wait for progress, then background the app (launching any other app is enough). The server should keep receiving reads for as long as the transfer needs, and the app should report completion when it comes back. On `main` the connection dies about a second after the app is suspended.
2. **Resumes after a cancel.** Start the download, abort it via `signal` partway, then call `download()` for the same URL again. The second attempt should report progress starting near the abort point, and the server should see a `Range: bytes=N-` request rather than a second full one.
3. **forceDownload really restarts.** Abort partway, then `download(..., { forceDownload: true })`. The server should see a request with no `Range`.

Android is unchanged; the three new spec methods reject there as iOS-only.

### Screenshots

Measured on an iPhone 16 (iOS 26.5), server-side log:

```
suspended at 12.8%      -> SENT continues to 311 MB of 314 MB, +129 s, no DROPPED
                        -> app reports p=1.0000, DONE
control (old path)      -> DROPPED after 42,991,616 bytes, +1 s
abort at p=0.0793       -> GET /medium.bin Range=bytes=25134704-   (= 0.0799)
                        -> app reports resumed at p=0.0799
```

### Related issues

Part of #1366. Stacked on #1368, which is stacked on #1365.

### Checklist

### Additional notes

A transfer that finishes while the app is **killed** needs the host app to forward `application:handleEventsForBackgroundURLSession:completionHandler:`. Surviving suspension works without it. I left that as a follow-up rather than adding an AppDelegate integration here, but say the word if it belongs in this PR.

The download speed part of #1366 is still to come, in its own PR.


barhanc (commented):
Let's not add native code to patch a missing feature in fetching library. The fetcher inside our lib is more like a helper addition, not a core functionality and there are already react-native libs that only do background fetching, so let's keep our helper simple. If a user needs more than that they can just use one of these libs for fetching.

msluszniak:
> Let's not add native code to patch a missing feature in fetching library. The fetcher inside our lib is more like a helper addition, not a core functionality and there are already react-native libs that only do background fetching, so let's keep our helper simple. If a user needs more than that they can just use one of these libs for fetching.

You're right, I'll change it to make it easy to plug an optional dependency with background download here.

packages/react-native-executorch/src/fetcher/fetcher.ts
barhanc, line 1: Maybe we can now unify the behavior on iOS and Android to both use the background downloader if present and else just not support background downloading, otherwise the behavior is silently inconsistent between platforms. Plus, I think it would simplify the code drastically if we didn't have multiple different fetching mechanisms.

msluszniak: Ok, that make sense to use this mechanism to android as well.

msluszniak: @barhanc I couldn't remove download manager fallback for android because it wouldn't work at all. blob utils had a bug that is fixed on main, but no release after this time. There is an issue for that to bump this package and remove redundant fallback to download manager then, see: https://github.com/software-mansion/react-native-executorch/issues/1401

barhanc (approved):

PR #1376, opened 2026-08-21
[RNE Rewrite] chore!: miscellaneous jsdoc, export improvements and small fixes

### LINKED ISSUE #1380: [RNE Rewrite] Passing a model family to download() fetches every variant
## Problem

Model families in `src/models.ts` spread their default variant into the family object so it doubles as a usable config:

```ts
EFFICIENTNET_V2_S: {
  ...EFFICIENTNET_V2_S_XNNPACK_INT8,
  XNNPACK_INT8: EFFICIENTNET_V2_S_XNNPACK_INT8,
  XNNPACK_FP32: EFFICIENTNET_V2_S_XNNPACK_FP32,
  COREML_FP16: EFFICIENTNET_V2_S_COREML_FP16,
},
```

`download()` walks any nested structure of plain objects and collects every string leaf matching `/^https?:\/\//`, then fetches all of them. So passing a family object rather than a variant downloads **every variant in that family**, not just the default the spread advertises.

The convenience default therefore costs a multiple of what a caller would expect, and it pulls artifacts for backends the device cannot even run.

## Impact

`models.ocr.PADDLE.PPOCRV6_SMALL`, measured at the `v0.10.0` tag:

| file | size |
| --- | --- |
| `pp_ocrv6_xnnpack_int8.pte` | 23.9 MB |
| `pp_ocrv6_coreml_int8.pte` | 8.3 MB |
| `pp_ocrv6_vulkan_fp16.pte` | 26.2 MB |
| **total fetched** | **58.3 MB** |

Against 23.9 MB for the XNNPACK variant alone: 2.4x the bytes, and an Android device downloads the CoreML artifact while an iOS device downloads the Vulkan one.

This is not specific to OCR. **62 families across the registry use the spread**, so the same applies to classification, style transfer, segmentation and the rest, in proportion to how many variants each ships.

## Why this is worth fixing

The failure is silent. Nothing errors, nothing warns, the app just spends several times the bandwidth and storage it should on a first run, on metered connections and low-end devices included. The API shape actively invites it: `PPOCRV6_SMALL` looks like the simplest thing to pass, and it is the most expensive.

## Possible directions

1. Have `download()` only follow a config's own model fields rather than recursing into nested variant objects. Most targeted, but changes documented `download()` behaviour, which is deliberately generic.
2. Skip URLs for backends the running platform cannot use, which caps the waste without changing the registry.
3. Drop the spreads and require callers to name a variant. Cleanest semantics, but a breaking change to the registry shape.

Option 2 is the smallest fix that removes the cross-platform nonsense; option 1 or 3 is needed to fix the multiple-variant case properly.

## Context

Surfaced while reviewing #1322: https://github.com/software-mansion/react-native-executorch/pull/1322#discussion_r3840968663

barhanc:
## Description

Refactor the public API surface, add package subpath exports for domain-specific imports, and standardize TSDoc documentation across all modules in preparation for documentation rewrite.

### Package Subpaths

Add subpath exports in package.json for cv, llm, nlp, speech, math, and schema, enabling:
```ts
import { ... } from 'react-native-executorch/cv';
import { ... } from 'react-native-executorch/llm';
```
Add corresponding tsconfig.json path aliases for all example apps.

### API Changes
- Flatten exports: remove re-export leaks, flatten CV operations into cv namespace
- Rename ops: ops/boxes → ops/box, ops/points → ops/point (singular consistency)
- Move preprocessing: preprocessing → cv/utils/imagePreprocessor
- Export constants directly: `export * from './constants'` instead of `export * as constants`
- Namespace types under domain modules: app imports now use `llm.LLMGenerationStats`, `nlp.PiiEntity`, `schema.ConcreteDim`, etc.
- Model registry: models with variants now use `.DEFAULT` accessor (e.g., `models.objectDetection.SSDLITE320_MOBILENET_V3_LARGE.DEFAULT`)

### Documentation
- Add @packageDocumentation and @category tags to all modules
- Standardize TSDoc annotations (error throws, tensor shapes/dtypes, worklet threading)
- Add module-level docs to index.ts, models.ts, constants.ts, utils.ts
- Link imperative APIs and types in hook JSDoc

### Hooks
- Use switch statement for TTS factory resolution in useTextToSpeech
- Format JSDoc type links onto dedicated lines to prevent Prettier wrapping
- Use undefined for absent state instead of null

### Core
- Return raw ConcreteDim for untyped dim accessors
- Mark getRegisteredBackends as worklet

### Implementation
- math.ts: Extract `RandomNormalOptions` type; use `Math.random()` instead of `Date.now()` for seed default
- schema.ts: Extract resolveSymbolDim and createSpecMatch helpers from validateSpec to reduce duplication and improve readability

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

- [ ] Rebuild all example apps to make sure nothing is broken

### Related issues

Closes #1380 

### Checklist


msluszniak:
I'm wondering how do we want to communicate the fact that default model variant is probably rarely the optimal one? I think that our default behaviour will be including all backends etc. and if you want to exclude them, you need to do it manually. This way we would be able to set default models per platform which guarantees the best performance. Otherwise, I think that many people can miss huge advantage from backend acceleration.

barhanc:
> I'm wondering how do we want to communicate the fact that default model variant is probably rarely the optimal one? I think that our default behaviour will be including all backends etc. and if you want to exclude them, you need to do it manually. This way we would be able to set default models per platform which guarantees the best performance. Otherwise, I think that many people can miss huge advantage from backend acceleration.

Agreed. This will require changes in the download scripts and opt-in logic, right?

msluszniak:
> > I'm wondering how do we want to communicate the fact that default model variant is probably rarely the optimal one? I think that our default behaviour will be including all backends etc. and if you want to exclude them, you need to do it manually. This way we would be able to set default models per platform which guarantees the best performance. Otherwise, I think that many people can miss huge advantage from backend acceleration.
> 
> Agreed. This will require changes in the download scripts and opt-in logic, right?

I need to double check before any changes in the model registry regarding changed defaulting behaviour.

msluszniak:
Four consistency gaps found while reviewing this branch.

**1. Agent skills still document the old API.** `.agents/skills/add-task-pipeline/SKILL.md:103` imports from `'./preprocessing'` (now `../utils/imagePreprocessor`), and `.agents/skills/model-schema-validation/SKILL.md` uses `constr` / `constr.eq` in 9 places.

**2. Eight `@category` values are missing from `categoryOrder` in `docs/typedoc.json`:** `CV / Functions`, `Core / Schema / Functions`, `Core / Schema / Types`, `LLM / Functions`, `Math / Functions`, `Math / Types`, `NLP / Functions`, `Speech / Functions`. They land in the trailing `*` bucket, so `X / Types` renders above `X / Functions` on the cv/llm/nlp/speech pages while math and schema render the opposite way. Deriving this list and `VALID_CATEGORIES` in `.eslintrc.js` from one source would stop them drifting again.

**3. `.DEFAULT` is not applied to the deepest variant groups.** 74 groups have it, 33 do not: every `YOLO26.*.SIZE_384/512/640` under `objectDetection` and `instanceSegmentation`, plus `YOLO26_POSE.SIZE_*`. `YOLO26.NANO.DEFAULT` works, but `YOLO26.NANO.SIZE_512` still requires `.XNNPACK_FP32`.

**4. Two new typedoc warnings** (three others are fixed by this PR): `Failed to resolve link to "Tokenizer"` in `createTokenizer`, since `Tokenizer` is not imported in `tokenization.ts`; and `TtsHookResult` is no longer exported, so `useTextToSpeech`'s return type drops out of the API reference.

msluszniak:
Do we want to add the target `default value` mechanism in this PR or in a separate one?

msluszniak (approved):
Tested all apps and besides one comment with some nits everything look cool.

barhanc:
> Four consistency gaps found while reviewing this branch.
> 
> **1. Agent skills still document the old API.** `.agents/skills/add-task-pipeline/SKILL.md:103` imports from `'./preprocessing'` (now `../utils/imagePreprocessor`), and `.agents/skills/model-schema-validation/SKILL.md` uses `constr` / `constr.eq` in 9 places.
> 
> **2. Eight `@category` values are missing from `categoryOrder` in `docs/typedoc.json`:** `CV / Functions`, `Core / Schema / Functions`, `Core / Schema / Types`, `LLM / Functions`, `Math / Functions`, `Math / Types`, `NLP / Functions`, `Speech / Functions`. They land in the trailing `*` bucket, so `X / Types` renders above `X / Functions` on the cv/llm/nlp/speech pages while math and schema render the opposite way. Deriving this list and `VALID_CATEGORIES` in `.eslintrc.js` from one source would stop them drifting again.
> 
> **3. `.DEFAULT` is not applied to the deepest variant groups.** 74 groups have it, 33 do not: every `YOLO26.*.SIZE_384/512/640` under `objectDetection` and `instanceSegmentation`, plus `YOLO26_POSE.SIZE_*`. `YOLO26.NANO.DEFAULT` works, but `YOLO26.NANO.SIZE_512` still requires `.XNNPACK_FP32`.
> 
> **4. Two new typedoc warnings** (three others are fixed by this PR): `Failed to resolve link to "Tokenizer"` in `createTokenizer`, since `Tokenizer` is not imported in `tokenization.ts`; and `TtsHookResult` is no longer exported, so `useTextToSpeech`'s return type drops out of the API reference.

Fixed.

PR #1386, opened 2026-08-26
[RNE Rewrite] fix(ts): release native resources when a create<Task> fails

msluszniak:
## Description

Fixes the finding recorded in #1355: a `create<Task>` that throws part-way through construction abandoned everything it had already allocated.

A factory allocates as it goes, and only hands back a `dispose` at the very end. Anything that threw in between left the caller with no reference to what was already there, and native memory is not garbage collected, so it stayed alive for the rest of the process. Depending on the task that is a model, a tokenizer, a phonemizer, an LLM runner, or a whole nested pipeline (Whisper owns both a tokenizer and a VAD). `useModel` re-runs its factory whenever the config changes, so an app pointed at a mismatched model leaked a full resource set per attempt.

`src/core/lifetime.ts` adds `createResourceScope`, which gives a factory one teardown path for both outcomes: it tracks each resource as it is created, releases them in reverse order, and the same function becomes the pipeline's `dispose`. Every factory now allocates through a scope and wraps its body in `try`/`catch`:

```typescript
const scope = createResourceScope();
const dispose = scope.dispose;

try {
  const model = scope.track(await wrapAsync(loadModel, runtime)(modelPath));
  const { dims } = validateSpec(model.schema, { ... }); // may throw
  const tensors = [tensor('float32', outShape)] as const;
  tensors.forEach(scope.track);
  return { classify, classifyWorklet, dispose };
} catch (error) {
  dispose();
  throw error;
}
```

`createKokoroTextToSpeech` already did this with a local array and moves to the shared helper. That also closes a smaller hole it had: its two models were loaded before the `try` block, so one load rejecting stranded the other. Parallel loads elsewhere now track inside each promise for the same reason.

`createTokenizer` is unchanged. It loads a tokenizer and returns with nothing in between that can throw, so it has no window to leak through.

The helper is internal, so the public export surface is unchanged.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

```sh
yarn workspace react-native-executorch test
yarn typecheck
yarn lint
```

Expected: 28 suites, 3117 tests, 4 snapshots, 0 skipped; typecheck and lint clean.

To see the tests bite, stash `src/` and re-run: every construction-failure case fails.

### Screenshots

N/A

### Related issues

Follows up the finding recorded in #1355.

### Checklist

### Additional notes

Stacked on `@ms/api-tests` (#1355), because that is where the suite recording this behavior lives. Merge #1355 first and this retargets to `rne-rewrite` cleanly.

`tasks/constructionFailure.test.ts` was written to record the leak as current behavior so it would fail loudly the day a factory started cleaning up. That is this PR, so it now asserts the opposite, and covers all fifteen factories rather than five. It checks every kind of handle, so a failure names the factory and the resource rather than just reporting that something leaked.

The per-pipeline suites and `hooks/taskHooks.test.ts` drop the `allowNativeLeaks()` calls they needed for the same reason. Nothing in the suite leaks any more, so the setup file's global leak check now asserts this on every construction-failure test for free.

The diff is large mostly because indenting a factory body inside `try` touches every line of it. The behavioral change per file is the scope, the `track` calls and the `catch`.

### DIFF
--- a/packages/react-native-executorch/src/core/lifetime.ts
+++ b/packages/react-native-executorch/src/core/lifetime.ts
@@ -1,0 +1,74 @@
+/**
+ * Construction-time ownership of native resources.
+ *
+ * A `create<Task>` factory allocates as it goes: it loads a model, maybe a
+ * tokenizer or a phonemizer, validates the schema, pre-allocates its execution
+ * tensors, and only at the end hands the caller a `dispose`. Anything that
+ * throws in between leaves the caller with no reference to what was already
+ * allocated, and native memory is not garbage collected, so it stays alive for
+ * the rest of the process.
+ *
+ * A scope closes that window: every resource is tracked as it is created, and
+ * one `dispose` releases whatever exists at the moment it is called. The same
+ * function is the factory's own `dispose`, so there is a single teardown path
+ * rather than one for failure and one for success.
+ * @module Core/Lifetime
+ */
+
+/** Anything holding native memory that has to be released explicitly. */
+export type Disposable = { dispose: () => void };
+
+/**
+ * A set of native resources with a single teardown.
+ * @category Core / Types
+ */
+export type ResourceScope = {
+  /**
+   * Takes ownership of a resource and returns it unchanged, so it can wrap an
+   * allocation in place.
+   */
+  readonly track: <D extends Disposable>(resource: D) => D;
+  /**
+   * Releases every tracked resource, most recently allocated first. Safe to
+   * call more than once: a second call has nothing left to release.
+   */
+  readonly dispose: () => void;
+};
+
+/**
+ * Creates a {@link ResourceScope} for a factory to allocate into.
+ *
+ * Wrap the factory body in `try`/`catch`, `track` each resource as it is
+ * created, and return `scope.dispose` as the pipeline's `dispose`:
+ * ```typescript
+ * const scope = createResourceScope();
+ * const dispose = scope.dispose;
+ * try {
+ *   const model = scope.track(await wrapAsync(loadModel, runtime)(modelPath));
+ *   const { dims } = validateSpec(model.schema, { ... }); // may throw
+ *   const tensors = [scope.track(tensor('float32', shape))];
+ *   return { run, dispose };
+ * } catch (error) {
+ *   dispose();
+ *   throw error;
+ * }
+ * ```
+ * @category Core / Functions
+ * @returns A scope that tracks resources and releases them on `dispose`.
+ */
+export function createResourceScope(): ResourceScope {
+  const allocated: Disposable[] = [];
+
+  return {
+    track: <D extends Disposable>(resource: D): D => {
+      allocated.push(resource);
+      return resource;
+    },
+    // Reverse order, so a resource is released before whatever it was built
+    // from. `splice` empties the list as it goes, which is what makes a second
+    // call a no-op rather than a double dispose.
+    dispose: (): void => {
+      for (const resource of allocated.splice(0).reverse()) resource.dispose();
+    },
+  };
+}
--- a/packages/react-native-executorch/src/core/lifetime.ts
+++ b/packages/react-native-executorch/src/core/lifetime.ts
@@ -1,21 +1,22 @@
 /**
  * Construction-time ownership of native resources.
  *
- * A `create<Task>` factory allocates as it goes: it loads a model, maybe a
- * tokenizer or a phonemizer, validates the schema, pre-allocates its execution
- * tensors, and only at the end hands the caller a `dispose`. Anything that
- * throws in between leaves the caller with no reference to what was already
- * allocated, and native memory is not garbage collected, so it stays alive for
- * the rest of the process.
+ * A factory that builds a pipeline allocates as it goes: it loads models,
+ * validates their schemas, pre-allocates execution tensors, and only at the end
+ * hands the caller a `dispose`. Anything that throws in between leaves the
+ * caller with no reference to what was already allocated, and native memory is
+ * not garbage collected, so it stays alive for the rest of the process.
  *
  * A scope closes that window: every resource is tracked as it is created, and
  * one `dispose` releases whatever exists at the moment it is called. The same
- * function is the factory's own `dispose`, so there is a single teardown path
+ * function is the pipeline's own `dispose`, so there is a single teardown path
  * rather than one for failure and one for success.
- * @module Core/Lifetime
  */
 
-/** Anything holding native memory that has to be released explicitly. */
+/**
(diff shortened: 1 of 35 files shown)


packages/react-native-executorch/src/core/lifetime.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * Construction-time ownership of native resources.
```
barhanc, line 2: Please change the docs so that they don't mention `create<Task>` at all. The extensions depend on core no the other way around so nothing in core should reference extensions.

packages/react-native-executorch/src/core/lifetime.ts
```diff
@@ -0,0 +1,74 @@
...
+ * tensors, and only at the end hands the caller a `dispose`. Anything that
+ * throws in between leaves the caller with no reference to what was already
+ * allocated, and native memory is not garbage collected, so it stays alive for
+ * the rest of the process.
+ *
+ * A scope closes that window: every resource is tracked as it is created, and
+ * one `dispose` releases whatever exists at the moment it is called. The same
+ * function is the factory's own `dispose`, so there is a single teardown path
+ * rather than one for failure and one for success.
+ * @module Core/Lifetime
```
barhanc:
```suggestion
```
These are not used in docs actually.

packages/react-native-executorch/src/core/lifetime.ts
```diff
@@ -0,0 +1,74 @@
...
+ * A set of native resources with a single teardown.
+ * @category Core / Types
+ */
+export type ResourceScope = {
+  /**
+   * Takes ownership of a resource and returns it unchanged, so it can wrap an
+   * allocation in place.
+   */
+  readonly track: <D extends Disposable>(resource: D) => D;
+  /**
```
barhanc:
```suggestion
  readonly track: <D extends Disposable>(resource: D) => D;
  
  /**
```

packages/react-native-executorch/src/core/lifetime.ts
```diff
@@ -0,0 +1,74 @@
...
+ * const dispose = scope.dispose;
+ * try {
+ *   const model = scope.track(await wrapAsync(loadModel, runtime)(modelPath));
+ *   const { dims } = validateSpec(model.schema, { ... }); // may throw
+ *   const tensors = [scope.track(tensor('float32', shape))];
+ *   return { run, dispose };
+ * } catch (error) {
+ *   dispose();
+ *   throw error;
+ * }
```
barhanc: Same as above. Let's not mention extension context at all here, it's just for semi-automatic lifetime management.

packages/react-native-executorch/src/core/lifetime.ts
barhanc, line 1: This should also be added to public API, so users can follow this pattern in their custom pipelines.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
```diff
@@ -87,83 +88,87 @@ export async function createTextEmbedder(
...
-  return { embed, embedWorklet, dispose };
+  const scope = createResourceScope();
+  const dispose = scope.dispose;
+
+  try {
+    const { modelPath, tokenizerPath, defaultPrompt } = config;
+    const [model, tokenizer] = await Promise.all([
+      wrapAsync(loadModel, runtime)(modelPath).then(scope.track),
+      wrapAsync(loadTokenizer, runtime)(tokenizerPath).then(scope.track),
+    ]);
```
barhanc: Using `Promise.all` can still leak if I'm not mistaken, because it will reject on the first reject, but won't cancel the other enqueued loads. Safer to just rewrite this to sequantial await calls.

packages/react-native-executorch/src/core/lifetime.ts
barhanc, line 62: Put this example at the bottom under `@example` tag.

packages/react-native-executorch/src/core/lifetime.ts
msluszniak: Done. One thing this exposes: `Disposable` ships with it, and it shadows the built-in TS `Disposable` for anyone importing from the package root. Worth renaming to `NativeResource`, or leave it?

packages/react-native-executorch/src/core/lifetime.ts
msluszniak: Done.

packages/react-native-executorch/src/extensions/nlp/tasks/textEmbedding.ts
msluszniak: Agreed, promise leaks as well, handled it as suggested.

packages/react-native-executorch/src/core/lifetime.ts
msluszniak: Removed redundant comment.

packages/react-native-executorch/src/core/lifetime.ts
msluszniak: Removed all mentions.

packages/react-native-executorch/src/core/lifetime.ts
barhanc: Yes, let's rename it.

msluszniak: Ok, renamed.

packages/react-native-executorch/src/extensions/nlp/tasks/privacyFilter.ts
```diff
@@ -130,161 +131,165 @@ export async function createPrivacyFilter<Label extends string>(
...
-    static: method(
-      'forward', // prettier-ignore
-      [i64(1, 'S'), i64(1, 'S')],
-      [f32(1, 'S', numLabels)]
-    ),
-  });
+    // Loaded one at a time on purpose: `Promise.all` rejects on the first
+    // failure while the others keep loading, so their resources would land
+    // after the scope has already been disposed.
+    const model = scope.track(await wrapAsync(loadModel, runtime)(modelPath));
```
barhanc:
```suggestion
    const model = scope.track(await wrapAsync(loadModel, runtime)(modelPath));
```
And similarly for other tasks. This is the straight-forward approach, I'd rather see an explaining comment in the Promise.all approach.

barhanc (approved):
Just a style comment. Otherwise looks solid.

packages/react-native-executorch/src/extensions/nlp/tasks/privacyFilter.ts
msluszniak: Applied to all 4 places.

PR #1392, opened 2026-08-27
[RNE Rewrite] feat(models)!: resolve the DEFAULT model variant per platform

### LINKED ISSUE #1389: [RNE Rewrite] Guarantee that default model is chosen based on platform
Right now, we should xnnpack models as default ones because they are universal. This is not correct approach. We need to change it to implement it in the way described in this comment: https://github.com/software-mansion/react-native-executorch/pull/1376#issuecomment-5411790144.

msluszniak:
## Description

`models.<task>.<MODEL>.DEFAULT` pointed at one fixed export, almost always XNNPACK, so an app that did nothing special ran on the CPU while the accelerator sat idle.

`DEFAULT` is now resolved at load time from the platform (iOS: Core ML, MLX, XNNPACK; Android: Vulkan, XNNPACK; iOS simulator: XNNPACK only), the backends the binary was linked with, and declaration order as a tie-break. Naming a variant directly still overrides it. 96 of 117 groups change on an iOS device; on Android the embedders, whisper-tiny, Supertonic, PP-OCRv6, LFM2.5-VL and Gemma 4 E2B move to Vulkan.

Where a model ships both Core ML and MLX the winner is a benchmark result, not something ordering can state, so those 9 groups pin their choice at the call site with the reason and a test fails until a new such group does the same. Core ML wins every measured pair (1.5x to 4.1x) except distiluse, which goes to MLX.

Auditing the 48 referenced HF repos turned up 41 published `.pte` files no entry pointed at; 38 are now wired. Left out: a pre-rename duplicate and the two Llama 3.2 QAT+LoRA builds, which the rewrite does not support.

Vulkan margins over XNNPACK, per execution on device: `all-MiniLM-L6-v2` 2.4x, whisper encode 1.6x to 1.9x across all six sizes, Supertonic 3.1x on a Mali-G76 and 1.0x to 1.5x on an Adreno 840. Every Vulkan export was checked against the fp32 CPU reference and against run-to-run drift. Vulkan pays a one-time shader compile of roughly 300 ms (Adreno) to 700 ms (Mali) on first load.

Both mpnet embedders also ship an int8 weight-only Vulkan export, 133 MB against 218 MB, but **fp16 stays the Android default for both**. int8 runs 1.34x faster than fp16 on an Adreno 840 (59 ms against 79 ms at the 382-token bound) and 6 to 8 times slower on a Mali-G76 (25.4 s against 3.0 s at 382 tokens, 5.4 s against 1.17 s at 128). `linear_q8csw_tiled` unpacks the weights with `bitfieldExtract` inside the accumulation loop and ships with `TILE_M4 = TILE_N4 = TILE_K4 = 1`, so each unpacked texel feeds four FMAs; Adreno issues integer and float at comparable rates and still wins on memory traffic, Bifrost dual-issues fp16 on the FMA pipe but not the integer path. The registry resolves by backend rather than by GPU vendor, so one Android default has to serve both and fp16 is the safe one. `VULKAN_INT8` is reachable by name and is the better choice on Adreno.

Accuracy is not the reason: int8 holds cosine 0.9983 to 0.9994 against the fp32 reference for `all-mpnet-base-v2` and 0.9952 to 0.9993 for `multi-qa-mpnet-base-dot-v1`, measured on a Mali-G76 at 24 token counts from 1 to each model's bound, chosen to exercise every tile size the int8 kernel selects.

Two fixes the resolution depends on: `download-libs.js` omitted backends `DEFAULT` could then never resolve to (now held in sync by a test), and the demo apps did not declare features they use. Diff bulk is registry entries trading a hand written `DEFAULT` line for a `variants({ ... })` / `family({ ... })` wrapper.

- [x] Yes
- [ ] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

- [ ] Run `yarn workspace react-native-executorch test` and `yarn lint`.
- [ ] On an iOS device, load a Core ML backed model and an MLX backed one through `.DEFAULT`, and confirm the accelerated file is the one downloaded.
- [ ] On the iOS simulator, confirm the same calls fall back to the XNNPACK file and load.
- [ ] On Android, confirm the models listed above come up on Vulkan and everything else is unchanged.
- [ ] On an iOS device, spot-check the newly wired Core ML exports that have not been benchmarked: YOLO26-seg, YOLO26-pose and the text embedders.

### Related issues

Closes #1389

### Checklist

### Additional notes

The Vulkan exports need the runtime fixes on the labs fork branch `@ms/separate-backends-1.4.1`, published in `v0.10.0-libs-1.4.1`. Without them the file loses the Vulkan device, aborts on a missing operator, or returns NaNs. That release also carries six upstream fixes (pytorch/executorch#22348, #22349, #22409, #22410, #22411, #22430). The last one is what the int8 exports need: `aten._weight_int8pack_mm` dispatched to `linear_qcs8w_*` shaders that no longer exist, so before it every weight-only int8 model aborted on its first execute. Only the `vulkan-android-*` artifacts changed; core and xnnpack are untouched.


packages/react-native-executorch/src/modelVariants.ts
barhanc, line 1: No need for additional file in root for this imo. This can be compressed to around 80 lines and put directly inside `model.ts` as local helpers. Something like
```ts
// =============================================================================
// MODEL REGISTRY
// =============================================================================

/** All supported ExecuTorch backend identifiers. */
const ALL_BACKENDS = ['xnnpack', 'coreml', 'mlx', 'vulkan'] as const;

/** Canonical tag representing an ExecuTorch backend. */
type BackendTag = (typeof ALL_BACKENDS)[number];

/** Platform-specific backend priority (accelerators first with CPU fallback). */
const BACKENDS_PREFERENCE: Record<'ios' | 'android', readonly BackendTag[]> = {
  ios: rnexecutorchJsi.isEmulator ? ['xnnpack'] : ['coreml', 'mlx', 'xnnpack'],
  android: ['vulkan', 'xnnpack'],
};

function getCandidateBackends(): readonly BackendTag[] {
  if (Platform.OS !== 'ios' && Platform.OS !== 'android') {
    return ALL_BACKENDS;
  }

  let registered: string[] = [];
  try {
    registered = getRegisteredBackends().map((name) => name.toLowerCase());
  } catch {
    registered = [];
  }
  if (registered.length === 0) registered = [...ALL_BACKENDS];
  return BACKENDS_PREFERENCE[Platform.OS].filter((backend) =>
    registered.some((name) => name.includes(backend))
  );
}

/**
 * Candidate backends for default resolution on this platform, ordered best-first
 * (accelerators before CPU fallback). Falls back to all preferred backends if
 * the native runtime cannot be queried.
 */
const CANDIDATE_BACKENDS = getCandidateBackends();

/**
 * Resolves the platform-optimal `DEFAULT` variant for a group of model exports.
 * @param options Map of named model variants.
 * @param pinned Optional per-platform override for models where a specific
 * export is preferred.
 * @returns The variant map augmented with a resolved `DEFAULT` alias.
 */
function variants<T extends Record<string, unknown>>(
  options: T,
  pinned?: Partial<Record<'ios' | 'android', keyof T>>
): T & { readonly DEFAULT: T[keyof T] } {
  if (Platform.OS !== 'ios' && Platform.OS !== 'android') {
    return { ...options, DEFAULT: options[Object.keys(options)[0] as keyof T] };
  }

  const backendOf = (k: string) => k.toLowerCase().split('_')[0] as BackendTag;

  const pin = pinned?.[Platform.OS];
  if (pin !== undefined && CANDIDATE_BACKENDS.includes(backendOf(pin as string))) {
    return { ...options, DEFAULT: options[pin] };
  }

  for (const backend of CANDIDATE_BACKENDS) {
    const match = Object.keys(options).find((k) => backendOf(k) === backend);
    if (match !== undefined) {
      return { ...options, DEFAULT: options[match as keyof T] };
    }
  }

  return { ...options, DEFAULT: options[Object.keys(options)[0] as keyof T] };
}

```

packages/react-native-executorch/src/models.ts
```diff
@@ -887,18 +1116,45 @@ const SUPERTONIC_3_MLX_FP32: SupertonicTtsModel<SupertonicDefaultVoiceName> = {
...
+const KOKORO_STANDARD_PATHS = kokoroModelPaths('std', 'standard');
+const KOKORO_POLISH_PATHS = kokoroModelPaths('pl', 'polish');
+const KOKORO_GERMAN_PATHS = kokoroModelPaths('de', 'german');
+const KOKORO_STANDARD_COREML_PATHS = kokoroModelPaths('std', 'standard', 'coreml');
+const KOKORO_POLISH_COREML_PATHS = kokoroModelPaths('pl', 'polish', 'coreml');
+const KOKORO_GERMAN_COREML_PATHS = kokoroModelPaths('de', 'german', 'coreml');
+
+// A language differs from its Core ML twin only in which pair of .pte files it
+// loads: the voices, the phonemizer and the voice union are all backend-free.
+const kokoroCoreMl = <V extends string>(
```
barhanc: This reverts the changes from the kokoro PR review.

packages/react-native-executorch/src/models.ts
```diff
@@ -2358,63 +2615,61 @@ export const models = {
...
      * Kokoro — a lightweight phoneme-driven Text-to-Speech model. Each language
      * entry bundles the matching model weights, grapheme-to-phoneme assets and
      * the voices available for that language, nested per backend.
+     *
+     * The Core ML builds are the registry's only fp32 Core ML exports — that is
+     * what is published, and fp32 keeps them off the Neural Engine, which is
+     * fp16-only. They still default on iOS because the duration predictor runs
+     * 14-21x faster warm there than the XNNPACK one on an iPhone 16. The cost
+     * is a one-time per-method Core ML compile on first use (~13s), cached
+     * across launches; reach for `XNNPACK_FP32` explicitly to avoid it.
```
barhanc: Shouldn't this be an internal comment like in other places, instead of public API doc?

packages/react-native-executorch/src/models.ts
```diff
@@ -1693,40 +1898,59 @@ export const models = {
...
+      // marks invisible drift further, up to 15.7 px, but their coordinates are
+      // undefined when `vis` is 0.)
+      //
+      // That fp16 build took a while to exist and is fragile in a specific way,
+      // so do not "simplify" it. It is correct on device only with both the CPU
+      // and the Neural Engine excluded — measured against fp32's 0.863, `all`
+      // gives 0.421, `cpu_only` 0.411, `cpu_and_ne` 0.422 and `cpu_and_gpu`
+      // 0.862 — and on macOS only the CPU path is wrong, so a host check passes
+      // a build the device gets wrong. It also needs an iOS17 deployment target;
+      // iOS18 degrades every build of this model at either precision. Both
```
barhanc: I think just a gist of the explanation would be enough. Similarly in other places.

packages/react-native-executorch/src/models.ts
```diff
@@ -1898,72 +2112,115 @@ export const models = {
...
-          DEFAULT: WHISPER_TINY_EN_XNNPACK_FP32,
-          XNNPACK_FP32: WHISPER_TINY_EN_XNNPACK_FP32,
-          COREML_FP16: WHISPER_TINY_EN_COREML_FP16,
-          MLX_BF16: WHISPER_TINY_EN_MLX_BF16,
-          MLX_INT8: WHISPER_TINY_EN_MLX_INT8,
-        },
+        TINY: variants(
+          {
+            XNNPACK_FP32: WHISPER_TINY_EN_XNNPACK_FP32,
+            XNNPACK_INT8: WHISPER_TINY_EN_XNNPACK_INT8,
```
barhanc, line 2328: Is fp32 above int8 correct here?

barhanc (commented):
I will test the models later today.

packages/react-native-executorch/src/modelVariants.ts
msluszniak: Done, folded into `models.ts` as local helpers and deleted the file. The `variants()` unit tests now assert the same four rules against the real registry instead of a synthetic map.

packages/react-native-executorch/src/models.ts
msluszniak: Reverted. `kokoroModelPaths(backend, variant, dir)` is back with the explicit first arg, and the Core ML presets are literals again.

packages/react-native-executorch/src/models.ts
msluszniak: Yes, moved it to an internal comment above `KOKORO`.

packages/react-native-executorch/src/models.ts
msluszniak: Trimmed: RF-DETR 19 lines to 6, Whisper 10 to 4, distiluse 7 to 4.

packages/react-native-executorch/src/models.ts
msluszniak: Not for `BASE`/`SMALL`, both now list int8 first. Over 250 LibriSpeech test-clean clips int8 moves base.en 4.84% to 5.20% WER and small.en 3.42% to 3.38%, too little against halving the download (247 MB vs 399, 448 vs 1129). `TINY` keeps fp32 first, where int8 costs 6.08% to 7.77%.

packages/react-native-executorch/src/models.ts
```diff
@@ -49,6 +57,127 @@ import {
...
+// variants are declared in. A group whose best export does not follow from
+// that order pins one per platform — see the second argument of `variants`.
+
+/** The backend prefix a variant key starts with. */
+type BackendTag = 'XNNPACK' | 'COREML' | 'MLX' | 'VULKAN';
+
+/** The platforms the registry resolves defaults for. */
+type TargetPlatform = 'ios' | 'android';
+
+const ALL_BACKENDS: readonly BackendTag[] = ['XNNPACK', 'COREML', 'MLX', 'VULKAN'];
```
barhanc: Better to write 
```ts
const ALL_BACKENDS = [ ... ] as const;
type BackendTag = (typeof ALL_BACKENDS)[number]
```
so there is no duplication.

packages/react-native-executorch/src/models.ts
```diff
@@ -49,6 +57,127 @@ import {
...
+
+/**
+ * Backends linked into this binary and usable on this device.
+ * @returns The usable tags — every one of them when the native runtime cannot
+ * be asked, so that a missing answer widens the choice rather than emptying it.
+ */
+function usableBackends(): ReadonlySet<BackendTag> {
+  let registered: readonly string[] = [];
+  try {
+    registered = rnexecutorchJsi.getExecuTorchRegisteredBackends();
```
barhanc: Use `utils/getRegisteredBackends` not the untyped version attached to jsi.

packages/react-native-executorch/src/models.ts
```diff
@@ -49,6 +57,127 @@ import {
...
+  const names = registered.map((name) => name.toLowerCase());
+  const usable = ALL_BACKENDS.filter((tag) =>
+    names.some((name) => name.startsWith(tag.toLowerCase()))
+  );
+
+  // The simulator links the Core ML backend but cannot run it: no Neural
+  // Engine, and MPSGraph refuses the compiled models. MLX only ever ships a
+  // device slice, so it drops out of `registered` on its own.
+  const onSimulator = PLATFORM === 'ios' && rnexecutorchJsi.isEmulator === true;
+  return new Set(onSimulator ? usable.filter((tag) => tag !== 'COREML' && tag !== 'MLX') : usable);
```
barhanc: Maybe cleaner to wire this directly into `BACKEND_ORDER` so it's immediately visible what backends are available on what platform / emulator in one place, like in the example snippet.

packages/react-native-executorch/src/models.ts
```diff
@@ -49,6 +57,127 @@ import {
...
+};
+
+const PLATFORM: TargetPlatform = Platform.OS === 'ios' ? 'ios' : 'android';
+
+/**
+ * Backends linked into this binary and usable on this device.
+ * @returns The usable tags — every one of them when the native runtime cannot
+ * be asked, so that a missing answer widens the choice rather than emptying it.
+ */
+function usableBackends(): ReadonlySet<BackendTag> {
```
barhanc: Name is misleading because if `registered` fails or is empty then we don't really know what is usable. `getCandidateBackends` is imo more clear.

packages/react-native-executorch/src/models.ts
```diff
@@ -49,6 +57,127 @@ import {
...
+  try {
+    registered = rnexecutorchJsi.getExecuTorchRegisteredBackends();
+  } catch {
+    registered = [];
+  }
+  if (registered.length === 0) return new Set(ALL_BACKENDS);
+
+  const names = registered.map((name) => name.toLowerCase());
+  const usable = ALL_BACKENDS.filter((tag) =>
+    names.some((name) => name.startsWith(tag.toLowerCase()))
```
barhanc: `.toLowerCase` wouldn't be needed if backends were lowercase already.

packages/react-native-executorch/src/models.ts
```diff
@@ -49,6 +57,127 @@ import {
...
+
+const USABLE_BACKENDS = usableBackends();
+
+/**
+ * Reads the backend out of a variant key.
+ * @param key The variant key, e.g. `COREML_FP16`.
+ * @returns The backend the key names, or `undefined` when it names none.
+ */
+const backendOf = (key: string): BackendTag | undefined =>
+  ALL_BACKENDS.find((tag) => key === tag || key.startsWith(`${tag}_`));
```
barhanc: Imo can be just defined inside function where it is used, it's a simple one liner. Also wouldn't just `key.includes(tag)` be enough here?

barhanc (commented):
Tested the new CoreML and Vulkan models and they work great. Just some additional comments regarding the default variant resolution code.

packages/react-native-executorch/src/models.ts
msluszniak: Done, `ALL_BACKENDS` is `as const` and `BackendTag` is derived from it.

packages/react-native-executorch/src/models.ts
msluszniak: Switched to `getRegisteredBackends` from `utils`.

packages/react-native-executorch/src/models.ts
msluszniak: Moved into `BACKEND_ORDER`: `ios: isEmulator ? [xnnpack] : [coreml, mlx, xnnpack]`, and the post-filter is gone. It also fixes a case where an empty `registered` skipped the carve-out and let the simulator pick Core ML.

packages/react-native-executorch/src/models.ts
msluszniak: Renamed to `getCandidateBackends`, and the constant to `CANDIDATE_BACKENDS`.

packages/react-native-executorch/src/models.ts
msluszniak: Right, tags are lowercase now, so only the registered names get lowercased.

packages/react-native-executorch/src/models.ts
msluszniak: Inlined into `pickVariant` as `key.toLowerCase().split("_")[0]`. Kept the prefix split over `includes` so only the leading segment can name a backend.

barhanc (approved):

msluszniak:
@barhanc forgot to ask if MRs 18 and 20 from gitlab are ok since they were used to introduce some of delegates from this PR.

PR #1397, opened 2026-08-30
feat(speech): accept a whisper model exported at a fixed audio length

msluszniak:
## Description

`createWhisperSpeechToText` declared `encode` as taking a dynamic `T_audio`, so a model exporting it at one fixed sample count was rejected at load with `Symbol 'T_audio' is 'constant', expected 'dynamic'`. The Vulkan delegate has no dynamic-shape support, so a fixed length is the only shape it can export.

Adds a second spec variant instead of a backend flag: `validateSpec` already tries variants in order and reports which one matched, so the model selects its own path from the schema it ships, the way the CV tasks do it. When the static variant matches, the chunk is zero-padded to the declared length before each `encode`.

Padding is free semantically: whisper's mel preprocessor pads or truncates to a fixed 3000 frames regardless, so a fixed window is the shape the model already works in internally. The dynamic path is untouched and allocates nothing extra, since `fitToLength` returns its input when the length already matches.

Note that the task feeds chunks of at most `CHUNK_LENGTH_SECONDS` (29 s), and as little as `MIN_CHUNK_SIZE`, so a static model pays for a full window on every chunk. That is inherent to exporting without dynamic shapes, not something the padding introduces.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

`yarn typecheck`, `yarn lint` (from the repo root) and `yarn test` in `packages/react-native-executorch`. Two tests were added to `__tests__/tasks/remainingTasks.test.ts`: one accepting a model whose `encode` is exported at a constant 480000 samples, one rejecting an `encode` signature that matches neither variant.

Not yet exercised on device: the Vulkan whisper model this unblocks is still in review in export-scripts !20.

### Checklist

### Additional notes

The static variant binds `StateDim` the same way the dynamic one does, so it also accommodates a wider encoder output than 384 without further changes.

### DIFF
--- a/packages/react-native-executorch/__tests__/tasks/remainingTasks.test.ts
+++ b/packages/react-native-executorch/__tests__/tasks/remainingTasks.test.ts
@@ -298,6 +298,36 @@ describe('createWhisperSpeechToText', () => {
     stt.dispose();
   });
 
+  it('accepts a model whose encode is exported at one fixed sample count', async () => {
+    register();
+    // Vulkan has no dynamic-shape support, so its encode takes a constant length.
+    fakeJsi.registerModel(MODEL_PATH, {
+      schema: exported({
+        ...method('encode', [f32(480000)], [f32(1, 1500, 3072)]),
+        ...method('decode', [i64(1, 1), i64(1), f32(1, 1500, 3072)], [f32(1, 1, 51865)]),
+      }),
+    });
+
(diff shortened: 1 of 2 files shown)


msluszniak:
Not valid, vulkan successfully handles dynamic shapes.

PR #1398, opened 2026-08-31
feat(speech): add Vulkan whisper-tiny variants

msluszniak:
## Description

Registers the fp16 and int8 Vulkan exports of multilingual whisper-tiny as `WHISPER.TINY.VULKAN_FP16` / `WHISPER.TINY.VULKAN_INT8`, adds `vulkan` to the `speechToText` entry in the `download-libs.js` feature map so the backend lib is fetched, and offers both in the speech demo on Android.

Encode runs on Vulkan (mel preprocessor falls back to XNNPACK); decode stays on XNNPACK, where it measured faster. On an Adreno 840 the encode is about 1.6x the XNNPACK time and decode is 4.08 ms, down from 11.96 ms.

The exports require the ExecuTorch 1.4.1 Vulkan runtime published in `v0.10.0-libs-1.4.1`. Older Vulkan libs lack the gelu tanh clamp, the reduce-shader barrier fix and the Adreno local-work-group fix, and produce NaNs or a device loss.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

1. Run `apps/speech` on an Android device.
2. Open Transcribe Audio File, pick Tiny Multilingual (Vulkan fp16), transcribe the sample file, and check the transcript carries no `<|0.00|>` timestamp tokens.
3. Repeat with Tiny Multilingual (Vulkan int8).
4. Repeat both on Live Transcription (Mic).

### Checklist

### Additional notes

Verified on a Galaxy S26 Ultra against the exact `libvulkan_executorch_backend.so` published in `v0.10.0-libs-1.4.1` (md5 `3aba5c82...`, matched between the installed APK and the release tarball).

The `speechToText` feature map lists `xnnpack`, `coreml` and now `vulkan`, but not `mlx`, even though the registry exposes `MLX_BF16` / `MLX_INT8` whisper variants. That looks like a pre-existing gap and is left alone here.

### DIFF
--- a/apps/speech/app/audio-file-transcription/index.tsx
+++ b/apps/speech/app/audio-file-transcription/index.tsx
@@ -74,6 +74,16 @@ const MODELS = [
     disabled: Platform.OS !== 'ios',
   },
   {
+    name: 'Tiny Multilingual (Vulkan fp16)',
+    config: models.speechToText.WHISPER.TINY.VULKAN_FP16,
+    disabled: Platform.OS !== 'android',
+  },
+  {
+    name: 'Tiny Multilingual (Vulkan int8)',
+    config: models.speechToText.WHISPER.TINY.VULKAN_INT8,
+    disabled: Platform.OS !== 'android',
+  },
+  {
     name: 'Base Multilingual (CPU)',
     config: models.speechToText.WHISPER.BASE.XNNPACK_FP32,
   },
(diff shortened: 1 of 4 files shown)


msluszniak:
Folded into #1392, which already carries the registry variant wiring.

PR #1402, opened 2026-09-01
[RNE Rewrite] feat(core): eagerly load and compile model methods during loadModel

### LINKED ISSUE #1393: [RNE Rewrite] Move model compilation (first run) into model loading rather than first run
We should warm up models, especially delegated to CoreML first, because the first run can take up to a few seconds.

barhanc:
## Description

Adds an `eagerLoadMethods` option (defaulting to `true`) to `loadModel()`:
- Calls `etModule_->load_method(methodName)` during `ModelHostObject` initialization for all exported methods.
- Pre-compiles backend delegates and allocates execution buffers upfront while the model is loading, eliminating the first-inference lag spike.
- Allows opting out via `loadModel(path, { eagerLoadMethods: false })`.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

- [ ] Build and run example apps to make sure nothing is broken.

### Related issues

Closes #1393

### Checklist

### DIFF
--- a/.agents/skills/verify-and-build/SKILL.md
+++ b/.agents/skills/verify-and-build/SKILL.md
@@ -97,7 +97,7 @@ Compiles staged (or explicitly passed) `cpp/` sources with the same flags clangd
 ```bash
 # The script takes file paths as positional arguments and exits 0 immediately
 # with none — it has no built-in file discovery. Pass files explicitly:
-find cpp/ -name '*.cpp' -o -name '*.h' | \
+find cpp/ -path cpp/tests -prune -o \( -name '*.cpp' -o -name '*.h' \) -print | \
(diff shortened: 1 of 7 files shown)


msluszniak:
I tested this and it worked correctly. However, I spotted two non-related issues. The first one is that demo apps currently links to the variants that are no longer on HF, we should update them. The second one is a tiny shift to the right on every instance segmentation I tested on rfdetr coreml variant. Could please double check if you have the same effect?

msluszniak (approved):
Created a separate issue for RF-DETR: https://github.com/software-mansion/react-native-executorch/issues/1406, let's fix demo apps in a separate PR.

PR #1428, opened 2026-09-07
docs(release): match the release flow to models.ts

msluszniak:
The v0.10 rewrite removed every path and constant the version-tag section named, so RELEASE.md now describes files that do not exist.

| RELEASE.md said | on `main` |
| --- | --- |
| `src/constants/modelUrls.ts` | gone, model URLs are in `src/models.ts` |
| `src/constants/versions.ts` | gone (only under `legacy/`) |
| `LIB_VERSION` derives the tag | no such constant; tags are literals |
| `VERSION_TAG` = in-development | it is the **published** tag (`resolve/v0.9.0`) |
| `PREVIOUS_VERSION_TAG` | does not exist; in-dev tag is `NEXT_VERSION_TAG` |

The constants were renamed **and their meanings swapped**, which is the dangerous part. Old step 2 says to make `VERSION_TAG` resolve to the upcoming release; doing that today repoints all 288 already-shipped model URLs onto an unpublished tag. That gets a caution block, not just a correction.

Also in this PR:

- adapter package paths corrected (`packages/bare-resource-fetcher`, `packages/expo-resource-fetcher`), plus a note that the publish workflow reads the version from `packages/react-native-executorch/package.json`
- a new **HuggingFace artifacts** section. The flow depended on moving release tags but never described it: publishing a `.pte` to a model repo changes nothing for users until the tag moves, and model repos are tagged independently of npm
- link to the model card standard in the spec repo

### Two things found while checking, not fixed here

Both are code, not docs, so they are left for separate PRs:

1. `packages/react-native-executorch/package.json` has `version: 0.0.0` on `main`, and that is the field the publish workflow reads.
2. `.github/workflows/npm-publish.yml` calls `./scripts/create-package.sh`, which does not exist on `main` (it is still on `release/0.9`). A stable publish from `main` looks like it would fail.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,20 +2,58 @@
 
 ## Version tag conventions
 
-Model URLs in `packages/react-native-executorch/src/constants/modelUrls.ts` are built from two constants in `versions.ts`:
+Model URLs in `packages/react-native-executorch/src/models.ts` are built from
+`BASE_URL` plus one of two tag constants defined in the same file:
 
-- `VERSION_TAG` — the **in-development** version (`resolve/v${LIB_VERSION}`). On `main` it points at the upcoming release; HuggingFace files under this tag may not be published until release day.
-- `PREVIOUS_VERSION_TAG` — the **latest published stable** release. Use for back-compat references that should follow the last shipped version.
+- `NEXT_VERSION_TAG` — the **in-development** tag (`resolve/v${MAJOR}.${NEXT_MINOR}.0`). Models
+  whose files are new or re-exported for the upcoming release point here. The
+  HuggingFace files under this tag may not exist until release day.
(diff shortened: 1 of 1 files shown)


barhanc (approved):

barhanc:
> One more thing here are the links to github we have in documentation like the ones at the bottom of each task
> ```
>Source Code
>View the implementation on GitHub:
>
>[src/extensions/cv/tasks/keypointDetection.ts ↗](https://github.com/software-mansion/react-native->executorch/blob/main/packages/react-native-executorch/src/extensions/cv/tasks/keypointDetection.ts)
>```
>these should probably be swapped on release to point to release tag, not main.

Nvm, the `build-versioned-docs.sh` already does that automatically.

PR #1450, opened 2026-09-09
fix(build): fail with the cause when the setup is wrong

msluszniak:
## Description

Setup failures that reach the user as an error pointing somewhere else. All reproduced against the published 0.10.0.

**Guards.** The native artifacts are downloaded by the postinstall hook, which pnpm 10 blocks by default (`Ignored build scripts`), as do `--ignore-scripts` and `npm ci --ignore-scripts`. `pod install` still succeeds, because CocoaPods never checks that a vendored framework exists, and the build dies later on `Build input files cannot be found`. The podspec and `build.gradle.kts` now check first and name `pnpm approve-builds` / `npm rebuild`.

**Config block in a monorepo.** It was read only from `INIT_CWD`, the workspace root, so a block in `apps/mobile/package.json` was ignored and an app declaring `"libs": []` got OpenCV anyway. The lookup now also walks up from the installed package and logs which manifest it read.

**Android `minSdkVersion`.** Default was 21, docs said Android 13+, nothing enforced either. `.note.android.ident` reads API 26 in all three shipped `.so` files, so anything lower shipped a library the linker is not guaranteed to load. Default is now 26.

**Docs.** A Troubleshooting page keyed by the exact error text covers the rest:

| setup | symptom |
| --- | --- |
| `use_frameworks!` dynamic | `transitive dependencies that include statically linked binaries` |
| another OpenCV pod | `frameworks with conflicting names: opencv2.xcframework` |
| Intel Mac, Rosetta, Intel CI | `None of the architectures in ARCHS (x86_64) are valid` |
| Expo without `expo-build-properties` | iOS 17 requirement is a warning only |
| default `reactNativeArchitectures` | `armeabi-v7a` / `x86` splits ship without the `.so` |

The OpenCV conflict has a verified escape hatch: `"libs": ["phonemis"]` drops `opencv-rne` and `pod install` then succeeds, at the cost of every vision task.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

`pod ipc spec react-native-executorch.podspec` in a checkout without artifacts prints the missing paths and the fix, and passes once they exist. 7 tests in `__tests__/api/nativeLibsConfig.test.ts` cover the config lookup. Docs build and lint clean.

### Checklist

### Additional notes

**The Gradle guard has not been run.** It configures at the top of `build.gradle.kts`, so please build an Android app on this branch before merging. Raising `minSdkVersion` can also break an app currently on 21-25; that build was already unsound, but it is a behaviour change.

### DIFF
--- a/docs/docs/01-fundamentals/01-getting-started.md
+++ b/docs/docs/01-fundamentals/01-getting-started.md
@@ -85,7 +85,9 @@ React Native ExecuTorch requires:
 
 - **New Architecture** enabled
 - **React Native 0.81+** or **Expo SDK 54+** with [Development Builds](https://docs.expo.dev/develop/development-builds/introduction/) (**Expo Go is not supported** due to custom C++ native libraries)
-- **iOS 17.0+** / **Android 13+**
+- **iOS 17.0+** / **Android 13+** (the shipped native libraries are built
+  against API 26, so `minSdkVersion` must be at least 26; the library defaults
+  it to that)
 
 For supported React Native versions, see the [Compatibility
 table](../05-other/01-compatibility.mdx). If an install or a build fails, see
--- a/docs/docs/01-fundamentals/01-getting-started.md
+++ b/docs/docs/01-fundamentals/01-getting-started.md
@@ -85,9 +85,7 @@ React Native ExecuTorch requires:
 
 - **New Architecture** enabled
 - **React Native 0.81+** or **Expo SDK 54+** with [Development Builds](https://docs.expo.dev/develop/development-builds/introduction/) (**Expo Go is not supported** due to custom C++ native libraries)
-- **iOS 17.0+** / **Android 13+** (the shipped native libraries are built
-  against API 26, so `minSdkVersion` must be at least 26; the library defaults
-  it to that)
+- **iOS 17.0+** / **Android 13+** (`minSdkVersion` >= 26)
 
 For supported React Native versions, see the [Compatibility
 table](../05-other/01-compatibility.mdx). If an install or a build fails, see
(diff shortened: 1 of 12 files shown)


docs/docs/01-fundamentals/01-getting-started.md
```diff
@@ -85,10 +85,14 @@ React Native ExecuTorch requires:
 
 - **New Architecture** enabled
 - **React Native 0.81+** or **Expo SDK 54+** with [Development Builds](https://docs.expo.dev/develop/development-builds/introduction/) (**Expo Go is not supported** due to custom C++ native libraries)
-- **iOS 17.0+** / **Android 13+**
+- **iOS 17.0+** / **Android 13+** (the shipped native libraries are built
+  against API 26, so `minSdkVersion` must be at least 26; the library defaults
+  it to that)
```
barhanc:
```suggestion
- **iOS 17.0+** / **Android 13+** (`minSdkVersion` >= 26)
```

docs/docs/05-other/02-troubleshooting.md
```diff
@@ -0,0 +1,147 @@
...
+
+```ruby
+use_frameworks! :linkage => :static
+```
+
+## `frameworks with conflicting names: opencv2.xcframework`
+
+Another pod in the app vendors its own OpenCV under the same framework name —
+`react-native-fast-opencv` (via `FastOpenCV-iOS`) is the common one. CocoaPods
+cannot install both.
```
barhanc: I'm wondering if we don't want to patch it some other way so it is possible to use rne with react-native-fast-opencv, as it wouldnt be that weird if someone wanted to use our lower level API for nn inference and opencv for some image transformations we don't support.

msluszniak: Ok, I solve this by linking to fast open cv in case of both fast open cv and rne are available.

barhanc (approved):

PR #1455, opened 2026-09-09
feat(llm): split a prefill that exceeds the model's per-call window

msluszniak:
## Description

Stacked on #1454, which it needs to see the real error. Draft: the native half is unverified on device, see below.

A dynamic-shape export bounds one prefill call to `get_max_seq_len`, the decoder's per-call window. That is not the conversation budget the model advertises as `get_max_context_len`, so a conversation well inside its context can still hand one call more tokens than the graph holds. The runtime rejects it:

```
LLMRunner.prefill: Failed: Error::NotSupported
Attempted to resize a bounded tensor with a maximum capacity of 262144 elements to 274432 elements
```

262144 is 256x1024 and 274432 is 268x1024: the call carried 268 tokens into a 256 window.

| model | get_max_seq_len | get_max_context_len | enable_dynamic_shape |
|---|---|---|---|
| lfm2.5-vl-450m vulkan | 256 | 2048 | true |
| lfm2.5-vl-1.6b vulkan | 256 | 2048 | true |
| gemma4-e2b vulkan | 128 | 2048 | true |
| lfm2.5-vl-450m xnnpack | 2048 | 2048 | false |

Prefill already appends at the runner's current position, so the same prompt delivered as several calls builds the identical cache. **Nothing in the upstream ExecuTorch runner changes.** The split is driven from TypeScript over two additions to our own binding:

- `getKVCacheState` gains `maxPrefillLen`. `maxSeqLen` resolves to the context, so a caller could not otherwise see the smaller number it must stay inside.
- `parsePrompt` accepts `{ kind: 'tokens' }`. A split is only faithful on token boundaries: cutting the string and letting the runtime re-encode the pieces is not equivalent when BPE merges across the cut, and a chunk could still overflow the window it was cut to fit. `MultimodalInput::Type::TOKENS` already exists upstream, so this is a passthrough.

A prompt that fits is sent in one call, byte for byte as today, and a model that declares no window loads no tokenizer. Every model currently working is on its existing path.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [ ] Android

### Testing instructions

`yarn test` in `packages/react-native-executorch`. Two suites, both verified to fail without the change:

- `__tests__/tasks/prefillChunking.test.ts` pins the split itself, including that the token stream is identical across it.
- `__tests__/tasks/llmChatSession.test.ts` drives a session against a runner that rejects an over-long call, which is the device failure. Reverting the chunking makes it fail with `Error::NotSupported (10 tokens over the 4 window)`.

The fake runner gained the window and enforces it; without that it accepted any prompt and could not show whether the caller stayed inside.

### Additional notes

**Why this is a draft.** The TS half is covered by tests, but the two C++ additions have not run on device: I have not rebuilt the native library against a Vulkan LLM and replayed the original failure. That is the check I would want before merge. The reproduction is a Galaxy S26 Ultra with `LFM2_5_VL_450M.VULKAN_8DA4W` and any conversation past 256 tokens.

**One judgement call worth review.** Tokenizing in TS means the ids come from our tokenizer rather than the one inside the runner, so a prompt could in principle encode differently. I confined that to prompts that would otherwise fail outright: a prompt that fits still goes down the text path untouched. If we would rather not have two tokenizers at all, the alternative is doing the same loop in C++ with the runner's own tokenizer, at the cost of moving the policy into the binding.

**Not addressed here.** `resetOnTurn` clears the KV cache but not `history`, so each turn re-prefills a longer conversation and grows the prompt (measured +85 tokens per turn on device). That is what pushed this model past its window on the fourth turn. Worth a separate look, since it also means a benchmark iteration is not a fixed workload.

### DIFF
--- a/packages/react-native-executorch/__tests__/support/fakeJsi.ts
+++ b/packages/react-native-executorch/__tests__/support/fakeJsi.ts
@@ -181,6 +181,11 @@ export type FakeRunnerProgram = {
   /** Context window the runner reports. Defaults to `128`. */
   maxSeqLen?: number;
   /**
+   * Per-call prefill window, as a dynamic-shape export declares. Defaults to
+   * `0`, meaning the model declares none and a prompt goes in one call.
+   */
+  maxPrefillLen?: number;
+  /**
    * Responses handed out in order, one per `generate` call. A runner that runs
    * past the end of the list repeats the last entry, so a test only has to
    * script the turns it cares about.
@@ -207,6 +212,17 @@ const promptText = (prompt: unknown): string =>
       ? prompt.filter((part) => typeof part === 'string').join('')
       : '';
 
+/** Token ids carried by a prompt's `tokens` parts, which text parts do not cover. */
+const promptTokenCount = (prompt: unknown): number =>
+  Array.isArray(prompt)
+    ? prompt.reduce(
+        (sum, part) =>
+          sum +
+          (part && typeof part === 'object' && part.kind === 'tokens' ? part.tokens.length : 0),
+        0
+      )
+    : 0;
+
 /**
  * Builds a fake LLM runner.
  *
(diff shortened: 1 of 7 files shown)


msluszniak:
Closing this in favour of #1458 comprehensive solution.

PR #1463, opened 2026-09-10
feat(segmentation): accept an index-map output from Core ML models

### LINKED ISSUE #1459: perf(segmentation): return an index map instead of 22.7 MB of per-class logits
Semantic segmentation returns full-resolution logits for every class. On the 21-class models that is **22.71 MB per inference**, against a 3.24 MB input. The task then reduces those logits away immediately, so nothing downstream ever sees them.

## Evidence that this is the bottleneck

Measured on `deeplab-v3-mobilenet-v3-large`:

- **It barely scales with hardware.** Core ML fp16 runs 41.38 ms on an iPhone SE 3 (A15), 30.59 on an iPhone 16 (A18), 30.08 on an iPhone 17 (A19). That is 1.38x over four generations, where XNNPACK int8 gains 1.92x over the same span. Compute scales, memory bandwidth does not.
- **The compute unit barely matters.** Sweeping all four Core ML compute units moves it 12% end to end (`CPU_AND_NE` 30.64, `ALL` 30.75, `CPU_AND_GPU` 33.77, `CPU_ONLY` 34.42). For comparison, the same sweep on `ssdlite320-mobilenet-v3-large` moves it 75%. If the math were the cost, moving the math would move the number.
- **Both backends pay it equally,** which is why Core ML is only on par with XNNPACK here (30.08 vs 29.39 ms on iPhone 17, the single family in the whole iOS estate where Core ML does not win). Same 520x520 input, same `(1,21,520,520)` output, both fully delegated: Core ML 1 delegate / 0 portable ops, XNNPACK 1 delegate / 1 portable op.

## The logits are already thrown away

`semanticSegmentation.ts` consumes all K channels only to reduce them:

```ts
tOutput
  .copyTo(tReshape)
  .through(toChannelsLast, tChanLast)
  .through(argmax, tMask, -1)          // the 21 channels die here
  .through(applyColormap, tRgba, colormapData);
```

and the public result carries no per-class scores:

```ts
type SemanticSegmentationResult<L> = {
  readonly buffer: ImageBuffer;
  readonly colormap?: ColorMap<L>;
};
```

So moving the argmax into the exported graph costs the caller nothing. It relocates a reduction that already happens, and additionally deletes the `toChannelsLast` transpose over 5.7M floats.

## Proposed change

Export an index map instead of logits: `(1, 520, 520)` int32 rather than `(1, 21, 520, 520)` float32. **22.71 MB -> 1.08 MB, a 21x cut.**

Affected models, all identical in shape:

| model | current output | size |
|---|---|---|
| deeplab-v3-mobilenet-v3-large | (1,21,520,520) f32 | 22.71 MB |
| deeplab-v3-resnet50 | (1,21,520,520) f32 | 22.71 MB |
| deeplab-v3-resnet101 | (1,21,520,520) f32 | 22.71 MB |
| fcn-resnet50 | (1,21,520,520) f32 | 22.71 MB |
| fcn-resnet101 | (1,21,520,520) f32 | 22.71 MB |
| lraspp-mobilenet-v3-large | (1,21,520,520) f32 | 22.71 MB |

`selfie-segmentation` is unaffected: it is single-channel at 256x256 (0.26 MB) and takes the sigmoid path.

## What has to change together

The runtime does not accept this shape today, so the export and the task must land together.

1. `validateSpec` currently pins the contract to `[f32(1,'K','H','W')]` / `[f32('K','H','W')]`. Needs a third variant for an index output.
2. A branch that skips `toChannelsLast` and `argmax` and goes straight to `applyColormap`.
3. The class count has to move into config. `K` is currently read off the output shape, and the `labels.length !== nClasses` guard depends on it.

## Trap to avoid

Do **not** emit the index map as `(1, 1, H, W)` float32. That validates against the existing spec, binds `nClasses = 1`, and silently takes the binary `sigmoid` / `normalize` / `GRAY2RGBA` branch, producing a greyscale ramp of class indices instead of a colour mask. It fails as wrong output, not as an error. An `int32` output fails loudly with `SCHEMA_MISMATCH`, which is the behaviour we want while the two sides are out of step.

## Open question to settle first

`argmax` over the channel dim has to lower to the delegate. If it does not, `lower_full_graph=False` in the segmentation exporters will silently leave it as a portable CPU op over 5.7M floats, which could cost more than the transfer it saves. Cheap to check: probe the exported `.pte` and confirm 1 delegate / 0 portable ops before measuring anything.

A smaller fallback if the contract has to keep logits: emit them as fp16 rather than float32, which halves the output to 11.4 MB for no accuracy the ANE is not already discarding.

## Measurements

iPhone 16 (A18, iOS 26.6.1), release build, RNE benchmark harness, one run per compute unit, pipeline median. Compute unit sweep exported on ExecuTorch 1.4.1 / coremltools 9.0. Device-generation figures are from the v0.10.0 estate runs in #1363.

msluszniak:
## Description

Core ML segmentation models can now return a `(1,H,W)` int32 index map instead of `(1,K,H,W)` float32 logits. The task already reduced those logits away immediately, so nothing downstream loses information.

Measured on an iPhone 16, all six Core ML segmentation models, 10 iterations after 2 warmups:

| model | before ms | after ms | speedup | peak MB | mem cut |
|---|---|---|---|---|---|
| lraspp-mobilenet-v3-large | 12.63 | 7.75 | 1.63x | 250.97 -> 177.72 | 29% |
| deeplab-v3-mobilenet-v3-large | 31.16 | 25.42 | 1.23x | 253.53 -> 195.25 | 23% |
| fcn-resnet50 | 53.66 | 47.95 | 1.12x | 320.86 -> 297.25 | 7% |
| fcn-resnet101 | 70.65 | 65.41 | 1.08x | 364.05 -> 302.50 | 17% |
| deeplab-v3-resnet50 | 106.29 | 100.85 | 1.05x | 309.00 -> 161.39 | 48% |
| deeplab-v3-resnet101 | 123.08 | 117.51 | 1.05x | 236.09 -> 162.69 | 31% |

The saving is a near constant 5 to 5.7 ms because the argmax cost depends only on the output shape, which is identical across all six. That is a large relative win on the light models and a small one on the heavy backbones, while the memory saving lands everywhere.

The re-exported artifacts are published under the `v0.11.0` tag; this branch points the six Core ML models at it. XNNPACK stays on `v0.10.0`.

### The part worth knowing

`torch.argmax` is the wrong way to do this and makes it **2x slower** (31 -> 65 ms). Core ML lowers it to `ios18.reduce_argmax`, which has no ANE kernel and is pinned to the CPU, so the ANE writes all 21 channels out anyway and the CPU reduces 5.7M elements. The compute plan shows it directly: 194 ANE ops and 1 CPU op.

Building the same reduction from `reduce_max` and elementwise ops keeps it on the ANE (200 ANE ops, 1 CPU op, a cast over 270K elements) and is pixel-exact against `torch.argmax`, ties included.

This is Core ML only. XNNPACK delegates neither `argmax` nor `amax`/`eq`, so the same rewrite splits the graph into 3 delegates and measures **1.19x slower** (160.04 -> 190.21 ms). XNNPACK variants keep the logits contract.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

With a local index-map `.pte`, `useSemanticSegmenter` resolves the `batchedIndex` schema variant and renders through the colormap branch without the JS argmax. Models emitting logits are unaffected. Please A/B against `main` to confirm the new models are actually faster.

### Related issues

Closes #1459, with a correction: the issue attributed the cost to output bandwidth, and its proposed implementation would have made things slower. The win comes from keeping the reduction on the ANE.

### DIFF
--- a/packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
+++ b/packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
@@ -253,18 +253,14 @@ export async function createSemanticSegmenter<L extends PropertyKey = string>(
 
         const colormapData = modelOpts.labels.map((l) => returnColormap![l]);
 
-        if (indexMap) {
-          // The graph already did the argmax, so there is nothing to reduce:
-          // reshape the (1,H,W) index map to the (H,W,1) the colormap wants.
-          // This also skips the toChannelsLast transpose over H*W*K floats.
-          tOutput.copyTo(tMask).through(applyColormap, tRgba, colormapData);
-        } else {
-          tOutput
-            .copyTo(tReshape)
-            .through(toChannelsLast, tChanLast)
-            .through(argmax, tMask, -1)
-            .through(applyColormap, tRgba, colormapData);
-        }
+        // An index-map model already did the argmax in the graph, so both the
+        // transpose over H*W*K floats and the reduction are skipped and the
+        // (1,H,W) indices go straight into the (H,W,1) the colormap wants.
+        tOutput
+          .copyTo(indexMap ? tMask : tReshape)
+          .throughIf(!indexMap, toChannelsLast, tChanLast)
+          .throughIf(!indexMap, argmax, tMask, -1)
+          .through(applyColormap, tRgba, colormapData);
       } else {
         tOutput
           .copyTo(tReshape)
--- a/packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
+++ b/packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
@@ -7,7 +7,7 @@ import type { WorkletRuntime } from 'react-native-worklets';
 
 import { tensor } from '../../../core/tensor';
 import { loadModel } from '../../../core/model';
-import { validateSpec, method, f32 } from '../../../core/schema';
+import { validateSpec, method, f32, i32 } from '../../../core/schema';
 import { wrapAsync } from '../../../core/runtime';
 
 import type { ImageBuffer } from '../image';
@@ -168,11 +168,40 @@ export async function createSemanticSegmenter<L extends PropertyKey = string>(
         [f32(3, 'H', 'W')],
         [f32('K', 'H', 'W')]
       ),
+      // Models that run the argmax inside the graph return the class index per
+      // pixel instead of a logit per class: (1,H,W) int32 rather than
+      // (1,K,H,W) float32, which is 22.71 MB -> 1.08 MB on the 21-class models.
+      // K cannot be read off the shape here, so it comes from the labels.
+      batchedIndex: method(
+        'forward', // prettier-ignore
+        [f32(1, 3, 'H', 'W')],
+        [i32(1, 'H', 'W')]
+      ),
+      unbatchedIndex: method(
+        'forward', // prettier-ignore
+        [f32(3, 'H', 'W')],
+        [i32('H', 'W')]
+      ),
     });
 
-    const [nClasses, H, W] = dims.constant('K', 'H', 'W');
-    const inpShape = { batched: [1, 3, H, W], unbatched: [3, H, W] }[variant];
-    const outShape = { batched: [1, nClasses, H, W], unbatched: [nClasses, H, W] }[variant];
+    // An index-map model has no K in its output shape, so the class count comes
+    // from the labels the caller configured. The logit models keep reading it
+    // off the shape, which also keeps the labels/classes guard meaningful.
+    const indexMap = variant === 'batchedIndex' || variant === 'unbatchedIndex';
+    const [H, W] = dims.constant('H', 'W');
+    const nClasses = indexMap ? modelOpts.labels.length : dims.constant('K')[0]!;
+    const inpShape = {
+      batched: [1, 3, H, W],
+      unbatched: [3, H, W],
+      batchedIndex: [1, 3, H, W],
+      unbatchedIndex: [3, H, W],
+    }[variant];
+    const outShape = {
+      batched: [1, nClasses, H, W],
+      unbatched: [nClasses, H, W],
+      batchedIndex: [1, H, W],
+      unbatchedIndex: [H, W],
+    }[variant];
 
     // Generate highly distinct, high-contrast colors, see:
     // https://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
@@ -189,7 +218,7 @@ export async function createSemanticSegmenter<L extends PropertyKey = string>(
     }
 
     const tensors = [
-      tensor('float32', outShape),
+      tensor(indexMap ? 'int32' : 'float32', outShape),
       tensor('float32', [nClasses, H, W]),
       tensor('float32', [nClasses, H, W]),
       tensor('float32', [H, W, nClasses]),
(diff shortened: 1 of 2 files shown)


packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -224,11 +253,18 @@ export async function createSemanticSegmenter<L extends PropertyKey = string>(
...
+          // The graph already did the argmax, so there is nothing to reduce:
+          // reshape the (1,H,W) index map to the (H,W,1) the colormap wants.
+          // This also skips the toChannelsLast transpose over H*W*K floats.
+          tOutput.copyTo(tMask).through(applyColormap, tRgba, colormapData);
+        } else {
+          tOutput
+            .copyTo(tReshape)
+            .through(toChannelsLast, tChanLast)
+            .through(argmax, tMask, -1)
+            .through(applyColormap, tRgba, colormapData);
```
msluszniak: can we use throughIf two times here instead of that if?

msluszniak: Done in 46cc49ef8. throughIf returns the tensor unchanged when the predicate is false, so both paths fold into one chain:

```ts
tOutput
  .copyTo(indexMap ? tMask : tReshape)
  .throughIf(!indexMap, toChannelsLast, tChanLast)
  .throughIf(!indexMap, argmax, tMask, -1)
  .through(applyColormap, tRgba, colormapData);
```

The copyTo target still has to switch, since an index map lands straight in tMask while logits go through tReshape.

packages/react-native-executorch/src/models.ts
```diff
@@ -172,12 +172,16 @@ function family<V extends Record<string, { readonly DEFAULT: unknown }>>(
...
-// because noUnusedLocals rejects a constant nothing reads.
+// Models resolve through VERSION_TAG, the latest published stable tag, unless
+// they have been re-exported for the next release, in which case their URL moves
+// over to NEXT_VERSION_TAG. Both tags are pinned snapshots, so a model only
+// changes when its URL is moved here.
 const VERSION_TAG = 'resolve/v0.10.0';
+// Semantic segmentation models re-exported for 0.11: their Core ML variants now
+// emit an index map instead of per-class logits (see the `batchedIndex` and
+// `unbatchedIndex` schema variants). The XNNPACK variants are unchanged and stay
+// on VERSION_TAG.
```
```suggestion
```

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -168,11 +168,40 @@ export async function createSemanticSegmenter<L extends PropertyKey = string>(
...
+      ),
     });
 
-    const [nClasses, H, W] = dims.constant('K', 'H', 'W');
-    const inpShape = { batched: [1, 3, H, W], unbatched: [3, H, W] }[variant];
-    const outShape = { batched: [1, nClasses, H, W], unbatched: [nClasses, H, W] }[variant];
+    // An index-map model has no K in its output shape, so the class count comes
+    // from the labels the caller configured. The logit models keep reading it
+    // off the shape, which also keeps the labels/classes guard meaningful.
+    const indexMap = variant === 'batchedIndex' || variant === 'unbatchedIndex';
```
barhanc: I think a name like `returnsIndex` would be better here.

packages/react-native-executorch/src/extensions/cv/tasks/semanticSegmentation.ts
```diff
@@ -168,11 +168,40 @@ export async function createSemanticSegmenter<L extends PropertyKey = string>(
...
+    const [H, W] = dims.constant('H', 'W');
+    const nClasses = indexMap ? modelOpts.labels.length : dims.constant('K')[0]!;
```
barhanc: Here maybe better use the single dim accessor.

barhanc (approved):

PR #1465, opened 2026-09-11
feat(models): serve the Core ML text embedders as int8

### LINKED ISSUE #1419: Test int8 variants for model exported to CoreML with `CPU_ONLY` compute unit
As in the title, need to sweep over these models and check if int8 variants are faster or not.

msluszniak:
## Description

The five CPU_ONLY Core ML text embedders gain an int8 variant: weight-only, per-channel, including the embedding table. `COREML_INT8` is declared ahead of `COREML_FP16`, so it is the Core ML variant `DEFAULT` picks. The fp16 variants stay on `v0.10.0` for anyone naming them explicitly.

distiluse is the exception: it pins `{ ios: 'MLX_INT8' }`, so its iOS default is unchanged and the int8 Core ML build is only what an explicit `COREML_INT8` gets. The pin is left alone, since nothing here re-measured MLX.

Measured on an iPhone 16, release build, 3 warmup + 20 timed:

| model | fp16 ms | int8 ms | fp16 MB | int8 MB | download |
|---|---|---|---|---|---|
| all-MiniLM-L6-v2 | 1.256 | 1.277 | 122.11 | 102.34 | 45.5 -> 22.8 MB |
| multi-qa-MiniLM-L6-cos-v1 | 1.283 | 1.268 | 123.24 | 91.56 | 45.5 -> 22.8 MB |
| paraphrase-multilingual-L12 | 3.004 | 2.979 | 431.55 | 324.83 | 235.6 -> 118.4 MB |
| distiluse-base-multilingual-v2 | 6.966 | 6.909 | 361.22 | 243.91 | 270.6 -> 135.6 MB |
| clip-vit-base-patch32-text | 5.152 | 5.190 | 213.16 | 152.44 | 127.4 -> 63.9 MB |

Latency is flat, peak memory falls 17-26%, every download halves. Cosine against the torch fp32 reference is 0.99940 or better on all five.

### The part worth knowing

ExecuTorch applies `op_linear_quantizer_config` with `op_type_configs={"gather": None}`, exempting the embedding table. Right for a transformer, backwards for a sentence embedder: paraphrase-L12 is 214.1 MB with the exemption and 118.4 MB without, and the table costs nothing to compress (latency flat, cosine 0.99990).

Upstream takes the opt-in as a compile spec in pytorch/executorch#22728. Until that lands, export-scripts strips the exemption itself and asserts the patch took.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [ ] Android

### Testing instructions

The published int8 artifacts were benchmarked on device, so the table above measures the exact files these URLs serve. Please A/B against `main` to confirm the new models are actually faster and smaller. Android is unaffected: only Core ML URLs moved.

### Related issues

Models published under the `v0.11.0` tag. Export side: export-scripts!25.


packages/react-native-executorch/src/models.ts
```diff
@@ -172,12 +172,14 @@ function family<V extends Record<string, { readonly DEFAULT: unknown }>>(
...
-// one by the first model re-exported for 0.11, and each such model's URL moves
-// over to it; the constant is absent until then rather than declared and unused,
-// because noUnusedLocals rejects a constant nothing reads.
+// Models resolve through VERSION_TAG, the latest published stable tag, unless
+// they have been re-exported for the next release, in which case their URL moves
+// over to NEXT_VERSION_TAG. Both tags are pinned snapshots, so a model only
+// changes when its URL is moved here.
 const VERSION_TAG = 'resolve/v0.10.0';
+// The Core ML text embedders re-exported for 0.11: weight-only int8 including
+// the embedding table, which halves every file at unchanged latency.
```
```suggestion
```

barhanc (approved):

PR #1466, opened 2026-09-11
fix(libs): stop shipping backends the app opted out of

### LINKED ISSUE #1464: Backend opt-out doesn't shrink the app — `core-*` artifacts ship every backend on both platforms
## Summary

The `backends` / `libs` / `features` opt-out in `package.json` gates **which tarballs are downloaded** and **which backends get linked**. It does not control what the `core-*` artifacts contain, and on Android it does not control what ships.

Both `core-android-<abi>.tar.gz` and `core-ios.tar.gz` already carry every backend, contradicting the contract documented in `download-libs.js`. From there the platforms diverge:

- **Android** — nothing prunes after extraction and `jniLibs.srcDirs` packages the directory, so an app declaring `backends: ["vulkan"]` ships **2 673 072 B (2.67 MB) of XNNPACK as dead weight**: linked by nothing, registered by nothing, loaded by nothing.
- **iOS** — the podspec gates at link time, so the extra binaries never reach the app. Instead **36.03 MB of device-slice `.a` is downloaded and never used**. Wasted bandwidth rather than wasted app size.

The iOS approach is the correct one. The concrete ask is to fix the artifacts on both platforms, and to make Android decide from the resolved config the way the podspec does.

Checked against `v0.10.0-libs`. Android figures are arm64-v8a; iOS figures are the `ios-arm64` slice unless noted.

---

## 1. The `core-*` artifacts contradict their documented contract

`scripts/download-libs.js` lines 12–14 describe them as backend-free:

```
core-android-arm64-v8a.tar.gz    -- executorch, pthreadpool, cpuinfo for arm64
core-android-x86_64.tar.gz       -- executorch for x86_64
core-ios.tar.gz                  -- ExecutorchLib.xcframework (without xnnpack/coreml)
```

Neither holds.

**Android** — `core-android-arm64-v8a.tar.gz` (8.35 MB gz):

```console
$ tar tzvf core-android-arm64-v8a.tar.gz | grep -v '/\._'
-rw-r--r-- msluszniak/staff     98123 2026-08-31 10:13 ./executorch.jar
-rwxr-xr-x msluszniak/staff   2673072 2026-08-31 10:13 ./executorch/arm64-v8a/libxnnpack_executorch_backend.so
-rwxr-xr-x msluszniak/staff  11736456 2026-08-31 10:13 ./executorch/arm64-v8a/libvulkan_executorch_backend.so
-rwxr-xr-x msluszniak/staff  13451672 2026-08-31 10:13 ./executorch/arm64-v8a/libexecutorch.so
```

Byte-identical across both `-libs` tags (same mtime), so not a one-off bad upload.

**iOS** — `core-ios.tar.gz` is 21.84 MB gz, **76.59 MB unpacked**, and contains the xcframework *plus* a full `libs/executorch/` tree with xnnpack, coreml, mlx and mps in it:

```
libbackend_mlx_ios.a                15 088 280 B   15.09 MB
ExecutorchLib (ios-arm64-simulator)  9 204 112 B    9.20 MB
ExecutorchLib (ios-arm64)            8 919 824 B    8.92 MB
libexecutorch_llm_simulator.a        6 686 704 B    6.69 MB
libexecutorch_llm_ios.a              6 676 168 B    6.68 MB
libkernels_optimized_simulator.a     6 675 648 B    6.68 MB
libkernels_optimized_ios.a           6 473 400 B    6.47 MB
libbackend_xnnpack_ios.a             3 788 232 B    3.79 MB
libbackend_xnnpack_simulator.a       3 641 696 B    3.64 MB
mlx.metallib                         1 068 302 B    1.07 MB
libbackend_coreml_ios.a                931 128 B    0.93 MB
libbackend_coreml_simulator.a          930 872 B    0.93 MB
libbackend_mps_ios.a                   405 328 B    0.41 MB
libbackend_mps_simulator.a             404 752 B    0.40 MB
… plus kleidiai, torchao, quantized and llm kernels, threadpool
```

## 2. The standalone backend artifacts are therefore redundant

On Android, `xnnpack-android-*` and `vulkan-android-*` only ever overwrite a file core already placed.

On iOS the duplication is exact — `xnnpack-ios`, `coreml-ios` and `mlx-ios` contain nothing new, just a different layout (xcframework vs. loose `.a`):

```
7290cadabf11b7bd548c3e39a3063bfd5292dd8e595483337054e6e6636af5aa
  core-ios/libs/executorch/libbackend_mlx_ios.a
  mlx-ios/MLXBackend.xcframework/ios-arm64/libMLXBackend.a

2eef3d55ca7bc697f81b6c8d83ad8035f01bcac635bceba54de4d3471ce7f3d9
  core-ios/libs/executorch/mlx.metallib
  mlx-ios/libs/executorch/mlx.metallib
```

Same sizes for the xnnpack and coreml pairs. An iOS app with everything enabled downloads `libMLXBackend.a` twice — 15 MB, twice.

## 3. Nothing prunes after extraction

`extract()` is a bare untar into `destDir`:

```js
function extract(tarball, destDir) {
  ensureDir(destDir);
  execSync(`tar -xzf "${tarball}" -C "${destDir}"`);
}
```

No filtering, no removal of backends absent from the resolved config, on either platform.

## 4. Android: Gradle packages the directory, not the link graph

`android/build.gradle.kts:97`

```kotlin
jniLibs.srcDirs("../third-party/android/libs/executorch")
```

Gradle bundles **every** `.so` under a `jniLibs` source dir. `-DRNE_ENABLE_XNNPACK=false` stops CMake from importing and linking it — correct and sufficient for behaviour, since the load-time constructor never runs and the backend is never registered — but it has no bearing on packaging.

The CMakeLists comment (lines 75–79) states the intended model:

> Backends ship as separate .so files. When enabled, CMake imports the .so and links libreact-native-executorch.so against it; Gradle packages the .so into the APK (via jniLibs srcDir) […]

"When enabled" holds for the linking half only. The packaging half is unconditional.

## 5. iOS: the podspec gets this right

`react-native-executorch.podspec` decides from the resolved config, at link time:

```ruby
if enable_xnnpack
  physical_ldflags  << "-force_load \"#{xnnpack_xcframework_path}/ios-arm64/libXnnpackBackend.a\""
  simulator_ldflags << "-force_load \"#{xnnpack_xcframework_path}/ios-arm64-simulator/libXnnpackBackend.a\""
end
```

`vendored_frameworks` is only `ExecutorchLib.xcframework`, and nothing under `libs/executorch/` is referenced except `libthreadpool_*.a`. The stray `.a` files from §1 are inert.

Net: 36.03 MB of unreferenced device-slice `.a` per install — bandwidth and disk, not app size.

This is the model §4 should adopt. The problem on Android isn't a missing check, it's that the shipping decision is delegated to "whatever happens to be in the directory."

## 6. Stale binaries survive a config change

Because nothing cleans the extraction dir and `isCacheValid()` short-circuits re-download, flipping a backend **off** after a previous install leaves the old binary in place. On Android it keeps shipping; the opt-out silently does nothing until someone deletes the tree by hand. Wiping the target dir before extracting core would cover both platforms.

## 7. Two different Vulkan binaries exist, and config decides which one ships

`libxnnpack_executorch_backend.so` is byte-identical between `core-` and `xnnpack-`:

```
c8eb1d09ef050a6ef521d7d066dcc2cc5b6edaf0d4ba09e916d3c85ed3e2698d  (both)
```

`libvulkan_executorch_backend.so` is not:

| source | size (B) | Build ID | sha256 |
| --- | --- | --- | --- |
| `core-android-arm64-v8a` | 11 736 456 | `056eaac5d089c55f153db23679f4c9686c2cb9cc` | `5da20453…` |
| `vulkan-android-arm64-v8a` | 11 737 528 | `3ab7f8ac3ed334e71b1d971d94a88e79529dc66d` | `4959cc48…` |

1072 B apart, 6 848 406 bytes differing. Since XNNPACK came out identical, separate CI runs don't explain it on their own — this looks like non-determinism in the Vulkan build, plausibly shader codegen / SPIR-V embedding order (`.rodata` is 74% of that library, §11).

Consequence: `getArtifacts()` pushes `core-` before `vulkan-` and the loop extracts in order, so with Vulkan **enabled** the standalone build wins and with Vulkan **disabled** the core copy is what sits there. A config flag that shouldn't affect binary identity decides which binary users get. The release notes already say a full rebuild isn't reproducible for every artifact; this is a concrete instance worth pinning down.

> **Root cause found, and it is not non-determinism.** The two libraries differ in **symbol names**, which no amount of build non-determinism produces. Symbol count is 4 109 either way:
>
> | only in `core-android-arm64-v8a` | only in `vulkan-android-arm64-v8a` |
> | --- | --- |
> | `vkcompute::add_linear_qcs8w_node` | `vkcompute::reduce_nworkers` |
> | `vkcompute::linear_qcs8w_local_wg_size` | `vkcompute::softmax_nworkers` |
> | `vkcompute::linear_qcs8w_global_wg_size` | `vkcompute::reduce2d_local_wg_size` |
>
> They are **two different source revisions of the fork**. The standalone artifact is the newer one: it carries the reduce/softmax workgroup-sizing work and has dropped the weight-only int8 linear that was found dead. `core-android` still holds the build from before all of that. The section deltas follow from it — `.text` −1 768, `.rodata` +2 344 (different shaders embedded), `.dynstr` −118 (the renamed symbols) — so the `.rodata`/SPIR-V-ordering hypothesis above is refuted.
>
> How it happened: a previous refresh republished only `vulkan-android-*` and left the stale copy inside `core-android-*`, which is the standing practice of touching only the tarballs a change affects. It went unnoticed precisely because of the extraction order described above: a Vulkan-enabled app always got the newer build on top.
>
> **Closed by §1 rather than fixed on its own.** Core carries no Vulkan binary now, so the standalone build is the only one that can ship and the stale copy is gone. Worth keeping in mind for future partial refreshes: while a backend lived in two tarballs, refreshing one silently left the other behind.

## 8. `-force_load` on iOS means the backend `.a` size *is* the cost

Force-loaded archives aren't dead-stripped, so the archive size transfers to the binary essentially whole:

| component | link mode | cost |
| --- | --- | --- |
| `ExecutorchLib.xcframework` | vendored framework, normal link | 8.92 MB on disk, dead-strippable |
| `libthreadpool_ios.a` | `-force_load` | 0.17 MB, full |
| XNNPACK | `-force_load` | 3.79 MB, full |
| CoreML | `-force_load` | 0.93 MB, full |
| MLX | `-force_load` + `s.ios.resource` | 15.09 MB + 1.07 MB metallib, full |

**MLX is the heaviest single artifact in the release** — larger than the Vulkan backend on Android, and none of it strippable. It is also on by default: an app with no `react-native-executorch` block gets every backend, so a default iOS install carries all 15 MB of it. But that's the intended behaviour. It should download all by default.

## 9. MPS ships with no toggle

`libbackend_mps_ios.a` (0.41 MB device / 0.40 MB simulator) is in `core-ios`, but:

```js
const ALL_BACKENDS = ['xnnpack', 'coreml', 'mlx', 'vulkan'];
```

No `mps`. The podspec doesn't reference it either, so it's inert — an undocumented backend riding along with no way to enable or exclude it. Either wire it up or drop it from the build.

## 10. iOS simulator slices are unconditional

> **Updated after the §1 fix (v0.10.1-libs).** Most of this section's weight was the simulator halves of the archives §1 removed anyway. Measured in the new `core-ios`, the simulator content is now **9.37 MB** (`ExecutorchLib` simulator slice 8.78 MB + `libthreadpool_simulator.a` 0.16 MB), down from 28.99 MB. The remaining prize is roughly 3 MB compressed, and only for someone who opts into device-only.
>
> Splitting what is left is also more invasive than it first looks: `ExecutorchLib.xcframework/Info.plist` enumerates both slices, so a split needs the plist rewritten per tarball and re-merged at extraction time when both are present. A wrong plist breaks every iOS build. Left open deliberately, at much reduced value.

`core-ios` carries 21.20 MB of `*_simulator.a` plus the simulator slice of `ExecutorchLib`. Every install pays for it, including device-only CI. An `RNET_DEVICE_ONLY=1` escape hatch, or splitting simulator slices into their own tarball, would mirror the per-backend split that already exists.

## 11. Section breakdown, for whoever picks up binary-size work

Android arm64-v8a, all three `.so` already `strip`ped:

```
libexecutorch.so                13 451 672 B   (gzip -9: 4.61 MB)
  .text                          9 553 248
  .dynstr                        1 098 541
  .eh_frame                        829 940
  .rodata                          550 096

libvulkan_executorch_backend.so 11 736 456 B   (gzip -9: 2.65 MB)
  .rodata                        8 731 588   ← 74%, effectively embedded SPIR-V
  .text                          2 020 296
  .eh_frame                        211 620
  .gcc_except_table                118 516

libxnnpack_executorch_backend.so 2 673 072 B   (gzip -9: 0.93 MB)
  .text                          1 755 376
  .eh_frame                        238 008
```

- **Vulkan is ~4.4× XNNPACK and almost entirely data.** Only ~2 MB is code. Size work belongs in shader variant count, compression, or lazy loading — not in the C++.
- **`.eh_frame` totals 1 279 568 B (~1.28 MB) across the three.** ExecuTorch 1.2 introduced `.eh_frame` suppression among its binary-size options and these are built from 1.4.1, so it may simply be unset here. `.gcc_except_table` is present too.

> **Measured, and decided against.** Counting `.eh_frame_hdr` and `.gcc_except_table` as well, all the unwind data comes to **1 781 000 B (1.70 MB)**:
>
> | library | `.eh_frame` | `.eh_frame_hdr` | `.gcc_except_table` | share of file |
> | --- | --- | --- | --- | --- |
> | `libexecutorch.so` | 829 940 | 126 748 | 156 932 | 8.3% |
> | `libvulkan_executorch_backend.so` | 211 620 | 32 684 | 118 516 | 3.1% |
> | `libxnnpack_executorch_backend.so` | 238 008 | 43 132 | 23 420 | 11.4% |
>
> That is 4.7% of the 36.38 MB stripped APK payload for xnnpack+vulkan, and unwind tables are highly redundant, so the compressed share is smaller again.
>
> **`.eh_frame` is what unwinders use to build native stack traces.** Suppressing it degrades every native crash report from Play Console and Crashlytics — on a library whose hard bugs are native — to buy well under half a megabyte of download. That is not a trade worth making, so this is closed as won't-do rather than left open.
>
> Two further findings from the assessment:
>
> - `libexecutorch.so` **defines and exports** `__cxa_throw` and `__cxa_begin_catch` as `GLOBAL FUNC`: it statically links libc++'s exception runtime and re-exports it. `libRnExecutorch.so` imports both, and its `DT_NEEDED` lists `libexecutorch_jni.so` *before* `libc++_shared.so`, so RNE's own `throw` can bind into libexecutorch.so's copy. Suppressing unwind info there could therefore break **RNE's** error paths, not just ExecuTorch's. Proven at the symbol level; the runtime binding is unverified. Two C++ ABI runtimes in one process is a hazard worth its own issue regardless of §11.
> - The Vulkan `.rodata` observation holds: 8 731 588 B of 11 737 528 is data, ~74%. Size work there belongs in shader variant count or compression, as the section says.

## 12. AppleDouble junk in every tarball

Packed on macOS without `COPYFILE_DISABLE`, so each carries `._*` sidecars:

| tarball | `._` entries | bytes |
| --- | --- | --- |
| `core-android-arm64-v8a` | 7 | 978 |
| `vulkan-android-arm64-v8a` | 4 | 489 |
| `xnnpack-android-arm64-v8a` | 4 | 489 |

The iOS tarballs have the same. Negligible in size, but they get extracted alongside the real libraries and are noise for any tooling that globs those directories. `COPYFILE_DISABLE=1 tar …` fixes it. The `msluszniak/staff` owner in the tar headers also suggests these are packed from a workstation rather than CI, which ties back to §7.

## 13. Sizes, and the docs gap

There's no bundle-size page for RNE (cf. [react-native-skia's](https://shopify.github.io/react-native-skia/docs/getting-started/bundle-size)). It's the most common question for an on-device AI library, and today the only way to answer it is to download release artifacts and run `size -A`.

**Android, measured** (arm64-v8a; `extractNativeLibs=false` is the RN 0.73+ default, so the uncompressed figure is what the install grows by; AAB delivery compresses roughly 3:1):

| config | in APK (uncompressed) | ≈ Play download |
| --- | --- | --- |
| xnnpack only | 16.12 MB | 5.54 MB |
| vulkan only | 25.19 MB | 7.26 MB |
| xnnpack + vulkan | 27.86 MB | 8.19 MB |

With the current artifacts **every one of these rows is actually 27.86 MB**, because of §1 + §3 + §4. The table is the post-fix state.

**iOS, estimated** — lower confidence, `ExecutorchLib` is the one dead-strippable piece and I haven't archived a build:

| config | est. binary contribution |
| --- | --- |
| xnnpack only | ~8–11 MB |
| xnnpack + coreml | ~9–12 MB |
| xnnpack + coreml + mlx | ~25–28 MB |

iOS starts leaner than Android because the runtime is dead-stripped, then MLX roughly triples it.

**What I could not measure.** OpenCV ships as static archives (14.68 MB of `.a` on Android: `imgproc` 5.62, `core` 4.92, `features2d` 1.27, `photo` 0.98, `video` 0.75, `highgui` 0.26, `kleidicv*` 0.88), so only referenced objects link into `libreact-native-executorch.so` — the 14.68 MB is not the cost. Phonemis has no artifact at all; it's built from the in-tree submodule and the flag only gates compilation. Both fold into the main library and need a real build to quantify. Rough guesses: OpenCV 3–6 MB, phonemis 0.5–3 MB, which would put the full Android spread at roughly 18–20 MB lean to 34–38 MB loaded.

I have a script that measures the Android matrix properly by building a release APK per config and reading `lib/<abi>/*.so` out of it — happy to contribute it if useful. iOS needs a different workflow (archive, Ad-Hoc distribute, read `App Thinning Size Report.txt`).

---

## Suggested fixes

1. Build **both** `core-android-<abi>.tar.gz` and `core-ios.tar.gz` without backend binaries. On iOS this removes ~36 MB of unreferenced download and makes the standalone backend tarballs non-redundant; on Android it fixes the APK bloat directly. (§1, §2)
2. Android: prune backends absent from the resolved config after extracting core — belt and braces if (1) lands. (§3)
3. Android: adopt the podspec's link-time gating. Either point `jniLibs.srcDirs` at a staging dir populated per-config, or drive `packaging { jniLibs.excludes }` from `rne-build-config.json`. (§4, §5)
4. Clear the extraction dir before untarring, both platforms. (§6)
5. ~~Investigate Vulkan build determinism; if shader codegen ordering is the cause, sorting it removes the §7 ambiguity.~~ (§7) — not a determinism problem: the two binaries are different source revisions, and §1 removes the duplicate copy. See the note in that section.
6. `COPYFILE_DISABLE=1` when packing, or pack in CI. (§12)
7. ~~Check whether `.eh_frame` suppression can be enabled for these builds.~~ (§11) — measured at 1.70 MB, 4.7% of the payload. Not worth the loss of native stack traces; closed as won't-do.
8. Add a bundle-size section to the docs, both platforms. (§13)
9. Decide what to do about MPS. (§9)
10. Consider making iOS simulator slices opt-out. (§10) — much smaller after §1 landed: 9.37 MB rather than 28.99 MB, see the note in that section.

Happy to open PRs — (1) through (4) and (6) are small and self-contained.

## Reproduction

```bash
TAG=v0.10.0-libs
BASE=https://github.com/software-mansion/react-native-executorch/releases/download/$TAG

for a in core-android-arm64-v8a vulkan-android-arm64-v8a xnnpack-android-arm64-v8a \
         core-ios xnnpack-ios coreml-ios mlx-ios; do
  curl -sL -o $a.tar.gz "$BASE/$a.tar.gz"
  mkdir -p $a && tar xzf $a.tar.gz -C $a
done

# §1 — core ships every backend, both platforms
tar tzvf core-android-arm64-v8a.tar.gz | grep -v '/\._'
find core-ios -name '*.a' ! -name '._*' -exec ls -l {} + | awk '{print $5, $9}' | sort -rn

# §2 — iOS standalone artifacts are byte-identical duplicates
sha256sum core-ios/libs/executorch/libbackend_mlx_ios.a \
          mlx-ios/MLXBackend.xcframework/ios-arm64/libMLXBackend.a \
          core-ios/libs/executorch/mlx.metallib \
          mlx-ios/libs/executorch/mlx.metallib

# §5 — what the podspec actually links
grep -n 'force_load\|vendored_frameworks\|s.ios.resource' \
  packages/react-native-executorch/react-native-executorch.podspec

# §7 — xnnpack identical, vulkan not
sha256sum */executorch/arm64-v8a/*.so

# §11 — section breakdown
size -A vulkan-android-arm64-v8a/executorch/arm64-v8a/libvulkan_executorch_backend.so
```

To confirm §4 end to end, build a release APK for an app declaring `backends: ["vulkan"]`:

```bash
unzip -l app-release.apk | grep 'lib/arm64-v8a/.*\.so'
```

## Caveats

I read the sources and inspected the published artifacts; I did not run an Android or iOS build. Specifically: §4 is inferred from `jniLibs.srcDirs` semantics rather than observed in an APK, and §8 and the iOS half of §13 are inferred from `-force_load` semantics rather than read off an App Thinning report. Worth confirming both before anyone acts on the numbers.

msluszniak:
## Description

Fixes the parts of #1464 that hold up under checking. The opt-out gated downloads and linking; it did not control what the `core-*` artifacts contain, and on Android it did not control what ships.

Artifacts republished as [`v0.10.1-libs`](https://github.com/software-mansion/react-native-executorch/releases/tag/v0.10.1-libs). **Repackaging only: every library byte is identical to v0.10.0-libs, nothing was recompiled**, so this cannot introduce a behaviour change or hit the Vulkan build's non-determinism.

| artifact | v0.10.0-libs | v0.10.1-libs |
|---|---|---|
| core-android-arm64-v8a | 7.96 MB | 4.49 MB |
| core-android-x86_64 | 8.50 MB | 4.50 MB |
| core-ios | 20.82 MB | 5.87 MB |

### What changed

- **`package-release-artifacts.sh`** stages the runtime only into core. MPS goes with it (§9): `libbackend_mps_*.a` had no toggle in `ALL_BACKENDS` and the podspec never referenced it.
- **`build.gradle.kts`** drives `packaging.jniLibs.excludes` from the resolved config (§4, §5), the Android counterpart of the podspec choosing its `-force_load` entries.
- **`download-libs.js`** prunes backends absent from the config after extracting (§3, §6), so turning one off after an install takes effect instead of silently shipping the old binary.
- **`COPYFILE_DISABLE=1`** when packing (§12).

### Verification

**§4 was inferred in the issue; it is now observed.** apps/legacy/bare-rn, release, arm64-v8a, `backends: ["vulkan"]`, old core staged so XNNPACK is present on disk:

| build | merged_native_libs | libxnnpack |
|---|---|---|
| without the gradle fix | 269,127,104 B | present, 2,673,072 B |
| with `jniLibs.excludes` | 266,454,032 B | absent |

Delta is exactly the XNNPACK library. Read from `merged_native_libs/release/.../arm64-v8a`, the tree the APK packages.

**iOS**: a Release link of the same app against the reduced `core-ios` succeeds. Everything the linker takes from `third-party/ios`:

```
libs/executorch/libthreadpool_ios.a
XnnpackBackend.xcframework/ios-arm64/libXnnpackBackend.a   (-force_load)
CoreMLBackend.xcframework/ios-arm64/libCoreMLBackend.a     (-force_load)
MLXBackend.xcframework/ios-arm64/libMLXBackend.a           (-force_load)
+ ExecutorchLib.framework
```

None of the 21 removed archives appears, so §5 is confirmed by the linker rather than by reading the podspec.

**§6**: turning Vulkan on, then off, now removes the stale binary instead of leaving it to ship.

### One claim in the issue does not reproduce

**§12: there are no AppleDouble sidecars in v0.10.0-libs.** `tar tzf | grep -c '\._'` is 0 for all eight tarballs checked, against the 7 the issue reports. `COPYFILE_DISABLE=1` is added anyway, since it costs nothing and the failure mode is real on macOS.

### Not addressed here

§7 root cause (the two Vulkan binaries differ; fixing §1 removes the *consequence*, since the standalone build is now the only one that can ship, but not the cause), §8 and §11 (inferred from `-force_load` and section sizes, no archive build run), §10 simulator split, §13 docs.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

`nativeLibsVersion` is bumped to 0.10.1, so a fresh install pulls the new artifacts. Declare `backends: ["vulkan"]` in an app's package.json, build a release APK, and check `unzip -l app-release.apk | grep 'lib/arm64-v8a/.*\.so'`: `libxnnpack_executorch_backend.so` should be absent. Flip a backend off after an install and confirm the postinstall log reports removing it.

### Related issues

Refs #1464.

### DIFF
--- a/docs/docs/03-core-and-advanced/08-native-libraries.md
+++ b/docs/docs/03-core-and-advanced/08-native-libraries.md
@@ -93,7 +93,7 @@ Specifying a task under `features` is shorthand: it automatically expands to the
 | `imageEmbeddings`      | xnnpack, coreml, mlx, vulkan | opencv              |
 | `classification`       | xnnpack, coreml              | opencv              |
 | `objectDetection`      | xnnpack, coreml              | opencv              |
-| `keypointDetection`    | xnnpack, coreml              | opencv              |
+| `keypointDetection`    | xnnpack, coreml, mlx         | opencv              |
 | `semanticSegmentation` | xnnpack, coreml              | opencv              |
 | `instanceSegmentation` | xnnpack, coreml              | opencv              |
 | `ocr`                  | xnnpack, coreml, vulkan      | opencv              |
--- a/docs/docs/03-core-and-advanced/08-native-libraries.md
+++ b/docs/docs/03-core-and-advanced/08-native-libraries.md
@@ -102,3 +102,60 @@ Specifying a task under `features` is shorthand: it automatically expands to the
 | `textToImage`          | xnnpack, coreml              | opencv              |
 | `segmentAnything`      | xnnpack, coreml              | opencv              |
 | `tokenizer`            | —                            | —                   |
+
+## Binary size
+
+What the backends actually cost, measured rather than estimated. Both tables come
+from `apps/legacy/bare-rn` built in Release with OpenCV and phonemis enabled, so
+the absolute figures move with which features you compile in. **The deltas between
+rows are the transferable part.**
+
+### Android
+
+`arm64-v8a`, symbols stripped as the APK ships them. Covers `libexecutorch.so`,
+the backend libraries, and `libRnExecutorch.so`; it excludes React Native's own
+libraries, which do not change with this configuration.
+
+| backends             | in the APK | ≈ Play download |
+| -------------------- | ---------- | --------------- |
+| `xnnpack`            | 25.18 MB   | 8.16 MB         |
+| `vulkan`             | 33.83 MB   | 9.79 MB         |
+| `xnnpack` + `vulkan` | 36.38 MB   | 10.68 MB        |
+
+Per library:
+
+| library                            | stripped | gzipped |
+| ---------------------------------- | -------- | ------- |
+| `libexecutorch.so`                 | 12.82 MB | 4.39 MB |
+| `libvulkan_executorch_backend.so`  | 11.19 MB | 2.52 MB |
+| `libRnExecutorch.so`               | 9.80 MB  | 2.87 MB |
+| `libxnnpack_executorch_backend.so` | 2.54 MB  | 0.88 MB |
+
+`extractNativeLibs=false` is the default from React Native 0.73, so the install
+grows by the uncompressed figure; the Play column is the compressed transfer.
+
+Vulkan is ~4.4x XNNPACK and almost entirely data: 8.73 MB of its 11.19 MB is
+`.rodata`, effectively embedded SPIR-V, against 2.02 MB of `.text`. It also
+compresses far better than its size suggests.
+
+### iOS
+
+Mach-O size of the app binary, `arm64`, device slice. Each row is a full relink.
+
+| backends                     | app binary | added by the last backend |
+| ---------------------------- | ---------- | ------------------------- |
+| none                         | 19.33 MB   | —                         |
+| `xnnpack`                    | 20.75 MB   | +1.41 MB                  |
+| `xnnpack` + `coreml`         | 21.12 MB   | +0.37 MB                  |
+| `xnnpack` + `coreml` + `mlx` | 26.07 MB   | +4.95 MB                  |
+
+MLX also ships `mlx.metallib` (1.07 MB) as a bundle resource, which is on top of
+the binary.
+
+A backend costs roughly a third of its archive on disk: XNNPACK is a 3.79 MB
+archive for 1.41 MB linked, MLX 15.09 MB for 4.95 MB. The backends are attached
+with `-force_load`, which defeats archive member selection but **not**
+dead-stripping, so the linker still drops what your app cannot reach.
+
+iOS starts leaner than Android because the ExecuTorch runtime is dead-stripped
+into the binary rather than shipped as a standalone library.
--- a/docs/docs/03-core-and-advanced/08-native-libraries.md
+++ b/docs/docs/03-core-and-advanced/08-native-libraries.md
@@ -93,7 +93,7 @@ Specifying a task under `features` is shorthand: it automatically expands to the
 | `imageEmbeddings`      | xnnpack, coreml, mlx, vulkan | opencv              |
 | `classification`       | xnnpack, coreml              | opencv              |
 | `objectDetection`      | xnnpack, coreml              | opencv              |
-| `keypointDetection`    | xnnpack, coreml, mlx         | opencv              |
+| `keypointDetection`    | xnnpack, coreml              | opencv              |
 | `semanticSegmentation` | xnnpack, coreml              | opencv              |
 | `instanceSegmentation` | xnnpack, coreml              | opencv              |
 | `ocr`                  | xnnpack, coreml, vulkan      | opencv              |
--- a/docs/docs/05-other/02-troubleshooting.md
+++ b/docs/docs/05-other/02-troubleshooting.md
@@ -164,3 +164,15 @@ app to the supported ABIs:
 ```properties
 reactNativeArchitectures=arm64-v8a,x86_64
 ```
+
+## `global.loadOCR is not a function`
+
+The deprecated legacy API registers one `load*` global per task, and the tasks
+backed by a native third-party library are only registered when that library is
+compiled in. Opting out of `opencv` therefore drops the legacy vision hooks
+(`useOCR`, `useVerticalOCR`, `useClassification`, `useObjectDetection`,
+`usePoseEstimation`, `useStyleTransfer`, `useSemanticSegmentation`,
+`useInstanceSegmentation`, `useImageEmbeddings`, `useTextToImage`), and opting
+out of `phonemis` drops `useTextToSpeech`. They build fine and throw this on
+first use. Add the library back under `libs`, or list a `feature` that expands
+to it, in [Native Libraries](../03-core-and-advanced/08-native-libraries.md).
--- a/packages/react-native-executorch/__tests__/api/modelVariants.test.ts
+++ b/packages/react-native-executorch/__tests__/api/modelVariants.test.ts
@@ -470,8 +470,8 @@ describe('feature map', () => {
     for (const [, feature, backends, libs] of doc.matchAll(
       /^\|\s*`(\w+)`\s*\|\s*([^|]*?)\s*\|\s*([^|]*?)\s*\|$/gm
     )) {
-      if (FEATURE_MAP[feature])
-        documented.set(feature, { backends: cell(backends), libs: cell(libs) });
+      if (FEATURE_MAP[feature!])
+        documented.set(feature!, { backends: cell(backends!), libs: cell(libs!) });
     }
 
     expect([...documented.keys()].sort()).toEqual(Object.keys(FEATURE_MAP).sort());
--- a/packages/react-native-executorch/__tests__/api/modelVariants.test.ts
+++ b/packages/react-native-executorch/__tests__/api/modelVariants.test.ts
@@ -394,4 +394,27 @@ describe('feature map', () => {
 
     expect(offenders.sort()).toEqual([]);
   });
+
+  it('provisions no backend the registry does not publish for that feature', () => {
+    // The mirror of the case above, and the one that actually bites: a feature
+    // naming a backend nothing in its family exports makes every app using that
+    // feature download a library it can never load. keypointDetection carried
+    // `mlx` this way, which no keypoint model has ever published.
+    const offenders: string[] = [];
+
+    for (const [category, node] of Object.entries(registry)) {
+      const feature = FEATURE_OF_CATEGORY[category]!;
+      // `llm` is the one category two features cover, so neither entry alone has
+      // to justify every backend it names.
+      if (feature === 'llm') continue;
+
+      const published = publishedBackends(node);
+      for (const backend of FEATURE_MAP[feature].backends) {
+        if (!published.has(backend))
+          offenders.push(`${feature} provisions ${backend}, unpublished`);
+      }
+    }
+
+    expect(offenders.sort()).toEqual([]);
+  });
 });
--- a/packages/react-native-executorch/__tests__/api/modelVariants.test.ts
+++ b/packages/react-native-executorch/__tests__/api/modelVariants.test.ts
@@ -369,6 +369,46 @@ describe('feature map', () => {
     return found;
   }
 
+  /** The Hugging Face repo slugs a registry category's URLs point at. */
+  function repoSlugs(node: unknown): Set<string> {
+    const found = new Set<string>();
+    const walk = (value: unknown): void => {
+      if (typeof value === 'string') {
+        const match = value.match(/react-native-executorch-([^/]+)\//);
+        if (match) found.add(match[1]!);
+      } else if (Array.isArray(value)) value.forEach(walk);
+      else if (isObject(value)) Object.values(value).forEach(walk);
+    };
+    walk(node);
+    return found;
+  }
+
+  // `features` provisions for the whole package, and the legacy API keeps model
+  // URLs the new registry has since dropped: the RF-DETR keypoint MLX export
+  // removed in #1418 is still reachable as
+  // RF_DETR_KEYPOINT_PREVIEW_MLX_FP32_MODEL. Those count as published, matched
+  // to a feature through the repo slug they share with the new registry; a
+  // legacy-only slug cannot be classified that way and is skipped.
+  const legacyBackendsBySlug = (() => {
+    const source = readFileSync(join(__dirname, '../../legacy/src/constants/modelUrls.ts'), 'utf8');
+    const bySlug = new Map<string, Set<string>>();
+    for (const [, slug, backend] of source.matchAll(
+      /\$\{URL_PREFIX\}-([^/`]+)\/[^`]*?\/(xnnpack|coreml|mlx|vulkan)\//g
+    )) {
+      if (!bySlug.has(slug!)) bySlug.set(slug!, new Set());
+      bySlug.get(slug!)!.add(backend!);
+    }
+    return bySlug;
+  })();
+
+  /** New-registry exports for a category, plus the legacy ones for its repos. */
+  function allPublishedBackends(node: unknown): Set<string> {
+    const found = publishedBackends(node);
+    for (const slug of repoSlugs(node))
+      for (const backend of legacyBackendsBySlug.get(slug) ?? []) found.add(backend);
+    return found;
+  }
+
   const registry = registryFor({ os: 'ios' });
 
   it('names a feature for every registry category', () => {
@@ -395,11 +435,10 @@ describe('feature map', () => {
     expect(offenders.sort()).toEqual([]);
   });
 
-  it('provisions no backend the registry does not publish for that feature', () => {
+  it('provisions no backend neither API publishes for that feature', () => {
     // The mirror of the case above, and the one that actually bites: a feature
     // naming a backend nothing in its family exports makes every app using that
-    // feature download a library it can never load. keypointDetection carried
-    // `mlx` after the one keypoint MLX export was dropped in #1418.
+    // feature download a library it can never load.
     const offenders: string[] = [];
 
     for (const [category, node] of Object.entries(registry)) {
@@ -408,7 +447,7 @@ describe('feature map', () => {
       // to justify every backend it names.
       if (feature === 'llm') continue;
 
-      const published = publishedBackends(node);
+      const published = allPublishedBackends(node);
       for (const backend of FEATURE_MAP[feature].backends) {
         if (!published.has(backend))
           offenders.push(`${feature} provisions ${backend}, unpublished`);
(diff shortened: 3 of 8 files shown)


docs/docs/05-other/02-troubleshooting.md
```diff
@@ -164,3 +164,15 @@ app to the supported ABIs:
 ```properties
 reactNativeArchitectures=arm64-v8a,x86_64
 ```
+
+## `global.loadOCR is not a function`
```
barhanc: Not sure we want the legacy stuff in troubleshooting.

docs/docs/03-core-and-advanced/08-native-libraries.md
```diff
@@ -102,3 +102,60 @@ Specifying a task under `features` is shorthand: it automatically expands to the
 | `textToImage`          | xnnpack, coreml              | opencv              |
 | `segmentAnything`      | xnnpack, coreml              | opencv              |
 | `tokenizer`            | —                            | —                   |
+
+## Binary size
```
barhanc, line 106: In the current form I feel this section is not very accessible to a normal user. A simple table: this backends adds this much weight would be better imo.

msluszniak: Ok, added simple table instead.

docs/docs/05-other/02-troubleshooting.md
msluszniak: Removed.

barhanc (approved):

PR #1471, opened 2026-09-11
fix(build): gate the legacy sources on phonemis and opencv

### LINKED ISSUE #1470: Opting out of a native lib breaks the build: phonemis and opencv both
Any app that opts out of text-to-speech fails to build for a device, on both platforms.

The legacy C++ is globbed into the library unconditionally, but phonemis is only compiled and linked when `RNE_ENABLE_PHONEMIS` is on, so `legacy/cpp/rnexecutorch/models/text_to_speech/kokoro/Kokoro.cpp` is built without it.

Android stops at the include:

```
legacy/cpp/rnexecutorch/models/text_to_speech/kokoro/Kokoro.h:14:10:
fatal error: 'phonemis/base/pipeline.h' file not found
```

iOS keeps the phonemis `src` dir on `HEADER_SEARCH_PATHS`, so it compiles and dies at link instead:

```
Undefined symbols for architecture arm64:
  "phonemis::Pipeline::preprocess(...)", referenced from: ... Kokoro.o
```

The non-legacy tree already handles this: `phonemizer.cpp` is filtered out of the sources and its callers sit behind `#ifdef RNE_ENABLE_PHONEMIS`. The legacy tree never got the same treatment.

### Reproduction

Build `apps/computer-vision` or `apps/nlp` for a device. Neither asks for `textToSpeech`, so phonemis is off for both. `apps/speech` builds because it asks for TTS, and `apps/legacy` builds because it declares no config block at all and defaults everything on.

### Impact

This hits anyone following [Native libraries](https://docs.swmansion.com/react-native-executorch/docs/core-and-advanced/native-libraries) to trim their binary, which is the documented reason to declare `features` in the first place.

msluszniak:
## Description

Opting out of a native lib broke the build on both platforms. `legacy/cpp/**` is globbed into the library unconditionally, but phonemis and opencv are only compiled and linked under their flags, so the legacy sources were built against libraries that were not there. Android stopped at the include, iOS at the link.

Both now follow the toggle the current API already uses: the sources leave the build, the registration sites are guarded, and the `load*` globals for those tasks are absent, the same way the current API omits `speech.createPhonemizer` and the whole `cv` namespace. `GlobalThreadPool` was the awkward one, core infrastructure that needed opencv for a single `setNumThreads` call.

Also fixes the legacy test suite's `PACKAGE_ROOT`, which has pointed one directory short since the pre-0.10 tree moved under `legacy/cpp/` and left the suite unable to configure at all.

Only sources are excluded, not headers: the `Types.h` files next to them are opencv-free and `JsiConversions.h` includes them unconditionally.

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [x] iOS
- [x] Android

### Testing instructions

Build `apps/speech` (`opencv=false, phonemis=true`). It failed before this on both platforms. Then build `apps/computer-vision` (`opencv=true, phonemis=false`) and an all-libs-on config to check the enabled paths.

### Related issues

Closes #1470

### Checklist

### Additional notes

`apps/speech` builds clean both platforms: Android native lib and debug APK, iOS pod and app for simulator. The ninja graph drops from 116 objects to 84, with no opencv archives on the link line and no undefined `cv::` symbols in the `.so`. The all-libs-on build (compile and link only, no runtime run of the legacy vision path) still registers every legacy global: the enabled `.so` carries all ten `loadOCR`/`loadStyleTransfer`/... symbols and the opted-out one carries none. The podspec exclusions are inert when opencv is on.

A legacy LLM built with `capabilities: ["vision"]` now throws a named error when opencv is off, rather than failing to link.

### DIFF
--- a/docs/docs/03-core-and-advanced/08-native-libraries.md
+++ b/docs/docs/03-core-and-advanced/08-native-libraries.md
@@ -77,8 +77,6 @@ Hardware backends provide optimized execution kernels for specific processors an
 - **[OpenCV](https://opencv.org/)** — High-performance computer vision library providing image transformations, color space conversions, resizing, and pixel format operations (used by vision pipelines and multimodal LLMs). Provided on iOS via CocoaPods and on Android as static prebuilt libraries.
 - **[phonemis](https://github.com/IgorSwat/Phonemis)** — High-performance C++ library for Grapheme-to-Phoneme (G2P) conversion, delivering universal IPA phonemization as the frontend for [Text-to-Speech](../02-extensions/speech/02-text-to-speech.md) pipelines. Compiled from source on both Android and iOS when enabled.
 
-Opting a library out also drops the matching hooks from the deprecated legacy API, which then throw at runtime (`global.loadOCR is not a function`) rather than failing the build.
-
 ### Features
 
 Specifying a task under `features` is shorthand: it automatically expands to the union of backends and native libraries required by the pre-exported models in that domain.
--- a/docs/docs/03-core-and-advanced/08-native-libraries.md
+++ b/docs/docs/03-core-and-advanced/08-native-libraries.md
@@ -77,6 +77,8 @@ Hardware backends provide optimized execution kernels for specific processors an
 - **[OpenCV](https://opencv.org/)** — High-performance computer vision library providing image transformations, color space conversions, resizing, and pixel format operations (used by vision pipelines and multimodal LLMs). Provided on iOS via CocoaPods and on Android as static prebuilt libraries.
 - **[phonemis](https://github.com/IgorSwat/Phonemis)** — High-performance C++ library for Grapheme-to-Phoneme (G2P) conversion, delivering universal IPA phonemization as the frontend for [Text-to-Speech](../02-extensions/speech/02-text-to-speech.md) pipelines. Compiled from source on both Android and iOS when enabled.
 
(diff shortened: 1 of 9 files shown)


docs/docs/03-core-and-advanced/08-native-libraries.md
```diff
@@ -77,6 +77,8 @@ Hardware backends provide optimized execution kernels for specific processors an
 - **[OpenCV](https://opencv.org/)** — High-performance computer vision library providing image transformations, color space conversions, resizing, and pixel format operations (used by vision pipelines and multimodal LLMs). Provided on iOS via CocoaPods and on Android as static prebuilt libraries.
 - **[phonemis](https://github.com/IgorSwat/Phonemis)** — High-performance C++ library for Grapheme-to-Phoneme (G2P) conversion, delivering universal IPA phonemization as the frontend for [Text-to-Speech](../02-extensions/speech/02-text-to-speech.md) pipelines. Compiled from source on both Android and iOS when enabled.
 
+Opting a library out also drops the matching hooks from the deprecated legacy API, which then throw at runtime (`global.loadOCR is not a function`) rather than failing the build.
```
msluszniak: I would remove this comment from here and add note to troubleshooting in PR 1466. This PR will be merged first.

msluszniak:
Merging this without approval due to urgency of this change. It was tested and verified by myself pretty well.

PR #1363, opened 2026-08-16
[RNE Rewrite] feat: benchmark harness for every registry variant

msluszniak:
## Description

Adds `apps/benchmarks`, a headless Expo app that measures load time, inference latency and peak memory for every published registry variant (163 Android / 234 iOS, generated from `models.ts`), plus a collector, a summarizer and a comparator.

Notes:
- Execute time is measured in-band via a new `getExecutionProfile()` API, so `Execute %` reflects the shapes and call counts the pipeline actually used.
- Memory is a separate pass (reading PSS costs milliseconds) and uses process footprint, which counts mmap'd `.pte` pages.
- Inputs are synthetic and deterministic, so two devices feed identical bytes.
- Runs are release builds and gated on device temperature before every measurement.

Protocol for other devices: `apps/benchmarks/BENCHMARK_SPEC.md`.

### Introduces a breaking change?

- [ ] Yes
- [x] No

### Type of change

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Documentation update (improves or adds clarity to existing documentation)
- [ ] Other (chores, tests, code style improvements etc.)

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

```bash
cd apps/benchmarks
yarn bench --platform android --suite quick --label v0.10.0 --max-temp-c 37
yarn bench:summary ~/rne-benchmarks/v0.10.0-android-SM-S948B.jsonl
```

### Related issues

Closes #1078

### Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation accordingly
- [x] My changes generate no new warnings

### Additional notes

Not yet run on iOS.

PR #1479, opened 2026-09-18
feat(models): wire the published Vulkan exports and bump nativeLibsVersion

msluszniak:
## Description

Wires the 44 Vulkan `.pte` files published under the HF `v0.11.0` tag: style transfer (4), YOLO26 detection (15), YOLO26-seg (15), YOLO26-pose (3), RF-DETR detector/segmentation/keypoint, FastSAM s/x, BlazeFace and SDXS. They become the Android default through the existing `variants` resolution.

New URLs resolve through `NEXT_VERSION_TAG` (`v0.11.0`), since a tag carries only the files cut under it and each model's older exports stay on `VERSION_TAG`.

`nativeLibsVersion` moves to `0.10.4`: its core carries the constant-segment fix without which RF-DETR fails to load (`0x23`), and its Vulkan backend is 1.8-3.8x faster on Mali for the conv models added here.

`FEATURE_MAP` and the native-libraries doc gain `vulkan` for `objectDetection`, `keypointDetection`, `instanceSegmentation`, `styleTransfer`, `textToImage` and `segmentAnything`.

Five published Vulkan exports are deliberately **not** wired, because each measured slower than the Android default it would replace: `efficientnet-v2-s` (fp32 export, 2.4x slower than XNNPACK int8 on Adreno), `selfie-segmentation` and its landscape checkpoint (1.2x slower on Adreno), `lraspp` (0.86x vs int8 on Adreno) and `ssdlite320` (loses on both GPUs).

- [ ] Yes
- [x] No

### Type of change

### Tested on

- [ ] iOS
- [x] Android

### Testing instructions

Every one of the 44 URLs was HEAD-checked against HF for a 200 and a plausible content-length. `yarn workspace react-native-executorch test` (4082 tests), `tsc --noEmit` and `yarn lint` pass; the registry's own `modelVariants` invariants cover the Android default resolution.

### Checklist


barhanc (approved):
I tested selected models from each model family and they work correctly. One note that might be important is that on S20+ I didn't observe any speedup compared to xnnpack version and in some cases Vulkan is noticeably slower (e.g. FastSAM ~800ms on cpu vs ~1s on Vulkan). 

I guess we can merge it after conflict resolution.

msluszniak:
> I tested selected models from each model family and they work correctly. One note that might be important is that on S20+ I didn't observe any speedup compared to xnnpack version and in some cases Vulkan is noticeably slower (e.g. FastSAM S ~800ms on cpu vs ~1s on Vulkan).
> 
> I guess we can merge it after conflict resolution.

You didn't observe any speedup at all on any model, or there were some models with regressions and no speedup on s20+?

barhanc:
Some models were slightly faster, e.g. the Yolo variants across different tasks (on the order of ~10% and the speedup was stable and consistent). I didn't observe any huge speedups (like the CoreML vs xnnpack case on iOS) and on the large models e.g. FastSAM vulkan was slower + I observed that the UI (of the whole phone not just the app) was slightly laggy when the heavy inference was running (probably because of using the same GPU resources as rendering).

barhanc:
@msluszniak I've added the S20+ benchmark harness results to #1363 PR comment and they also seem to reproduce the general behavior:
- Vulkan Supertonic is noticealby slower than xnnpack (~5.6s vs ~5s).
- Similar speed regression on large Whispers.
- On small model like PaddleOCR there is a speedup of the order 10-20%.

